PR #105, opened 2021-05-16
Remove some verified email banners

lfaucon:
This PR removes the verified email banner, because it is in very non-user friendly places. It now only appears on the home page


sergeivolodin:
I remember @lenhoanglnh wanted them everywhere - @lenhoanglnh do you confirm they should only be on the front page?

lenhoanglnh:
I confirm that they should only appear in the front page.

sergeivolodin (approved):
LGTM

PR #121, opened 2021-05-18
Remove titles for pending and rejected emails domains

lfaucon:
The API has been removed for this, so the titles are now useless


sergeivolodin (commented):
lgtm

PR #122, opened 2021-05-18
Add some statistics on the home page

### LINKED ISSUE #40: Non-certified contributors also in home page
The home page should also display both the number of certified contributors and the total number of contributors.
Also, the number of ratings displayed on the home page should be the total number of ratings (both from certified and non-certified contributors).

(@sergeivolodin we probably need an API from the back end to get these statistics)

lfaucon:
Closes #40 

@sergeivolodin can you make sure that text under the numbers is correct?


sergeivolodin:
- `weekly_active_ratings` shows the number of ratings changed/created within the last week, not the number of active contributors
- Maybe instead of "users" we should say "total contributors"?
- Maybe make it 1 row instead of 2?

image

sergeivolodin (changes requested):
Waiting for the issues mentioned in the comment to be resolved

https://github.com/tournesol-app/tournesol/pull/122#issuecomment-843963989

lfaucon:
Thanks for the review and corrections! I'll make the changes

lfaucon:
@sergeivolodin Because @TheDuckWhisperer developped a complete redesign of the front page, I think it is not a big deal to have two line for the statistics temporarily. Also it looks good, doesn't it?

sergeivolodin (approved):

Issue #124, opened 2021-05-19
Fix privacy policy formatting


sergeivolodin:
![Untitled](https://user-images.githubusercontent.com/1012270/118802412-62fea380-b8a2-11eb-879d-00a26ed1cee6.png)

PR #129, opened 2021-05-19
fix the HTML formatting of privacy policy page

GresilleSiffle:
related to #124

### DIFF
--- a/frontend/src/components/PrivacyPolicy.js
+++ b/frontend/src/components/PrivacyPolicy.js
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { makeStyles } from '@material-ui/core/styles';
+import Typography from '@material-ui/core/Typography';
 
 const useStyles = makeStyles(() => ({
   root: {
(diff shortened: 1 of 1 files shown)


sergeivolodin (commented):
Looks good, can you attach a screenshot with how it looks now? This one gives one screenshot for the whole page: http://gofullpage.com/

GresilleSiffle:
I saw the frontend uses Material UI, so I think the usage of the [Typography](https://material-ui.com/components/typography/) component instead of the classic `<p>` tag should be better.

This way, the look and feel of the Privacy Policy page will be consistent with the rest of the site.

What do you think ?

GresilleSiffle:
The result with the `<p>` tag

privacy_policy_with_p

GresilleSiffle:
Here is a new screen shot with the usage of `<Typography paragraph component='p'>` instead of `<p>`.

The result is consistent with the other paragraphs in the application.

privacy_policy_with_typography

sergeivolodin:
Screenshots look good, thank you!

Right now tests fail because of lint issues, you can fix them with `npm run eslintFix`.
Generally, you can go to the workflow Details page and scroll down to the last message

image

GresilleSiffle:
> Right now tests fail because of lint issues, you can fix them with npm run eslintFix.

Thanks for the trick ! Sorry I didn't know there were a process involving eslint.

Are there other commands I should run before making a commit? (to be sure I didn't break anything)

sergeivolodin:
> > Right now tests fail because of lint issues, you can fix them with npm run eslintFix.
> 
> Thanks for the trick ! Sorry I didn't know there were a process involving eslint.
> 
> Are there other commands I should run before making a commit? (to be sure I didn't break anything)

Generally you just run `./tests.sh` (does the same thing as github CI) and manually test that everything works (test wouldn't catch things like broken layout)

sergeivolodin:
Thank you for the edit @GresilleSiffle !

Issue #118, opened 2021-05-18
Move "all_video_ratings.csv" into the public database


aidanjungo:
Right now, this file is in the user data when downloading them from "My profile" page, I propose to:

- rename this file "all_video_scores.csv" 
- add it in the public database instead of the user data
- add a column for the Tournesol score

GresilleSiffle:
Hello @aidanjungo

The modifications are now online on https://tournesol.app : ]

Is the result satisfying?

PR #132, opened 2021-05-20
move all video ratings CSV to the public database

GresilleSiffle:
related to #118 

edit: the pull request is now ready, with a new test and without any `flake8` warning

Hi !

This pull request is not ready yet, but I'm posting it nevertheless to open the discussion.

I'm still exploring the Tournesol code base and I didn't find how to add the Tournesol score to the `all_video_scores.csv` yet. In fact, I don't know where to look : >

Any hint?


lfaucon:
There are two places on the website where a user can download data. From the homepage, a user can download the public data base and from a user personnal page one can download one's own submitted comparisons and some additional data. (see two downloaded folders on screenshot) 

Screenshot 2021-05-20 at 21 30 19

What's wrong is that the file "all_video_ratings.csv" is not personnal data, but is public data. I believe it contains the aggregated scores for all videos on Tournesol.

lfaucon:
I had never looked at this part of the code previously, but I think you made the correct change. 

@sergeivolodin Can you confirm and merge if ok?

GresilleSiffle:
> I had never looked at this part of the code previously, but I think you made the correct change.

Yes, it also seems to me that the code which move from the personal download link to the public download link is fine.

But, there is one bullet point in the #118 this pull request doesn't resolve yet:

> add a column for the Tournesol score

Because I'm not familial enough with this notion. Is the score stored in the database or computed on the fly, I don't know : ] I think I need a little walkthrough in the backend models to understand how Tournesol works.

lfaucon:
Hey @GresilleSiffle !

Sorry, I did not understand before what you needed help with.

If you look at the code here, you will see how the tournesol score is computed: https://github.com/tournesol-app/tournesol/blob/f5000e3c85b4384c22afe2ca7ac62ec46898ccb2/backend/backend/api_v2/videos.py#L371

It is only a linear combination of the other features. Currently it the sum over the rating criterias of 50 * criteria_score. Best to use the proper constants when computing the score for the csv export

GresilleSiffle:
Hi @lfaucon 

Thanks for the hint.

I found a bug in `django-pandas` 0.6.2 (the currently used version) that pevents me to add the tournesol score computation within the `read_frame` call. So I've done this in two steps:

1. build the data frame without the score
2. add the score to the existing data frame with `Dataframe.insert()`

You can see the code in this commit: https://github.com/tournesol-app/tournesol/pull/132/commits/d624d916d356fe5b3d93d0d10538c003e300daa8

I'm going to try to upgrade `django-pandas` from 0.6.2 to latest version 0.6.4 to see if it helps. If not, I'll fix the integration tests I just broke : p , then I'll submit the code "as is" for review.

GresilleSiffle:
`django-pandas` 0.6.4 doesn't allow the usage of `QuerySet.annotate()` out of the box too.

I saw there were already a workaround in the file I was editing so I used it:

```python
# TODO: fix it
import django
django.db.models.fields.FieldDoesNotExist = django.core.exceptions.FieldDoesNotExist
```

Now I'm going to update the integration tests to work with the new zip files content.

@sergeivolodin @lfaucon if you want me to squash all my commits at the end, I'll do.

lfaucon:
My recommendation is to use this option (see screenshot when merrging every PR. 

Screenshot 2021-05-24 at 21 02 35

GresilleSiffle:
The branch is now ready for the review : )

I also took the time to re-order the import of the two files I edited to follow the Django coding style:
https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports

The import sections should be more readable now.

sergeivolodin (approved):

Issue #117, opened 2021-05-18
Add social media links on the homepage


aidanjungo:
Something like that, somewhere on the homepage.

![image](https://user-images.githubusercontent.com/53896963/118676267-8b7f9280-b7fb-11eb-944b-4796fe96d817.png)

GresilleSiffle:
Hi @aidanjungo 

One solution could be to add the icons in the yellow `<AppBar>` present in the top of the screen, so they will always be visible. Like on this site : https://material-ui.com

I don't know if it's really appropriate because I often see this kind of bars used to provide functionality related to the application like accessing my profile, see the notifications or the current pending actions (like the current number of " rate later " videos). Actions that keep the user on the same web site.

Nevertheless, it's an option : )

GresilleSiffle:
A classic implementation with standard Material UI components, without any CSS added looks like this.


![appbar_socialmedia](https://user-images.githubusercontent.com/39056254/118887324-7dfd0200-b8fa-11eb-8cb8-ffefd3ee1402.png)

aidanjungo:
Yes, it looks nice to me. If you want to work on the implementation of that, feel free 😃

sergeivolodin:
@aidanjungo please close if it looks good (live on tournesol.app in ~10mins)

PR #133, opened 2021-05-20
add social media links on the home page

GresilleSiffle:
related to #117 

I also added a `sectionDesktop` CSS class allowing to hide the social media links when the screen width is too small. It prevents the icons from stacking on each other on small screens, making the app bar height three times its original height.

If we still want to display these icons on small screens (i.e. width < 960 px), I can polish the display to make it look good.

Let me know : )

Oh and by the way, feel free to add, remove or re-order the social media links. I've added only those I was aware of.


sergeivolodin (commented):
Awesome! Looks good, **can you add a screenshot**? Thank you!

GresilleSiffle:
Sure :

appbar_socialmedia

sergeivolodin:
Awesome!

PR #169, opened 2021-10-07
[frontend]: the users can change their own password

GresilleSiffle:
**integration in the main UI**

- [x] add a settings page
- [x] add a link to the settings page in the top bar

**the settings sub-menu**

- [x] create a new component `SettingsMenu` in the features folder
- [x] add two items, profile and account
- [x] items can be selected
- [x] make the navigation work between items

The profile item will display all informations related to the person and not related to the user's account, like the first name, the last name, etc. On the other hand the account item will display all technical fields like the username, password, and so on.

**the settings > profile page**

- [x] add a " Under construction " message

**the settings > account page**

- [x] add the previously created sub-menu
- [x] make the page responsive
- [x] add a change password form
- [x] username and e-mail update will be in another pull request 
- [x] make the form send real HTTP requests
- [x] display success message
- [x] display all errors if needed
- [x] use `notistack` to stack messages on to of each other

**Also...**

The is no test for those components yet. I have to learn how to add them, following the `Login.spec.tsx` example.

I added in  `utils/notifications` a `contactAdministrator` function diplaying an error message followed by an invitation to contact the administrator. I'm using it as a fallback in the `PasswordForm` if the server return an error `HTTP 500`. It can be used anywhere in the application as long as the calling component uses `notistack`.  See:

settings_contact_admin

### DIFF
--- a/frontend/src/features/settings/account/PasswordForm.spec.tsx
+++ b/frontend/src/features/settings/account/PasswordForm.spec.tsx
@@ -1,0 +1,177 @@
+import React from 'react';
+
+import fetchMock from 'fetch-mock-jest';
+import { SnackbarProvider } from 'notistack';
+import { act } from 'react-dom/test-utils';
+import { Provider } from 'react-redux';
+import { MemoryRouter, Route, Switch } from 'react-router-dom';
+import configureStore, {
+  MockStoreCreator,
+  MockStoreEnhanced,
+} from 'redux-mock-store';
+import { AnyAction } from '@reduxjs/toolkit';
+import thunk, { ThunkDispatch } from 'redux-thunk';
+import { render, screen, fireEvent } from '@testing-library/react';
+
+import PasswordForm from './PasswordForm';
+import { initialState } from '../../login/loginSlice';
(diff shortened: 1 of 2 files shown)


GresilleSiffle:
> The settings actions could be moved to a floating menu, as it was suggested in the mock-ups. The expected behavior on mobile would need to be specified though.

Exactly. To keep the scope of this PR concise we have decided with Louis to plan it after the merge. I don't know if the floating menu will have exactly the same items as in the account submenu, in any case it will be a good UX addition.

> The app uses multiple mechanisms to handle error messages in forms. For example the account creation uses an uncontrolled form with errors assigned to each individual field, whereas this PR introduces "notistack". This would need to be standardized at some point, both for UX consistency and easier maintenance.

I fully agree. I tried `notistack` because it allows to stack several messages easily and nicely but it's indeed better if we use the same mechanism everywhere. I take the point, let's discuss about it in a meeting : )

GresilleSiffle:
I'm gonna add the three missing tests today, and this PR will be ready.

lfaucon:
Please merge @GresilleSiffle whenever you are happy with it :)

PR #180, opened 2021-10-08
Implement n_contributors per video

### LINKED ISSUE #199: Feature: VideoCard should show "rated by {N} contributors"
This should be implemented on both the backend and the frontend. Be careful aboiut performance because this feature caused performance issues in the previous version of Tournesol.

sandre35:
Closes #199


sandre35:
@amatissart do you have any idea on the frontend swagger behaviour ?

amatissart:
> @amatissart do you have any idea on the frontend swagger behaviour ?

The swagger itself looks ok, right? But the generated openapi "services" were not updated as we currently rely on a committed schema (in "frontend/scripts/tmp/openapi.yaml"). This behavior was changed as a temporary solution, to be able to run the CI on "frontend" and "backend" independently, but it's not satisfying to have to update this file manually.  (See #265)

In the short term, I have just updated this "openapi.yaml" file, with other minor adjustments to adapt the implementation. A single issue remains: the frontend uses a non-existing field `video.n_rating`.

amatissart (approved):

PR #244, opened 2021-10-25
Add simple end-to-end tests, based on dev-env and Cypress, and run them in CI

### LINKED ISSUE #200: Set up end-to-end tests

amatissart:
Closes #200 

A running [dev-env](https://github.com/tournesol-app/tournesol/tree/e2e-cypress/dev-env) is required to run the tests, as some setup operations currently depend on `docker` commands (initializing the database, fetch email content from logs, etc.)


lfaucon:
Do you think it is time to remove the duplication of tests for every PR now that you added a test that takes 5-6 minutes to run?

amatissart:
> Do you think it is time to remove the duplication of tests for every PR now that you added a test that takes 5-6 minutes to run?

What exactly would you recommend? We could disable e2e tests on "push" to keep them on "pull_request" only. And possibly disable backend tests for branches where  the "backend" folder is unchanged.  
It would not affect the time spent in CI though, as the different jobs run in parallel.

tests/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Requirements
+
+* A local instance of Tournesol, as provided by [dev-env](../dev-env/README.md)
+* Node.js 14
+
+### Install the dependencies
+
+```
+yarn install --frozen-lockfile
```
GresilleSiffle, line 13: We can precise that this command must be run from `/tests/`.

GresilleSiffle (approved):

tests/README.md
GresilleSiffle: I disagree with the me from the past.

GresilleSiffle:
Looks good to me. It's well documented and works on my laptop.

lfaucon:
> > Do you think it is time to remove the duplication of tests for every PR now that you added a test that takes 5-6 minutes to run?
> 
> What exactly would you recommend? We could disable e2e tests on "push" to keep them on "pull_request" only. And possibly disable backend tests for branches where the "backend" folder is unchanged. It would not affect the time spent in CI though, as the different jobs run in parallel.

I didn't realize that everything runs in parallel. I was mostly worried about delay before knowing a PR passes the tests.

Another worry I have is about limits for free compute time of GitHub actions. If I am not mistaken it was a problem with Tournesol 1.0.

I would recommend that on all PR each pipeline runs only one time.

No need to block this work for now. We can easily change this later

amatissart:
> I didn't realize that everything runs in parallel. I was mostly worried about delay before knowing a PR passes the tests.

Yes we need to keep an eye on it. I've just created issue #290 about this.  
The good news is most of the time is spent in the setup, and not in tests themselves. So it should not increase too much with additional tests. 

> Another worry I have is about limits for free compute time of GitHub actions. If I am not mistaken it was a problem with Tournesol 1.0.

Github Actions are completely free for public repositories, so this is not a worry, at least in the short term:
https://github.com/features/actions#pricing-details

> I would recommend that on all PR each pipeline runs only one time.

I updated the configuration to run the e2e workflow on "push_request" only (in addition to "main" branch).  
We'll see if we lack not having them on new branches before the PR is opened.

> No need to block this work for now. We can easily change this later

I will merge then! :champagne:

PR #251, opened 2021-10-27
[front] users can now update their username

### LINKED ISSUE #246: Enable changing username

GresilleSiffle:
Fixes #246 

**ProfileForm**
- [x] create a new `ProfileForm` component
- [x] retrieve the username from the API to initialize the form ( instead of querying the store ) 
- [x] use the new `showSuccessAlert` and `showErrorAlert` functions 
- [x] update the store after a username update
- [ ] add tests

**hooks.useLoginState**
- [x] now provides the `updateUsername` action
- [x] use `useCallback` to create the exported functions ( to avoid infinite loop when using those functions as `useEffect` dependencies ) 

**utils.nofitications**
- [x] the " contact an administrator " message now also suggest to retry later 

**notes**

The way I handle the generic OpenAPI success and error messages in the components `ProfileForm` and `PasswordForm` is exactly the same. It may be a good idea to factorize the code in order to:
- simplify the code of every form component
- avoid redundant code ( less code, less bugs )
- make every form reacts to rejected promises in the same way

```javascript 

      // this code is used in both components

        const newErrorMessages = Object.values(reason['body']).flat();
        newErrorMessages.map((msg) => showErrorAlert(enqueueSnackbar, msg));
      } else {
        contactAdministrator(
          enqueueSnackbar,
          'error',
          'Sorry, an error has occurred, cannot update the profile.'
        );
      }

     // ... and this one too

      if ('detail' in response) {
        showSuccessAlert(enqueueSnackbar, response['detail']);
      } else {
        showSuccessAlert(enqueueSnackbar, 'Profile changed successfully');
      }
```


GresilleSiffle:
I'll add an end-to-end after the PR https://github.com/tournesol-app/tournesol/pull/244 is merged.

amatissart (approved):

Issue #232, opened 2021-10-18
Disallow creating multiple accounts with the same email


lfaucon:
Currently it is possible to create multiple accounts with the same email address. This should not be possible.

lfaucon:
We are two weeks past the deadline given to users for updating their non-unique emails. I propose that we move on to close this ticket. Probably adding a PR that finds and update duplicated emails, then enforce the unique option.

PR #254, opened 2021-10-28
Update UserAdmin config to help maintenance and support on user accounts

amatissart:
Among the features brought by the standard `UserAdmin`, it provides a search bar to find users by email address.

This will help with the manual operations that are required by #232.

### DIFF
--- a/backend/core/admin.py
+++ b/backend/core/admin.py
@@ -2,7 +2,10 @@
 Defines Admin interface for Tournesol's core app
 """
 
+from typing import List, Tuple
 from django.contrib import admin
+from django.contrib.admin.utils import flatten_fieldsets
+from django.contrib.auth.admin import UserAdmin as DjangoUserAdmin
 from .models import (
     User,
     UserPreference,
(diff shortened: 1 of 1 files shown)


backend/core/admin.py
```diff
@@ -25,8 +28,27 @@ class DegreeAdmin(admin.ModelAdmin):
...
+    ordering = ['-date_joined']
+
+    def get_fieldsets(self, request, obj) -> List[Tuple]:
+        fieldsets = super().get_fieldsets(request, obj=obj)
+        if not obj:
+            return fieldsets
+
+        # For existing user, show all fields from our custom User model that
+        # are not already present on the page, in a separate fieldsets.
+        existing_fields = flatten_fieldsets(fieldsets)
```
lfaucon, line 41: Slightly more performant by calling `set(...)` here.

backend/core/admin.py
```diff
@@ -25,8 +28,27 @@ class DegreeAdmin(admin.ModelAdmin):
...
 @admin.register(User)
-class UserAdmin(admin.ModelAdmin):
-    pass
+class UserAdmin(DjangoUserAdmin):
+    ordering = ['-date_joined']
+
+    def get_fieldsets(self, request, obj) -> List[Tuple]:
+        fieldsets = super().get_fieldsets(request, obj=obj)
+        if not obj:
+            return fieldsets
```
lfaucon, line 37: This doesn't seem necessary. We might prefer it to raise an error if it happens?

lfaucon (approved):
The user admin got a lot nicer! Great. Approved with suggestions

backend/core/admin.py
amatissart: :+1:
Done.

backend/core/admin.py
amatissart: This case is actually necessary, in order to initialize the admin form used to create a new user.  
(Via the "Add User" button, or directly on http://localhost:8000/admin/core/user/add/ )

Issue #226, opened 2021-10-17
Improve keyword search


lfaucon:
Currently text based search is very minimalistic as it only checks that the search keyword is contained in the video title or description. There exist more advanced search algorithm that we would want to use instead.

- [x]  Add video tags on video endpoint
- [x]  Explore possibility to use `SearchVectorField` on backend --> Not possible for our needs

PR #257, opened 2021-10-29
Add tags to videos

sandre35:
Linked issue : https://github.com/tournesol-app/tournesol/issues/226
In this PR, we aim to solve the first task of the linked issue


backend/tournesol/views/video.py
```diff
@@ -41,8 +41,11 @@ def list(self, request, *args, **kwargs):
         queryset = Video.objects.all()
         search = request.query_params.get('search')
         if search:
-            queryset = queryset.filter(Q(name__icontains=search) |
-                                       Q(description__icontains=search))
+            queryset = queryset.filter(
+                Q(name__search=search) |
```
lfaucon, line 45: Could you link to a doc that explains what this `__search` does ? It is probably useful to leave a comment here too.

backend/tournesol/views/video.py
```diff
@@ -41,8 +41,12 @@ def list(self, request, *args, **kwargs):
...
         search = request.query_params.get('search')
         if search:
-            queryset = queryset.filter(Q(name__icontains=search) |
-                                       Q(description__icontains=search))
+            #  search keywork is about the same as icontains but is more efficient with a psql db
+            queryset = queryset.filter(
+                Q(name__search=search) |
+                Q(description__search=search) |
+                Q(tags__name__search=search)
+            )
```
amatissart, line 49: In what sense is it more efficient exactly?  
As far as I can see, it is indeed more robust to different query terms, as a tokenizer is used by Postgres to compare the data with the query.  But it also seems to be significantly slower than `__icontains`.

According to [Django documentation](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/search/#performance):
> Special database configuration isn’t necessary to use any of these functions, however, if you’re searching more than a few hundred records, you’re likely to run into performance problems. [...]

To make this approach work on production data, we would probably need to index the search vectors, for example using a [`SearchVectorField`](https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/search/#searchvectorfield).

amatissart (approved):
Looks good to me, except a comment is to be removed.

PR #260, opened 2021-10-30
feat: front end deployments now generate alerts

GresilleSiffle:
As does the back end.

I also added some " style " in the alerts to put an emphasis on the host name affecter by the deployment.


lfaucon (approved):
github.com new pull request approved

PR #262, opened 2021-11-01
Update EmailDomainAdmin for easier editing by admins

amatissart:
image

* Domains can be filtered by `status`, and the `status` can updated directly from the list.
* The search bar allows to filter by domain (or any substring of `domain`)
* It's possible to select multiple domains to mark them as "accepted" or "rejected" all at once, via the "action" menu:
  image


lfaucon (approved):
Awesome!

PR #263, opened 2021-11-01
Restrict ML to trusted comparisons, based on accepted email domains

amatissart:
### Changes

* A signal `pre_save` is used to create the `EmailDomain` related to any new or updated `User` email
* `User.trusted_users()` is a queryset containing only the users whose `email` is associated with an accepted domain.  
As the query is implemented using the Django ORM, the filter is applied at the database level, and it can be used efficiently in all contexts where a queryset of `User` is accepted.
* In the User admin, "trusted" or "non-trusted" users can be filtered from the list.
* ML: the input dataset defined in `fetch_data()` is restricted to comparisons created by trusted users.   
As a result, the video scores will be derived from these comparisons only, and non-trusted users won't have any `ContributorRatingCriteriaScore` associated with them.

### A question of terminology
I used the term **"trusted user"**, for consistency with "trusted domain" that is specified in the UI messages to be implemented. But in older versions **"certified user"** was also used, should we stick to it?

Anyway, this notion should not be confused with **"verified user"**, that is already used to describe a user whose email has been verified after registration (meaning that the email address is valid, without any other guarantee).


backend/core/tests/test_trusted_users.py
```diff
@@ -0,0 +1,25 @@
...
+
+class TestTrustedUsers(TestCase):
+    def test_user_trusted_based_on_email_domain(self):
+        user = User.objects.create_user(username="test-user", email="test@example.test")
+        email_domain = EmailDomain.objects.filter(domain="@example.test").first()
+        self.assertIsNotNone(email_domain)
+        self.assertFalse(user.is_trusted)
+
+        email_domain.status = EmailDomain.STATUS_ACCEPTED
+        email_domain.save()
```
sandre35, line 14:
```suggestion
        email_domain.update(status=EmailDomain.STATUS_ACCEPTED)
```

sandre35 (commented):
What confuses me a bit is that the `@` is registered inside the `domain`. I would suggest to remove it.

backend/core/signals.py
```diff
@@ -0,0 +1,27 @@
+import logging
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+from .models import User, EmailDomain
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(pre_save, sender=User)
+def ensure_that_email_domain_exists(sender, instance, update_fields, **kwargs):
```
GresilleSiffle, line 10: Is there any benefit here of using a signal instead of overwriting the `save` method of our `User` model?

Signals are really handy when it comes to make uncoupled Django apps react to each other, but they have the downside of creating scattered and less easy to maintain code.

In our case, both `EmailDomain` and `User` models are part of the same app so, overwriting the `User`'s `save` method won't make uncoupled folders coupled. 

I personally tend to avoid signals as much as I can (i.e. when I don't explicitely need them) so that all behaviours related to a model are visible in the model's file. I was also inspired by the authors of the book [Two Scoops of Django](https://www.feldroy.com/books/two-scoops-of-django-3-x) who ecourage, in the version 2.x of the book, using the model's hooks depending on your use case.

GresilleSiffle (commented):
Hi @amatissart 

I added a simple comment so that we can discuss about signals. I didn't hit the request changes button because I wanted to know your opinion before : )

backend/core/admin.py
```diff
@@ -27,9 +28,28 @@ class DegreeAdmin(admin.ModelAdmin):
...
 
 
+class IsTrustedFilter(admin.SimpleListFilter):
+    title = 'is trusted'
+    parameter_name = 'is_trusted'
+
+    def lookups(self, request, model_admin):
+        return (
+            ('1', 'Yes'),
+            ('0', 'No'),
```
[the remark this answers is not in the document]
amatissart: Thy need to be strings: the tuples are used to specify values used by the filter in the Django Admin. The first element is the coded value that will appear in the URL query, and the second item is the option that appear in the menu on the right.   
I will add a comment about this, that's indeed a bit surprising without more context.

backend/core/tests/test_trusted_users.py
amatissart: `.update()` can be used on a `QuerySet` but not on a model instance, it seems.

backend/core/signals.py
amatissart: I have no strong opinion about this.  

I agree that too many signals may lead to surprising behaviour and hard-to-maintain code.  
In this case, overriding `User.save()` might also be a bit confusing, as the side-effect affects another model `EmailDomain`. But I see that this approach was already used for the (unused) `VerifiableEmail`, so I will make the change and remove the signal.

backend/core/models/user.py
```diff
@@ -181,21 +187,67 @@ def user_preferences(self):
...
-    def is_domain_rejected(self):
-        """Check if the user's email is certified. See #152"""
-        any_rejected = VerifiableEmail.objects.filter(
-            user=self, is_verified=True, domain_fk__status=EmailDomain.STATUS_REJECTED
-        ).count()
-        return True if any_rejected else False
+    def is_trusted(self):
+        return User.trusted_users().filter(pk=self.pk).exists()
+
+    def ensure_that_email_domain_exists(self):
```
lfaucon, line 231: The `_that_` in the method's name is rather unusual

lfaucon (approved):
LGTM

PR #267, opened 2021-11-03
Implement "Change email address" form

amatissart:
* A new section is added on the "Account" page to ask for a verification link and change the email address associated to the current account:
image

* The verification flow is similar to the one used after registration, and reuses the same component `VerifyUser`, renamed to `VerifySignature`.

* The API rejects any existing email address.
* The OpenAPI initialization is refactored: `OpenAPI.BASE` and `OpenAPI.TOKEN` are defined in the top level `App`. Other components don't need to interact with them anymore, and can use `useLoginState()` if necessary (to check if the current user is logged in, etc.)  
Another PR will follow to clean up existing components. It would add a lot of noise here.


frontend/src/pages/settings/account/Account.tsx
```diff
@@ -7,20 +7,55 @@ import ContentHeader from '../../../components/ContentHeader';
...
+  let sectionTitle;
+  if (typeof title === 'string') {
+    sectionTitle = (
+      <Typography variant="h4" color="secondary">
+        {title}
+      </Typography>
+    );
+  } else {
+    sectionTitle = title;
+  }
```
lfaucon, line 34: I prefer the following:

```suggestion
  const sectionTitle = (typeof title === 'string') 
    ?  (<Typography variant="h4" color="secondary">
        {title}
      </Typography>) 
    : title
```

The two advantages I find are 1. more concise and 2. using `const` instead of `let` reassures me that I don't need to read more code to know what `sectionTitle` is.

frontend/src/features/settings/account/EmailAddressForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+      setIsLoading(false);
+    };
+    loadProfile();
+  }, []);
+
+  if (isSuccess) {
+    return (
+      <Box marginBottom={2}>
+        <Typography>
+          A verification email will be sent to confirm your new email address.
```
lfaucon, line 53: "has been sent" ?

frontend/src/features/settings/account/EmailAddressForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </Typography>
+      </Box>
+    );
+  }
+
+  const formError = apiError?.status === 400 ? apiError.body : null;
+
+  return (
+    <>
+      {isLoading && <CircularProgress />}
```
lfaucon, line 63: I'd recommend not rendering anything else if `isLoading` otherwise a user clicking rapidly or with slow network may trigger the form twice.

frontend/src/features/settings/account/EmailAddressForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+              formError={formError}
+            />
+          </Grid>
+          <Grid item md={6}>
+            <Button
+              type="submit"
+              color="secondary"
+              fullWidth
+              variant="contained"
+              disabled={isLoading}
```
lfaucon, line 92: Ok, you may ignore the above comment. I'm not sure what's the best practice in general.

Maybe the best UX may be that the whole form should keep the same height when submitted.

frontend/src/pages/signup/Verify.tsx
```diff
@@ -0,0 +1,84 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const title =
+    verify === 'user' ? 'Verify registration' : 'Verify new email address';
+
+  const successMessage =
+    verify === 'user'
+      ? '✅ Your account is now verified.'
+      : '✅ Your new email address is now verified.';
```
lfaucon, line 68:
```suggestion
  const [executeVerify, title, success] = (verify === 'user')
    ? [executeVerifyUser, 'Verify registration', '✅ Your account is now verified.']
    :  [executeVerifyEmail, 'Verify new email address', '✅ Your new email address is now verified.']
  
  useEffect(() => {
    const verifyParams = async () => {
      try {
        await executeVerify(searchParams);
        setVerificationState('success');
      } catch (err) {
        console.error(err);
        setVerificationState('fail');
      }
    };
    verifyParams();

    // Let's make sure the verification runs only once,
    // even if searchParams are re-evaluated.
    //
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

frontend/src/pages/signup/Verify.tsx
```diff
@@ -0,0 +1,84 @@
...
+        } else if (verify === 'email') {
+          await executeVerifyEmail(searchParams);
+        }
+        setVerificationState('success');
+      } catch (err) {
+        console.error(err);
+        setVerificationState('fail');
+      }
+    };
+    verifyParams();
```
lfaucon, line 54: I never knew this trick after years of JavaScript 🤯

lfaucon (approved):
Only optinoal code style comments. Thanks for the feature. It works perfectly locally.

frontend/src/App.tsx
```diff
@@ -19,8 +19,25 @@ import VideoRecommendationPage from './pages/videos/VideoRecommendation';
...
 
+import { OpenAPI } from 'src/services/openapi';
+import { LoginState } from './features/login/LoginState.model';
+
+const API_URL = process.env.REACT_APP_API_URL;
+
+const initializeOpenAPI = (loginState: LoginState) => {
+  OpenAPI.BASE = API_URL ?? '';
+  OpenAPI.TOKEN = async () => loginState.access_token ?? '';
+  return OpenAPI;
```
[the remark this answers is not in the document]
amatissart: :+1:

frontend/src/App.tsx
```diff
@@ -19,8 +19,25 @@ import VideoRecommendationPage from './pages/videos/VideoRecommendation';
...
 import ResetPassword from './pages/login/ResetPassword';
 
+import { OpenAPI } from 'src/services/openapi';
+import { LoginState } from './features/login/LoginState.model';
+
+const API_URL = process.env.REACT_APP_API_URL;
+
+const initializeOpenAPI = (loginState: LoginState) => {
+  OpenAPI.BASE = API_URL ?? '';
+  OpenAPI.TOKEN = async () => loginState.access_token ?? '';
```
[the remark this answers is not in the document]
amatissart: I am not certain this is necessary, depending on React rerender cycle, but it feels safer to me to use a function to fetch the latest `access_token` on every request, even if the `useEffect` did not run after the token has been updated.   
And async is required as OpenAPI expects a `Promise`.

frontend/src/features/settings/account/EmailAddressForm.tsx
amatissart: Good point, I wrapped the content in a `<Box>` with "minHeight", to reduce flickering and keep the same height in most cases. This is not a perfect solution.  

In some cases, [ "skeleton components"](https://v4.mui.com/components/skeleton/#skeleton) and animations can also used to prevent abrupt changes in layout.

frontend/src/pages/signup/Verify.tsx
amatissart: If you refer to using an "async" function in a sync block, it's actually a common pattern in `useEffect()`.  
In general, that's not a silver bullet though: it works here as there is no return value, and the function catches all errors, to avoid unhandled promise rejections.

Issue #269, opened 2021-11-04
Add feedback notification on action buttons


amatissart:
* Add to rater later
* Delete video from rate later list
* Comparison submit
...

`showSuccessAlert` can be used:
https://github.com/tournesol-app/tournesol/blob/main/frontend/src/utils/notifications.ts#L22-L29

PR #271, opened 2021-11-05
Add visual feedback for actions buttons in rate later list and comparison page

sandre35:
Closes https://github.com/tournesol-app/tournesol/issues/269
Add multiple responses with pop to inform users that request are handled, and given them feedback on the result of their request.

- Recommendation page : alert print if video added/already exists in rate Later List.
- Rate Later List : pop-up when video deleted from rate later
- Comparison page : pop-up when comparison is submitted

### DIFF
--- a/frontend/src/utils/action.tsx
+++ b/frontend/src/utils/action.tsx
@@ -26,28 +26,29 @@ export const CompareNowAction = ({ videoId }: { videoId: string }) => {
 export const AddToRateLaterList = ({ videoId }: { videoId: string }) => {
   const video_id = videoId;
   const { enqueueSnackbar } = useSnackbar();
+  const handleCreation = async () => {
+    try {
+      const response = await UsersService.usersMeVideoRateLaterCreate({
+        video: { video_id } as Video,
+      });
+      console.log(response);
+      showSuccessAlert(
+        enqueueSnackbar,
+        'The video has been added to your rate later list.'
+      );
+    } catch (error) {
+      showInfoAlert(
+        enqueueSnackbar,
+        'The video is already in your rate later list.'
+      );
+    }
+  };
   return (
     <Tooltip title="Rate later" placement="left">
       <IconButton
         size="medium"
         color="secondary"
-        onClick={async () => {
-          try {
-            const response = await UsersService.usersMeVideoRateLaterCreate({
-              video: { video_id } as Video,
-            });
-            console.log(response);
-            showSuccessAlert(
-              enqueueSnackbar,
-              'The video has been added to your rate later list.'
-            );
-          } catch (error) {
-            showInfoAlert(
-              enqueueSnackbar,
-              'The video is already in your rate later list.'
-            );
-          }
-        }}
+        onClick={handleCreation}
         style={{ color: '#CDCABC' }}
       >
         <AddIcon />
--- a/frontend/src/utils/action.tsx
+++ b/frontend/src/utils/action.tsx
@@ -64,6 +64,7 @@ export const AddToRateLaterList = ({ videoId }: { videoId: string }) => {
 
 export const RemoveFromRateLater = (asyncCallback?: () => void) => {
   const RemoveFromRateLaterComponnent = ({ videoId }: { videoId: string }) => {
+    const { enqueueSnackbar } = useSnackbar();
     const video_id = videoId;
     return (
       <Tooltip title="Remove" placement="left">
@@ -72,6 +73,10 @@ export const RemoveFromRateLater = (asyncCallback?: () => void) => {
           onClick={async () => {
             await UsersService.usersMeVideoRateLaterDestroy(video_id);
             if (asyncCallback) await asyncCallback();
+            showSuccessAlert(
+              enqueueSnackbar,
+              'The video has been deleted from your rate later list.'
+            );
           }}
           style={{ color: '#CDCABC' }}
         >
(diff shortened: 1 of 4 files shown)


lfaucon (changes requested):
The TODO that you propose already exist as `usersMeVideoRateLaterRetrieve` if it returns content it means that the video is in the video_rate_later list.

frontend/src/utils/action.tsx
```diff
@@ -23,16 +25,35 @@ export const CompareNowAction = ({ videoId }: { videoId: string }) => {
...
+        onClick={async () => {
+          let flag = false;
+          const rateLaterList = await UsersService.usersMeVideoRateLaterList();
+          rateLaterList.results?.forEach((rateLaterVideo) => {
+            if (rateLaterVideo.video.video_id === video_id) {
+              flag = true;
+            }
+          });
+          if (!flag) {
+            await UsersService.usersMeVideoRateLaterCreate({
```
amatissart, line 43: `UsersService.usersMeVideoRateLaterCreate()` will raise an `ApiError` with status 409, if the video already exists in the rate later list. So calling `UsersService.usersMeVideoRateLaterList()` should not be necessary (and rigorously several pages may be necessary to fetch the whole rate later list, which would be impractical).  

And by catching the API error, a single request to API will be executed.

frontend/src/utils/action.tsx
```diff
@@ -51,6 +73,10 @@ export const RemoveFromRateLater = (asyncCallback?: () => void) => {
           onClick={async () => {
             await UsersService.usersMeVideoRateLaterDestroy(video_id);
             if (asyncCallback) await asyncCallback();
+            showSuccessAlert(
+              enqueueSnackbar,
+              'The video has been deleted of your rate later list.'
```
amatissart, line 78: Wording
```suggestion
              'The video has been deleted from your rate later list.'
```

frontend/src/utils/action.tsx
```diff
@@ -23,16 +25,35 @@ export const CompareNowAction = ({ videoId }: { videoId: string }) => {
...
+            });
+            showSuccessAlert(
+              enqueueSnackbar,
+              'The video has been added to your rate later list.'
+            );
+          } else {
+            showErrorAlert(
+              enqueueSnackbar,
+              'The video is already in your rate later list.'
+            );
```
amatissart, line 54: Is that really an error? The message in red might be unnecessarily frightening. Maybe an "info" alert would be more appropriate.

amatissart (commented):
The notifications mechanism look good to me.  

The case of the video that already exists in the rate later list can be simplified with a single request though.

sandre35:
Screenshot from 2021-11-12 15-09-09
This what I have currently when trying to add a video that already exists in my rate later list

frontend/src/utils/action.tsx
```diff
@@ -51,6 +67,10 @@ export const RemoveFromRateLater = (asyncCallback?: () => void) => {
           onClick={async () => {
             await UsersService.usersMeVideoRateLaterDestroy(video_id);
             if (asyncCallback) await asyncCallback();
+            showSuccessAlert(
+              enqueueSnackbar,
+              'The video has been deleted from your rate later list.'
+            );
```
lfaucon, line 73: Do we never expect the Destroy to fail ? If it can fail, then a similar logic as for the Create would be best.

frontend/src/utils/action.tsx
```diff
@@ -23,16 +25,29 @@ export const CompareNowAction = ({ videoId }: { videoId: string }) => {
...
+              enqueueSnackbar,
+              'The video has been added to your rate later list.'
+            );
+          } catch (error) {
+            showInfoAlert(
+              enqueueSnackbar,
+              'The video is already in your rate later list.'
+            );
+          }
+        }}
```
lfaucon, line 50: For code readability, I usually like to keep this sort of method, out of the  so that we separate "how something is rendered" and "what the interactions do". 

```JavaScript
export const AddToRateLaterList = ({ videoId }: { videoId: string }) => {
  const video_id = videoId;
  const { enqueueSnackbar } = useSnackbar();

  const handleCreate = async () => {
    [...]
  }

  return (
    <Tooltip title="Rate later" placement="left">
      <IconButton
        size="medium"
        color="secondary"
        onClick={handleCreate}
        style={{ color: '#CDCABC' }}
      >
              <AddIcon />
      </IconButton>
    </Tooltip>
```

lfaucon (approved):

frontend/src/utils/action.tsx
sandre35: The only solution I see is that the user has two parallel pages and he decides two press the button delete on both pages.
I implemented the logic got add button because the button appears even if it is already in the rate later list

frontend/src/utils/action.tsx
```diff
@@ -23,16 +25,30 @@ export const CompareNowAction = ({ videoId }: { videoId: string }) => {
 
 export const AddToRateLaterList = ({ videoId }: { videoId: string }) => {
   const video_id = videoId;
+  const { enqueueSnackbar } = useSnackbar();
+  const handleCreation = async () => {
+    try {
+      const response = await UsersService.usersMeVideoRateLaterCreate({
+        video: { video_id } as Video,
+      });
+      console.log(response);
```
amatissart, line 34: This `console.log()` can be removed.

amatissart (approved):
Looks good to me :+1:

PR #281, opened 2021-11-05
Refactor actions and add style to VideoCards

lfaucon:
This PR improve the style of the video card:
- Adds border and shadow frmo Figma's mock-up
- Changes color of Video card actions
- Changes layout of actions to be a row on small screens and a column on large screens
- Add tooltip for Video card actions

Additionally I moved the code of the actions we have, previously they were defined inside other components which makes them impossible to re-use.

Before:
Screenshot 2021-11-06 at 00 31 29

After:
Screenshot 2021-11-06 at 00 31 09

### DIFF
--- a/frontend/src/features/videos/VideoList.tsx
+++ b/frontend/src/features/videos/VideoList.tsx
@@ -1,15 +1,12 @@
 import React from 'react';
 
-import { Typography, makeStyles, Grid, IconButton } from '@material-ui/core';
-import { Add as AddIcon } from '@material-ui/icons';
+import { Typography, makeStyles, Grid } from '@material-ui/core';
 import type {
   PaginatedVideoSerializerWithCriteriaList,
   VideoSerializerWithCriteria,
-  Video,
 } from 'src/services/openapi';
 import VideoCard from '../videos/VideoCard';
-import { CompareNowAction } from 'src/utils/action';
-import { UsersService } from 'src/services/openapi';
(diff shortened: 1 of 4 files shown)


frontend/src/features/videos/VideoList.tsx
```diff
@@ -26,37 +23,17 @@ function VideoList({
...
     <div>
       {videos.results?.length ? (
         videos.results.map((video: VideoSerializerWithCriteria) => (
           <Grid container className={classes.card} key={video.video_id}>
             <Grid item xs={12}>
-              <VideoCard
-                video={video}
-                actions={[CompareNowAction, AddToRateLaterList]}
-              />
+              <VideoCard video={video} actions={actions} />
```
sandre35, line 36: I don't fully understand this change. If the user is not logged in, he can not add videos to his rate later list, neither compare it

sandre35 (changes requested):

frontend/src/features/videos/VideoList.tsx
lfaucon: I did not change from the previous behaviour. But you are right. I'll remove the link to comparison

sandre35 (approved):
Great job! It works very well !!

frontend/src/features/videos/VideoList.tsx
sandre35: Indeed it was my mistake at the beginning

Issue #273, opened 2021-11-05
Enable donwloading a user's own data


lfaucon:
I would say minimum metadata and list of all comparisons made by the user.

PR #283, opened 2021-11-07
Add endpoint for downloading all comparisons as a CSV

lfaucon:
Part of the backend for ticket: #273

### DIFF
--- a/backend/tournesol/tests/test_api_user.py
+++ b/backend/tournesol/tests/test_api_user.py
@@ -3,6 +3,7 @@ from rest_framework import status
 from rest_framework.test import APIClient
 
 from core.models import User
+from tournesol.models import Comparison, Video, ComparisonCriteriaScore
 
 
 class UserDeletionTestCase(TestCase):
@@ -99,3 +100,34 @@ class UserRegisterNewEmailTest(TestCase):
             text="email address already exists",
             status_code=400
         )
+
+class UserDataDump(TestCase):
+    def setUp(self) -> None:
+        self.user_with_comparisons = User.objects.create_user(username="user_with", email="user_with@example.com")
+        
+        self.video1 = Video.objects.create(video_id="test_all_data_1", name="test_all_data_1")
+        self.video2 = Video.objects.create(video_id="test_all_data_2", name="test_all_data_2")
+        self.comparison = Comparison.objects.create(user=self.user_with_comparisons, video_1=self.video1, video_2=self.video2)
+        ComparisonCriteriaScore.objects.create(comparison=self.comparison,score=5,criteria="largely_recommended")
+        self.user_without_comparisons = User.objects.create_user(username="user_without", email="user_without@example.com")
+        self.client = APIClient()
+
+    def test_not_authenticated_cannot_download(self):
+        resp = self.client.get("/users/me/all_data/")
+        self.assertEqual(resp.status_code, 401)
+
+    def test_authenticated_with_comparisons_can_download(self):
+        self.client.force_authenticate(self.user_with_comparisons)
+        resp = self.client.get("/users/me/all_data/")
+        self.assertEqual(resp.status_code, 200)
+        # Ensures the csv contains the information about the single comparison criteria score
+        self.assertIn(b'test_all_data_1', resp.content)
+        self.assertIn(b'largely_recommended', resp.content)
+
+    def test_authenticated_without_comparisons_can_download(self):
+        self.client.force_authenticate(self.user_without_comparisons)
+        resp = self.client.get("/users/me/all_data/")
+        self.assertEqual(resp.status_code, 200)
+        # Ensures the csv does not contain information about comparison criteria score
+        self.assertNotIn(b'test_all_data_1', resp.content)
+        self.assertNotIn(b'largely_recommended', resp.content)
--- a/backend/tournesol/tests/test_api_user.py
+++ b/backend/tournesol/tests/test_api_user.py
@@ -100,34 +100,3 @@ class UserRegisterNewEmailTest(TestCase):
             text="email address already exists",
             status_code=400
         )
-
-class UserDataDump(TestCase):
-    def setUp(self) -> None:
-        self.user_with_comparisons = User.objects.create_user(username="user_with", email="user_with@example.com")
-        
-        self.video1 = Video.objects.create(video_id="test_all_data_1", name="test_all_data_1")
-        self.video2 = Video.objects.create(video_id="test_all_data_2", name="test_all_data_2")
-        self.comparison = Comparison.objects.create(user=self.user_with_comparisons, video_1=self.video1, video_2=self.video2)
-        ComparisonCriteriaScore.objects.create(comparison=self.comparison,score=5,criteria="largely_recommended")
-        self.user_without_comparisons = User.objects.create_user(username="user_without", email="user_without@example.com")
-        self.client = APIClient()
-
-    def test_not_authenticated_cannot_download(self):
-        resp = self.client.get("/users/me/all_data/")
-        self.assertEqual(resp.status_code, 401)
-
-    def test_authenticated_with_comparisons_can_download(self):
-        self.client.force_authenticate(self.user_with_comparisons)
-        resp = self.client.get("/users/me/all_data/")
-        self.assertEqual(resp.status_code, 200)
-        # Ensures the csv contains the information about the single comparison criteria score
-        self.assertIn(b'test_all_data_1', resp.content)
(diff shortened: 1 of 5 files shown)


backend/tournesol/views/user.py
```diff
@@ -23,3 +30,29 @@ def delete(self, request):
...
+
+
+
+
+class UserDataDumpView(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename="export.csv"'
```
amatissart, line 42: Don't we want to wrap the file in a .zip archive, to be able to include multiple .csv files in the future?

backend/tournesol/views/user.py
```diff
@@ -23,3 +30,29 @@ def delete(self, request):
...
+        writer = csv.DictWriter(response, fieldnames=['video_a', 'video_b', 'criteria', 'weight', 'score'])
+        writer.writeheader()
+                
+        comparison_queryset = Comparison.objects.filter(user=request.user)
```
amatissart, line 46: Suggestion to reduce the number of potential SQL queries required to build the export:
```suggestion
        comparison_queryset = Comparison.objects\
            .filter(user=request.user)\
            .select_related("video_1", "video_2")\
            .prefetch_related("criteria_scores")

```

backend/tournesol/tests/test_api_user.py
```diff
@@ -99,3 +100,34 @@ def test_cannot_submit_existing_email(self):
...
+        self.assertIn(b'test_all_data_1', resp.content)
+        self.assertIn(b'largely_recommended', resp.content)
+
+    def test_authenticated_without_comparisons_can_download(self):
+        self.client.force_authenticate(self.user_without_comparisons)
+        resp = self.client.get("/users/me/all_data/")
+        self.assertEqual(resp.status_code, 200)
+        # Ensures the csv does not contain information about comparison criteria score
+        self.assertNotIn(b'test_all_data_1', resp.content)
+        self.assertNotIn(b'largely_recommended', resp.content)
```
sandre35, line 133: I would recommend to use a library that verifies directly inside the csv that information are correctly inside it.
Information are for sure inside the reponse, but we are not sure about the readability of the csv

backend/tournesol/views/user.py
```diff
@@ -23,3 +30,29 @@ def delete(self, request):
...
+
+
+class UserDataDumpView(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename="export.csv"'
+        writer = csv.DictWriter(response, fieldnames=['video_a', 'video_b', 'criteria', 'weight', 'score'])
+        writer.writeheader()
```
sandre35, line 44: We need also to test if headers are correctly written, and readable for a human

backend/tournesol/views/user.py
```diff
@@ -23,3 +30,29 @@ def delete(self, request):
         logout(request)
         logger.info("User '%s' with email '%s' has been deleted.", user.username, user.email)
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+
```
sandre35, line 35: There are 3 too many empty lines

sandre35 (changes requested):

backend/tournesol/views/user.py
lfaucon: I thought it my be nice to also have the possibility to download only the comparisons as a csv. 

Not I added a second route:
- /exports/comparisons : downloads a csv with comparisons
- /exports/all : downloads a zip archive (currently only comparisons in it)

backend/tournesol/views/user.py
lfaucon: Thanks for the optimization! Made the change as you recommend

backend/tournesol/views/user.py
lfaucon: Fixed

backend/tournesol/tests/test_api_user.py
lfaucon: Fixed

backend/tournesol/views/user.py
lfaucon: As we now use the csv.DictReader in the test, the test includes parsing the column names, then makes sure that "scores" is one of the columns

backend/tournesol/tests/test_exports.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_authenticated_with_comparisons_can_download_all(self):
+        self.client.force_authenticate(self.user_with_comparisons)
+        resp = self.client.get("/users/me/exports/all/")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+
+    def test_authenticated_without_comparisons_can_download_all(self):
+        self.client.force_authenticate(self.user_without_comparisons)
+        resp = self.client.get("/users/me/exports/all/")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
```
sandre35, line 57: There is one line missing at the end of the file

backend/tournesol/tests/test_exports.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_authenticated_with_comparisons_can_download_comparisons(self):
+        self.client.force_authenticate(self.user_with_comparisons)
+        resp = self.client.get("/users/me/exports/comparisons/")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        # Ensures that the csv contians a single row with the correct score
+        csv_file = csv.DictReader(io.StringIO(resp.content.decode()))
+        comparison_list = [row for row in csv_file]
+        self.assertEqual(len(comparison_list), 1)
+        self.assertEqual(float(comparison_list[0]['score']), 5)
```
sandre35, line 34: It is very much about details but `5` is a int not a float. So maybe we can replace it by 5.0 ?

backend/tournesol/tests/test_exports.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_authenticated_without_comparisons_can_download_comparisons(self):
+        self.client.force_authenticate(self.user_without_comparisons)
+        resp = self.client.get("/users/me/exports/comparisons/")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        # Ensures the csv does not contain information about comparison criteria score
+        csv_file = csv.DictReader(io.StringIO(resp.content.decode()))
+        comparison_list = [row for row in csv_file]
+        self.assertEqual(len(comparison_list), 0)
```
sandre35, line 43: From what I can read. This file is completely empty. So, the file doesn't have headers.
It would be more readable for the user if he knows what columns are referring to ?

sandre35 (changes requested):

backend/tournesol/tests/test_exports.py
lfaucon: The file looks like this for a user who has submitted 0 comparisons

```
❯ cat ~/Downloads/export.csv
video_a,video_b,criteria,weight,score
```

backend/tournesol/tests/test_exports.py
lfaucon: Fixed. Do you want to look into adding this check in the automated syntax checks ? This way you will not have to check for it yourself. 

This is also the sort of thing that may be added in a git commit hook

backend/tournesol/views/exports.py
```diff
@@ -0,0 +1,67 @@
...
+    def get(self, request):
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename="export.csv"'
+        write_comparisons_file(request, response)
+        return response
+
+
+class ExportAllView(APIView):
+
+    def get(self, request):
```
amatissart, line 49: DRF Spectacular is logging errors because it can't find any serializer for these views: the response format can be defined as `OpenApiTypes.BINARY` from `drf_spectacular.types`:
```suggestion
    @extend_schema(
        description="Download current user data in .zip file",
        responses={200: OpenApiTypes.BINARY}
    )
    def get(self, request):
```

amatissart (approved):
It looks good to me.

Note that [`openapi-typescript-codegen`](https://github.com/ferdikoomen/openapi-typescript-codegen) used by "frontend" to generate the "services" code based on the OpenAPI schema, does not implement specific behaviour when `Content-Disposition` header is present in the response: only the body is accessible as a string.  
So we may need to implement the download ourselves when integrating these exports links in the UI.

PR #284, opened 2021-11-07
Optimize video list query by computing the total score in the database

amatissart:
It has been noticed that requests to `/video/` could be relatively slow in production (~ 1.5 seconds). It's mostly due to how all `VideoCriteriaScore` items need to fetched in this loop:
https://github.com/tournesol-app/tournesol/blob/7c20129409a8bd41acab62badee2afd713468717/backend/tournesol/views/video.py#L78-L80

There are currently more than 10 000 `VideoCriteriaScore`  items. Even if `prefetch_related` allows them to be fetched in a single SQL query, instantiating these objects in Python has a significant overhead. And it would get worse as the number of videos on the platform increases.

This branch suggests to compute the total score in Postgres, and only fetch the details about the criteria scores that are actually needed in the response.
On my local environment the response times on `GET /video/` drops from 1sec to less than 100ms.


lfaucon (approved):
LGTM

PR #286, opened 2021-11-08
Video list: avoid serializer initialization overhead with 'many=True'

amatissart:
I was not completely convinced with the performance gain visible on staging after deploying #284 

There is actually another easy optimization: when serializing multiple objects, `many=True` should be preferred to avoid the overhead of the serializer initialization:
See https://www.django-rest-framework.org/api-guide/serializers/#dealing-with-multiple-objects

On my local environment this brings another 3x speedup.


sandre35 (approved):
Great notice

PR #291, opened 2021-11-10
Show indication about email status on account page + Add page about trusted domains

### LINKED ISSUE #191: Email domain verification
In order to ensure that Tournesol cannot be manipulated, we cannot allow fake account. Our initial solution for this consist in maintaining a list of verified email domains.

TODO:

- [x] We can manually tag email domains as verified or not using the admin interface. This already works but should be improved by searching for pending domains and making editing easier.
#262 

- [x]  ML: tournesol scores use only verified users comparisons should be included. 
#263 

- [x] Backend, new accounts or email changes should trigger the creation of a new email domain if necessary
Included in #263

- [x] UX/frontend: Users can change their emails to use one that is from a verified domain 
#267 

- [ ] UX/frontend: Show an indication about the state of their email domain. 

- [ ] Add an explanation in a Tooltip on the frontend: 
> In order to protect Tournesol from fake accounts, by default, Tournesol assigns a limited voting right to newly created accounts. 
You can gain significantly more voting rights by validating an email address from a trusted domain. 
We are currently working on designing additional means for contributors to gain voting rights.
In any case, your contributions are valuable to us, as they will motivate further research in safe and ethical algorithms.

- [ ]  List of Accepted email domains. We show only accepted ones because rejected or pendeing domain reveal private information.

- [ ] Ensure that emails are unique
 Addressed by #232

amatissart:
Closes #191 

* Indication about the email status on account page: "Trusted" or "Non-trusted":
  A boolean `is_trusted` has been added in the response of `GET /accounts/profile/`

* A page "/about/trusted_domains" explaining the motivation for this mechanism:
  A new route `GET /domains/` has been implemented to fetch the list of accepted email domains

|/settings/account| /about/trusted_domains|
|-|-|
|Capture d’écran du 2021-11-10 15-05-42|image|

### DIFF
--- a/frontend/src/features/settings/account/EmailAddressForm.tsx
+++ b/frontend/src/features/settings/account/EmailAddressForm.tsx
@@ -1,19 +1,68 @@
 import React, { useState, useEffect } from 'react';
 
-import Button from '@material-ui/core/Button';
-import Grid from '@material-ui/core/Grid';
-import { CircularProgress, Box, Typography } from '@material-ui/core';
+import {
+  CircularProgress,
+  Box,
+  Typography,
+  Theme,
+  Grid,
+  Button,
+} from '@material-ui/core';
+import { Link as RouterLink } from 'react-router-dom';
 
 import { useSnackbar } from 'notistack';
 
 import { showErrorAlert } from '../../../utils/notifications';
 import { FormTextField } from 'src/components';
 
-import { AccountsService, ApiError } from 'src/services/openapi';
+import { AccountsService, ApiError, UserProfile } from 'src/services/openapi';
+import { Lens as LensIcon, HelpOutline as HelpIcon } from '@material-ui/icons';
+import { useTheme } from '@material-ui/styles';
+
+const TrustStatus = ({ isTrusted }: { isTrusted: boolean }) => {
+  const theme = useTheme<Theme>();
+
+  return (
+    <Box display="flex" flexDirection="row" flexWrap="wrap" alignItems="center">
+      <Typography>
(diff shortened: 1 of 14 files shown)


frontend/src/features/settings/account/EmailAddressForm.tsx
```diff
@@ -1,19 +1,68 @@
...
+          size="small"
+          style={{ fontSize: 13, color: '#777' }}
+          startIcon={<HelpIcon />}
+        >
+          Learn more about trusted domains
+        </Button>
+      </Box>
+    </Box>
+  );
+};
```
lfaucon, line 61: I slightly dislike the excessive amount of Css Boxes into Css Boxes. Something like this seems to do the same job:

```
      <Typography style={{display: 'flex', alignItems: "center"}}>
        Email status:
        <LensIcon style={{ fontSize: 16, color: statusColor, marginLeft: 8 }} />
        <Box color={statusColor} fontWeight="bold">{isTrusted ? 'Trusted' : 'Non-trusted'}</Box>
      </Typography>
```

From my point of view I love having less code to scroll through, but maybe there are advantages that I fail to see?

Also, I find the layout nicer with a `justifyContent="space-between"` on the top Box instead of the `marginLeft="auto"` for medium screens

lfaucon (approved):
Awesome work! 🟢 Trusted

frontend/src/features/settings/account/EmailAddressForm.tsx
amatissart: Thanks for the simplification, after a few iterations my solution was a bit convoluted indeed.
I added your suggestions :+1:

PR #294, opened 2021-11-11
admin: Define list_display and search_fields in VideoAdmin

amatissart:
As suggested by @aidanjungo, it should now be easier to look for a specific video, e.g in order to correct its "language".

image

### DIFF
--- a/backend/tournesol/admin.py
+++ b/backend/tournesol/admin.py
@@ -3,6 +3,7 @@ Defines Tournesol's backend admin interface
 """
 
 from django.contrib import admin
+from django.utils.html import format_html
 
 from .models import (
     Video,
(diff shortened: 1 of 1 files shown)


backend/tournesol/admin.py
```diff
@@ -16,7 +17,23 @@
...
+        'publication_date',
+        'rating_n_ratings',
+        'rating_n_contributors',
+        'language',
+        'link_to_youtube',
+    )
+    search_fields = ('video_id', 'name', 'uploader')
+    list_filter = (
+        ('language', admin.AllValuesFieldListFilter),
+    )
```
lfaucon, line 33: Why not having language as `list_editable`, that would serve our usecase even better?

lfaucon (approved):
LGTM

backend/tournesol/admin.py
amatissart: Adding `list_editable = ('language',)` would add a `<select>` input on each line (with currently more than 8000 language choices). And it makes the page load extremely slow.

For now I will keep it as it is. If there is a need to update multiple video languages at once, we could probably find a more involved solution to adapt the layout of the admin page.

PR #297, opened 2021-11-13
Multiple improvements in Swagger docs

lfaucon:
* Use default pagination in VideoViewSet

* Remove Update and Destroy from VideoViewSet

* Fix warning about duplicated operation_id

* Fix warnings about missing .queryset

* Renamed ComparisonListOnlyApi -> ComparisonListFilteredApi


amatissart (approved):
Looks good to me.

The OpenAPI could be completed further to define all query parameters accepted by `GET /video/` (?language=, ?search=, etc.). But this is probably not a very high priority.

lfaucon:
Yes! I may do this as next steps. Also the /video/{id} needs fixing to be usable from the automatically generated code

PR #298, opened 2021-11-14
(frontend) Enable to export all data in Settings>Account

lfaucon:


frontend/src/components/Section.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
```
lfaucon, line 1: Moved this component here from Settings>Account so that it may be used in Settings>Profile too

frontend/src/features/settings/account/EmailAddressForm.tsx
```diff
@@ -134,33 +134,30 @@ export const EmailAddressForm = () => {
...
-      </>
+        </Grid>
+      </Grid>
     );
   };
 
   return <Box minHeight="180px">{getContent()}</Box>;
 };
+
+export default EmailAddressForm;
```
lfaucon, line 163: Re-structured a little bit this component so that it is more similar to other forms on the same page. Previously it was the only component that may take more than `md={6}` on large screens

frontend/src/features/settings/account/ExportAllDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+      ? await OpenAPI.TOKEN(options)
+      : OpenAPI.TOKEN;
+  const headers = { Authorization: `Bearer ${access_token}` };
+  const response = await fetch(`${api_url}${options.path}`, {
+    credentials: 'include',
+    method: options.method,
+    headers: headers,
+  });
+  return response.blob();
+}
```
lfaucon, line 35: Discussed this with @amatissart and decided to implement our own access to the API endpoint because the autogenerated code does not support `response.blob()`

frontend/src/features/settings/account/ExportAllDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const zipFileBlob = await usersMeExportsAllRetrieveBlob();
+    setDownloadUrl(window.URL.createObjectURL(zipFileBlob));
+    setLoading(false);
+  };
+
+  const prepareExport = () => {
+    setLoading(true);
+    setDownloadUrl('');
+    try {
+      downloadExport();
```
amatissart, line 52: As `downloadExport()` is async, we need to await the promise to be able to catch potential errors.  
Otherwise any error in the async call would cause an unhandled rejection, and would never be caught by the "catch" block.

```suggestion
  const prepareExport = async () => {
    setLoading(true);
    setDownloadUrl('');
    try {
      await downloadExport();
```

frontend/src/features/settings/account/ExportAllDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+        ) : (
+          <Button
+            type="button"
+            fullWidth
+            color="secondary"
+            variant="contained"
+            onClick={prepareExport}
+          >
+            Download data
+          </Button>
```
amatissart, line 92: The wording of these 2 buttons might be a bit confusing for the user, especially as the API request is typically very quick. 
What about "Prepare export" and then "Download file"?

frontend/src/features/settings/account/ExportAllDataForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const prepareExport = () => {
+    setLoading(true);
+    setDownloadUrl('');
+    try {
+      downloadExport();
+    } catch (error) {
+      showErrorAlert(
+        enqueueSnackbar,
+        'An error seems to have occured. If this error persist, please contact us.'
```
amatissart, line 56: typo
```suggestion
        'An error seems to have occurred. If this error persists, please contact us.'
```

frontend/src/features/settings/account/EmailAddressForm.tsx
amatissart: Nothing critical, but it was actually intentional to allow the information message to fill the remaining width.

With `md={6}` on the whole section, the message  block may be unnecessarily narrow on some screens:
image

lfaucon: We can see from your screenshot that it is not the only problem that we have for the layout on md screens.

I propose to go ahead with the current change proposed in this PR, then clean-up the whole layout. I created a ticket for it #306

amatissart (approved):

PR #299, opened 2021-11-15
Add scripts and documentation to restore Postgres database from a backup

### LINKED ISSUE #201: Enable restoring back-up and share knowledge
Ensure we are well able to restore back-ups. Probably Jst, Oizo, Amatissart and Lfaucon should know how to restore a backup

amatissart:
Closes #201

2 distinct mechanisms are implemented:

1. Restore **an entire backup from the same host:**
To be used after an incident, or in order to restore the database to a previous state (e.g after a temporary experiment on a non-production environment).

2. Fetch and import a backup **from a different host** (e.g for development purposes). 
In this case, some data on the target database is preserved to keep OAuth applications functional.

See [the updated README](https://github.com/tournesol-app/tournesol/blob/restore-pg-backup/infra/README.md#restore-postgres-backup) for more details.

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -40,7 +40,6 @@
 
 ## TODO
 
-- Backup automation and monitoring (database mainly?)
 - Application artifacts retrieval with proper triggers on updates (CI integration, CD?) (for now ansible clones, builds and deploys during each run)
 - CI/CD design
 - IDS/IPS? WAF?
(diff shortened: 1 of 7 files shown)


lfaucon (commented):
Right after restoring the db, I want Django migrations to run, isn't it? I don't see what part of the recipe would do this.

lfaucon (commented):
Our db is still super small so these operations take seconds, but I expect it will get slower over time. Do we have any risk of inconsistency / weird behaviour if users are using the instance during the restore operation? Do we want to do something like "switch to maintenance state (=no API & frontend shows a temporary maintenance screen)" then "restore db" then "switch back to normal state". Not blocking this work but I am interested in your thoughts on this.

amatissart:
> Right after restoring the db, I want Django migrations to run, isn't it? I don't see what part of the recipe would do this.

Good point. There could be situations where we'd like to avoid re-running buggy migrations, but then a rollback could probably be deployed beforehand. So yes, running migrations seems like what what we would want here. I made the necessary change.

> Our db is still super small so these operations take seconds, but I expect it will get slower over time. Do we have any risk of inconsistency / weird behaviour if users are using the instance during the restore operation? Do we want to do something like "switch to maintenance state (=no API & frontend shows a temporary maintenance screen)" then "restore db" then "switch back to normal state". Not blocking this work but I am interested in your thoughts on this.

The import process of data from an external backup uses `pg_restore --single-transaction`, but it can't be used when the database needs to be created from scratch.  
So Gunicorn is now stopped before the restore process. It is indeed safer, and easy enough to implement.

lfaucon (approved):
LGTM. But may be best to also get a review by @jstainer-kleis

amatissart:
@GresilleSiffle Have you had a look at this PR? Do you think that the instructions about import and restore are clear enough?

GresilleSiffle (approved):

PR #300, opened 2021-11-15
fix(frontend): Improvements for the style of the comparison List

lfaucon:
* Add VideoThumb component

* Refactor VideoCardFromId into useVideoMetadata hook

* Add caching in getVideoInformation

* Refactor getVideoInformation to use await

* Style of VS link

Screenshot 2021-11-19 at 21 11 08


frontend/src/features/frame/components/sidebar/SideBar.tsx
```diff
@@ -121,6 +129,7 @@ const SideBar = () => {
     >
       <List onClick={isSmallScreen ? () => dispatch(closeDrawer()) : undefined}>
         {menuItems.map(({ targetUrl, IconComponent, displayText }) => {
+          if (!IconComponent || !targetUrl) return <Divider />;
```
amatissart, line 132: This causes a minor React warning (relevant only for optimization purposes): 
```console
Warning: Each child in a list should have a unique "key" prop.
```

`key={index}` could be used in the Divider.

amatissart (approved):
Looks good to me.  
Just one small (non-blocking) warning in the comments.

Issue #225, opened 2021-10-17
Deploy extension on Firefox


lfaucon:
Needs to set-up CORS on prod and staging and testing the the Firefox extension works out of the box

PR #304, opened 2021-11-17
Browser extension: compatibility with Manifest v2 and Firefox

amatissart:
Related to #225 

* Back to Manifest v2, as Firefox does not support v3 yet
  https://blog.mozilla.org/addons/2021/05/27/manifest-v3-update/
  https://bugzilla.mozilla.org/show_bug.cgi?id=1578284

* More robust access token handling:
  * The token is fetched from extension storage on each request to the API
  * The token is updated in the extension storage during navigation on tournesol.app, without explicit reload. 

* Error handling and feedback in UI: buttons are disabled after the first error, error messages are visible on most actions.

* A few cosmetic changes in the popup:
image

### DIFF
--- a/browser-extension/README.md
+++ b/browser-extension/README.md
@@ -1,4 +1,4 @@
-# tournesol-chrome-extension
+# Tournesol browser extension
 
 The extension allows to quickly rate a video from youtube on the Tournesol expert rating page.
 
--- a/browser-extension/addRateLaterButton.css
+++ b/browser-extension/addRateLaterButton.css
@@ -12,6 +12,10 @@ button#tournesol-rate-button {
   font-weight: 500;
 }
 
+button#tournesol-rate-button:disabled {
+  opacity: 0.5;
+}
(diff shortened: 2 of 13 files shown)


lfaucon:
@aidanjungo want to test it?

aidanjungo:
For me, it seems good on Firefox, maybe we should just add the "build" script also for Firefox and deploy a version, to be sure everything works fine.

amatissart:
> For me, it seems good on Firefox, maybe we should just add the "build" script also for Firefox and deploy a version, to be sure everything works fine.

The same .zip file should be accepted by both Chrome Web Store and Firefox Add-ons.

lfaucon (commented):
I'll spend more time testing it this weekend. Thanks for this work. I didn't see anything wrong reading through the code except for the addRateLater call mentioned on discord.

lfaucon (approved):
🚢

PR #305, opened 2021-11-18
Upgrade drf-spectacular, and fix default auth parameters in Swagger

amatissart:
Due to how a custom template was used to serve the Swagger UI, the OAuth configuration for DRF Specatular was ignored.  
It does not seem there was any reason for this template, so let's keep it simple and use the default one.

So now, we just have to fill in a username and a password in the Swagger login form to use all authenticated endpoints. OAuth parameters (client_id, client_secret, etc.) will be provided automatically.


lfaucon (approved):
Works for me. 

What are the risks about exposing these Client ID and Client Secret publicly?

amatissart:
> What are the risks about exposing these Client ID and Client Secret publicly?

It would be exactly the same behaviour as the frontend, since the Swagger and the Frontend are defined as 2 OAuth applications with similar configuration. So I don't think that exposing these credentials is an issue.  
@jstainer-kleis Do you confirm?

It could actually make sense to define these applications with ["client_type"](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1)  `public` (instead of `confidential`), as using a client secret does not provide any additional security in our use case.

lfaucon:
> > What are the risks about exposing these Client ID and Client Secret publicly?
> 
> It would be exactly the same behaviour as the frontend, since the Swagger and the Frontend are defined as 2 OAuth applications with similar configuration. So I don't think that exposing these credentials is an issue. @jstainer-kleis Do you confirm?
> 
> It could actually make sense to define these applications with ["client_type"](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1) `public` (instead of `confidential`), as using a client secret does not provide any additional security in our use case.

That's my understanding too

jstainer-kleis:
> > What are the risks about exposing these Client ID and Client Secret publicly?
> 
> It would be exactly the same behaviour as the frontend, since the Swagger and the Frontend are defined as 2 OAuth applications with similar configuration. So I don't think that exposing these credentials is an issue. @jstainer-kleis Do you confirm?
> 
> It could actually make sense to define these applications with ["client_type"](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1) `public` (instead of `confidential`), as using a client secret does not provide any additional security in our use case.

That's indeed the case, client IDs and secrets aren't easily kept secret, they're exposed to any client. What's ensuring security is the restriction on redirect URIs.

jstainer-kleis:
> Due to how a custom template was used to serve the Swagger UI, the OAuth configuration for DRF Specatular was ignored. It does not seem there was any reason for this template, so let's keep it simple and use the default one.
> 
> So now, we just have to fill in a username and a password in the Swagger login form to use all authenticated endpoints. OAuth parameters (client_id, client_secret, etc.) will be provided automatically.

The issue I worked around on swagger auth at the time was that authenticated endpoints were not identified in the spec. I forced auth with manually specified parameters in order to be able to test the endpoints but that's not the proper way to do it.

amatissart:
> The issue I worked around on swagger auth at the time was that authenticated endpoints were not identified in the spec. I forced auth with manually specified parameters in order to be able to test the endpoints but that's not the proper way to do it.

It seems like yet another issue, unrelated to the changes introduced by this PR.

Do you refer to this setting in `SPECTACULAR_SETTINGS`?
https://github.com/tournesol-app/tournesol/blob/3d38e32c53fc2a7c9a18822c082725b8439a980f/backend/settings/settings.py#L282-L284
As far as i understand, with the current implementation of the API, this setting is required for the "security" section to appear in the schema for each endpoint:
```
security:
- oauth2:
  - read write groups
```

This is explained by how rest-spectacular uses the DRF `permission_classes` to define which oauth scopes are required on each endpoint. We are currently using `"rest_framework.permissions.IsAuthenticated"` as the only permission class. So OAuth scopes are not used, and as a consequence the "oauth2" section is absent in the schema.
Perhaps [`TokenHasReadWriteScope`](https://django-oauth-toolkit.readthedocs.io/en/latest/rest-framework/permissions.html#tokenhasreadwritescope) could be a good fit.

Anyway, I would suggest to address this in a separate issue, especially if granular permissions are desired (e.g based on read/write scopes).

PR #310, opened 2021-11-19
fix(extension): Remove `cookies` from extension's permissions

lfaucon:


amatissart (approved):
I confirm that it does work without this permission on both Firefox and Chromium.

PR #313, opened 2021-11-21
feat: Extensions and Contribute sections on home page

lfaucon:
Changes to the HomePage based on PolyConseil's feedback.

Next steps may be to include the comparison interface and correctly handle signing up the user when a comparison is submitted. Currently clicking "compare videos now" leads to the login/signup page

Screenshot 2021-11-21 at 11 42 04

### DIFF
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -31,7 +31,7 @@ const renderComponent = (drawerOpen: boolean) =>
     </Provider>
   );
 
-test('renders Tournesol banner', () => {
-  const { getByText } = renderComponent(true);
-  expect(getByText(/Tournesol/i)).toBeInTheDocument();
+test('Home page renders and contains the word "Tournesol"', () => {
+  const { getAllByText } = renderComponent(true);
+  expect(getAllByText(/Tournesol/i).length).toBeGreaterThan(0);
 });
--- a/frontend/src/pages/home/ContributeSection.tsx
+++ b/frontend/src/pages/home/ContributeSection.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import { Link } from 'react-router-dom';
 import { Typography, Button, Box } from '@material-ui/core';
 
 // ContributeSection is a paragraph displayed on the HomePage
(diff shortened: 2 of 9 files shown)


frontend/src/App.test.tsx
```diff
@@ -31,7 +31,7 @@ const renderComponent = (drawerOpen: boolean) =>
     </Provider>
   );
 
-test('renders Tournesol banner', () => {
-  const { getByText } = renderComponent(true);
-  expect(getByText(/Tournesol/i)).toBeInTheDocument();
+test('Home page renders and contains the word "Tournesol"', () => {
+  const { getAllByText } = renderComponent(true);
+  expect(getAllByText(/Tournesol/i).length).toBeGreaterThan(0);
```
lfaucon, line 36: Fixed this test, because it was failing because the word "Tournesol" now appears more than once on the Home page.

frontend/src/theme.ts
```diff
@@ -29,6 +29,19 @@ export const theme = responsiveFontSizes(
...
+        'BlinkMacSystemFont',
+        '"Segoe UI"',
+        'Roboto',
+        '"Helvetica Neue"',
+        'Arial',
+        'sans-serif',
+        '"Apple Color Emoji"',
+        '"Segoe UI Emoji"',
+        '"Segoe UI Symbol"',
+      ].join(','),
```
amatissart, line 44: Nice for consistency.  
Note that using a custom font as the default font is sometimes discouraged, as it may be rendered slightly differently by different systems and be uncomfortable for the eyes, especially for longer text content. We'll see how the change is appreciated on different devices.

lfaucon: How is it different if the font Poppins is the default rather than setting it manually on every components?

frontend/src/pages/home/ContributeSection.tsx
```diff
@@ -0,0 +1,37 @@
...
+        watched. First copy paste two links to videos, then tell us which video
+        you think should be largely recommended. If you want to you can also
+        compare the videos in more details based on the multiple comparison
+        criterias.
+      </Typography>
+      <Button
+        color="primary"
+        variant="contained"
+        component="a"
+        href="/comparisons"
```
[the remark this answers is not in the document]
lfaucon: No, it's a typo

frontend/src/theme.ts
amatissart: It's not different, except after this change the font is now applied in every form, Typography blocks, etc. (where a system font was used until now).

amatissart (approved):

Issue #274, opened 2021-11-05
(feat) Public data


lfaucon:
# Context

This is a rather large feature. 

At any time this list of tasks can be split into multiple tickets. It is also recommended to split the work in as many independent units as possible

In order to make this feature simple both for users and for implementation we want only to enable that a user can specify "Everything about video V is private". This means a Comparison will be public if it was marked as public for both videos.

# Subtasks:
- [x] Update models to specify an attribute `is_public` per user & video. This could either re-use the ContributorRating model or use a new model
- [x] Open and API endpoint or modify an existing one enable fetching and updating the `is_public` information
#315 
#320

- [x] Add interface on the frontend that allows users to control the is_public for their list of videos. It would be useful if users can find their list of private & public video and if user can choose to make all their video public or all their video private with a single interaction.
Partially addressed in #322 

- [x] Deprecate unused code related to making comparison and aggregated ratings public

- [x] An api endpoint and a link somewhere on the frontend to download the database of all public comparisons, contributor ratings and videos (with criteria scores)

PR #315, opened 2021-11-23
Extend API to handle `is_public` update in ContributorRating

amatissart:
Related to #274 

The boolean `is_public` was already defined in `ContributorRating`. This flag represents whether the contributions of a user related to a specific video should be public (included to the public dataset, associated with the username), or private (only used to compute aggregated scores, without publishing the details about the ratings, comparisons, etc.).

## Changes

* `is_public` can be updated via  `PATCH /users/me/contributor_ratings/<video_id>/`
* `ContributorRating` objects, related to the pair (user, video) will be created when a new comparison is submitted, if the relevant objects don't exist yet. For now, the default value for `is_public` is always "false".
* A custom migration has been implemented to create missing `ContributorRating` objects, related to all `Comparisons` that already exist in the database.
* The list of contributor ratings (`GET /users/me/contributor_ratings/`) can be filtered with `?is_public=true` or `is_public=false`.

## Questions

* Should an additional endpoint be implemented to allow a user to update `is_public` values for all their ratings at once?
* Should we also persist in the database a user preference, whose value would be used as a default value for `is_public` when a new comparison is created?

### DIFF
--- a/backend/tournesol/views/comparison.py
+++ b/backend/tournesol/views/comparison.py
@@ -4,6 +4,7 @@ API endpoints to interact with the contributor's comparisons
 
 from django.db.models import ObjectDoesNotExist, Q
 from django.http import Http404
+from django.db import transaction
 from drf_spectacular.utils import extend_schema
 
 from rest_framework import generics, mixins, status, exceptions
@@ -89,6 +90,7 @@ class ComparisonListApi(
         """List all comparisons made by the logged user."""
         return self.list(request, *args, **kwargs)
 
+    @transaction.atomic
     def perform_create(self, serializer):
         if self.comparison_already_exists(self.request):
             raise exceptions.ValidationError(
--- a/backend/tournesol/views/comparison.py
+++ b/backend/tournesol/views/comparison.py
@@ -6,10 +6,10 @@ from django.db.models import ObjectDoesNotExist, Q
 from django.http import Http404
 from drf_spectacular.utils import extend_schema
 
-from rest_framework import generics, mixins, status
+from rest_framework import generics, mixins, status, exceptions
 from rest_framework.response import Response
 
-from ..models import Comparison, Video
+from ..models import Comparison, ContributorRating
 from ..serializers import ComparisonSerializer, ComparisonUpdateSerializer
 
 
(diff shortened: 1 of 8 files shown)


lfaucon:
> Should an additional endpoint be implemented to allow a user to update is_public values for all their ratings at once?

Yes, the option "make my information about all videos public" would be very useful. And it would be fair to users to enable making all information private too without clicking every video one by one.

> Should we also persist in the database a user preference, whose value would be used as a default value for is_public when a new comparison is created?

Mmmm, yes that would be nice, although it is probably not a priority.

backend/tournesol/views/ratings.py
```diff
@@ -33,4 +42,15 @@ class ContributorRatingList(generics.ListAPIView):
...
+        ratings = ContributorRating.objects.filter(user=self.request.user)
+        is_public = self.request.query_params.get('is_public')
+        if is_public:
+            if is_public == "true":
+                ratings = ratings.filter(is_public=True)
+            elif is_public == "false":
+                ratings = ratings.filter(is_public=False)
+            else:
+                raise exceptions.ValidationError(
+                    "'is_public' query param should be 'true' or 'false'"
```
GresilleSiffle, line 54: We can enforce this valudation by using the word **must** instead of **should**.

Like:
> 'is_public' query param must be either 'true' or 'false'

amatissart: Done :+1:

backend/tournesol/views/comparison.py
```diff
@@ -89,19 +89,22 @@ def get(self, request, *args, **kwargs):
         """List all comparisons made by the logged user."""
         return self.list(request, *args, **kwargs)
 
+    def perform_create(self, serializer):
```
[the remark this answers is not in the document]
amatissart: Good idea. It's a very small price to pay, especially to ensure that a `ContributorRating` exists for each video compared by the user.

(The ratings statistics, updated by `.update_n_ratings()` are less critical. We could tolerate them being out of date.)

Your suggestion is committed.

GresilleSiffle (approved):
:+1:

PR #316, opened 2021-11-23
Add a a privacy policy page

aidanjungo:
It replaces the pdf file.


aidanjungo:
@lfaucon I don't know where is stored the pdf, maybe we could delete it to avoid confusions.

lfaucon:
Probably in a folder called 'public' or (second guess) 'static'. Find it ;)

lfaucon (approved):
LGTM

amatissart (approved):

Issue #282, opened 2021-11-05
Display duration of videos on VideoCard


lfaucon:
Update the backend endpoint if needed.

- [x] Retrieve information from youtube API
- [x] Process data to register time in duration field
- [x] Update swagger
- [x] Display information on Video Card

PR #318, opened 2021-11-25
Back: add duration of video

sandre35:
Linked issue : https://github.com/tournesol-app/tournesol/issues/282
This PR tackles steps 1 and 2

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -17,7 +17,7 @@ pytest==6.2.4
 PyYAML==5.4.1
 tqdm==4.62.1
 langdetect==1.0.8
-isodate
+isodate==0.6.0
 
 #API Youtube data
 google-api-python-client==2.15.0
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -17,6 +17,7 @@ pytest==6.2.4
 PyYAML==5.4.1
 tqdm==4.62.1
 langdetect==1.0.8
+isodate
 
 #API Youtube data
 google-api-python-client==2.15.0
--- a/backend/tournesol/serializers.py
+++ b/backend/tournesol/serializers.py
@@ -102,31 +102,11 @@ class VideoCriteriaScoreSerializer(ModelSerializer):
         fields = ["criteria", "score", "uncertainty", "quantile"]
 
 
-class VideoSerializerWithCriteria(ModelSerializer):
+class VideoSerializerWithCriteria(VideoSerializer):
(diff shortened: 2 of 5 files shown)


backend/tournesol/serializers.py
```diff
@@ -91,8 +101,15 @@ class Meta:
...
             "rating_n_ratings",
             "rating_n_contributors",
+            "duration",
         ]
 
+    # Convert duration to seconds to facilitate use of Humanize package
+    def get_duration(self, obj):
+        if obj.duration:
+            return int(obj.duration.total_seconds())
+        return 0
```
lfaucon, line 111: @sandre35 @GresilleSiffle @amatissart 

There is some code duplication between `VideoSerializer` and `VideoSerializerWithCriteria`


Could this be refactored into 

```
class VideoSerializerWithCriteria(VideoSerializer):
    criteria_scores = VideoCriteriaScoreSerializer(many=True)
    
    class Meta(VideoSerializer.Meta):
        fields = VideoSerializer.Meta.fields + ['criteria_scores'] 
```

backend/requirements.txt
```diff
@@ -17,6 +17,7 @@ pytest==6.2.4
 PyYAML==5.4.1
 tqdm==4.62.1
 langdetect==1.0.8
+isodate
```
lfaucon, line 20: Can you document it in `python_dependency.md`? Explain where we use it and why.

lfaucon (approved):
Looks good to me. Approving with suggestion & question

backend/requirements.txt
amatissart, line 20: The version used should be specified, to avoid potential breakage when a new version is available:
```suggestion
isodate==0.6.0
```

backend/tournesol/serializers.py
```diff
@@ -38,8 +40,15 @@ class Meta:
...
```
amatissart, line 50: A type hint is necessary, so as to the correct type appears in the OpenAPI schema.  
Also, I would recommend to return `null` when the duration is not defined, to make sure the frontend handles this case explicitly.
```suggestion
    # Convert duration to seconds to facilitate use of Humanize package
    def get_duration(self, obj) -> Optional[int]:
        if obj.duration:
            return int(obj.duration.total_seconds())
        return None
```

amatissart (changes requested):

amatissart:
@sandre35 This branch needs to be rebased on main, as there are minor conflicts after changes introduced by #257.

amatissart (approved):

PR #319, opened 2021-11-25
(#194) [front] feat: revoke access token at logout

### LINKED ISSUE #194: Login out: token invalidation
Invalidate tokens in the backend when a user chooses to logout.

GresilleSiffle:
Fixes #194 

I added a `logoutProcess` function, in the login / logout component, responsible of calling all logout routines. At the moment there are only two steps: revoking the token, and logout the currently logged user.

Django OAuth Toolkit doc: https://django-oauth-toolkit.readthedocs.io/en/latest/tutorial/tutorial_04.html?highlight=revoke#setup-a-request

### DIFF
--- a/frontend/src/features/frame/components/topbar/AccountInfo.tsx
+++ b/frontend/src/features/frame/components/topbar/AccountInfo.tsx
@@ -1,12 +1,15 @@
 import React from 'react';
+import { Link } from 'react-router-dom';
 import clsx from 'clsx';
+import { useSnackbar } from 'notistack';
+
 import { makeStyles } from '@material-ui/core/styles';
 import { Button, Grid } from '@material-ui/core';
-import { Link } from 'react-router-dom';
 import { AccountCircle } from '@material-ui/icons';
 
 import { useLoginState } from 'src/hooks';
 import { revokeAccessToken } from '../../../login/loginAPI';
+import { contactAdministratorLowSeverity } from '../../../../utils/notifications';
 
 const useStyles = makeStyles({
   AccountInfo: {
@@ -31,12 +34,19 @@ const useStyles = makeStyles({
 });
 
 const LoggedInActions = () => {
+  const { enqueueSnackbar } = useSnackbar();
+
   const classes = useStyles();
   const { logout, loginState } = useLoginState();
 
   const logoutProcess = async () => {
     if (loginState.access_token) {
-      await revokeAccessToken(loginState.access_token);
+      await revokeAccessToken(loginState.access_token).catch(() => {
+        contactAdministratorLowSeverity(
+          enqueueSnackbar,
+          'A non impacting error occurred during your logout.'
+        );
+      });
     }
     logout();
   };
--- a/frontend/src/features/frame/components/topbar/AccountInfo.tsx
+++ b/frontend/src/features/frame/components/topbar/AccountInfo.tsx
@@ -40,8 +40,8 @@ const LoggedInActions = () => {
   const { logout, loginState } = useLoginState();
 
   const logoutProcess = async () => {
-    if (loginState.access_token) {
-      await revokeAccessToken(loginState.access_token).catch(() => {
+    if (loginState.refresh_token) {
+      await revokeAccessToken(loginState.refresh_token).catch(() => {
         contactAdministratorLowSeverity(
           enqueueSnackbar,
           'A non impacting error occurred during your logout.'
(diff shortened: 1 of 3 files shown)


GresilleSiffle:
If there are any error during the token revocation, the user is still logged out, but an information is displayed so that the application doesn't fail silently.

screencapture-localhost-3000-2021-11-25-20_11_00

frontend/src/features/frame/components/topbar/AccountInfo.tsx
```diff
@@ -30,14 +34,28 @@ const useStyles = makeStyles({
...
 
 const LoggedInActions = () => {
+  const { enqueueSnackbar } = useSnackbar();
+
   const classes = useStyles();
   const { logout, loginState } = useLoginState();
 
+  const logoutProcess = async () => {
+    if (loginState.access_token) {
+      await revokeAccessToken(loginState.access_token).catch(() => {
```
amatissart, line 44: As far as I can see, the refresh_token remains valid after the access_token has been revoked. I think we would expect both to be revoked, to avoid any credentials from a previous session to be reused.

Calling `/o/revoke_token` with the refresh_token would actually be enough, as the API will take care of deleting the associated access_token.

frontend/src/utils/notifications.ts
```diff
@@ -22,6 +22,27 @@ export const contactAdministrator = (
...
+  display: ProviderContext['enqueueSnackbar'],
+  message?: string
+) => {
+  if (!message) {
+    message = 'Oops, a non impacting error occurred.';
+  }
+  display('Please, contact an administrator to report the issue.', {
+    variant: 'warning',
+  });
+  display(message, { variant: 'info' });
```
amatissart, line 43: This is really minor, but I find a bit disturbing to trigger 2 messages with different severity levels. As the error is not really severe, could "info" be used instead of "warning" here?

frontend/src/features/frame/components/topbar/AccountInfo.tsx
GresilleSiffle: > As far as I can see, the refresh_token remains valid after the access_token has been revoked.

That's true.

> I think we would expect both to be revoked, to avoid any credentials from a previous session to be reused.

You're right, for a reason I thought the `/o/revoke_token/` was only for access tokens.

frontend/src/utils/notifications.ts
GresilleSiffle: Of course :+1:

frontend/src/features/frame/components/topbar/AccountInfo.tsx
GresilleSiffle: According to the [RFC 7009 section 2.1](https://datatracker.ietf.org/doc/html/rfc7009#section-2.1), revoking a refresh token " should " also revoke the access token.

>  If the particular token is a refresh token and the authorization server supports the revocation of access tokens, then the authorization server SHOULD also invalidate all access tokens based on the same authorization grant (see Implementation Note).  If the token passed to the request is an access token, the server MAY revoke the respective refresh token as well.

But, Django OAuth Toolkit seems to only revoke the access token, even if I use the refresh token in the request.

Example, I expected the following request to delete the refresh token `75U5OZPTCeXC2jnfpkj2qHJz6mNg4y`, but it deletes the related access token instead.

```http
POST /o/revoke_token/ HTTP/1.1
Content-Type: application/x-www-form-urlencoded
token=75U5OZPTCeXC2jnfpkj2qHJz6mNg4y&client_id=[...]&client_secret=[...]&token_type_hint=refresh_token
```

I'm gonna do more tests.

amatissart: > Example, I expected the following request to delete the refresh token 75U5OZPTCeXC2jnfpkj2qHJz6mNg4y, but it deletes the related access token instead.

 The refresh_token is also revoked, but contrary to the access_token it's not deleted from the database immediately. It's only marked as revoked, via the field `revoked`.

GresilleSiffle: Thanks!

frontend/src/utils/notifications.ts
GresilleSiffle: Fixed by d63b49c10f9bb4ef67522571e7ed04d231069a66.

frontend/src/features/frame/components/topbar/AccountInfo.tsx
GresilleSiffle: Fixed by d63b49c10f9bb4ef67522571e7ed04d231069a66.

amatissart (approved):

PR #320, opened 2021-11-26
feat(backend): Implement ContributorRating initialization endpoint

amatissart:
:warning: Depends on #315 

* New endpoint `POST /users/me/contributors_ratings/` to initialize a rating about a specific video, with an optional `is_public` preference.  
This endpoint will typically be called before a user submits their first comparison about a video.

* The `ContributorRating` serializers include a new field `n_comparisons` (= the count of comparisons by the authenticated user about the current video)

### DIFF
--- a/backend/tournesol/serializers.py
+++ b/backend/tournesol/serializers.py
@@ -302,6 +302,14 @@ class ContributorRatingCreateSerializer(ContributorRatingSerializer):
             video = Video.objects.get(video_id=video_id)
         except Video.DoesNotExist:
             raise ValidationError(f"Video with video_id '{video_id}' does not exist")
+
+        user = self.context["request"].user
+        if user.contributorvideoratings.filter(video=video).exists():
+            raise ValidationError(
+                "A ContributorRating already exists for this (user, video)",
+                code='unique',
+            )
+
         attrs["video"] = video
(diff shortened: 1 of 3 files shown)


GresilleSiffle:
Hello @amatissart 

The new code is good and well structured for me :+1: 

I found only one bug with `POST /users/me/contributor_ratings`: calling it several times with the same  `video_id` raises the exception `IntegrityError` instead returning a JSON body with a parsable message.

```
IntegrityError at /users/me/contributor_ratings/
ERREUR:  la valeur d'une clé dupliquée rompt la contrainte unique « tournesol_contributorvideorating_user_id_video_id_1717b1fc_uniq »
DETAIL:  La clé « (user_id, video_id)=(1, 1) » existe déjà.
``` 

I also found one minor issue not introduced by this PR, but still related to ratings. We can either fix it now, or after everything related to public / private ratings is merged.

The `GET`, `PUT` and `PATCH` endpoints share the same description in the generated documentation, which is wrong for `PUT` and `PATCH`
> Retrieve the logged in user's ratings for a specific video (computed automatically from the user's comparisons) 

We could use something like that instead:

> Update public / private status of the logged in user's rating for a specific video.

What do you think?

GresilleSiffle (approved):

PR #322, opened 2021-11-26
feat(frontend): Public status toggle on comparison page

amatissart:
## Description

On the comparison page, the number of existing comparisons by the current user is displayed below each video.
And a toggle is available to update the visibility of these comparisons:

### Desktop

image

A message is also visible near the "submit" button to inform about the user the current comparison is or will be public:
image

---
### Mobile

|Default|Expanded|
|-|-|
|image|image


GresilleSiffle:
Hi,

We took the time to test the UI with @lfaucon. As there was no direction indicated in the related issue, we would like to propose the text:

> You have compared it X time(s)

instead of:

> 1 comparison(s)

The later confused me a bit, because I didn't know if this number was counting only my comparisons or all the comparisons in the database. The sentence is longer, and may not appear nicely on mobile for now, here are some ideas for consideration:

1. hide all additional video information on mobile. It will fit the current design at the cost of a mobile UI displaying less information.
2. make the two video stack on each other (with the component `Grid`), it requires a little re-organization of the component JSX, and will display as much information on desktop than on mobile.

amatissart:
@lfaucon I implemented your suggestions:
 * A card is wrapping the description of each video, including the text field.
 * The font size used in VideoCard is smaller on xs screens
 * By default, the toggle value is "public" for videos that have not been compared by the user yet
 * The bug where settings were hidden in VideoCard after the window is resized is fixed
 
Bonus: a slight animation is used in `<EmptyVideoCard>` when the content is loading  :sunflower:

frontend/src/features/videos/VideoList.tsx
```diff
@@ -1,46 +1,35 @@
...
-        videos.results.map((video: VideoSerializerWithCriteria) => (
-          <Grid container className={classes.card} key={video.video_id}>
-            <Grid item xs={12}>
-              <VideoCard video={video} actions={actions} />
-            </Grid>
-          </Grid>
+    <>
+      {videos.length ? (
+        videos.map((video: VideoObject) => (
+          <Box key={video.video_id} margin={1}>
```
lfaucon, line 23: You have reduced the margin between the video cards on the recommendation page and I think it's going the opposite way as what we should do. I remember Laure reporting that it lacks a bit of space. I appreciate the extra padding though.

lfaucon (approved):
Really good, thanks for addressing all recommendations! Let's ship it

Issue #241, opened 2021-10-21
Define a public description 


sandre35:
The file `index.html` in the folder public of frontend is empty. We could provide a description for people who are mentioning tournesol, and also appears in the research criterias of google

aidanjungo:
@lfaucon Should it be a short or a long description?

I would propose:

- if short: 
`Collaborative Content Recommendations`

if long:
`Tournesol is an open source platform which aims to collaboratively identify top videos of public utility by eliciting contributors' judgements on content quality. We hope to contribute to making today's and tomorrow's large-scale algorithms robustly beneficial for all of humanity.`

aidanjungo:
Also I would like to change that logo, but I don't know where I have to change it...

![logogoogle](https://user-images.githubusercontent.com/53896963/143683745-22c27e28-394f-443d-b47e-508184d0db37.png)

lfaucon:
I can look for how to change the logo. I'd say long description

PR #330, opened 2021-11-27
Define public description

aidanjungo:
related to #241


aidanjungo:
I change a logo that was pointing to a none existing one, but I am not sure if we should change something else.
Also I don't know how to test that .

lfaucon (approved):

lfaucon:
I don't think you changed the icon related to what appears in Google search. It's probably something related to how the website looks on some apple tablet of phone. A good change nontheless.

PR #333, opened 2021-12-02
Improvements to ContributorRatings admin

lfaucon:
* link_to_youtube moved to Video model
* link_to_youtube opens a new tab
* Clean-up ContributorRating admin
* CR admin displays and can filter on is_public

### DIFF
--- a/backend/tournesol/admin.py
+++ b/backend/tournesol/admin.py
@@ -3,7 +3,6 @@ Defines Tournesol's backend admin interface
 """
 
 from django.contrib import admin
-from django.utils.html import format_html
 
 from .models import (
     Video,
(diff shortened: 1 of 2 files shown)


backend/tournesol/admin.py
```diff
@@ -43,7 +39,18 @@ class VideoCriteriaScoreAdmin(admin.ModelAdmin):
...
 class ContributorRatingAdmin(admin.ModelAdmin):
-    pass
+    list_display = (
+        'user',
+        'video',
+        'link_to_youtube',
+        'is_public',
+    )
+    list_filter = (
+        ('is_public', admin.AllValuesFieldListFilter),
```
amatissart, line 49: `admin.AllValuesFieldListFilter` is useful to build a list from the values present in the database . It was relevant for the `language` field in Video since the enumeration of all accepted languages would be very large.

As `is_public` is a boolean, no explicit filter class is required:
```suggestion
        'is_public',
```

amatissart (approved):
Approved, with a minor comment.

PR #336, opened 2021-12-03
Add possibility to filter on uploader

sandre35:


frontend/src/features/recommendation/UploaderFilter.tsx
```diff
@@ -0,0 +1,33 @@
...
+          type="submit"
+          size="large"
+          variant="contained"
+          color="primary"
+          onClick={props.onChange}
+        >
+          <span>Reset</span>
+        </Button>
+      </Box>
+    </FilterSection>
```
amatissart, line 29: Instead of adding a new `<FilterSection>` that is a bit cumbersome in the existing layout, it could be a very good usecase for a deletable [`<Chip>`](https://v4.mui.com/components/chips/#chip-2).  

Then, it could be displayed above other filters, with minimal modification:
image



And the implementation is simple enough:
```tsx
    <Chip
      label={
        <Typography variant="body2">
          <strong>Uploader: </strong>
          {props.value}
        </Typography>
      }
      color="secondary"
      onDelete={props.onChange}
    />
```

What do you think @sandre35?

frontend/scripts/openapi.yaml
```diff
@@ -976,6 +976,10 @@ paths:
         name: language
         schema:
           type: string
+      - in: query
+        name: uploader
+        schema:
+          type: string
```
amatissart, line 982: Has the schema been generated automatically? On this branch, the new `OpenApiParameter` has not been defined in [`views/video.py`](https://github.com/tournesol-app/tournesol/blob/c484c58d4a8c176c7d5dcdf1436086f040f0b5e3/backend/tournesol/views/video.py#L33-L60)

frontend/src/features/recommendation/UploaderFilter.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Box, Button } from '@material-ui/core';
+import { FilterSection } from 'src/components';
+
+interface Props {
+  value: string;
+  onChange: () => void;
```
amatissart, line 7: `onDelete` may be a more relevant name, as the only possible action is to delete the existing filter.

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -218,9 +220,18 @@ function VideoCard({
               {video.publication_date}
             </span>
           )}
-          {video.uploader && (
-            <span className={classes.channel}>{video.uploader}</span>
-          )}
+          <Tooltip title="Filter by channel" placement="bottom">
```
amatissart, line 223: For consistency with the `UploaderFilter`, "Filter by uploader" could be a more appropriate title.

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -218,9 +220,18 @@ function VideoCard({
...
+            <div>
+              {video.uploader && (
+                <Link
+                  to={`/recommendations?uploader=${video.uploader}`}
```
amatissart, line 227: The uploader name needs to be URL-encoded here, for example using `encodeURIComponent`, or `URLSearchParams`. Otherwise special characters could lead to invalid query parameters in the URL.

amatissart (commented):
Nice! The behavior is

frontend/scripts/openapi.yaml
sandre35: Fixed

backend/tournesol/views/video.py
```diff
@@ -86,6 +88,10 @@ def get_queryset(self):
         request = self.request
         queryset = self.queryset
 
+        uploader = request.query_params.get('uploader')
+        if uploader:
+            queryset = queryset.filter(uploader=unquote(uploader))
```
amatissart, line 93: `request.query_params` is already responsible for decoding URL params, so `unquote` is unnecessary here and could event lead to incorrect values.

frontend/src/features/recommendation/RecommendationApi.tsx
```diff
@@ -56,6 +56,7 @@ export const getRecommendedVideos = async (searchString: string) => {
       offset: getNumberValue('offset'),
       search: params.get('search') ?? undefined,
       language: params.get('language') ?? undefined,
+      uploader: encodeURI(params.get('uploader') ?? ''),
```
amatissart, line 59: The generated client implements the necessary function to encode these params correctly. So the values should not be encoded here either.

You can have a look at the `getQueryString()` function in "services/openapi/core.request.ts" if you want to learn more about how it works.

frontend/src/features/videos/VideoCard.tsx
amatissart: However here, the URL is passed directly to the History API, and will update the current URL. An uploader name such `"a&b"` would be injected as `?uploader=a&b` and would be parsed as 2 distincts parameters `uploader` and `b`.  

That's why the parameter needs to be encoded explicitly here ;)

```suggestion
                  to={`/recommendations?uploader=${encodeURIComponent(video.uploader)}`}
```

frontend/src/features/recommendation/SearchFilter.tsx
```diff
@@ -65,6 +66,12 @@ function SearchFilter() {
         Filters
       </Button>
       <Collapse in={expanded} timeout="auto" unmountOnExit>
+        {searchParams.get('uploader') && (
+          <UploaderFilter
+            value={searchParams.get('uploader') ?? ''}
+            onDelete={() => setFilter('uploader', '')}
+          />
```
amatissart, line 73: What about adding a little margin-bottom here?
```suggestion
          <Box marginBottom={1}>
            <UploaderFilter
              value={searchParams.get('uploader') ?? ''}
              onDelete={() => setFilter('uploader', '')}
            />
          </Box>
```

amatissart (commented):
Nice!  

There is still an issue about the "uploader" value that is injected in the URL. My suggestion about encoding the uploader name was a bit confusing.  
Sorry about that. I hope the comments below will be clearer.

frontend/src/features/recommendation/SearchFilter.tsx
```diff
@@ -23,9 +24,15 @@ function SearchFilter() {
...
-    <Box color="text.secondary">
+    <Box color="text.secondary" marginBottom={1}>
       <CollapseButton expanded={expanded} onClick={handleExpandClick} />
       <Collapse in={expanded} timeout="auto" unmountOnExit>
+        {filterParams.get('uploader') && (
+          <UploaderFilter
+            value={filterParams.get('uploader') ?? ''}
+            onDelete={() => setFilter('uploader', '')}
+          />
+        )}
```
amatissart, line 35: Sorry to bother you again about the margin. With the current branch, the padding of the other filters conflicts with the "delete" button, which is only partially clickable:
image

I think that `marginButtom` should be applied around the `UploaderFilter` specifically:
```jsx
          <Box marginBottom={1}>
            <UploaderFilter
              value={searchParams.get('uploader') ?? ''}
              onDelete={() => setFilter('uploader', '')}
            />
          </Box>
```

amatissart (approved):

PR #340, opened 2021-12-04
[front] feat: make PasswordForm disabled during HTTP requests

GresilleSiffle:
This PR adds a little improvment to `PasswordForm` component UI/UX.

The submit button is now disabled when an HTTP request is processing, as is the submit button of the login form. The tests have been updated accordingly.

Disabling the submit button adds a visual hint indicating that something is going on in the background and informs the user that his/her action has been correctly submitted.

It also prevents users to click several time in a row on the submit button and might prevent unexpected API or UI behaviours.

I also made a similar PR for the `ProfileForm` here: #341

### DIFF
--- a/frontend/src/features/settings/account/PasswordForm.tsx
+++ b/frontend/src/features/settings/account/PasswordForm.tsx
@@ -19,12 +19,14 @@ const PasswordForm = () => {
   const [oldPassword, setOldPassword] = useState('');
   const [password, setPassword] = useState('');
   const [passwordConfirm, setPasswordConfirm] = useState('');
+  const [disabled, setDisabled] = React.useState(false);
 
   const passwordConfirmMatches =
     password !== '' && password === passwordConfirm;
 
   const handleSubmit = async (event: React.FormEvent) => {
     event.preventDefault();
+    setDisabled(true);
 
     const response: void | Record<string, string> =
       await AccountsService.accountsChangePasswordCreate(
@@ -60,6 +62,7 @@ const PasswordForm = () => {
       setPasswordConfirm('');
       (document.activeElement as HTMLElement).blur();
     }
+    setDisabled(false);
   };
 
   return (
@@ -117,7 +120,13 @@ const PasswordForm = () => {
           />
         </Grid>
         <Grid item>
-          <Button type="submit" color="secondary" fullWidth variant="contained">
+          <Button
+            fullWidth
+            type="submit"
+            color="secondary"
+            variant="contained"
+            disabled={disabled}
+          >
             Update password
           </Button>
         </Grid>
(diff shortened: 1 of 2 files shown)


GresilleSiffle:
Here is how it looks like:

https://user-images.githubusercontent.com/39056254/144706509-fe2fc62d-62da-45ee-9be2-510044a417cd.mp4

frontend/src/features/settings/account/PasswordForm.tsx
```diff
@@ -19,12 +19,14 @@ const PasswordForm = () => {
   const [oldPassword, setOldPassword] = useState('');
   const [password, setPassword] = useState('');
   const [passwordConfirm, setPasswordConfirm] = useState('');
+  const [disabled, setDisabled] = React.useState(false);
```
lfaucon, line 22: `useState` is imported. See usage on line 21.

frontend/src/features/settings/account/PasswordForm.tsx
```diff
@@ -117,7 +120,13 @@ const PasswordForm = () => {
           />
         </Grid>
         <Grid item>
-          <Button type="submit" color="secondary" fullWidth variant="contained">
+          <Button
+            fullWidth
+            type="submit"
+            color="secondary"
+            variant="contained"
+            disabled={disabled}
```
lfaucon, line 128: Unrelated to the PR, but would it be better with `diasbled || !passwordConfirmMatches` ?

lfaucon (approved):

frontend/src/features/settings/account/PasswordForm.tsx
GresilleSiffle: Thanks, I'll fix that.

frontend/src/features/settings/account/PasswordForm.tsx
GresilleSiffle: We could indeed disable the submit button if the whole form is not valid. In our case it means:
- it's not waiting for an HTTP response ;
- password and confirmation matches ;
- and the old password field is not empty.

This way the form will perform a full validation before enabling the button, which would be more user friendly.

Nevertheless, my personal preference would be to keep the form as is, because for me, disabled buttons appear like either disabled features, temporary unavailable features, or features with strong impact like deleting an account. I don't feel invited to use forms with disabled submit buttons, as I don't feel invited to delete my account on Tournesol. On the other hand, I feel invited to change my e-mail, my password, etc.

It's a really small preference, but still a preference : ) I'm curious about your opinion on this.

frontend/src/features/settings/account/PasswordForm.tsx
GresilleSiffle: Fixed by db5c1e6475c264dca426f878eeba2d05971b7f9a.

frontend/src/features/settings/account/PasswordForm.tsx
lfaucon: My suggestion was a question full of doubts :) I understand and find your reasons convincing.

PR #341, opened 2021-12-04
[front] feat: make ProfileForm disabled during HTTP requests

GresilleSiffle:
It's like #340 but for the `ProfileForm` component.


lfaucon (approved):
Nice improvement 👍

Issue #338, opened 2021-12-03
Update logo before public announcement


aidanjungo:
I think the new Tournesol logo should be changed on:

- [x] Github
- [x] Discord
- [x] Wiki
- [x] Twitter
- [x] Twitter bots (fr/en)
- [x] LinkedIn
- [x] Firefox extension store
- [x] Chrome extension store
- [x] Paypal
- [x] Instagram

PR #342, opened 2021-12-04
Update Tournesol's logos used by the extension

lfaucon:
Related to #338


browser-extension/manifest.json
```diff
@@ -1,20 +1,21 @@
...
+    "512": "Logo512.png"
   },
   "background": {
     "page": "background.html",
     "persistent": true
   },
   "browser_action": {
     "default_icon": {              
-      "16": "favicon.png"
+      "16": "Logo16.png"
```
amatissart, line 18: It would be useful to define a larger alternative here, for high-dpi displays:
```suggestion
      "16": "Logo16.png",
      "64": "Logo64.png"
```

browser-extension/manifest.json
```diff
@@ -34,6 +35,6 @@
     }
   ],
   "web_accessible_resources": [
-	  "rate_now_icon.png"
+	  "Logo128.png"
```
amatissart, line 38: This filename also needs to be updated in addRateLaterButton.js : 
https://github.com/tournesol-app/tournesol/blob/3a1ff590e693dad3035f9b4a7ca395a830efac1f/browser-extension/addRateLaterButton.js#L56

GresilleSiffle:
I don't know if it's related to these changed but I noticed a broken image link on YouTube after installing the extension from this branch.

See the image bellow, on Chromium:
 
capture

The same error on Firefox.

Images in Chromium and Firefox toolbars are OK.

lfaucon:
Thanks @GresilleSiffle & @amatissart for checking, I indeed forgot to update one of the reference to the old icons that I removed. It should be good now.

amatissart (approved):

PR #343, opened 2021-12-05
fix(backend): Server error on video creation when statistics are missing in Youtube API response

amatissart:
For some videos, such as [this one](https://www.youtube.com/watch?v=JMpuxLIgjPs), the `viewCount` is not provided by Youtube. So this field needs to be considered as optional when parsing the API response.

This PR also simplifies a bit the implementation of `POST /video/`, to take advantage of the mechanism provided by the DRF mixins that were already defined. Especially, this makes sure that the Youtube API is not called before the request data has been properly validated.

### DIFF
--- a/backend/tournesol/views/video.py
+++ b/backend/tournesol/views/video.py
@@ -8,11 +8,10 @@ from django.utils import timezone, dateparse
 from django.db.models import Q, Case, When, Sum, F
 from django.conf import settings
 
-from rest_framework import mixins, status
+from rest_framework import mixins
 from rest_framework.pagination import LimitOffsetPagination
-from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
-from rest_framework.exceptions import ValidationError
+from rest_framework.exceptions import ValidationError, NotFound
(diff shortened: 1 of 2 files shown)


backend/tournesol/views/video.py
```diff
@@ -139,32 +141,16 @@ def get_serializer_class(self):
...
-        if not video_id:
-            return Response('No video_id given', status=status.HTTP_400_BAD_REQUEST)
-        if len(video_id) != 11:
-            return Response("video_id is not valid", status=status.HTTP_400_BAD_REQUEST)
-        if Video.objects.filter(video_id=video_id):
-            Response(
-                "The video with the id : {id} is already registered".format(
-                    id=request.data["video_id"]),
-                status=status.HTTP_400_BAD_REQUEST
-            )
```
amatissart, line 156: A `return` statement was missing here to avoid a call to the Youtube API when the video already exists in our database. 

By overriding `perform_create()` instead of `create()`, the `VideoSerializer` is initialized automatically, and we can avoid redundant validation here.

lfaucon (approved):
Works as expected. We were indeed pinging the youtube API for already existing videos. Very important fix scaling up.

PR #344, opened 2021-12-06
refactor(frontend): Define reusable components for search filters, tweak appearance

amatissart:
### Description

This is a prerequisite to implement the new page "My rated videos", that will also include similar filters (on public/private ratings, etc.)

2 new components: 
* **`FilterSection`**: a generic filter section with an underlined title
* **`ChoicesFilterSection`**: a filter column, with multiple checkboxes defined by a list of `choices`.   
Used by `LanguageFilter` and `DateFilter`.

### Screenshots

|Before|After|
|-|-|
|image|image
|image|image

### DIFF
--- a/frontend/src/components/filter/FilterSection.tsx
+++ b/frontend/src/components/filter/FilterSection.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { Typography } from '@material-ui/core';
+
+interface Props {
+  title: string;
+  children: React.ReactNode;
+}
+
+export const FilterSection = ({ title, children }: Props) => {
+  return (
+    <>
+      <Typography
+        variant="h6"
+        style={{ borderBottom: '1px solid #E7E5DB', marginBottom: '0.3em' }}
+      >
+        {title}
+      </Typography>
+      {children}
+    </>
+  );
+};
--- a/frontend/src/components/filter/FilterSection.tsx
+++ b/frontend/src/components/filter/FilterSection.tsx
@@ -1,5 +1,13 @@
 import React from 'react';
 import { Typography } from '@material-ui/core';
+import { makeStyles } from '@material-ui/styles';
+
+const useStyles = makeStyles({
+  filterTitle: {
+    borderBottom: '1px solid #E7E5DB',
+    marginBottom: '0.3em',
+  },
+});
 
 interface Props {
   title: string;
@@ -7,12 +15,10 @@ interface Props {
 }
 
 const FilterSection = ({ title, children }: Props) => {
+  const classes = useStyles();
   return (
     <>
-      <Typography
-        variant="h6"
-        style={{ borderBottom: '1px solid #E7E5DB', marginBottom: '0.3em' }}
-      >
+      <Typography variant="h6" className={classes.filterTitle}>
         {title}
       </Typography>
       {children}
--- a/frontend/src/components/filter/FilterSection.tsx
+++ b/frontend/src/components/filter/FilterSection.tsx
@@ -6,7 +6,7 @@ interface Props {
   children: React.ReactNode;
 }
 
-export const FilterSection = ({ title, children }: Props) => {
+const FilterSection = ({ title, children }: Props) => {
   return (
     <>
       <Typography
@@ -19,3 +19,5 @@ export const FilterSection = ({ title, children }: Props) => {
     </>
   );
 };
+
+export default FilterSection;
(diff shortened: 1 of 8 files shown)


frontend/src/components/filter/FilterSection.tsx
```diff
@@ -0,0 +1,23 @@
...
+  title: string;
+  children: React.ReactNode;
+}
+
+const FilterSection = ({ title, children }: Props) => {
+  return (
+    <>
+      <Typography
+        variant="h6"
+        style={{ borderBottom: '1px solid #E7E5DB', marginBottom: '0.3em' }}
```
GresilleSiffle, line 14: I would not recommend using inline style in our components.

The day we will add a Content Security Policy in the front end, inline style will be blocked if not explicity authorized. And authorizing inline style would be a security flaw.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#unsafe_inline_styles

frontend/src/features/recommendation/SearchFilter.tsx
```diff
@@ -31,45 +21,55 @@ function SearchFilter() {
...
-        in={expanded}
-        timeout="auto"
-        unmountOnExit
-      >
-        <Grid container>
-          <CriteriaFilter setFilter={setFilter} />
-          <DateFilter setFilter={setFilter} />
-          <LanguageFilter setFilter={setFilter} />
+      <Collapse in={expanded} timeout="auto" unmountOnExit>
+        <Grid container spacing={4} style={{ marginBottom: '8px' }}>
```
GresilleSiffle, line 54: Avoidable inline style.

frontend/src/features/recommendation/SearchFilter.tsx
```diff
@@ -31,45 +21,55 @@ function SearchFilter() {
...
+    <Box color="text.secondary">
       <Button
-        color="secondary"
         size="large"
-        className={classes.filter}
         startIcon={!expanded ? <ExpandMore /> : <ExpandLess />}
         aria-expanded={expanded}
         aria-label="show more"
         onClick={handleExpandClick}
+        style={{
```
GresilleSiffle, line 44: Avoidable inline style.

GresilleSiffle, line 44: Avoidable inline style.

frontend/src/features/recommendation/SearchFilter.tsx
GresilleSiffle, line 54: Avoidable inline style.

GresilleSiffle (changes requested):
The new design is great!

I found one issue related to the security, and one error in the browser console. Can you confirm if you have the same error?

I didn't take the time to spot where the error was triggered, but in the console, moving a slider display this message:

> Material-UI: A component is changing the default value state of an uncontrolled Slider after being initialized. To suppress this warning opt to use a controlled Slider.

I confirm that, even if the message considers itself as a warning, it appears as an error.

capture

**edit:** we saw with @amatissart that the error is also present in `main`, thus not introduced by this PR, and could be fixed by using the prop `value` instead of `defaultValue` in the component `src/features/recommendation/CriteriaFilter.tsx`

See: https://v4.mui.com/api/slider/#props

frontend/src/components/filter/FilterSection.tsx
amatissart: I updated the PR to use `className` instead of inline `style`.  

Although your point about Content-Security-Policy is valid, enforcing such a policy everywhere would probably be constraining. Usage of inline styles is quite deeply baked into Material UI: see https://github.com/mui-org/material-ui/issues/19938

GresilleSiffle: You're right.

I'm curious about how many mui components use inline style.

GresilleSiffle (approved):

amatissart:
> capture
> 
> **edit:** we saw with @amatissart that the error is also present in `main`, thus not introduced by this PR, and could be fixed by using the prop `value` instead of `defaultValue` in the component `src/features/recommendation/CriteriaFilter.tsx`
> 
> See: https://v4.mui.com/api/slider/#props

As discussed on Discord, updating the `CustomSlider` to behave as a controlled input require additional changes. As it's not related to this PR and the current uncontrolled sliders seem to work as expected despite the warning in development mode, I would suggest to address this issue separately.

PR #352, opened 2021-12-08
fix(backend): Fix missing 'nullable' attributes in OpenAPI schema

amatissart:
Some fields were no longer nullable in the generated API schema, due to how `VideoSerializerWithCriteria` inherits from the main `VideoSerializer`. 

See the comments for more details.


GresilleSiffle (approved):
LGTM

PR #353, opened 2021-12-08
feat(backend): Implement route to update is_public status for all ContributorRating

amatissart:
Related to #274 

The new route `PATCH /users/me/contributor_ratings/_all/`  will be used on the page "My rated videos" to allow marking of all existing user contributions as "public" or "private".

### DIFF
--- a/backend/tournesol/urls.py
+++ b/backend/tournesol/urls.py
@@ -12,7 +12,11 @@ from .views.exports import ExportComparisonsView, ExportAllView
 from .views.video import VideoViewSet
 from .views.video_rate_later import VideoRateLaterDetail, VideoRateLaterList
 from .views.user import CurrentUserView
-from .views.ratings import ContributorRatingList, ContributorRatingDetail
+from .views.ratings import (
+    ContributorRatingList,
+    ContributorRatingDetail,
+    ContributorRatingUpdateAll,
+)
 from .views.email_domains import EmailDomainsList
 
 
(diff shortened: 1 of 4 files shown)


GresilleSiffle (approved):
LGTM

backend/tournesol/urls.py
```diff
@@ -70,6 +74,11 @@
         ContributorRatingList.as_view(),
         name="ratings_me_list",
     ),
+    path(
+        "users/me/contributor_ratings/_all/",
```
lfaucon, line 78: Do we need this route? Can't we just use the patch on the route `"users/me/contributor_ratings/"` and rely on the ContributorRatingList view?

amatissart: I actually hesitated with this solution. Implementing `PATCH` on a collection of resources does not seem to be a very common pattern in RESTful APIs. And DRF mixins are not suitable for this case, so a custom implementation would be necessary anyway.

I also find that defining a special identifier makes the intent a bit more explicit: `_all` can be PATCHed just like any other instance. And any future action/route on individual ratings could also be defined for this identifier too.

But it's mostly a matter of taste. If you prefer to stick closer to "standard" REST I can change it.

lfaucon: Sounds good! Let's go with your initial proposition 👍

lfaucon (approved):

PR #354, opened 2021-12-08
frontend: Update OpenAPI schema, generate API client with named arguments

amatissart:
Now that the `/video/` endpoints are correctly described in the schema (#332), the corresponding routes can be called using the generated client. However, `openapi-typescript-codegen` uses unnamed arguments by default. It would make the usage of this endpoint (currently with 16 optional query parameters) very impractical.

That's why this PR sets [`--useOptions`](https://github.com/ferdikoomen/openapi-typescript-codegen#argument-style-vs-object-style---useoptions) in the "openapi" command line, to use options objects as arguments: parameters must be named explicitly, and no specific order is required. Moreover, the body is passed as `requestBody`. As a result, this PR includes many adjustments in how the generated client is used. But the behavior should be unchanged.

### DIFF
--- a/frontend/src/features/recommendation/RecommendationApi.tsx
+++ b/frontend/src/features/recommendation/RecommendationApi.tsx
@@ -1,4 +1,8 @@
 import { VideoService } from 'src/services/openapi';
+import { allCriteriaNamesObj } from 'src/utils/constants';
+import { snakeToCamel } from 'src/utils/string';
+
+const CRITERIA_KEYS = Object.keys(allCriteriaNamesObj);
 
 export const getRecommendedVideos = async (searchString: string) => {
   const dayInMillisecondes = 1000 * 60 * 60 * 24;
@@ -53,16 +57,9 @@ export const getRecommendedVideos = async (searchString: string) => {
       search: params.get('search') ?? undefined,
       language: params.get('language') ?? undefined,
       dateGte: params.get('date_gte') ?? undefined,
-      largelyRecommended: getNumberValue('largely_recommended'),
-      reliability: getNumberValue('reliability'),
-      importance: getNumberValue('importance'),
-      engaging: getNumberValue('engaging'),
-      pedagogy: getNumberValue('pedagogy'),
-      laymanFriendly: getNumberValue('layman_friendly'),
-      diversityInclusion: getNumberValue('diversity_inclusion'),
-      backfireRisk: getNumberValue('backfire_risk'),
-      betterHabits: getNumberValue('better_habits'),
-      entertainingRelaxing: getNumberValue('entertaining_relaxing'),
+      ...Object.fromEntries(
+        CRITERIA_KEYS.map((c) => [snakeToCamel(c), getNumberValue(c)])
+      ),
     });
   } catch (err) {
     console.error(err);
--- a/frontend/src/features/recommendation/RecommendationApi.tsx
+++ b/frontend/src/features/recommendation/RecommendationApi.tsx
@@ -1,13 +1,6 @@
-import type { PaginatedVideoSerializerWithCriteriaList as PaginatedVideoList } from 'src/services/openapi';
+import { VideoService } from 'src/services/openapi';
 
-const api_url = process.env.REACT_APP_API_URL;
-const client_id = process.env.REACT_APP_OAUTH_CLIENT_ID || '';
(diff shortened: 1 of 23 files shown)


frontend/src/features/videos/VideoApi.tsx
```diff
@@ -1,12 +1,11 @@
...
+  VideoService,
+  VideoSerializerWithCriteria as Video,
+} from 'src/services/openapi';
 
-const api_url = process.env.REACT_APP_API_URL;
-const client_id = process.env.REACT_APP_OAUTH_CLIENT_ID || '';
-const client_secret = process.env.REACT_APP_OAUTH_CLIENT_SECRET || '';
-
-const _inMemoryCache: { [videoId: string]: Video } = {};
+const _inMemoryCache: { [videoId: string]: Promise<Video> } = {};
```
amatissart, line 8: Not directly related to this PR, but nice to have: the cache values are now  `Promise<Video>`. As soon as a request is executed, the promise is present in cache, and the multiple requests to the same video are no longer duplicated.

frontend/src/utils/video.ts
```diff
@@ -70,7 +70,9 @@ export async function getVideoFromRateLaterListForComparison(
   otherVideo: string | null,
   currentVideo: string | null
 ): Promise<string | null> {
-  const rateLaterResult = await UsersService.usersMeVideoRateLaterList(99, 0);
+  const rateLaterResult = await UsersService.usersMeVideoRateLaterList({
+    limit: 99,
+  });
```
sandre35, line 75: We could maybe keep offset parameter for more clarity

frontend/src/utils/video.ts
```diff
@@ -87,11 +89,11 @@ export async function getVideoFromPreviousComparisons(
   currentVideo: string | null
 ): Promise<string | null> {
   const comparisonCount: number =
-    (await UsersService.usersMeComparisonsList(0, 0))?.count || 0;
-  const comparisonVideoResult = await UsersService.usersMeComparisonsList(
-    99,
-    Math.floor(Math.random() * comparisonCount)
-  );
+    (await UsersService.usersMeComparisonsList({ limit: 0 }))?.count || 0;
```
sandre35, line 92: same here

frontend/src/utils/video.ts
```diff
@@ -131,7 +133,7 @@ export async function getVideoForComparison(
     );
     if (videoFromComparisons) return videoFromComparisons;
   }
-  const videoResult = await VideoService.videoList(100, 0);
+  const videoResult = await VideoService.videoList({ limit: 100 });
```
sandre35, line 136: here too

sandre35 (commented):
Great job, this PR is very dense

frontend/src/features/recommendation/RecommendationApi.tsx
```diff
@@ -48,30 +41,34 @@ export const getRecommendedVideos = (
...
+      largelyRecommended: getNumberValue('largely_recommended'),
+      reliability: getNumberValue('reliability'),
+      importance: getNumberValue('importance'),
+      engaging: getNumberValue('engaging'),
+      pedagogy: getNumberValue('pedagogy'),
+      laymanFriendly: getNumberValue('layman_friendly'),
+      diversityInclusion: getNumberValue('diversity_inclusion'),
+      backfireRisk: getNumberValue('backfire_risk'),
+      betterHabits: getNumberValue('better_habits'),
+      entertainingRelaxing: getNumberValue('entertaining_relaxing'),
```
lfaucon, line 65: I think this could be automatically generated from the `allCriteriaNames` constant. Unless Typescript complains about it?

frontend/src/features/videos/VideoApi.tsx
lfaucon: Thanks! Looks good

lfaucon (commented):
I wouldn't be very confident about all the changes if we were not using TypeScriopt. That's really nice.

frontend/src/features/recommendation/RecommendationApi.tsx
amatissart: Done, using a new util function to convert params from `snake_case` to `camelCase`.

frontend/src/utils/video.ts
amatissart: Done

Issue #355, opened 2021-12-08
feat: API endpoint to download public data


lfaucon:
The goal of this feature is to enable anyone to download Tournesol's public data.

## Backend
Implement an API endpoint `exports/comparisons/` similar to `users/me/exports/comparisons/` except that comparisons are not filtered for the users, but all comparisons that are public are returned.  
Note that a comparison is public if both videos are tagged as public by the user (i.e `is_public` is True in the `ContributorRating` objects related to each video).

## Frontend 
Add a section at the bottom of the Home page with the text:

>Contributors on Tournesol can decide to make their data public. You may download the dataset of all data publicly available. We hope this important data will prove useful for researchers on ethics of algorithms and large scale recommender systems.

and with a yellow button that downloads the csv file from the `exports/comparisons/` endpoint. You may use a similar implementation as for the component `ExportAllDataForm`

anikeith22:
Working on the frontend component of this issue.

PR #360, opened 2021-12-12
API endpoint to download public data (Completed)

YJ-Hannah:
This PR is for issue #355

### DIFF
--- a/backend/tournesol/views/exports.py
+++ b/backend/tournesol/views/exports.py
@@ -62,8 +62,7 @@ def write_public_comparisons_file(request, write_target):
             "video_b": comparison["video_b"]["video_id"],
             **criteria_score
         }
-        for public_username in public_usernames
-        for comparison in serialized_comparisons
+        for (public_username, comparison) in zip(public_usernames, serialized_comparisons)
         for criteria_score in comparison['criteria_scores']
     )
 
--- a/backend/tournesol/views/exports.py
+++ b/backend/tournesol/views/exports.py
@@ -47,23 +47,23 @@ def write_public_comparisons_file(request, write_target):
     serialized_comparisons = []
     public_videos = set((rating.user, rating.video) for rating in public_data)
     comparisons = Comparison.objects.all()\
-        .select_related("video_1", "video_2")\
+        .select_related("video_1", "video_2", "user")\
         .prefetch_related("criteria_scores")
     public_comparisons = [
         comparison for comparison in comparisons
         if ((comparison.user, comparison.video_1) in public_videos
             and (comparison.user, comparison.video_2) in public_videos)]
-    serialized_comparisons = [
-        (comparison.user.username, ComparisonSerializer(comparison).data)
-        for comparison in public_comparisons]
-
+    public_usernames = [comparison.user.username for comparison in public_comparisons]
+    serialized_comparisons = ComparisonSerializer(public_comparisons, many=True).data
+    
     writer.writerows(
         {
-            "public_username": comparison[0],
+            "public_username": public_username,
             "video_a": comparison[1]["video_a"]["video_id"],
             "video_b": comparison[1]["video_b"]["video_id"],
             **criteria_score
         }
+        for public_username in public_usernames
         for comparison in serialized_comparisons
         for criteria_score in comparison[1]['criteria_scores']
     )
--- a/backend/tournesol/views/exports.py
+++ b/backend/tournesol/views/exports.py
@@ -3,7 +3,6 @@ import zipfile
 from io import StringIO
 
 from django.http import HttpResponse
-from backend.tournesol.models.ratings import ContributorRating
 from rest_framework.views import APIView
 from rest_framework.permissions import IsAuthenticated
 from drf_spectacular.utils import extend_schema, OpenApiTypes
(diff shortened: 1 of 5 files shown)


backend/tournesol/views/exports.py
```diff
@@ -35,6 +35,43 @@ def write_comparisons_file(request, write_target):
...
+    writer.writeheader()
+    public_data = ContributorRating.objects.filter(is_public=True).select_related("user", "video")
+    serialized_comparisons = []
+    public_videos = set()
+    users = set()
+    for public_video in public_data:
+        public_videos.add((public_video.user, public_video.video))
+        users.add(public_video.user)
+    for user in users:
+        comparisons = Comparison.objects.filter(user=user)\
```
lfaucon, line 54: This is going to be a performance bottleneck because it does a database query for each user (that means a few thousands of database queries right now).

```Python
comparisons = Comparison.objects.all().select_related("video_1", "video_2").prefetch_related("criteria_scores")

public_comparisons = [comparison for comparison in comparisons if (
    (comparison.user, comparison.video_1) in public_videos and (comparison.user, comparison.video_2) in public_videos
)]
```

lfaucon, line 54: This is going to be a performance bottleneck because it does a database query for each user (that means a few thousands of database queries right now).

```Python
comparisons = Comparison.objects.all().select_related("video_1", "video_2").prefetch_related("criteria_scores")

public_comparisons = [comparison for comparison in comparisons if (
    (comparison.user, comparison.video_1) in public_videos and (comparison.user, comparison.video_2) in public_videos
)]
```

backend/tournesol/views/exports.py
```diff
@@ -35,6 +35,43 @@ def write_comparisons_file(request, write_target):
...
+    fieldnames = ['public_username', 'video_a', 'video_b', 'criteria', 'weight', 'score']
+    writer = csv.DictWriter(write_target, fieldnames=fieldnames)
+    writer.writeheader()
+    public_data = ContributorRating.objects.filter(is_public=True).select_related("user", "video")
+    serialized_comparisons = []
+    public_videos = set()
+    users = set()
+    for public_video in public_data:
+        public_videos.add((public_video.user, public_video.video))
+        users.add(public_video.user)
```
lfaucon, line 52: May be simplified:

```Python
public_videos = set((rating.user, rating.video) for rating in public_data)
users = set(rating.user for rating in public_data)
```

backend/tournesol/views/exports.py
```diff
@@ -49,6 +86,18 @@ def get(self, request):
         return response
 
 
+class ExportPublicComparisonsView(APIView):
+    @extend_schema(
+        description="Download public data in .zip file",
```
lfaucon, line 91: I see this description is incorrect, as is the description in `ExportComparisonsView` so that's mnot on you. The file extension is csv.

lfaucon (changes requested):
Thanks for this contribution!

I requested some changes, let me know if you have questions or doubts.

Additionally, could you add a test for the new endpoint similar to the tests we have in `tournesol/tests/test_exports.py`?

Optionally, I find that we start having slight code duplications between the methods `write_comparisons_file` and `write_public_comparisons_file` and the views `ExportComparisonsView` and `ExportPublicComparisonsView`. If you easily find a way to avoid the code duplication, that would be great, but if it is too complicated, I don't mind keeping the code as it is now.

backend/tournesol/views/exports.py
YJ-Hannah: Should I change the description into `.csv` in `ExportComparisonsView` as well?

lfaucon: Yes please. That would be great

lfaucon (commented):
Thanks for addressing my comments.

@GresilleSiffle and @amatissart could one of you look at the code and approve it?

anikeith22:
I will be adding the frontend component of this feature soon.

frontend/src/pages/home/PublicDownloadSection.tsx
```diff
@@ -0,0 +1,97 @@
...
+            </Button>
+          ) : (
+            <Button
+              color="primary"
+              variant="contained"
+              onClick={prepareExport}
+            >
+              Click to Download
+            </Button>
+      )}
```
amatissart, line 92: As `/exports/comparisons/` is a public endpoint provided by the API, I think it would acceptable and significantly simpler to remove the "prepare" phase completely.  
Loading a blob in the browser is not necessary, a direct link to the .csv file would work:

```suggestion
      <Button
        component="a"
        href={`${api_url}/exports/comparisons/`}
        download="export.csv"
        color="primary"
        variant="contained"
      >
        Download
      </Button>
```

amatissart (commented):
Thanks for this PR! I think it's very close to being mergeable.  

I made a few suggestions to improve the performance of the export generation, and to simplify the integration of the download button on the home page.  

One of the unit test case is currently failing on the frontend due to how `notistack` is not currently configured to be used in `App.test.tsx`. But it should no longer be relevant after the comments are addressed, so you can probably ignore it.

However it would be nice to run the linter in the "frontend" directory. `yarn lint:fix` should do the trick and guide you about potential warnings. Please let me know if you need help to configure your local environment.

frontend/src/pages/home/PublicDownloadSection.tsx
anikeith22: fixed with the commit a935d31, need to fix linting errors

amatissart: Great, thanks!

backend/tournesol/views/exports.py
```diff
@@ -35,6 +35,39 @@ def write_comparisons_file(request, write_target):
...
+    writer.writerows(
+        {
+            "public_username": public_username,
+            "video_a": comparison["video_a"]["video_id"],
+            "video_b": comparison["video_b"]["video_id"],
+            **criteria_score
+        }
+        for public_username in public_usernames
+        for comparison in serialized_comparisons
+        for criteria_score in comparison['criteria_scores']
```
amatissart, line 67: A bug slipped through here. The nested loop is incorrect: each comparison would be duplicated and appear with all usernames, instead of being associated with the correct username. Perhaps the test case could be extended to cover this situation (with multiple usernames in the public dataset).

A possible solution:
```suggestion
        for (public_username, comparison) in zip(public_usernames, serialized_comparisons)
        for criteria_score in comparison['criteria_scores']
```

anikeith22: Resolved with commit e025b17

lfaucon (approved):
Awesome 😎

PR #362, opened 2021-12-13
[front] security: add a content security policy

GresilleSiffle:
**todo**
- [x] fix the YT videos not playing anymore
- [x] test on staging with the team to check if some pages are broken (use != browsers)
- [x] add a test ensuring the CSP header is present and configured 

---

This PR add a content security policy in the front end :rocket: 

Adding a [Content Security Policy][CSP] is a major security improvment as it allows the Tournesol front end to define a list of trusted sources from which resources can be loaded. To paraphrase Mozilla:

> The use of this header is the best method to prevent cross-site scripting (XSS) vulnerabilities. Due to the difficulty in retrofitting CSP into existing websites, CSP is mandatory for all new websites and is strongly recommended for all existing high-risk sites.

In our case, I disabled unsafe inline JavaScript. It should not break anything as our JavaScript sources are compiled and loaded from a file and not embedded in the HTML.

I also allowed HTTP requests to only the remote sources currently used:
- our API
- https://i.ytimg.com, the YouTube CDN providing the thumbnails of YT videos
- https://noembed.com, a third party service returning formatted data allowing to easily display YT videos
- https://www.youtube.com to make the videos playable

As a result, any subsequent remote sources will be blocked by the browser by default, and will require to be explicitly added to the CSP to be considered safe. The new variables `REACT_APP_CSP_CONNECT_SRC` and `REACT_APP_CSP_IMAGE_SRC` can contain as many sources as we want. 

[CSP]: https://infosec.mozilla.org/guidelines/web_security#content-security-policy


infra/ansible/inventory.yml
```diff
@@ -26,6 +26,10 @@ all:
           nvm_version: "v0.38.0"
           npm_lts_version: fermium
           frontend_scheme: http
+          frontend_csp_connect_src: https://noembed.com
+          frontend_csp_frame_src: https://www.youtube.com
+          frontend_csp_image_src: "https://i.ytimg.com https://www.paypal.com https://www.paypalobjects.com"
+          frontend_csp_script_src: "https://www.youtube.com/iframe_api https://www.youtube.com/s/player/"
```
amatissart, line 32: Perhaps it would be reasonable to simply set `https://www.youtube.com` here, for easier maintenance and to be more robust against potential changes in Youtube implementation?

amatissart (commented):
Looks good to me, on Firefox as well as Chromium.

infra/ansible/inventory.yml
GresilleSiffle: The YouTube domain is indeed more maintainable than a list of authorized URLs.

I was about to update the PR and I was discouraged to use directly `https://www.youtube.com` by a [CSP evaluator](https://csp-evaluator.withgoogle.com/) and this [particular slide](https://lbrunet.github.io/Plongee-dans-les-CSP/#32) of a  [CSP conference](https://lbrunet.github.io/Plongee-dans-les-CSP/#2). 

> /!\  www.youtube.com is known to host JSONP endpoints which allow to bypass this CSP.

See:

capture

I'm not familiar with JSONP, so I can't guarantee that `https://www.youtube.com/s/player/` doesn't also provide JSONP endpoints as `https://www.youtube.com` does.

I propose keeping the CSP as is for now, because:
- regarding the evaluator it might help mitigating the risk of JavaScript attacks
- if the YouTube ecosystem evolves, it will be easy for us to quickly update the CSP to update the authorized scripts
- if the YouTube ecosystem evolves too often, we can easily revert this decision of keeping the CSP as is

What do you think @amatissart?

amatissart: Ok. Thanks for looking into this!

amatissart (approved):

PR #365, opened 2021-12-15
dev-env: Initialize database with test data including sample accounts, videos and scores

### LINKED ISSUE #325: Create a better database for the dev-env
It would be good to have a "real" database made of public data (e.g. from Lê, Louis and me) for the dev-env.
Maybe also have the possibility to be connected as different user, for example to test some functions which required to have a lot of comparison on the account.

amatissart:
Fixes #325 

As described in [`README`](https://github.com/tournesol-app/tournesol/blob/dev-env-dataset/dev-env/README.md), 5 sample accounts are included, with usernames `user1`, `user2`,..., in addition to the superuser `user`.

This PR also contains adjustements in `dev-env/run-docker-compose.sh` to wait for all services to be ready, including frontend.


GresilleSiffle (approved):
It works :+1:

PR #366, opened 2021-12-15
fix(infra): Configure certbot service to authenticate via nginx

amatissart:
During the provisioning, TLS certificates are generated with `certbot certonly --standalone` and this "standalone" setting is persisted in the letsencrypt configuration.  

On renewal, the certbot.service should use existing "nginx" instance as the authenticator, instead of using a standalone server (which would fail as the port is not available). Hence this configuration that forces `certbot.service` to use the `--nginx` option.

### DIFF
--- a/infra/ansible/roles/nginx/tasks/main.yml
+++ b/infra/ansible/roles/nginx/tasks/main.yml
@@ -81,4 +81,12 @@
   notify:
     - Reload Nginx
 
+- name: Override certbot service to authenticate via nginx
+  template:
+    src: certbot.service
+    dest: /lib/systemd/system/certbot.service
+  notify:
+    # Make sure that systemd reloads any changes
+    - Reload Nginx
+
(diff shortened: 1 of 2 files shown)


GresilleSiffle (approved):
Perfect.

I don't know if a `certbot` packet upgrade will overwrite the service in `/lib/systemd/system`. Let's keep an eye on that.

amatissart:
> I don't know if a `certbot` packet upgrade will overwrite the service in `/lib/systemd/system`. Let's keep an eye on that.

Yes, it's possible. I am not too worried about this: the service file will be checked on each deploy, and if I'm not mistaken the `--nginx` option is actually necessary only for the first renewal (then, the settings are persisted in `/etc/letsencrypt`). 

And once we are confident that the settings are correct on all environments, we could consider update the provisioning rules to use the `nginx` plugin from the beginning, and remove this custom template.

PR #369, opened 2021-12-16
fix(frontend): Update CSP to allow inline script used by development environment

amatissart:
The development server uses inline script to implement hot reloading and handle compilation errors.

Currently, automatic reloading is failing with an error and the application is frozen:

>Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https://www.youtube.com/iframe_api https://www.youtube.com/s/player/".
...

It seems it's related with how `react-error-overlay`  injects an iframe to display potential errors:  
https://github.com/facebook/create-react-app/blob/9673858a3715287c40aef9e800c431c7d45c05a2/packages/react-error-overlay/src/index.js#L139-L154


GresilleSiffle:
Thanks for spotting this.

GresilleSiffle (approved):

PR #370, opened 2021-12-16
View count display

### LINKED ISSUE #288: (style) Display number of views as human readable
We currently show on our interface numbers like `21520041 views`. I recommend that we use a common library to display human-readable numbers so that it looks like `21.5M views` or `21,520,041 views` (like on Youtube)

glerzing:
I would prefer to implement solution 4 since dots are not conspicuous and need a bit of focus to be seen.
Capture d’écran de 2021-12-12 20-04-47
But then as we said if I don't know the browser's default localization, I can't tell in what format it will be rendered and what I should expect.
I found a regex that does the job for all the languages : /(1.?54.?988|١٥٤٬٩٨٨|۱۵۴٬۹۸۸) views/
It is complicated and uses non-ascii caracters, so /1.?54.?988 views/  is also pretty good except it doesn't work for arabic.

Closes #288

### DIFF
--- a/backend/scripts/dataset-import/.gitignore
+++ b/backend/scripts/dataset-import/.gitignore
@@ -1,3 +1,4 @@
 /metadata
 /public-dataset
 /dump.sql
+/venv-import-dataset
--- a/frontend/src/features/videos/VideoCard.tsx
+++ b/frontend/src/features/videos/VideoCard.tsx
@@ -210,7 +210,7 @@ function VideoCard({
         <div className={classes.youtube_complements}>
           {video.views && (
             <span className={classes.youtube_complements_p}>
-              {video.views.toLocaleString()} views
+              {video.views.toLocaleString('en-US')} views
             </span>
           )}
           {video.publication_date && (
--- a/frontend/src/features/videos/VideoCard.tsx
+++ b/frontend/src/features/videos/VideoCard.tsx
@@ -210,7 +210,7 @@ function VideoCard({
         <div className={classes.youtube_complements}>
           {video.views && (
             <span className={classes.youtube_complements_p}>
-              {video.views} views
+              {video.views.toLocaleString()} views
             </span>
           )}
           {video.publication_date && (
--- a/frontend/src/features/videos/VideoCard.spec.tsx
+++ b/frontend/src/features/videos/VideoCard.spec.tsx
@@ -20,9 +20,8 @@ describe('VideoCard content', () => {
     expect(screen.getByTestId('video-card-info')).toHaveTextContent(
       'Video title'
     );
-    // handle all the possible number notations of 154988 with a regex
     expect(screen.getByTestId('video-card-info')).toHaveTextContent(
-      /(1.?54.?988|١٥٤٬٩٨٨|۱۵۴٬۹۸۸) views/
+      '154,988 views'
     );
     expect(screen.getByTestId('video-card-ratings')).toHaveTextContent(
       '9 comparisons by 4 contributors'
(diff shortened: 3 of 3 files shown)


glerzing:
for the tests I used this script :

var languages = ["af", "sq", "ar-SA", "ar-IQ", "ar-EG", "ar-LY", "ar-DZ", "ar-MA", "ar-TN", "ar-OM",
 "ar-YE", "ar-SY", "ar-JO", "ar-LB", "ar-KW", "ar-AE", "ar-BH", "ar-QA", "eu", "bg",
 "be", "ca", "zh-TW", "zh-CN", "zh-HK", "zh-SG", "hr", "cs", "da", "nl", "nl-BE", "en",
 "en-US", "en-EG", "en-AU", "en-GB", "en-CA", "en-NZ", "en-IE", "en-ZA", "en-JM",
 "en-BZ", "en-TT", "et", "fo", "fa", "fi", "fr", "fr-BE", "fr-CA", "fr-CH", "fr-LU",
 "gd", "gd-IE", "de", "de-CH", "de-AT", "de-LU", "de-LI", "el", "he", "hi", "hu", 
 "is", "id", "it", "it-CH", "ja", "ko", "lv", "lt", "mk", "mt", "no", "pl",
 "pt-BR", "pt", "rm", "ro", "ro-MO", "ru", "ru-MI", "sz", "sr", "sk", "sl", "sb",
 "es", "es-AR", "es-GT", "es-CR", "es-PA", "es-DO", "es-MX", "es-VE", "es-CO", 
 "es-PE", "es-EC", "es-CL", "es-UY", "es-PY", "es-BO", "es-SV", "es-HN", "es-NI", 
 "es-PR", "sx", "sv", "sv-FI", "th", "ts", "tn", "tr", "uk", "ur", "ve", "vi", "xh",
 "ji", "zu"];
 
 var n = "154988";
 for (l of languages)
 {
 	var m = n.toLocaleString().match(/1.?54.?988|١٥٤٬٩٨٨|۱۵۴٬۹۸۸/);
 	if (m){console.log(l, "OK", n.toLocaleString());}
 	else{console.log(l, "!!! NOK !!!", n.toLocaleString());}
 }

backend/scripts/dataset-import/.gitignore
```diff
@@ -1,3 +1,4 @@
 /metadata
 /public-dataset
 /dump.sql
+/venv-import-dataset
```
amatissart, line 4: Side note: we recently merged a change related to the dev-env (#365), to include a broader database. 
Hopefully it will make future development easier, if you are using Docker to instantiate a local environment. Some data related to videos and scores are provided out-of-the-box.

amatissart (commented):
Thanks for having looked into this. As your test shows, `toLocaleString()` can lead to very different results. I would tend to keep it simple.

I agree the current font and color are not very suitable for displaying dots. We could also consider changing them for these element (currently in the class called `youtube_complements`). Not necessarily in this PR though.

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -210,7 +210,7 @@ function VideoCard({
         <div className={classes.youtube_complements}>
           {video.views && (
             <span className={classes.youtube_complements_p}>
-              {video.views} views
+              {video.views.toLocaleString()} views
```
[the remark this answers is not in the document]
glerzing: It will feel a bit weird for French users, but I guess there will be a French version of Tournesol soon enough. Agreed.

backend/scripts/dataset-import/.gitignore
glerzing: Ok, thanks, that may be very useful.

lfaucon:
Note that I renamed this Pull Request and mentioned the Issue number in the PR description instead.

glerzing:
I see no automatic squash and rebase option. What if I want to have a single, clean commit pulled into main ?

lfaucon (approved):

PR #374, opened 2021-12-19
Kleis logo changed for the horizontal version

jstainer-kleis:
before: 
image

after:
image


lfaucon (changes requested):
Much better indeed. Could you also change the height attribute on the image to be 64px like PolyConseil's logo? I originally gave Kleis Logo higher height because the logo was vertical

jstainer-kleis:
sure

lfaucon (approved):

PR #321, opened 2021-11-26
Front : Display video duration

sandre35:
Warning: this PR needs to be merged first : https://github.com/tournesol-app/tournesol/pull/318
Linked issue : https://github.com/tournesol-app/tournesol/issues/282
This PR tackles steps 3 and 4.

PR #376, opened 2021-12-20
[front] fix: warning about missing ref in ReactPlayer

amatissart:
After a wrapper has been introduced in #321, a warning in React dev tools passed unnoticed.

> Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?  
Check the render method of `ReactPlayer`.
...

Using [`React.forwardRef()`](https://reactjs.org/docs/react-api.html#reactforwardref), the wrapper ref can be forwarded to the inner `Box`.


lfaucon (approved):
Looks good

Issue #349, opened 2021-12-07
[extension] feat: Use frontent chosen languages for extension


lfaucon:
- [ ] Use the languages configuration choice persisted by the frontend
- [ ] Deprecate language choice from the extension
- [ ] If no frontend, resort to default navigator.languages

glerzing:
Do you think I should do something for the browser extension too ?
I considered adding parameters for the languages, here is how it could look like :
![Capture d’écran de 2022-01-02 11-36-08](https://user-images.githubusercontent.com/47578089/147892864-d7c2ed0c-35fe-4a47-ac75-eac5fa6c857f.png)

Another potential improvement would be in the youtube recommendations, refreshing only Tournesol recommandations when clicking in on the current language button, and not the whole page.

But I don't know if it's not more relevant for users to stick to their favorite language. What do you think ?

Besides, is it possible to load the Tournesol recommandations before Youtube recommandations? Because with a relatively slow internet, Tournesol recommandations can take a few seconds to appear, even if they are at the top. There can also be an in-between image : 
![Capture d’écran de 2022-01-03 01-15-26](https://user-images.githubusercontent.com/47578089/147893055-e780853b-8eca-4298-99bb-faf8b587e908.png)

lfaucon:
> But I don't know if it's not more relevant for users to stick to their favorite language. And it's weird that all the menus are in english if other languages are supported. Maybe allowing only one language, and updating the menus accordingly is wiser. What do you think ?

I see this as two separate things: 1. Multiple possible languages for video recommendations; 2. One chosen language for the interface. That is quite similar to how Youtube works now (the user selects one language for the interface, but the recommendations can cover multiple languages). As for our priorities, I think point 1 is slightly more valuable to us right now and easier to implement. @amatissart has started working on point 2. for the main website (https://github.com/tournesol-app/tournesol/issues/329), maybe we could tackle the extension at the same time.

> Another potential improvement would be in the youtube recommendations, refreshing only Tournesol recommandations when clicking in on the current language button, and not the whole page.

Agreed. That's very sloppy work that I'm guilty of. Created a ticket: https://github.com/tournesol-app/tournesol/issues/404

> Besides, is it possible to load the Tournesol recommandations before Youtube recommandations? Because with a relatively slow internet, Tournesol recommandations can take a few seconds to appear, even if they are at the top. There can also be an in-between image (~ 0,5 seconds) while the tournesol recommendations are loading

[...]

PR #377, opened 2021-12-20
Languages filter

glerzing:
Enable querying multiple languages to video API #349

- set the API to accept a list of languages separated by commas in a string (e.g. "lang1,lang2,lang3"). One advantage is that it stays compatible with the old format, it just extends it.
- set a backend test
- renamed language => languages

I could probably go a step further, and change the frontend button to be able to check multiple language in the filters in Recommandations. It was not obvious to me how to modify the Checkbox in ChoicesFilterSection.tsx to be able to check multiple languages. Should I work on it ?

### DIFF
--- a/backend/tournesol/tests/test_api_video.py
+++ b/backend/tournesol/tests/test_api_video.py
@@ -364,6 +364,29 @@ class VideoApi(TestCase):
         self.assertEqual(bad_response.status_code, status.HTTP_200_OK)
         self.assertEqual(bad_response.data["count"], len(self._list_of_videos))
 
+    def test_get_video_languages_filter(self):
+        client = APIClient()
+
+        # Add 1 video in French and 2 videos in English
+        video1 = Video.objects.create(language="fr", video_id="youtube1233")
+        video2 = Video.objects.create(language="en", video_id="youtube1238")
+        video3 = Video.objects.create(language="en", video_id="youtube1239")
+        VideoCriteriaScore.objects.create(video=video1, criteria="engaging", score=1)
+        VideoCriteriaScore.objects.create(video=video2, criteria="engaging", score=1)
+        VideoCriteriaScore.objects.create(video=video3, criteria="engaging", score=1)
+
+        resp = client.get("/video/?languages=")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        self.assertEqual(resp.data["count"], len(self._list_of_videos) + 3) # not filtered, 3 videos were added
+
+        resp = client.get("/video/?languages=fr")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        self.assertEqual(resp.data["count"], 1) # 1 French video
+
+        resp = client.get("/video/?languages=de,en,fr")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        self.assertEqual(resp.data["count"], 3) # 1 French and 2 English videos
+
     def test_get_video_date_filters(self):
         client = APIClient()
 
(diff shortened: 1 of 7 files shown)


backend/tournesol/views/video.py
```diff
@@ -110,9 +112,10 @@ def get_queryset(self):
                 queryset = queryset.filter(publication_date__gte=date_gte)
             except ValueError:
                 raise ValidationError('"date_gte" is an invalid datetime')
-        language = request.query_params.get('language') \
-            if request.query_params.get('language') else ""
-        queryset = queryset.filter(language=language) if language else queryset
+
+        languages = request.query_params.get('languages').split(",") \
+            if request.query_params.get('languages') else []
+        queryset = queryset.filter(language__in=languages) if languages else queryset
```
amatissart, line 118: Not introduced by this PR, but I think this could be made slightly more readable:
```suggestion
        language = request.query_params.get("language")
        if language:
            queryset = queryset.filter(language__in=language.split(","))
```

backend/tournesol/tests/test_api_video.py
```diff
@@ -364,6 +364,30 @@ def test_cannot_get_video_with_particular_request(self):
...
+        # not filtered, 3 videos were added
+        self.assertEqual(resp.data["count"], len(self._list_of_videos) + 3)
+
+        resp = client.get("/video/?languages=fr")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        self.assertEqual(resp.data["count"], 1) # 1 French video
+
+        resp = client.get("/video/?languages=de,en,fr")
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        self.assertEqual(resp.data["count"], 3) # 1 French and 2 English videos
```
amatissart, line 389: Great :+1:

backend/tournesol/views/video.py
```diff
@@ -32,7 +32,9 @@
         description="Retrieve a list of recommended videos, sorted by decreasing total score.",
         parameters=[
             OpenApiParameter("search"),
-            OpenApiParameter("language"),
+            OpenApiParameter("languages",
+                             description="Accepted languages separated by commas "
+                                         "(e.g. 'en,fr,de'). If empty, accept all languages"),
```
amatissart, line 37: _Side note about OpenAPI and Swagger_ (you can ignore it :stuck_out_tongue: ):

This parameter is currently typed as single string. 
image
(Visible on http://localhost:8000/docs/#/video/video_list )


Ideally, we would like to describe that this parameter accepts multiple values using the options of `OpenAPIParameter`.  
However, this comma-separated representation is not the default one [supported by the OpenAPI standard](https://swagger.io/docs/specification/serialization/#query), so we would need to set `explode=False` which is [not currently supported](https://github.com/ferdikoomen/openapi-typescript-codegen/issues/654) by the library we are using to generate the client code used by the frontend.  

So once again, let's keep it simple. Your current description is perfect for what we want to achieve. We'll just need to handle this comma-separated format ourselves in the frontend code.

amatissart (commented):
>I could probably go a step further, and change the frontend button to be able to check multiple language in the filters in Recommandations. It was not obvious to me how to modify the Checkbox in ChoicesFilterSection.tsx to be able to check multiple languages. Should I work on it ?

Yes, it would be nice to integrate this feature on the frontend. Here or in another PR.
As you noticed, the current filter accepts only one value at once. The `Checkbox` component could be used for this purpose. But they are currently controlled to be `checked` only if the value found in the URL matches one of the choices:
https://github.com/tournesol-app/tournesol/blob/c8923b02b61d7d8e496ab10f7e4b9cee36b42975/frontend/src/components/filter/ChoicesFilterSection.tsx#L26-L38

So we need to extend this `ChoicesFilterSection` component to handle multiple values, as an option (for example, the `DateFilter` uses choices that are exclusive from one another, and should still accept a single value).

We can discuss this together in the coming days if necessary.

lfaucon (commented):
Awesome work. Thanks @glerzing 

The next step is indeed to update the frontend to allow multiple languages, but I would recommend (quite strongly) doing it in another PR after completing the backend.

As a third step, we may also add the multi language option to the extension :)

glerzing:
As suggested, I'll make the rest in another PR

amatissart (approved):

PR #380, opened 2021-12-21
[back] chore: remove obsolete fields and properties related to private/public contributions

amatissart:
The `Video` model used several computed fields to keep track of the number of private and public contributions. Their implementation was obsolete and their potential purpose is not currently specified. So this PR suggests to remove them, to avoid any confusion.

This also includes some code cleanup, related to how these properties were computed.


backend/tournesol/models/comparisons.py
```diff
@@ -196,18 +190,6 @@ def sample_video(username, only_rated=False):
...
         return video
 
-    def weights_vector(self):
-        """How important are the individual scores, according to the rater?"""
-        return self._features_as_vector_fcn(suffix="_weight")
-
-    def save(self, *args, **kwargs):
-        """Save the object data."""
-        if not kwargs.pop("ignore_lastedit", False):
-            self.datetime_lastedit = timezone.now()
```
amatissart, line 206: This is no longer necessary, since `auto_now` has been set on the field `datetime_lastedit` in https://github.com/tournesol-app/tournesol-backend/pull/55

backend/core/utils/constants.py
```diff
@@ -2,35 +2,6 @@
...
-    "reliability": "#6573c3",
-    "pedagogy": "#8561c5",
-    "importance": "#ff784e",
-    "engaging": "#ffc107",
-    "backfire_risk": "#e91e63",
-    "layman_friendly": "#50f22c",
-    "diversity_inclusion": "#2ceff2",
-    "better_habits": "#c6eb34",
-    "entertaining_relaxing": "#6ba4ff",
-}
```
lfaucon, line 32: Unrelated to the PR, quick thought: 

If we want multiple frontends to share the same constants, would a good manner to achieve this be to serve the constants from the backend API? 

It may be a good design for Tournesol-As-A-Service, if we want different frontends from the same code being configured differently simply by connecting to a different backend.

backend/tournesol/migrations/0015_auto_20211221_1114.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 3.2.10 on 2021-12-21 11:14
```
lfaucon, line 1: I ususally try to give more explicit names to automatically generated migrations (it makes it slightly easier to work with afterwards).

lfaucon (approved):
Well needed clean-up 👍

backend/core/utils/constants.py
amatissart: Yes, fetching such context from the API would make sense.   
It could probably include more than just constants though: names/icons of the criteria, whether they are optional or not, available filters, etc. Why not even configurable via some kind of back-office. Some discussion would be needed to find the right level of abstraction here.

lfaucon (approved):

PR #384, opened 2021-12-22
[backend] fix: ignore nan when saving criteria scores after ml_train

amatissart:
Storing `nan` values in the database causes serialization errors in the API when `criteria_scores` are included in the response.  
We could also handle this case in the serializers, but it seems preferable to me to avoid including invalid scores in the database during the ml_train job.

The root cause is unclear. The issue has been observed on staging, after a user submitted comparisons where every criteria score has value `-10`. Are the computed local scores undefined when the criteria scores are constant?

### DIFF
--- a/backend/ml/management/commands/ml_train.py
+++ b/backend/ml/management/commands/ml_train.py
@@ -1,4 +1,5 @@
 import logging
+import numpy as np
 
 from tournesol.models import (
     ComparisonCriteriaScore,
@@ -88,6 +89,7 @@ def save_data(video_scores, contributor_rating_scores):
                 uncertainty=uncertainty,
             )
             for video_id, criteria, score, uncertainty in video_scores
+            if not np.isnan(score)
         ]
     )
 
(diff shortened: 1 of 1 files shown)


OscarVlld:
I see no reason why there should be NaNs when the criteria scores are constant. What's more, a NaN during the training is likely to "spread", which is concerning. I think not saving them cannot be a long-term fix.

amatissart:
> I see no reason why there should be NaNs when the criteria scores are constant. What's more, a NaN during the training is likely to "spread", which is concerning. I think not saving them cannot be a long-term fix.

Yes, this bug may reveal a deeper issue with the current implementation.   

On a closer look, a similar case exists in production data. For only a single user, on a single criteria, all the `ContributorRatingCriteriaScore` scores are `nan`, although the comparison criteria scores are defined correctly between -10 and 10. Weird.

lfaucon:
Completely agree with @OscarVlld 

@amatissart Would it be easy to isolate and save the problematic input somewhere maybe for a future unit test and helping investigate and fix the issue?

lfaucon:
Closing because this is stale and will not be necessary as we are soon refactoring the ML algorithm

PR #385, opened 2021-12-23
[front] Set up i18n and define translation keys for frontend content

amatissart:
This PR integrates [react-i18next](https://react.i18next.com/) as the internationalization framework used to load and render translated messages in the application.

### Language detection and supported languages

The language detector provided by i18next is configured to use `localStorage` and `navigator` in this order.

This means that if no language has been saved in localStorage on a previous visit, the selected language will be derived from the accepted languages configured on the browser, and selected among the list of supported languages set in the [i18n configuration](https://github.com/tournesol-app/tournesol/blob/react-i18n/frontend/src/i18n.ts) 

The translations related to the selected language will be fetched by a HTTP request during the app initialization.

English (`en`) is currently the only supported language. So no additional translation is required at this point to deploy this branch on production.

### Usage in frontend code
The `useTranslation()` hook is used to integrate translated content in React components.

2 main use-cases:

1. For basic text, without interpolation or plurals, use the `t` function:
```tsx
return (
  <div>
    {t('components.translationKey')}
  </div>
);
```

2. For advanced cases, use the [`Trans` component](https://react.i18next.com/latest/trans-component):
```tsx
return (
  <Trans t={t} i18nKey="account.userMessagesUnread" count={count}>
      Hello <strong>{{name}}</strong>, you have {{count}} unread message
  </Trans>
);
```

When the `count` prop is used, multiple translation keys will be generated to handle plurals.

### Update translations

[`i18next-parser`](https://github.com/i18next/i18next-parser) is used to parse the code and extract the translation keys automatically.

```
yarn run i18n:parse
```

The catalogs present in public/locales will be updated with the new keys found in code, and obsolete keys will be removed.

Then, translators can complete and commit the translations directly in the .json files, or possibly via a translation management platform (to be defined).

### Potential workflow when new features are introduced

1. Implement feature in frontend code, with new translation keys
2. Update catalogs with `yarn run i18n:parse`
3. Complete translations in all supported languages
4. Any PR is merged after all translations are present and approved.

Note that this workflow may not be manageable after more languages are supported, and external translators need to be involved.


GresilleSiffle (approved):
Excellent!

Can't wait to see the application translated into other langages :rocket:

PR #386, opened 2021-12-26
Languages filter frontend

glerzing:
part 2/3 of issue #349 (backend, **frontend**, extension)

If you lack time, don't hesitate to assign it to another reviewer.

### Modifications
- added an option in the component ChoicesFilterSection to enable checkboxes that can have multiple options checked at once
- added tests that cover the language filters and date filters rendering

I could also make a test for the criteria filters. But I don't know if it is really useful. It may just be inconvenient if someone ever wants to modify it. And currently the default URL parameters can be unexpectedly displayed in the URL (remark 1 below).

### (unrelated) minor remarks : 
1 - the criteria parameters like "reliability=50" are not removed from the URL, even when they are set at their default value (50).
2 - weird but probably normal : Could not verify the attribute "checked" of the checkboxes, as it doesn't appear in the DOM. I think it is updated in the React component, but never updated in the DOM (in HTML, it is just a default value).
3 - 1 warning in VideoCard.spec.tsx ([warning.txt](https://github.com/tournesol-app/tournesol/files/7776428/warning.txt)). Probably nothing important, but a bit unsettling.

### DIFF
--- a/frontend/src/components/ChoicesFilterSection.tsx
+++ b/frontend/src/components/ChoicesFilterSection.tsx
@@ -1,14 +1,19 @@
 import React from 'react';
 
-import { Typography, FormControlLabel, Checkbox, Box } from '@material-ui/core';
-import { CheckCircle, CheckCircleOutline } from '@material-ui/icons';
 import { TitledSection } from 'src/components';
+import { Typography, FormControlLabel, Checkbox, Box } from '@material-ui/core';
+import {
+  CheckCircle,
+  CheckCircleOutline,
+  CheckBox,
+  CheckBoxOutlineBlank,
+} from '@material-ui/icons';
 
 interface Props {
   title: string;
   value: string;
   choices: Record<string, string>;
-  exclusiveChoice: boolean;
+  multipleChoice: boolean;
   onChange: (value: string) => void;
 }
 
@@ -16,18 +21,11 @@ const ChoicesFilterSection = ({
   title,
   value,
   choices,
-  exclusiveChoice,
+  multipleChoice,
   onChange,
 }: Props) => {
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    if (exclusiveChoice) {
-      // Only one value allowed
-      if (event.target.checked) {
-        onChange(event.target.name);
-      } else {
-        onChange('');
-      }
-    } else {
+    if (multipleChoice) {
       // Multiple values allowed
       let all_values = value ? value.split(',') : [];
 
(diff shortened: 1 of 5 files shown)


frontend/src/components/ChoicesFilterSection.tsx
```diff
@@ -8,23 +8,48 @@ interface Props {
...
   };
 
   return (
     <TitledSection title={title}>
       <Box display="flex" flexDirection="column">
         {Object.entries(choices).map(([choiceValue, choiceLabel]) => {
-          const checked = value === choiceValue;
+          const checked = exclusiveChoice
+            ? value === choiceValue
+            : value != '' && value.split(',').includes(choiceValue);
```
glerzing, line 51: I could actually remove 'value != '' && '

frontend/src/features/recommendation/SearchFilter.spec.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { render, screen, fireEvent } from '@testing-library/react';
+import { Router } from 'react-router-dom';
+import { createMemoryHistory } from 'history';
+
+import SearchFilter from './SearchFilter';
+import { dateChoices } from './DateFilter';
+import { languageChoices } from './LanguageFilter';
+import { mainCriteriaNames } from 'src/utils/constants';
+
+describe('Filters feature', () => {
```
glerzing, line 13: tests durations :
  Filters feature
    ✓ Can open and close the filters menu (328 ms)
    ✓ Check that all the filter checkboxes and sliders are present (197 ms)
    ✓ Can only check one date filter at once (409 ms)
    ✓ Can check multiple language filters at once (290 ms)

frontend/src/components/ChoicesFilterSection.tsx
glerzing: I'll fix it in the same commit as other incoming remarks

lfaucon: The split would be computed once per object. Could you instead have sooner in the component `const valuesList = exclusiveChoice ? [value] : value.split(',')` and use this instead here?

frontend/src/components/ChoicesFilterSection.tsx
```diff
@@ -35,6 +60,7 @@ const ChoicesFilterSection = ({ title, value, choices, onChange }: Props) => {
                   onChange={handleChange}
                   name={choiceValue}
                   size="small"
+                  data-testid={title + ': ' + choiceLabel}
```
lfaucon, line 63: Commenting about line 57: `icon={<CheckCircleOutline />}`. Could you change the icon to a square when it is possible to select multiple? (I believe that is the usual UI for this type of multiple choice form)

lfaucon (changes requested):
Nice work!

frontend/src/components/ChoicesFilterSection.tsx
```diff
@@ -8,23 +8,48 @@ interface Props {
...
+  exclusiveChoice: boolean;
   onChange: (value: string) => void;
 }
 
-const ChoicesFilterSection = ({ title, value, choices, onChange }: Props) => {
+const ChoicesFilterSection = ({
+  title,
+  value,
+  choices,
+  exclusiveChoice,
```
lfaucon, line 19: I find it slightly more natural to have the opposite property `multipleChoice` but it is probably just an opinion.

glerzing: I think it is more intuitive too, fixed

lfaucon (approved):

frontend/src/components/ChoicesFilterSection.tsx
```diff
@@ -1,40 +1,77 @@
...
+  multipleChoice: boolean;
   onChange: (value: string) => void;
 }
 
-const ChoicesFilterSection = ({ title, value, choices, onChange }: Props) => {
+const ChoicesFilterSection = ({
+  title,
+  value,
+  choices,
+  multipleChoice,
```
amatissart, line 24: Very minor: a default value for `multipleChoice` may improve the usability of this component.  
```suggestion
  multipleChoice = false,
```

amatissart (approved):

PR #392, opened 2021-12-29
feat(backend): Adds caching to ExportPublicComparisonsView

### LINKED ISSUE #383: Hourly caching for public dataset download
Add caching the the public dataset download (it take ~5sec to return a result and overloads the server if asked 20 times in a minute)

Another improvement may be to temporarily disable the public download button when it is clicked.

lfaucon:
Closes #383

### DIFF
--- a/backend/tournesol/views/exports.py
+++ b/backend/tournesol/views/exports.py
@@ -3,6 +3,8 @@ import zipfile
 from io import StringIO
 
 from django.http import HttpResponse
+from django.utils.decorators import method_decorator
+from django.views.decorators.cache import cache_page
 from rest_framework.views import APIView
 from rest_framework.permissions import IsAuthenticated, AllowAny
 from drf_spectacular.utils import extend_schema, OpenApiTypes
@@ -84,6 +86,7 @@ class ExportComparisonsView(APIView):
 class ExportPublicComparisonsView(APIView):
     permission_classes = [AllowAny]
 
+    @method_decorator(cache_page(60*10))  # 10 minutes cache
     @extend_schema(
         description="Download public data in .csv file",
         responses={200: OpenApiTypes.BINARY}
--- a/backend/README.md
+++ b/backend/README.md
@@ -13,14 +13,13 @@ Django app for tournesol app.
 - Create a python env and install the requirements
   `pip install -r requirements.txt`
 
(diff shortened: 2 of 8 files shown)


amatissart (commented):
The cache definition looks good to me. 

_Important note though_: with the current configuration, we are still using the [local memory cache](https://docs.djangoproject.com/en/3.2/topics/cache/#local-memory-caching) used by default in Django, which is not really suitable for production. Its behavior has several drawbacks:
 * the cache storage is "per process". Unless successive requests are handled by the same process, the cached value that has just been computed may not be used.
 * as a consequence, different versions of the same file may be stored and served at a given moment, which may be confusing for users.
 * the cache storage is kept in RAM, and the memory being used is potentially multiplied by the number of processes.
 
 (Our deployments are currently running [4 Django processes, 1 per gunicorn worker](https://github.com/tournesol-app/tournesol/blob/0db7cd321b66c775b975cb2df54749fbc0d22459/infra/ansible/roles/django/templates/gunicorn.service.j2#L19))
 
 
So even if this configuration can be used in the very short term, I think that we could set up a proper cache with little additional effort. In particular, the [`DatabaseCache`](https://docs.djangoproject.com/en/3.2/topics/cache/#database-caching) wouldn't require any new component in the infrastructure, and could fit our needs well enough.

backend/tournesol/views/exports.py
```diff
@@ -84,6 +86,7 @@ def get(self, request):
 class ExportPublicComparisonsView(APIView):
     permission_classes = [AllowAny]
 
+    @method_decorator(cache_page(60*10))  # 10 minutes cache
```
amatissart, line 89: Not a problem for now, but to keep in mind:  
`cache_page` includes the current language code in the cache key. So we may need to find a workaround after multiple languages are enabled in the API (to get localized error messages, etc.) as it would be unexpected for this route to return different responses depending on the requested language.

amatissart (approved):

backend/tournesol/views/exports.py
lfaucon: I'm not so worried, this endpoint is supposed to return different results when time goes on. Any difference between languages would disappear in 10 minutes. I don't think that this will ever be a source of real problems.

PR #394, opened 2021-12-29
[backend] chore: enforce email uniqueness at the db level

### LINKED ISSUE #232: Disallow creating multiple accounts with the same email

amatissart:
Closes #232 

In addition to the update in `User` model, this PR includes a migration to guarantee that no duplicated address persists in the database before the uniqueness constraint is applied. As agreed, only the account with the highest number of comparisons will keep its email address, and the others will no longer be associated with a valid address.

After this change, the uniqueness constraint on `email` will be enforced by a database index. Blank values for `email` are no longer allowed, hence the adjustments in tests.

### DIFF
--- a/backend/core/migrations/0005_email_unique_in_db.py
+++ b/backend/core/migrations/0005_email_unique_in_db.py
@@ -1,4 +1,5 @@
 import logging
+
 from django.db import migrations, models
 from django.db.models import Count, Window, F
 from django.db.models.functions.window import RowNumber
--- a/backend/core/migrations/0005_email_unique_in_db.py
+++ b/backend/core/migrations/0005_email_unique_in_db.py
@@ -1,0 +1,62 @@
+import logging
+from django.db import migrations, models
+from django.db.models import Count, Window, F
+from django.db.models.functions.window import RowNumber
+
+
+def remove_duplicated_email(apps, schema_editor):
+    """
+    Find users with duplicated emails, and update email of the accounts associated
+    with the fewer comparisons (to preserve only 1 per original address).
+    Their 'email' value will be replaced with an invalid address derived
+    from the username to guarantee uniqueness in the database.
+    """
+    User = apps.get_model("core", "User")
+
+    duplicated_emails = (
+        User.objects.values("email").alias(count=Count("id")).filter(count__gte=2)
+    )
+    users_to_update = [
+        user
+        for user in User.objects.filter(email__in=duplicated_emails).annotate(
+            n_comparisons=Count("comparisons"),
+            rank_comparisons=Window(
+                expression=RowNumber(),
+                partition_by=[F("email")],
+                order_by=F("n_comparisons").desc(),
+            ),
+        )
+        if user.rank_comparisons > 1
+    ]
+
+    for u in users_to_update:
+        new_email = f"{u.username}@invalid"
+        logging.info(
+            'Updating email for user "%s", from "%s" to "%s"',
+            u.username,
+            u.email,
+            new_email,
+        )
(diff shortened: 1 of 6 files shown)


backend/core/migrations/0005_email_unique_in_db.py
```diff
@@ -0,0 +1,62 @@
...
+    ]
+
+    for u in users_to_update:
+        new_email = f"{u.username}@invalid"
+        logging.info(
+            'Updating email for user "%s", from "%s" to "%s"',
+            u.username,
+            u.email,
+            new_email,
+        )
```
sandre35, line 39: I would consider using a `logger.warning`, because the fact of creating two accounts with the same email is not recommended globally on websites.

backend/tournesol/tests/test_api_comparison.py
```diff
@@ -57,9 +57,9 @@ def setUp(self):
 
         At least 4 videos and 2 users with 2 comparisons each are required.
         """
-        user = User.objects.create(username=self._user)
-        user2 = User.objects.create(username=self._user2)
-        other = User.objects.create(username=self._other)
+        user = User.objects.create(username=self._user, email="user@test")
+        user2 = User.objects.create(username=self._user2, email="user2@test")
+        other = User.objects.create(username=self._other, email="other@test")
```
sandre35, line 62: With the adaptation of these fields in our tests, we should maybe consider using factories ?
This issue could be handled in another PR if you want.

GresilleSiffle: Hi @sandre35 

It's a good idea, it will make the tests development easier. It's indeed a feature that would be great in another pull request, as it will contains a refactor of several tests not related to the current PR.

Do you have any experience with test fixtures? I know there are several way to implement them:
- using the built-in fixtures provided by Django
- using a third-party module like [factory_boy](https://factoryboy.readthedocs.io/en/stable/index.html)

I only have used Django fixtures in my previous projects, and I liked the possibility to create and edit the fixtures directly in the Django admin interface, and dump them in JSON files afterward. I never used `factory_boy` or any similar tool (`Faker`?), but they seem good option to consider.

amatissart: I used "factory_boy" in a previous project, its integration with Django models to define factories is very handy:
https://factoryboy.readthedocs.io/en/stable/orms.html#django

Such factories are especially useful to focus on the fields that are relevant to the current test. So they are probably easier to maintain, compared to raw fixtures.

lfaucon: Because there seem to be an unanimous aggreement, I made the following ticket: https://github.com/tournesol-app/tournesol/issues/396. Any of you can feel free to take it.

backend/core/migrations/0005_email_unique_in_db.py
```diff
@@ -0,0 +1,62 @@
+import logging
+from django.db import migrations, models
```
GresilleSiffle, line 2: Very minor coding style suggestion.

As recommended by [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) and the Django [coding style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports), we should use a blank line between imports related to the Python standard library and the `django` module.

```suggestion

from django.db import migrations, models
```

GresilleSiffle (changes requested):
It works!

GresilleSiffle (approved):

PR #397, opened 2022-01-03
[review branch] to merge into `cache_public_dataset`

GresilleSiffle:
**important** this PR targets the feature branch `cache_public_dataset`, and must not be merged into `main`

- related to pull request #392 
- related to issue #383

---

As suggested by @amatissart , the Django command `createcachetable` was missing in our deployment routines. It is now automatically run by the development environment each time the API container is started. It is also run for each deployment made by Ansible.

### DIFF
--- a/infra/ansible/roles/django/tasks/main.yml
+++ b/infra/ansible/roles/django/tasks/main.yml
@@ -85,6 +85,7 @@
   become_user: gunicorn
   notify:
     - Migrate Django database
+    - Create database cache
     - Collect Django static assets
     - Restart Gunicorn
     - Notify backend upgrade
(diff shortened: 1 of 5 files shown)


infra/ansible/roles/django/tasks/main.yml
```diff
@@ -85,6 +85,7 @@
   become_user: gunicorn
   notify:
     - Migrate Django database
+    - Create database cache
```
amatissart, line 88: Similarly, I think that this handler should also be called in "infra/ansible/roles/restore/tasks/main.yml" after restore tasks.

GresilleSiffle: Thanks, fixed by aa79456.

Now the handler `Create database cache` is notified each time `Migrate Django database` is notified.

amatissart (approved):

lfaucon (approved):
Completely missed this! Good to have figured it out 😊

Issue #211, opened 2021-09-29
Throttling on sensitive requests


lfaucon:

lfaucon:
@GresilleSiffle may recommend to use Fail2Ban https://www.fail2ban.org/wiki/index.php/Main_Page

lfaucon:
We may limit to 1000 request per minute

100 request per minute seems to be an intense user behaviour (note that submitting a comparison generates about 10 requests)

amatissart:
Fail2Ban may be useful to protect the servers against attacks outside of the application itself (eg. ssh, etc.).  
To define limits on specific requests or actions in the API, I think it would be much easier to use the throttling mechanisms provided by Django Rest Framework: https://www.django-rest-framework.org/api-guide/throttling/#throttling

GresilleSiffle:
That's true, **DRF Throttling** and **fail2ban** are two different things.

The DRF Throttling will help to limit the number of requests a client can make. If I understood correctly, once the limit is reached, the clients get HTTP errors. It works at the application level, and reduce the load without banning anyone. For me, it's the first level load smoothing we should use in the application.

fail2ban is able to any log file, ssh, auth, or even http, and determines, according to a configured threshold, if successive failures must be transofrmed into an IP ban. It works at a system level. It's goal is to limit intrusion, DDoS, by banning IP addresses for a specific duration. In my opinion, fail2ban is also a first level of protection we should use in the system.

We could create two different tasks to adress those different issues.

last note: I think with fail2ban we can create specific rules ( jails ) that scan for successive login attempts through the login form, and ban the IP addresses that don't act like humans. We can suppose a human would have used the `Forgot your password?` page after a certain amount of attempt, or won't have filled a hidden form field in the HTML.

GresilleSiffle:
About DRF Throttling.

If we have different endpoints with different "cost", we can use different throttling configuration.

For instance, downloading the public dataset is an heavy operation, and could have its own throttling scope.

```python
class PublicDatasetThrottle(AnonRateThrottle):
    scope = 'public_dataset'
```

and in the project `SETTINGS.py`

```python
REST_FRAMEWORK = {
    'DEFAULT_THROTTLE_CLASSES': [
        'tournesol.throttles.PublicDatasetThrottle',
        # ...
    ],
    'DEFAULT_THROTTLE_RATES': {
        'public_dataset': '10/min',
        # ...
    }
}
```

Endpoints making calls to the YouTube API could have they own scope too.

[...]

Issue #193, opened 2021-08-28
Protect Video Creation endpoint


lfaucon:
**Context:**
The video creation endpoint is open to everyone without authentication. This endpoint uses our YoutubeAPI key to fetch video metadata and has a limit of 10k requests per day. Unfortunately, in its current state, it could easily be abused by someone who would consume our 

**Possible remediations:**
- Add a severe rate limit, for example 50 / hour / user (EDITED)
- Require authentication on the endpoint
- Moves the metadata API to be triggered only for videos in comparisons or user's rate later list

**Acceptance criteria:**
- The above solutions and possibly additional solutions have been discussed with the team. 
- The solution(s) agreed upon by the team has(ve) been implemented.

PR #405, opened 2022-01-04
[back][front] feat: define an HTTP requests throttling policy

GresilleSiffle:
**related to** #211 which is a sub-task of #193

---

The objective of this pull request is to define a requests throttling policy in the back end, using the built-in mechanism of [Django REST framework](https://www.django-rest-framework.org/api-guide/throttling/).

This policy must allow a comfortable anonymous or authenticated usage of the Tournesol application. Note that not all requests have the same "cost", some of them take more time to process like the public dataset construction and some of them use the YouTube API which allows a limited number of queries for a given period. 

### back end changes

**(1)**

- [x] propose a default global throttle rate, used everywhere

I defined two rates, one per minute and one per hour 
- **burst rate** - 120 per minute per user
- **sustained rate** - 3600 per hour per user

Currently both anonymous and authenticated users have the same rates, but they can be easily configured differently.

**(2)**

- [x] propose a specific throttle rate for the download of the public dataset

I set the rate to 10 download maximum per minute per user. We can further increase this rate now we have the downloaded result in cache.

Nevertheless I don't think there are use cases where a user would like to deliberately download the public dataset 10 times in a minute. It also acts a safeguard if the back end's cache is disabled for any reason. 

**(3)**

- [x] propose a specific throttle rate for the download of personnal data

I used the same rate as the public dataset download.

Even if this operation is super light compared to the download of the public dataset, having a rate here will allows us to add more data to the personal .zip file without worrying about how heavy the operation is.

**(4)**

- [x] propose a specific throttle rate for endpoints calling the YouTube API

I kept the rate defined in the issue #193, 50 per minute per user at the time of writing.

**(5)**

- [ ] get feedback from the team 

### front end changes

I had to refactor a bit the `ExportAllDataForm` in the front end. Originally when the maximum number of requests per minute was reached, the HTTP 429 error returned by the API was considered as a successful fulfilled Promise, which is OK, but no information was displayed in the UI to inform the user why its download hasn't been successful.

I also updated the `VideoSelector` component handle properly HTTP 429 too. It now doesn't create a contributor rating if the video has not been added to the database for any reason.

### future implications

Currently the front end pages display an empty content when the current user reach the maximum number of requests allowed. The UI could be improved to display a message or an alert informing the user to try again later.

### suggestions

In a following pull request, I would like to propose to define specific rates for the log in and the sign up forms.

The log in form can be the target of attacks aiming to steal users credentials and data. On the other hand using the sign up form 120 times in a minute looks more than an attack rather than normal accounts creation.

In a third pull request, I also suggest to install and configure `fail2ban` on our servers, which will allow us to ban IP addresses that continue targeting specific endpoints at an alarming rate even if they received an invitation to retry later with the response HTTP 429.


backend/tournesol/views/video.py
```diff
@@ -69,6 +74,14 @@ class VideoViewSet(mixins.CreateModelMixin,
     queryset = Video.objects.all()
     pagination_class = LimitOffsetPagination
     permission_classes = [IsAuthenticatedOrReadOnly]
+
+    throttle_classes = [
+        PostScopeRateThrottle,
+        BurstUserRateThrottle,
+        SustainedUserRateThrottle
+    ]
```
amatissart, line 82: So here so rate limit is applied for anonymous users on `GET /video/`. Is it deliberate?

GresilleSiffle: Fixed.

Issue #287, opened 2021-11-08
[front] duplicated code in forms possibly factorizable


GresilleSiffle:
The way we handle error and success messages in the components `ProfileForm` and `PasswordForm` is the same. It seems to be generic enough to be factorized in a `utils.openapi.response.js` file or similar.

We could create two functions:
- `displayErrorsFromResponse(response)` or `displayErrorMessages(response)`
- `displaySuccessMessage(response)`

It will provide an uniform way for all forms to handle messages, make the code smaller and less bug prone.

See the code at:

https://github.com/tournesol-app/tournesol/blob/f257db0ef54998405e18f053bd7100c686fe410e/frontend/src/features/settings/profile/ProfileForm.tsx#L49

and:

https://github.com/tournesol-app/tournesol/blob/f257db0ef54998405e18f053bd7100c686fe410e/frontend/src/features/settings/account/PasswordForm.tsx#L37

```javascript
        if (reason && 'body' in reason) {
          const newErrorMessages = Object.values(reason['body']).flat();
          newErrorMessages.map((msg) => showErrorAlert(enqueueSnackbar, msg));
        } else {
          contactAdministrator(
            enqueueSnackbar,
            'error',
            'Sorry, an error has occurred, cannot update password.'
          );
        }
```

PR #409, opened 2022-01-05
[front] refactor: simplify API responses handling in several forms

GresilleSiffle:
**related to:** #211 

**will fix the tech' debt:** #287 

**to merge before pull request:** #405 

---

In the front end not all forms handle the errors returned by the API in the same way, and it forces me to copy several times the same lines to handle properly the HTTP 429 Too Many Requests status in the PR #405.

This merge request should help by reducing the quantity of duplicated code, and will make the development of the previously mentioned PR a lot easier.

Fortunately we already have a rather generic piece of code able to treat the majority of forms as equal, so I decided to factorize it in a `src/utils/api/responses.ts` function.

This function:
- is documented
- uses `notistack`
- automatically handles `HTTP 429` with a nice " retry after X seconds "
- automatically handles all other HTTP status code
- displays a generic message when the API fails to deliver a standard ApiError object
- its behaviour can be easily overriden

For the moment, only `PasswordForm`, `RateLater` and `ProfileForm` have been updated to use this new function instead of using a custom code.

I also refactored `Login` to use `notistack` instead of `Alert` to keep it harmonized with the rest of the application. The only remaining component using `Alert` is now `Signup`.

Here is an example, in the PasswordForm component:

**before**

```typescript
      .catch((reason: { body: { [k: string]: string[] } }) => {
          const newErrorMessages = Object.values(reason['body']).flat();
          newErrorMessages.map((msg) => showErrorAlert(enqueueSnackbar, msg));
        } else {
          contactAdministrator(
            enqueueSnackbar,
            'error',
            'Sorry, an error has occurred, cannot update password.'
          );
        }
```

**after**

```typescript
      .catch((reason: ApiError) => {
        displayErrors(
          enqueueSnackbar,
          reason,
          'Sorry, an error has occurred, cannot update password.'
        );
      }
```

### DIFF
--- a/frontend/src/utils/api/response.ts
+++ b/frontend/src/utils/api/response.ts
@@ -1,0 +1,75 @@
+import { ProviderContext, VariantType } from 'notistack';
+import { ApiError } from 'src/services/openapi';
+import {
+  contactAdministrator,
+  showErrorAlert,
+  showInfoAlert,
+} from 'src/utils/notifications';
+
+/**
+ * Display all errors contained in an `ApiError` object.
+ *
+ * This function provides a generic way to display all error messages returned
+ * by an API call. It can be used in every form to harmonize the messages
+ * processing.
+ *
+ * Ex
+ *      const response = await addNewVideo({ video_id }).catch(
+ *        (reason: ApiError) => {
+ *          displayErrors(enqueueSnackbar, reason);
+ *        }
+ *      );
+ *
+ * The optional `overrideStatuses` parameter allows the developers to display
+ * in the UI custom messages for a specific set of HTTP response status code.
+ * They will override those contained in the response.
+ *
+ * Ex
+ *      const response = await addNewVideo({ video_id }).catch(
+ *        (reason: ApiError) => {
+ *          displayErrors(
+ *            enqueueSnackbar,
+ *            reason,
+ *            'Sorry, an error has occurred. The video cannot be added.',
+ *            [{ status: 409, variant: 'error', msg: 'Video already added.' }]
+ *          );
+ *        }
+ *      );
+ *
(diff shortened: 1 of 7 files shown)


GresilleSiffle:
I need a little help to make TypeScript with the following line : )

https://github.com/tournesol-app/tournesol/blob/c988d264e12d4bda61cde4ab753f40e01f848074/frontend/src/utils/api/response.ts#L71

frontend/src/utils/api/response.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    // HTTP 429 Too Many Requests are considered information
+    if (reason.status === 429) {
+      showInfoAlert(display, reason.body['detail']);
+    } else {
+      const newErrorMessages = Object.values(reason['body']).flat();
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      newErrorMessages.map((msg) => showErrorAlert(display, msg));
```
amatissart, line 71: It's always possible to use `as` to force a specific type when it's "unknown" for Typescript.  

And we have no guarantee about the content of `reason.body` here so it will be hard to be completely safe here.  
A bug had already been reported when the body is a raw string:
image
```
Due to the behavior of `Object.values()`:
> Object.values("Error")
Array(5) [ "E", "r", "r", "o", "r" ]
```

A slightly safer version would be:
```suggestion
      const body = reason.body;
      const newErrorMessages =
        typeof body === 'string'
          ? [body]
          : Object.values(reason['body']).flat();
      newErrorMessages.map((msg) => showErrorAlert(display, msg as string));
```

amatissart (approved):

PR #410, opened 2022-01-05
[extension] Fixes recommandations refreshing

### LINKED ISSUE #404: fix(extension): Don't refresh whole page when changing language options

FabienRoger:
Fixes #404 
+Replace language text by flags
+ Add a refresh button


GresilleSiffle:
Hello @FabienRoger 

Thanks for all these new additions.

This pull request includes four features and it will be easier for us, maintainers, to review them one by one in their individual pull request. It will allow us to give feedback about the refresh button for instance, without blocking the other features. In the future, it would be really helpful to create several pull request :pray: 

**about the refresh button**

I think it's a good choice to make the button disabled while the request in running. It will give the users an indication to wait before clicking a second time, and will prevent HTTP requests flooding. What do you think @amatissart @lfaucon ?

How about adding a different background color on hover, a set the cursor CSS to `pointer` too? Those kinds of little tweaks will improve the look'n'feel of the extension. I might be easier for visually impaired person to have the mouse cursor to change on hover.

**about the expand feature**

I personally like a lot the fact of having the possibility to display more Tournesol recommendations. As recommending a video to someone is impacting, should we @lfaucon @aidanjungo agree first on the criteria allowing the videos to be displayed? I think about having a minimum number of contributor ratings for some videos, defining how many recent / old / older videos we want to display, etc.

FabienRoger:
I have created these different pull requests!

lfaucon:
> I think it's a good choice to make the button disabled while the request is running. It will give the users an indication to wait before clicking a second time, and will prevent HTTP requests flooding. What do you think @amatissart @lfaucon ?

Yes, on top of this, it is a good feedback for the user that the click is working

> As recommending a video to someone is impacting, should we @lfaucon @aidanjungo agree first on the criteria allowing the videos to be displayed? I think about having a minimum number of contributor ratings for some videos, defining how many recent / old / older videos we want to display, etc.

The ML algorithm is such that if we have a sufficient number of contributors, a small number of users will not be able to pull too high the score of a video. Normally, we should not need to add a fxed minimum number of contributors. Apart from this, yes, it would be interesting to work on deciding what to recommend based on scores & recency & maybe something else :)

PR #415, opened 2022-01-06
[extension] Stop page from refreshing

### LINKED ISSUE #403: feat(extension): Faster loading of recommendation on Youtube home page.
Currently we wait for the page to load, then fetch Tournesol's recommendations, then display them on Youtube's home page. We could show the recommendations faster if we fetch them while the page is still loading, then wait for the page to finish loading to display them.

### LINKED ISSUE #404: fix(extension): Don't refresh whole page when changing language options

FabienRoger:
Fixes #404


browser-extension/addTournesolRecommendations.js
```diff
@@ -56,11 +56,10 @@ const getTournesolComponent = (data) => {
...
       lang_option = document.createElement('a');
       lang_option.className = 'language_option';
-      lang_option.href = '#';
       lang_option.append(lang);
       lang_option.onclick = () => {
         localStorage.setItem('tournesol_extension_config_language', lang);
-        window.location.reload();
+        loadRecommandations();
       };
       return lang_option;
```
amatissart, line 64: When `href` is absent, the browser no longer considers this element as interactive: no cursor change on hover, and no underline.  

So it would be more appropriate to keep this "href", describe its role as "button" (for accessibility), and disable the default click behavior with `preventDefault()`:

```diff
       lang_option = document.createElement('a');
       lang_option.className = 'language_option';
+      lang_option.href = '#';
+      lang_option.setAttribute('role', 'button');
       lang_option.append(lang);
-      lang_option.onclick = () => {
+      lang_option.onclick = (event) => {
+        event.preventDefault();
         localStorage.setItem('tournesol_extension_config_language', lang);
```

amatissart (commented):
Thanks for tackling this issue. 

I only have 1 suggestion before merging to keep the appearance of these languages choices as links.

Another small drawback of this change is that there is no feedback anymore on click: on long requests the user may be tempted to click multiple times which would make the problem worse.
But this is probably a wider issue in the extension, so this can be addressed separately.

Also note that this change may conflict with #412, so you may want to address these reviews one at a time.

browser-extension/addTournesolRecommendations.js
FabienRoger: Done!

amatissart:
@FabienRoger a conflict needs to be resolved on this branch, now that #412 is merged.

lfaucon (approved):

amatissart (approved):

PR #416, opened 2022-01-06
[extension] Replace language by flag

FabienRoger:
image


GresilleSiffle:
Hello again,

We add a short discussion today with three other members of the project about the pertinence of using country flags instead of language name on the web.

Currently we think using flags is not appropriate in our case. We only want users to select the language they would like to display.

One reason is that flags are unique to a country, and a specific country may not be the only ambassador of a language. A same language can be spoken by many people across the world, and there is no unique flag to represent them. For instance, french is spoken in a dozen of countries, like `Benin`, `Burkina Faso`, `Monaco`, `Quebec` and many others.

Another reason is not all people know by heart the association between a country and its spoken languages. At a first glance using flags can appear user friendly, but in fact it could make it difficult for a user to select a language.

I'm gonna leave the pull request as is for now, so that @lfaucon can have a look and maybe add feedback.

However, a sincere thank you for your implication and your contributions to the extension. And for having split you work in several pull requests :100:

lfaucon:
Thanks for all the recent contributions @FabienRoger 

I agree too with @GresilleSiffle comments about flags and languages. Too bad, because it does look nicer :)

PR #417, opened 2022-01-06
[extension] Add expand button

FabienRoger:
Proposal made in #404


FabienRoger:
I didn't bother with making the spacing right between the button and the rest, because it will change depending on which other branches (for example, the flags) will get merged.

amatissart:
There will be some details to be discussed about the appearance and the position of this button.

But more importantly for now, the behavior of this button is a bit unexpected in my opinion. Expanding or reducing the Tournesol container should not update the videos that are visible on the first row. It could be frustrating for the user to find exciting recommendations, and to lose them just because they clicked to see some more :slightly_smiling_face:

FabienRoger:
The expand button branch is now on top of the top of the refresh bottom branch as they share a lot of css code, and because they have many interactions concerning the loading are reloading of videos.

browser-extension/background.js
```diff
@@ -57,10 +57,10 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
       const threeWeeksAgo = getDateThreeWeeksAgo()
       const recent = await request_recommendations(`date_gte=${threeWeeksAgo}&language=${request.language}&limit=10`);
       const old = await request_recommendations(`date_lte=${threeWeeksAgo}&language=${request.language}&limit=50`);
-      const recent_sub = getRandomSubarray(recent, 3);
-      const old_sub = getRandomSubarray(old, 4 - recent_sub.length);
-      const videos = getRandomSubarray([...old_sub, ...recent_sub], 4);
-      return { data: videos };
+      const recent_sub = getRandomSubarray(recent, request.number - 1);
```
GresilleSiffle, line 60: I think we could discuss this algorithm together.

@aidanjungo @lfaucon @amatissart @FabienRoger 

If I understood correctly, when expanding the Tournesol recommendation panel, the code loads up to 7 recent videos, and 8 minus 7 older video(s).

So, when there is only one line of recommendations we have
- 3 recent videos (75 %)
- 1 older video (25 %)

When there is 3 lines of recommendations we expect to have:
- 10 recent videos (83%) 
- 2 older videos (17%)

The % of older videos keeps decreasing the more lines we loads. This leads me to several questions:

(1) Could this be a problem?

(2) Do we have enough pertinent contributions to trust the fact that the recommendations are still " quality recommendations " when we display three lines of videos with 83 % of recent videos?

browser-extension/addTournesolRecommendations.js
```diff
@@ -78,6 +86,45 @@ const getTournesolComponent = (data) => {
...
+      refresh_button.disabled = true;
+      loadRecommandations(loadVideos = true, loadAdditionalVideos = isExpanded);
+    };
+    inline_div.append(refresh_button);
+    // Expand button
+    expand_button = document.createElement('button');
+    expand_button.setAttribute('id', 'tournesol_expand_button');
+    // A new button is created on each video loading, the image must be loaded accordingly
+    if(isExpanded){
+      fetch(chrome.runtime.getURL('images/chevron-down.svg'))
```
GresilleSiffle, line 107: The `chevron-down.svg` and `chevron-up.svg` could be swapped.

Currently click on the up chevron will expand the panel downward, and clicking on the up chevron will expand the panel upward. We found with @amatissart that's it more natural to have the chevron pointing in the same direction where the panel will expand.

browser-extension/addTournesolRecommendations.css
```diff
@@ -28,25 +28,83 @@
...
+  position: relative;
+}
+
+.tournesol_simple_button svg {
+  position: absolute;
+  left: 0px;
+  top: 0px;
+  width: 20px;
+  height: 20px;
+  border-radius: 10px;
```
GresilleSiffle, line 79: The `border-radius` here crops the icons image, see:

**with border radius** the refresh icon doesn't look good

capture

**without border radius** the refresh icon looks better

capture2

To avoid cropping images, I recommend not using `border-radius` when there is no padding, or no background, or at least nothing to " round " : )

browser-extension/addTournesolRecommendations.js
```diff
@@ -3,8 +3,16 @@
 // This part is called on connection for the first time on youtube.com/*
 /* ********************************************************************* */
 
+const defaultVideoNumber = 4;
```
amatissart, line 6: This constant defines the number of videos per row, and not only the number of visible videos by default. So `videosPerRow` may be a better name ?

GresilleSiffle (changes requested):
Thanks for working on the improvement of the extension @FabienRoger 

I hit the request changes button because I saw some icons provided by Font Awesome were cropped by our custom CSS.

I also want to grab our attention on how the extension will recommends video based on the currently proposed algorithm. As the current pull request adds the possibility to recommend more videos than before, I'd feel more comfortable if @aidanjungo and @lfaucon could have a look : )

Thank you!

browser-extension/addTournesolRecommendations.css
```diff
@@ -28,25 +28,83 @@
...
 }
 
+.tournesol_simple_button {
+  fill: var(--yt-spec-text-secondary);
+  width: 20px;
+  height: 20px;
+  padding: 0px;
+  border-style: none;
+  background-color: transparent;
+  position: relative;
```
amatissart, line 70: Let's use a cursor pointer on these buttons for better UI:
```suggestion
  position: relative;
  cursor: pointer;
```

browser-extension/addTournesolRecommendations.js
GresilleSiffle: To keep the extension user-friendly, we could in a future task, decide where we want to put the refresh and the expand buttons.

With @amatissart we had the idea of moving the expand button below the recommendation panel. Right it's just an idea, but it raises the question where do we want to place the action buttons.

amatissart (commented):
Thanks for addressing the comments done in the previous review. It works great.

browser-extension/addTournesolRecommendations.js
amatissart: Yes I think the expand/reduce button would be better positioned below the panel. 

We could handle this in a separate PR though, with a cleaner layout for existing buttons. The current style with absolute positioning and floating elements is a bit difficult to work with.

browser-extension/background.js
FabienRoger: I agree. When I made this change I wasn't really thinking about the proportions of videos and I just did something "close enough". Thank you for catching that. I will revert to the old ratio and will properly implement something to make sure the ratio is respected both in the expanded and not-expanded form.

lfaucon: Because we are showing more videos, let's also take them from a larger set of possible videos to show.

Currently we pick from 10 recent and 50 older videos. 

We are not yet confident enough in the quality of very recent videos so I prefer keeping it at 10 or 15, but we could increase the number of old videos to 200. 

Then it may be best to vary the proportions: first 4 stay the same but next 8 videos could be 75% old and 25% recent.

Happy to discuss more if needed, but we will iteratively improve this later too.

GresilleSiffle: > I agree. When I made this change I wasn't really thinking about the proportions of videos and I just did something "close enough". 

That's perfectly ok. These kinds of consequences are hard to spot when you are discovering the code. It's something we can keep in mind when we are developing: " ok now that the code is working, what are its consequences? " .

FabienRoger: I fixed that !

FabienRoger:
I fixed and made explicit the parameters going into the recommendations request.
However, for now, the program assumes that there will be enough videos to match the requested oversampling ratio. If not, it sometimes doesn't load enough videos in the "expand" tab.
There is a choice to make to deal with this issue:
Either the extension lowers the oversampling ratio when there is not enough (recent) videos. Risking always having the same recent videos displayed.
Or the extension increase the number of old videos to make up for missing recent videos.
What do you think?

browser-extension/addTournesolRecommendations.js
lfaucon: Agreed!

browser-extension/addTournesolRecommendations.js
```diff
@@ -78,6 +86,45 @@ const getTournesolComponent = (data) => {
...
+    // A new button is created on each video loading, the image must be loaded accordingly
+    if(isExpanded){
+      fetch(chrome.runtime.getURL('images/chevron-up.svg'))
+        .then(r => r.text())
+        .then(svg => expand_button.innerHTML = svg);
+    }else{
+      fetch(chrome.runtime.getURL('images/chevron-down.svg'))
+        .then(r => r.text())
+        .then(svg => expand_button.innerHTML = svg);
+    }
```
lfaucon, line 114: Remove some code duplication with a simple ternary operator

```suggestion
      fetch(chrome.runtime.getURL(isExpanded ? 'images/chevron-up.svg' : 'images/chevron-down.svg'))
        .then(r => r.text())
        .then(svg => expand_button.innerHTML = svg);
```

browser-extension/background.js
```diff
@@ -53,14 +62,60 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
+      );
+      const old = await request_recommendations(
+        `date_lte=${threeWeeksAgo}&language=${request.language}&limit=${oldVideoToLoad+oldAdditionalVideoToLoad}`
+      );
+      
+      // Cut the response into the part for the videos and the one for the additional videos
+      const videoRecent = recent.slice(0,recentVideoToLoad);
+      const videoOld = old.slice(0,oldVideoToLoad);
+      const additionalVideoRecent = recent.slice(recentVideoToLoad);
+      const additionalVideoOld = old.slice(oldVideoToLoad);
```
lfaucon, line 88: That is not working as expected. The `recent` and `old` returned by the db are sorted so additional videos will always be taken from the bottom two thirds of the lists. Did you intend this?

I would propose:

```JavaScript
const numberOfRecentVideoToRespond = Math.round(request.videosNumber*recentVideoProportion);
const numberOfRecentAdditionalVideoToRespond = Math.round(request.additionalVideosNumber*recentVideoProportionForAdditionalVideos);

// fetch for both normal and additional videos
const recentVideos = getRandomSubarray(recent, numberOfRecentVideoToRespond + numberOfRecentAdditionalVideoToRespond)
// split the two lists
const recentVideos = recentVideos.slice(0, numberOfRecentVideoToRespond)
const additionalRecentVideos = recentVideos.slice(numberOfRecentVideoToRespond)

// TODO:
// - Compute number of required old and additionalOld videos based on lengths of lists obtained for recent and recentAdditional
// - Get the lists of videos with the same method from `old`
```

lfaucon (dismissed):
Thanks for working on this. It's a very cool feature to have!

In the next steps, I would try to add the `transition` css property to make the panel extend and collapse smoothly, but that requires a bit of code refactor so I much prefer that we get this first feature first, then improve it.

browser-extension/background.js
FabienRoger: I noticed that, but I believe it makes more sense like that. Most users will probably never click on the expand button, so I believe that it is crucial that this feature doesn't decrease the quality of the main recommendations. I believe that the current state of the branch doesn't change the main recommendations, and I think it is better if it remains so.

lfaucon: Ok! I'm convinced by your reasonning. Let's stay with your proposition for now and reassess it later if we feel the need or want to change more things. Thanks!

GresilleSiffle (commented):
Hi @FabienRoger 

I just tested the current code, the look and feel of the buttons has improved that's great.

If there are no more objection, we can merge these new features.

@lfaucon do you agree?

GresilleSiffle (dismissed):

lfaucon:
Yes, I agree 👍 Thanks for this feature @FabienRoger

lfaucon (approved):

PR #418, opened 2022-01-06
[extension] Add refresh button

FabienRoger:
Proposal made in #404


FabienRoger:
I didn't bother with making the spacing right between the button and the rest, because it will change depending on which other branches (for example, the flags) will get merged.

FabienRoger:
I didn't bother with making the spacing right between the button and the rest, because it will change depending on which other branches (for example, the flags) will get merged.

browser-extension/addTournesolRecommendations.css
```diff
@@ -28,25 +28,64 @@
...
+  -webkit-filter: opacity(100%);
+  filter: opacity(100%);
+  background: #bdbcb2;
+
+  -webkit-transition: all 0.1s ease;
+  -moz-transition: all 0.1s ease;
+  -o-transition: all 0.1s ease;
+  -ms-transition: all 0.1s ease;
+  transition: all 0.1s ease;
+}
```
lfaucon, line 87: The button style is barely visible on youtube dark mode 

Screenshot 2022-01-06 at 21 48 40

I believe you may use something like https://fontawesome.com/v5.15/icons/sync?style=solid for a similar icon and it would be modified by the css attribute `color: var(--yt-spec-text-primary);` like the title and text, unlike the png file that you currently use.

lfaucon (changes requested):
Thanks! It works very well

browser-extension/addTournesolRecommendations.css
FabienRoger: I need help to do this. Why this js extension it isn't as simple as doing "npm install" and my attempts failed...

amatissart: No dependency is necessary. Instead of a .png image, we need to use a .svg element where CSS rules can be applied.
There is a catch though: `` cannot by styled using CSS. We need to inject the `<svg>` element in the DOM.  

Here is a solution:
```js
refresh_button = document.createElement('button');
refresh_button.setAttribute('id', 'refresh_button');
fetch(chrome.runtime.getURL('images/refresh.svg'))
  .then(r => r.text())
  .then(svg => refresh_button.innerHTML = svg);
```

As the `<svg>` is now present inside the `<button>`, the CSS variable provided by youtube.com can be used to change the color:
```css
#refresh_button {
  fill: var(--yt-spec-text-primary);
}
```
Note that `fill` is used instead of `color` to color SVG elements.

browser-extension/addTournesolRecommendations.css
```diff
@@ -28,25 +28,74 @@
...
 #tournesol_link {
   color: var(--yt-spec-text-primary);
-  font-family: 'Roboto', 'Arial', sans-serif;
+  font-family: "Roboto", "Arial", sans-serif;
   position: absolute;
   right: 0px;
   top: 15px;
 }
 
+#refresh_button {
```
amatissart, line 45: To avoid potential conflicts with other elements we don't control on the page, it would be nice to use a `tournesol_` prefix (already used on previous items) in this id.

browser-extension/addTournesolRecommendations.css
```diff
@@ -28,25 +28,74 @@
...
+  margin-left: 12px;
+  margin-right: 12px;
+}
+
 #tournesol_icon {
   position: absolute;
   left: 0px;
 }
 
+.icon {
```
amatissart, line 57: Similarly, `icon` is a very generic name for a class. It may conflict with other items on the page, or items injected by other extensions.  

If I understand correctly, this class is applied on interactive icons. Something like `.tournesol_button_icon` could be more appropriate.

browser-extension/addTournesolRecommendations.js
```diff
@@ -78,6 +80,18 @@ const getTournesolComponent = (data) => {
...
     inline_div.append(tournesol_link);
 
+    // Refresh button
+    refresh_button = document.createElement('img');
+    refresh_button.setAttribute('id', 'refresh_button');
+    refresh_button.setAttribute('src', chrome.runtime.getURL('images/Refresh_icon.png'));
+    refresh_button.setAttribute('width', '24');
+    refresh_button.className = 'icon';
+    refresh_button.onclick = () => {
+      refresh_button.className = 'icon-disabled';
```
amatissart, line 90: By using a `<button>`, you could simply set `refresh_button.disabled = true`. And that it can be styled with [`.icon:disabled`](https://developer.mozilla.org/fr/docs/Web/CSS/:disabled)

The advantage is that disabled buttons can't trigger click events: no double clicks, no duplicated requests to the API.

FabienRoger:
I made those changes directly on the expand button branch, which sits on top of this branch and also need these changes.

PR #419, opened 2022-01-06
browser menu relooking

glerzing:
I tried some CSS on the extension menu, and I think it actually looks better with. Here is how it looks in chrome :

Capture d’écran de 2022-01-06 14-34-43

and in firefox:
Capture d’écran de 2022-01-06 14-43-27

### DIFF
--- a/browser-extension/menu.html
+++ b/browser-extension/menu.html
@@ -13,12 +13,12 @@
                 margin-bottom: 6px;
             }
             button {
+                border-width: 2px;
                 background-color: #506AD4;
                 color: white;
                 font-size: large;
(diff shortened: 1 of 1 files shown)


FabienRoger (dismissed):
I like it too!

amatissart:
Here is the design reference we are currently following:
https://www.figma.com/file/3uI7FiBhXp0hhgdTdHBSkj/Tournesol?node-id=630%3A93
image

So for consistency with the rest of the application, I wouldn't approve this change.

glerzing:
Ok, then I remove this PR

PR #420, opened 2022-01-06
Implement `useNotifications()` hook to translate all error messages

amatissart:
It also brings some simplification when using these alerts. Calling `useSnackbar` to get `enqueueSnackbar` is no longer necessary.


GresilleSiffle (approved):
:+1:

Issue #388, opened 2021-12-27
feat(frontend): Detailed score vizualization on video analysis page


lfaucon:
On a video analasys page (for example: http://localhost:3000/video/utWMGi8HTjY) 

Display the criteria scores already provided by the API (for example: https://api.tournesol.app/video/utWMGi8HTjY/) with a radar chart (more info: https://datavizcatalogue.com/methods/radar_chart.html)

PR #421, opened 2022-01-06
added a radar chart with npm, import it in pages/videocard_feed the a…

sintaz1992:
Related to issue #388 

Adds a radar chart to video card page to show criteria score from API.

The radar chart library (https://www.npmjs.com/package/react-svg-radar-chart)


GresilleSiffle (changes requested):
Hello @sintaz1992 

It seems you have added a `package-lock.json` in your files. As we are using `yarn` and `yarn.lock` this file is not required, so you can delete it.

Also, question from a person who have not integrated any charts in a React app: have you considered using a more complete package, like [Recharts](https://recharts.org/en-US/)?

If we plan in the future to integrate more charts in the application, let's say to display public statistics or so, it might be a good move to directly use `Recharts` or similar. Maybe @amatissart @lfaucon @aidanjungo you have a better understanding of our future plans : )

GresilleSiffle:
Here is how it looks like on my development environment: 

screencapture-localhost-3000-video-Ok5sKLXqynQ-2022-01-06-16_37_21

I would recommend to center the radar. Instead of inserting the radar in a `<div>`, we can use a Material UI component which comes with a minimal set of CSS that will help to automatically layout the page nicely ( like `<Container>`, see https://mui.com/components/container/ or `<Grid>`)

lfaucon:
> Also, question from a person who has not integrated any charts in a React app: have you considered using a more complete package, like Recharts instead of react-svg-radar-chart?

> If we plan in the future to integrate more charts in the application, let's say to display public statistics or so, it might be a good move to directly use Recharts or similar. Maybe @amatissart @lfaucon @aidanjungo you have a better understanding of our future plans : )

I'd say we are only planning to have just a few charts on the frontend. They are not super user-friendly (do you see lots of charts on Youtube?). I think it is fine to stick with `react-svg-radar-chart` for now. It will not cost much to change the library in a later iteration.

Short term the Video Analysis page is hidden from users (no links to it on the frontend, only a link from the extension). Let's keep it this way until we have worked more on it UX/UI-wise.

sintaz1992:
Thanks @GresilleSiffle for review.. I removed the packages.json. It was created when I added the package for radar chart. I agree with you about centering. As @lfaucon  says, because its not very used now by users, my suggestion is to merger it and then open a new ticket for centering it (adding container) . what do you think?

sintaz1992:
Also I have a conflict in yarn.lock.. what should I do? is it because of adding new package?

amatissart:
> Also I have a conflict in yarn.lock.. what should I do? is it because of adding new package?

"yarn.lock" is updated on your branch because a new package has been added, but other packages were added concurrently in #385 and also modified "yarn.lock". So this branch can't be merged automatically.  

The simplest way is probably to delete "yarn.lock" after rebasing your branch on "main". Then, all new packages will be defined in "package.json", and you can regenerate the file with `yarn install`.

GresilleSiffle:
Hello @lfaucon 

> I'd say we are only planning to have just a few charts on the frontend. They are not super user-friendly (do you see lots of charts on Youtube?). I think it is fine to stick with react-svg-radar-chart for now. It will not cost much to change the library in a later iteration.

Alright, I'm ok with continuing with `react-svg-radar-chart` for the moment.

I was thinking about charts non-related to a single video. For instance, we could have a dashboard with some metrics representing the community activity: like the evolution the non-aggregated number of comparisons made by month ( [with this style](https://recharts.org/en-US/examples/SimpleLineChart) ); or bar chart sorting videos by age, so that we can the " freshness " of the recommendations of Tournesol ( [like this](https://recharts.org/en-US/examples/StackedBarChart) ).

There were just examples, but I think those kind of metrics could help the association to see what's going on in the application, how much the community is active etc. I agree with you, it's another topic and a potential future work.

Hello @sintaz1992 

>  my suggestion is to merger it and then open a new ticket for centering it (adding container) . what do you think?

I confess that I'd prefer to add a least one or two Material UI tags so that the UI feels more clean, even if we don't aim for a perfect UX/UI the alignment of items is still an important criteria for me. But, if you strongly prefer to progress step by step, and let's say you're not that comfortable with Material UI right now, I'm open to merge the chart and the UI clean-up after.

What do you prefer @sintaz1992 ? : )

However there is still a regression in this pull request. I'd advise us to fix it before merging so that another person working on another feature in the same page will not have to address two issues in one.

Have you been able to reproduce it @sintaz1992 or anyone? Just add a new video to your rate later list, a video that doesn't already exist in the database, then check this page with its video id. I know this page is not used right now, and maybe too few people will notice, but I'm not comfortable with merging this pull request as is.

Hope it helps :pray:

sintaz1992:
I agree with you @GresilleSiffle .. I will consider the allignment issue and push again for your review

sintaz1992:
Thanks @amatissart ..wil do so

frontend/src/pages/videos/VideoCard.tsx
```diff
@@ -18,12 +21,52 @@ const useStyles = makeStyles(() => ({
...
+      ? Object.fromEntries(
+          video.criteria_scores.map((criteria) => [
+            criteria.criteria,
+            criteria.criteria,
+          ])
+        )
+      : {
+          d: 'd',
+        }
+    : null;
```
GresilleSiffle, line 38: This code works, but is a bit hard to read.

Let's make the `VideoCardPage` component easier to maintain by moving the `captions` calculation in a function outside of its body. There are several advantages:
- a human will read and understand what the component is doing faster
- the newly created function could be re-used by many other components

Here is a suggestion you can add in a single click : )

```suggestion
  const captions: any =
    video.criteria_scores?.length > 0 ? getCriteria(video.criteria_scores) : {};
```

frontend/src/pages/videos/VideoCard.tsx
```diff
@@ -18,12 +21,52 @@ const useStyles = makeStyles(() => ({
 function VideoCardPage() {
```
GresilleSiffle, line 21: These two function will be responsible of retrieving the criteria and the related scores. Having such functions outside of `VideoCardPage` page will ease the future development (see my other comments below).

They are not perfect, as they still contain the type `any` which should be replaced by a proper type definition if possible. 

```suggestion

const getVideoCriteria = (criteriaScores: Array<any>) => {
  return Object.fromEntries(
    criteriaScores.map((criteria: any) => [
      criteria.criteria,
      criteria.criteria,
    ])
  );
};

const getVideoMetaData = (criteriaScores: Array<any>) => {
  return {
    data: Object.fromEntries(
      criteriaScores.map((criteria: { criteria: string; score: number }) => [
        criteria.criteria,
        criteria.score,
      ])
    ),
    meta: { color: 'purple' },
  };
};

function VideoCardPage() {
```

frontend/src/pages/videos/VideoCard.tsx
```diff
@@ -18,12 +21,52 @@ const useStyles = makeStyles(() => ({
...
+        }
+      : {
+          data: {
+            d: 0,
+          },
+          meta: { color: 'blue' },
+        }
+    : null;
+  console.log(captions);
+  console.log(video.criteria_scores);
```
GresilleSiffle, line 59:
```suggestion

  const data: any =
    video.criteria_scores?.length > 0
      ? getVideoMetaData(video.criteria_scores)
      : {};

```

frontend/src/pages/videos/VideoCard.tsx
```diff
@@ -18,12 +21,52 @@ const useStyles = makeStyles(() => ({
...
   return (
     <div>
       <div className={classes.root}>
         <VideoCardFromId videoId={video_id} />
       </div>
+      <Container maxWidth="sm">
+        {video.criteria_scores && (
```
GresilleSiffle, line 66: I suggest to not display the `RadarChart` when there is no score for now.

The current code displays an empty chart with a "placeholder" value which is a bit hacky. Let's not display the cart if we have no proper way to display it empty.

```suggestion
        {video.criteria_scores?.length > 0 && (
```

GresilleSiffle (changes requested):
Hi @sintaz1992 

Thanks for fixing the issue, the page is now displayed correctly even for videos with no score.

Now that the code is working, I suggested in this review few adjustments to make the code easier to read and easier to maintain.

You can review them, and have the possibility to add them by simply clicking on the button `commit suggestion` .

After that, the only remaining work will be to make our code style checks happy by removing the type `any`, and using a correctly described objects.

If you need any help, do not hesitate to reach us on Discord.

lfaucon:
TODO:
- [ ] Figure out how to include image SVGs
- [ ] Center the chart again
- [ ] Clean-up the code :)
- [ ] Fix the `any` if there are some left
- [ ] Make sure all previous comments have been addressed

lfaucon (changes requested):
Starting to look good 👍

lfaucon:
Nice progress. Thank you @sintaz1992 

TODO:
- [ ] Color of svg
- [ ] Tootltips on the svg
- [ ] Separate RadarChart to make it a reusable component

sintaz1992:
@GresilleSiffle now CriteriaRadarChart is a component. Hopefull all your comments been addressed except issues with : any.. don't know how to remove them

lfaucon (approved):

frontend/src/App.tsx
```diff
@@ -17,7 +17,7 @@ import Frame from './features/frame/Frame';
 import ComparisonPage from './pages/comparisons/Comparison';
 import PublicRoute from './features/login/PublicRoute';
 import PrivateRoute from './features/login/PrivateRoute';
-import VideoCardPage from './pages/videos/VideoCard';
+import VideoAnalysisPage from './pages/videos/VideoAnalysis';
```
amatissart, line 20: As `recharts` is quite a big dependency (https://bundlephobia.com/package/recharts@2.1.8), and is only used on this single route, it would make sense to import this component lazily (it will be built in separate .js chunks, which won't be downloaded by most visitors).

```ts
const VideoAnalysisPage = React.lazy(
  () => import('./pages/videos/VideoAnalysis')
);
```

Also see https://reactjs.org/docs/code-splitting.html#route-based-code-splitting for more details.

GresilleSiffle (dismissed):

GresilleSiffle (commented):
Hello @lfaucon @sintaz1992 

The code looks good to me, I've dismissed my change request as all my previous comments are now outdated. Nice job!

Ready to merge after the last Adrien's comment have been reviewed.

amatissart (approved):

Issue #372, opened 2021-12-17
Enable gzip compression in Nginx


amatissart:
Responses that should be compressed:
 * static content from frontend: .js, .css, etc
 * JSON responses from backend

References:
* https://docs.nginx.com/nginx/admin-guide/web-server/compression/
* https://developer.mozilla.org/fr/docs/Web/HTTP/Headers/Content-Encoding

amatissart:
Fixed by #477 
Thanks @glerzing

PR #423, opened 2022-01-07
Infra readme

glerzing:
In the context of the #372 (Nginx compression), I made a separate PR for this. Personally I have had much fewer problems by renting a cloud server than by using a VM (problems with ansible, and creating/configuring the VM). And it is much more comfortable, since my computer is not slowed and I don't have to remake all the ansible process after a restart.
So I wanted to indicate this (maybe obvious) alternative for others. Tell me if you think it is an improvement. 
Also, there was a missing dependency for rsync.

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -1,7 +1,11 @@
 
-## Environment choice
-This section covers how to set a production-like environment. 
-For most developments, it is better to use the containers instead (see the dev-env directory). The infrastructure environment provides additional services like reverse-proxy (nginx) and monitoring, but is also usually not necessary and less practical for backend and frontend development tasks.
+# Infrastructure
+
+## Installing in a VM for Infrastructure Code Development
+
+This section covers how to set up a production-like environment in a virtual machine.
+
+This personal environment allows developppers to work on the Ansible deployment recipe and to test their modifications. If you want to work of the Tournesol application code, use the `dev-env` instead.
 
 There are 2 options, renting a server to a cloud provider, or creating a test VM for development:
 - Setting up a VM (development only). It is potentially safer, since it is free and not exposed to the internet. This may significantly slow your computer on runtime. A specific network configuration like port forwarding is also needed to communicate with the VM.
--- a/infra/README.md
+++ b/infra/README.md
@@ -41,13 +41,6 @@ There are 2 options, renting a server to a cloud provider, or creating a test VM
   
 If for any reason you're not able to set up a virtual machine on your computer - your hardware have missing virtualization capabilities, or is not powerful enough - you can still use a remote virtual machine from a Cloud provider. Some Cloud providers offer free credits for new users, but it can get costly if you rent a powerful server and forget to stop it after use. Note this installation method **is not supported** by the team and you might encouter unexpected issues. 
 
-- Choose a cloud provider (Vultr, Digital Ocean, AWS...). Most offer a free trial or free credit for new users.
-- Select a Debian server. 1 CPU cores and 2 GB of RAM should be sufficient, at least for development (to avoid fees, don't forget to close it after all your tests are done).
-- connect to the server via ssh
-- install `sudo` if not present: `apt install sudo`
-- create your user: `sudo useradd <username>` and set its password: `sudo passwd <username>`
-- add your user to the `sudo` group: `gpasswd -a <username> sudo`
-
 
 ## Provisioning
 
--- a/infra/README.md
+++ b/infra/README.md
@@ -7,10 +7,6 @@ This section covers how to set up a production-like environment in a virtual mac
 
 This personal environment allows developppers to work on the Ansible deployment recipe and to test their modifications. If you want to work of the Tournesol application code, use the `dev-env` instead.
 
-There are 2 options, renting a server to a cloud provider, or creating a test VM for development:
-- Setting up a VM (development only). It is potentially safer, since it is free and not exposed to the internet. This may significantly slow your computer on runtime. A specific network configuration like port forwarding is also needed to communicate with the VM.
-- Renting a server in the cloud is closer to the production environment. Most cloud providers (Vultr, Digital Ocean, AWS...) offer free credit for new users, but it can get costly if you rent a powerful server and forget to stop it after use.
-
 
 ## Option 1: Installing a VM for Infrastructure Code Development
 
--- a/infra/README.md
+++ b/infra/README.md
@@ -1,8 +1,25 @@
-## Fetching Debian Bullseye Image and Verifying It
 
-- `./base-image/fetch-debian-image.sh`
+## Environment choice
+This section covers how to set a production-like environment. 
+For most developments, it is better to use the containers instead (see the dev-env directory). The infrastructure environment provides additional services like reverse-proxy (nginx) and monitoring, but is also usually not necessary and less practical for backend and frontend development tasks.
 
-## Installing in a VM for Infrastructure Code Development
+There are 2 options, renting a server to a cloud provider, or creating a test VM for development:
+- Renting a server in the cloud is closer to the production environment. Most cloud providers (Vultr, Digital Ocean, AWS...) offer free credit for new users, but it can get costly if you rent a powerful server and forget to stop it after use.
+- Setting up a VM (development only). It is potentially safer, since it is free and not exposed to the internet. This may significantly slow your computer on runtime. A specific network configuration like port forwarding is also needed to communicate with the VM.
+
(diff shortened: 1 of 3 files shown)


glerzing:
(maybe in this case "tournesol-test" or "tournesol-dev" would be better than "tournesol-vm")

infra/README.md
```diff
@@ -1,8 +1,25 @@
-## Fetching Debian Bullseye Image and Verifying It
 
-- `./base-image/fetch-debian-image.sh`
+## Environment choice
+This section covers how to set a production-like environment. 
+For most developments, it is better to use the containers instead (see the dev-env repository). The infrastructure environment provides additional services like reverse-proxy (nginx) and monitoring, but is also much less practical.
 
-## Installing in a VM for Infrastructure Code Development
+There are 2 options, renting a server to a cloud provider, or creating a test VM for development:
+- Renting a server in the cloud is a simpler and faster way to get started. And closer to the production environment. Most cloud providers (Vultr, Digital Ocean, AWS...) offer free credit for new users, but it can get costly if you rent a powerful server and forget to stop it after use.
```
glerzing, line 7: je n'ai pas précisé mais vu que c'est du HTTP, tout est en clair et ça peut peut-être poser des problèmes de confidentialité

jstainer-kleis: with a public IP you can have HTTPS with Let's Encrypt certificates, it only requires setting an email address in the inventory (as specified in a comment)

jstainer-kleis: > Renting a server in the cloud is a simpler and faster way to get started
it's really subjective, I personally think it's the contrary

jstainer-kleis:
I'm not opposed to the addition, not really convinced by the value either, I think it's an obvious alternative. I'll let @amatissart decide whether or not to accept it.

infra/README.md
```diff
@@ -1,8 +1,25 @@
-## Fetching Debian Bullseye Image and Verifying It
 
-- `./base-image/fetch-debian-image.sh`
+## Environment choice
+This section covers how to set a production-like environment. 
+For most developments, it is better to use the containers instead (see the dev-env repository). The infrastructure environment provides additional services like reverse-proxy (nginx) and monitoring, but is also much less practical.
```
jstainer-kleis, line 4: dev-env is a directory not a respository

jstainer-kleis, line 4: "less practical", it depends for what. It's meant to test infrastructure modifications and software modification in a prod-like environment. It's less practical while developing because you need to redeploy instead of having hot reload.

infra/README.md
```diff
@@ -24,13 +41,15 @@
       - software selection: only SSH server and standard system utilities
 
 - Once the installation terminates and the VM has rebooted:
+  - if it is too slow, you can disable the graphical interface and work with command line: `sudo systemctl set-default multi-user` (becomes effective at restart)
```
jstainer-kleis, line 44: I'm not sure why the graphical interface would even be installed in the first place.
Even if it's the case, `systemctl disable gdm` should do the job, I'm not sure of the impact of changing the default boot target.

jstainer-kleis (changes requested):

glerzing:
> I'm not opposed to the addition, not really convinced by the value either, I think it's an obvious alternative. I'll let @amatissart decide whether or not to accept it.

I understand that it feels obvious when you have some experience with web deployment. If you think this won't help newcomers to get started more quickly, I can remove this PR.

infra/README.md
glerzing: indeed I meant directory

infra/README.md
glerzing: "software selection: only SSH server and standard system utilities" => I probably didn't uncheck something during the installation. I will remove the line I added.

GresilleSiffle:
> I'm not opposed to the addition, not really convinced by the value either, I think it's an obvious alternative. I'll let @amatissart decide whether or not to accept it.

I have the same feeling, but I think we can still get benefit from an improvement of the infrastructure documentation. If @glerzing had encountered issues during the environment set up, it may mean our documentation is not adapted.

I propose to have a discussion next week @glerzing , with @amatissart ( and @jstainer-kleis if you want ) so that you can explain us what exactly are the issues you faced, and how we can improve the `README.md`. Having used a remote virtual machine might mean our instructions regarding the deployment of a local one are incomplete.

What do you think @glerzing ?

jstainer-kleis:
> > I'm not opposed to the addition, not really convinced by the value either, I think it's an obvious alternative. I'll let @amatissart decide whether or not to accept it.
> 
> I have the same feeling, but I think we can still get benefit from an improvement of the infrastructure documentation. If @glerzing had encountered issues during the environment set up, it may mean our documentation is not adapted.
> 
> I propose to have a discussion next week @glerzing , with @amatissart ( and @jstainer-kleis if you want ) so that you can explain us what exactly are the issues you faced, and how we can improve the `README.md`. Having used a remote virtual machine might mean our instructions regarding the deployment of a local one are incomplete.
> 
> What do you think @glerzing ?

Fine with me, I agree that documentation of the infrastructure code may be improved and that people experiencing difficulties are in a good position to pinpoint the tricky parts.
My concerns revolve around the fact that the details of how to create the base VM vary a lot from one context to another, locally as well as remotely. The current documentation was written under the assumption that the user would be able to do this step in her own context.
Notify me if you organize a meeting, I'll try to join (but feel free to do without me in any case).

amatissart:
Thanks @glerzing for looking into this.

I agree with @jstainer-kleis that instantiating a remote VM is not necessarily simpler for newcomers. So it's probably not something we want to support or encourage as the "Option 1", especially as the configuration details may vary a lot depending on the provider.

It's nice to have a reminder here that "dev-env" should be preferred by contributors to test their changes. Perhaps the instructions about using a "cloud server" could be changed into a side note for people who know what they're doing?

glerzing:
I discussed with JST and agreed about the changes. I also set "renting a cloud server" as the option 2 following Adrien's comment.

glerzing:
The failed end-to-end test has nothing to do with this PR. I can reproduce this on my environment since I rebased. For some reason, the frontend container installation is just too slow and the script doesn't wait long enough for the frontend container (=> "frontend is unreachable")

infra/README.md
```diff
@@ -1,8 +1,16 @@
...
-- `./base-image/fetch-debian-image.sh`
+## Environment choice
+This section covers how to set a production-like environment. 
+For most developments, it is better to use the containers instead (see the dev-env directory). The infrastructure environment provides additional services like reverse-proxy (nginx) and monitoring, but is also usually not necessary and less practical for backend and frontend development tasks.
 
-## Installing in a VM for Infrastructure Code Development
+There are 2 options, renting a server to a cloud provider, or creating a test VM for development:
+- Setting up a VM (development only). It is potentially safer, since it is free and not exposed to the internet. This may significantly slow your computer on runtime. A specific network configuration like port forwarding is also needed to communicate with the VM.
+- Renting a server in the cloud is closer to the production environment. Most cloud providers (Vultr, Digital Ocean, AWS...) offer free credit for new users, but it can get costly if you rent a powerful server and forget to stop it after use.
+
```
GresilleSiffle, line 9: I suggest to remove this whole part also. The extra information contained in the " Renting " a server paragraph can be move below. See my next suggestions.

```suggestion
```

infra/README.md
```diff
@@ -1,8 +1,16 @@
...
+
+## Option 1: Installing a VM for Infrastructure Code Development
+
```
GresilleSiffle, line 12: Title already defined in a previous suggestion.

```suggestion
```

GresilleSiffle (changes requested):
Thanks @glerzing for highlighting the issues you've had with a local virtual machine installation, and for having proposed an alternative.

I didn't have enough time last week to schedule a meeting but after having discussed with @amatissart , here are my suggestions.

My objectives were to preserve the fact that there is an alternative to the local VM setup, and to explicitly notice the users that this method is no officialy supported (even if it should work).

I tried to factorize the installation instructions to avoid duplications, and I deliberately removed the mention of the Cloud providers name to not make any advertisement for them. I don't know them all, and I think recommending Cloud providers is an importact act as they have different policy regarding the data privacy.

glerzing:
I'm ok with GresilleSiffle's suggestions, it makes the Readme lighter.

GresilleSiffle (approved):

GresilleSiffle:
approved :+1: 

The PR is still waiting for the @jstainer-kleis approval.

Issue #396, opened 2022-01-01
feat(backend, testing): Set-up factory_boy to simplify writing tests


lfaucon:
See discussion: https://github.com/tournesol-app/tournesol/pull/394#discussion_r776635764

PR #424, opened 2022-01-07
Back/factory boy

sandre35:
Linked issue : https://github.com/tournesol-app/tournesol/issues/396

For the moment I only made changes on Video model.
I will adapt the other one in the hours to come

### DIFF
--- a/backend/tournesol/tests/test_api_ratings.py
+++ b/backend/tournesol/tests/test_api_ratings.py
@@ -2,8 +2,11 @@ from django.test import TestCase
 from rest_framework import status
 from rest_framework.test import APIClient
 
-from core.models import User
-from tournesol.models import Comparison, Video, ContributorRating, ContributorRatingCriteriaScore
+from tournesol.models import ContributorRating
+from tournesol.tests.factories.video import VideoFactory
+from tournesol.tests.factories.comparison import ComparisonFactory
+from core.tests.factories.user import UserFactory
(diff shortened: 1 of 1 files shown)


backend/core/factories/user.py
```diff
@@ -0,0 +1,15 @@
...
+class UserFactory(factory.django.DjangoModelFactory):
+
+    class Meta:
+        model = User
+
+    email = factory.LazyAttribute(
+        lambda a: '{}.{}@example.com'.format(a.first_name, a.last_name).lower()
+    )
+    first_name = name = factory.Sequence(lambda n: 'first name %s' % n)
+    last_name = name = factory.Sequence(lambda n: 'last name %s' % n)
```
amatissart, line 15: Not essential but nice to have, we can use Faker providers to generate realistic values 
https://factoryboy.readthedocs.io/en/stable/index.html?highlight=first_name#realistic-random-values

backend/tournesol/utils/video_id.py
```diff
@@ -0,0 +1,6 @@
+import random
+import string
+
+
+def generate_youtube_id():
+    return ''.join(random.choices(string.ascii_uppercase + string.digits, k=11))
```
amatissart, line 6: This functional will probably be used in tests only, so I would define it directly in the factory.

amatissart (commented):
It looks very good to me.   
Don't you think it would make sense to move the factories inside the "core/tests" and "tournesol/tests"?

amatissart (approved):

Issue #329, opened 2021-11-27
(feat) Multiple languages on the frontend


lfaucon:
- [x] Implement the internal mechanics required to support multiple languages
- [x] Translation to French
- [x] Language selector in top bar
- [x] Clear plan about how translation occurs when open source contributors implement a frontend feature
  #494 
- [ ] #455
- [x] Email verification and Password reset emails should be translated

lfaucon:
Updated TODO list with translation of emails sent to users

amatissart:
The last item will be specified in its own issue #455. I am closing.

PR #425, opened 2022-01-07
[back] feat: enable localized error messages in API

amatissart:
Related to #329 

Basically, this PR adds the `LocaleMiddleware` provided by Django, used for language detection:
https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#how-django-discovers-language-preference

As DRF provides translated error messages for its validation errors, they will be present in API responses and displayed by frontend consistently with the current language, based on the standard `Accept-Language` that has been configured in #385.

A small side-effect of this change is about caching. `cache_page` uses the current language in its cache key. So a custom decorator `cache_page_no_i18n` has been implemented to force the default locale on API views where caching a distinct response per language would be wasteful.

### DIFF
--- a/backend/tournesol/utils/cache.py
+++ b/backend/tournesol/utils/cache.py
@@ -1,6 +1,6 @@
 from django.conf import settings
-from django.views.decorators.cache import cache_page
 from django.utils import translation
+from django.views.decorators.cache import cache_page
 
 
 def cache_page_no_i18n(timeout: float):
--- a/backend/tournesol/utils/cache.py
+++ b/backend/tournesol/utils/cache.py
@@ -1,0 +1,25 @@
+from django.conf import settings
+from django.views.decorators.cache import cache_page
+from django.utils import translation
+
+
+def cache_page_no_i18n(timeout: float):
+    """
+    This decorator activates the default language before calling `cache_page`.
+
+    By default, `cache_page` includes the request language code in its cache key.
+    Some API responses (e.g public export files) don't depend on the language
+    and can be cached under a single key using this decorator.
+    """
+
+    def decorator(view):
+        cached_view = cache_page(timeout=timeout)(view)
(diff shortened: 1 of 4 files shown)


backend/tournesol/utils/cache.py
```diff
@@ -0,0 +1,25 @@
+from django.conf import settings
+from django.views.decorators.cache import cache_page
+from django.utils import translation
```
GresilleSiffle, line 3: We can swap these two import statements to keep them alphabetically ordered as recommended by `isort` ( which is recommended by Django ).

```suggestion
from django.utils import translation
from django.views.decorators.cache import cache_page
```

see: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports

GresilleSiffle (approved):
:+1: 

Approved with small coding style suggestion.

backend/tournesol/utils/cache.py
lfaucon: @GresilleSiffle Do you want to add isort to run automatically in Git-Hook / CI ? 

https://pycqa.github.io/isort/docs/configuration/git_hook.html

lfaucon (approved):

backend/tournesol/utils/cache.py
GresilleSiffle: Yes please : )

If no objection I would like too. I'll just finish the other tasks I'm working on before.

amatissart: At first I thought that could be enforced using `pylint`, but it seems pylint only checks that standard library modules are imported before external ones.

Still, a wider issue could be about enforcing `pylint` checks, possibly only a subset of the current rules if they are too constraining. For now only a global score is computed, with a threshold at 8.5 / 10.

PR #429, opened 2022-01-07
[extension] Fix a double alert bug

FabienRoger:
When the user was not connected to Tournesol, it received two consecutive alert asking him to connect. This fixes the issue.

### DIFF
--- a/browser-extension/utils.js
+++ b/browser-extension/utils.js
@@ -47,23 +47,23 @@ export const fetchTournesolApi = async (url, method, data) => {
 }
 
 export const addRateLater = async (video_id) => {
-  const addVideoResp = await fetchTournesolApi('video/', 'POST', {video_id: video_id});
-  if(!resp || resp.status !== 409){
+  const addVideoResponse = await fetchTournesolApi('video/', 'POST', {video_id: video_id});
+  if(!addVideoResponse || addVideoResponse.status === 409){
     return {
       success: false,
       message: 'Failed.'
     }
   }
 
-  const resp = 
+  const ratingStatusReponse = 
     await fetchTournesolApi('users/me/video_rate_later/', 'POST', {video: {video_id: video_id}});
-  if (resp && resp.ok) {
+  if (ratingStatusReponse && ratingStatusReponse.ok) {
     return {
       success: true,
       message: 'Done!'
     }
   }
-  if (resp && resp.status === 409) {
+  if (ratingStatusReponse && ratingStatusReponse.status === 409) {
     return {
       success: true,
       message: 'Already added.'
--- a/browser-extension/utils.js
+++ b/browser-extension/utils.js
@@ -48,7 +48,7 @@ export const fetchTournesolApi = async (url, method, data) => {
 
 export const addRateLater = async (video_id) => {
   const addVideoResponse = await fetchTournesolApi('video/', 'POST', {video_id: video_id});
-  if(!addVideoResponse || addVideoResponse.status === 409){
+  if(!addVideoResponse || addVideoResponse.status === 401){
     return {
       success: false,
       message: 'Failed.'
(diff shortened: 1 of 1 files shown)


browser-extension/utils.js
```diff
@@ -47,21 +47,23 @@ export const fetchTournesolApi = async (url, method, data) => {
...
-  const resp = await fetchTournesolApi('video/', 'POST', {video_id: video_id})
-    .then(() =>
-      fetchTournesolApi(
-        'users/me/video_rate_later/',
-        'POST',
-        {video: {video_id: video_id}}
-      )
-    );
+  const addVideoResp = await fetchTournesolApi('video/', 'POST', {video_id: video_id});
+  if(!resp || resp.status !== 409){
```
lfaucon, line 51: When logged in, I'm getting errors here, because `resp` is used before being defined. Did you mean `addVideoResp` (sidenote: I'd also prefer to use full names instead of abreviation `addVideoResponse`. I find that makes the code more readable)

```
Uncaught (in promise) ReferenceError: Cannot access 'resp' before initialization
```

lfaucon (changes requested):
FYI @GresilleSiffle  This is an unforseen side effect of recently adding required authentication on the POST /video/ endpoint.

Thanks for finding the issue and quickly adressing it. The code change looks good except for the small bug.

lfaucon:
@GresilleSiffle @amatissart 

Do our API response code make a lot of sense here?
- `POST /video/` with an existing video returns `400 {"video_id":["video with this video id already exists."]}`
- `POST /.../add_rate_later/` with an existing video already in rate later list returns `409 {"detail":"409 Conflict"}`

GresilleSiffle:
In our case, both HTTP code are correct, even if they are distinct.

I would use HTTP 409 Conflict, instead of 400 Bad Request, in `POST /video/` because 409 is more precise than a general " the request is not good ".

browser-extension/utils.js
GresilleSiffle: Should we use `.then()` here and `.catch()` to have the certitude that the promise has been fulfilled?

lfaucon: I find it easier to go with `try...catch` and `await`. And I dislike that we use sometimes one sometimes the other in our codebase currently

browser-extension/utils.js
```diff
@@ -47,21 +47,23 @@ export const fetchTournesolApi = async (url, method, data) => {
...
-    .then(() =>
-      fetchTournesolApi(
-        'users/me/video_rate_later/',
-        'POST',
-        {video: {video_id: video_id}}
-      )
-    );
-  if (resp && resp.ok) {
+  const addVideoResponse = await fetchTournesolApi('video/', 'POST', {video_id: video_id});
+  if(!addVideoResponse || addVideoResponse.status === 409){
```
GresilleSiffle, line 51: Hello @FabienRoger 

I managed to make the extension to work by using `401` instead of `409`. Currently the API automatically returns 401 Unauthorized when a call is made by an anonymous user.

```suggestion
  if(!addVideoResponse || addVideoResponse.status === 401) {
```

To be clear, the third-party Python package Django REST framework we used in the back end always returns a 401 or a 403 if a permission check fails.

If you want to better understand what's going on under the hood, you can check the code and its documentation.

- code  https://github.com/tournesol-app/tournesol/blob/main/backend/tournesol/views/video.py#L78
- docs  https://www.django-rest-framework.org/api-guide/permissions/#how-permissions-are-determined

GresilleSiffle (changes requested):

browser-extension/utils.js
FabienRoger: Ok, thank you for the explanation!

GresilleSiffle (approved):
It works!

Thanks @FabienRoger .

PR #438, opened 2022-01-10
[back] chore: set-up a single back end lint script

GresilleSiffle:
In the end, configuring `pre-commit` within our mono-repository configuration brings more complexity =/

`pre-commit` is a Python package, and it triggers the desired hooks before each commit

`isort`, `pylint` and `flake8` are Python packages that need to be run by `pre-commit`, so they need to be in the same virtual environment

it implies that `git commit` must always be run from a Python virtual environment (this doesn't feel right)

 different software in the mono-repository could requires different packages, so each sub-folder using Python hooks should maintain its own `requirements.dev.txt` with its own `.flake8` and `.pylintrc` configuration (it's not that hard to maintain)

BUT the front end doesn't require a virtual environment, and the `git commit` command fails if the pakcage `pre-commit` is not found

fine, let's manually patch `.git/hooks/pre-commit` to return 0 when the command `pre-commit` is not found

BUT manually editing is not sustainable as the modifications can be automatically reverted by `pre-commit` itself during the first installation and its subsequent migrations

So, we decided with @amatissart to keep the development process simple. This PR simply:
- adds `isort` to the development tools
- provides the `development` tools in the Docker API container
- describes in the back end `README.md` what tools are available, and how to run them

We also plan to integrate `isort` within the CI, but we need to fix the files before, otherwise all PR will be blocked.


backend/README.md
```diff
@@ -35,8 +57,70 @@ Both `dev-env/run-docker-compose.sh` and `dev-env/run-db-and-local-django.sh` de
...
+
+**automatic installtion with Docker**
+
+The commands should be run from the backend sub-folder.
+
+```shell
+docker exec tournesol-dev-api isort file1.py file2.py
+docker exec tournesol-dev-api pylint --rcfile=.pylintrc file1.py file2.py
+docker exec tournesol-dev-api flake8 --config=.flake8 file1.py file2.py
+```
```
amatissart, line 104: As discussed, I would suggest to add a lint script that could be called to run these tools in backend:
```
./scripts/lint.sh
```
or
```
docker exec tournesol-dev-api ./scripts/lint.sh
````

Or perhaps 2 different scripts `lint` and `lint-fix` similarly to the frontend scripts?

File: backend/README.md [hunk 2]
```diff
@@ -35,8 +57,70 @@ Both `dev-env/run-docker-compose.sh` and `dev-env/run-db-and-local-django.sh` de
...
+You can run them locally before creating a new commit, by running the
+following commands.
+
+**manual installation**
+
+The commands should be run from the root of the Git repository. 
+
+```shell
+# check file by file
+isort file1.py file2.py
```
amatissart, line 86: `isort` can also use a config file ".isort.cfg`

I think that something like this config would fit our needs:
```cfg
[settings]
skip = migrations
profile = black
line_length = 99
```

GresilleSiffle: Sure, I'll add this configuration in the next PR when I'll refactor all import statements.

backend/README.md
GresilleSiffle: I added `lint.sh`, and I plan to add `lint-fix` with the next pull request with `isort`.

backend/scripts/ci/lint.sh
```diff
@@ -0,0 +1,15 @@
...
+# return 0 if all checks return 0
+# 1 instead
+
+pylint --rcfile=.pylintrc ${@:-core tournesol}
+chk1=$?
+
+flake8 --config=.flake8 ${@:-}
+chk2=$?
+
+exit $chk1 && $chk2
```
amatissart, line 15: This does not seem to be correct. The exit code is always `$chk1`.

According to https://stackoverflow.com/a/16358989/5634396 a possible solution is
```suggestion
! (( chk1 || chk2 ))
```
```

GresilleSiffle: My bad, it's indeed not correct.

amatissart (approved):
Looks good to me. Perhaps the title of this PR should be updated before merging, as the lint script is not used as a pre-commit hook in the end.

PR #439, opened 2022-01-10
[backend] feat: Check freshness of video metadata on comparison creation

### LINKED ISSUE #277: [backend] feat: Mechanism for updating Video metadata
Displayed on Tournesol and may be outdated:

- Title
- Number of views
- Channel uploader

Missing metadata due to late implementation:

- duration

We seek to have a longterm solution that should work with our limit of 10k API calls per day, when we may assume 1k new videos added by users per day.

amatissart:
Closes #277 

* Refresh metadata about videos (fetched via Youtube API) on comparison creation, if metadata are **older than ~3 days~ 2 days**
* Implement an action available on the admin video list, to force the metadata refresh on selected videos
* Remove obsolete fields related to video metadata in `Video` 
* Replace `isodate` with `parse_duration` provided by Django


GresilleSiffle (approved):
:+1:

amatissart:
After discussion with @GresilleSiffle we reduced the expire delay from 3 days to 2 days.

Issue #413, opened 2022-01-05
PoC: Compare videos directly from Youtube with the extension


FabienRoger:
Do be honest, I find it difficult to have the motivation to go to the tournesol webpage and rate videos I marked as "rate later". Maybe I would use it more if in addition to the rate later button (which adds a video to the one in my library of seen videos), there was a "rate now" button, which would ask me to compare the current video to another I have seen on one or two criteria.
This idea might be tried out at first by making it an option of the extension  that isn't "on" by default.

What do you think about it? I didn't look into the app's security. Is there some security reason making this idea not feasible in practice?

lfaucon:
I am quite sure there is no security issue with this idea. We already have this button in the extension menu. It is just a matter of opening a new tab

FabienRoger:
I'm not speaking of opening Tournsol web app in a new tab, I think if you have to change tab you make the user make "too much of an effort". My idea would be something like this 
![image](https://user-images.githubusercontent.com/15524303/148291838-bdb0998e-4616-440a-85fa-26267af99f1d.png)

sintaz1992:
I think its a very good idea. I would say instead of having rate later button, I would prefer rate now and then a window can pop up for rating

lfaucon:
> I'm not speaking of opening Tournesol web app in a new tab, I think if you have to change tab you make the user make "too much of an effort".

Sorry for the misunderstanding. Yes, enabling comparinig videos from Youtube directly is a feedback we already received from a few users, so it would be great to initiate a PoC.

FabienRoger:
To make it easy to use, it could take the shape of a slim popup at the bottom of the page that appears when the user clicks on "rate now", like this (work in progress, of course, many things are missing from this popup, but this gives the general idea)
![image](https://user-images.githubusercontent.com/15524303/148646305-2b310657-4fe3-4cdd-989a-0445404edbe2.png)

lfaucon:
Looks good so far. 

It's fine to stick to only the "largerly recommended" criteria when not on Tournesol. (maybe with a link to open the comparison page on Tournesol). 

About Figma we don't stick to it too closely. It is more a guideline than a target so we do 20% of the work that gets us 80% of the style.

[...]

PR #459, opened 2022-01-14
Rate now button

FabienRoger:
See #413 

This feature is now added! But it needs a bit of polish. Help would be appreciated.


FabienRoger:
Some screenshots:
image

image

GresilleSiffle:
Hello @FabienRoger , hello everyone,

Thanks a lot for having designed and implemented this new feature :+1:  

The extension has evolved since this pull request has been open, and it's not possible to directly display the Tournesol application within an `iframe`. This open a lot of new possibilities.

Before adding more commits to this pull request, I suggest have a brainstorm with the rest of the team to design the comparison feature on YouTube. We will be able to decide where, why, and how we want users to compare videos.

Thanks!

PR #462, opened 2022-01-15
[front] Default recency set to 'month' in recommendations

lfaucon:
Discussed live with @GresilleSiffle , We think that makes the recommendation page more useful


GresilleSiffle (approved):
:+1:

PR #467, opened 2022-01-17
[backend] feat: Translate registration emails

amatissart:
Create templates for emails related to account registration, email update and password resets.  
The templates are derived from [rest_registration](https://github.com/apragacz/django-rest-registration/tree/0.7.0/rest_registration/templates/rest_registration), and translated in french only.

As they are the first translation and templates files introduced in our own Django app, I had to update the settings to configure the paths to these folders. And ensure that messages are compiled before deployment and during dev-env setup.

### DIFF
--- a/dev-env/docker-compose.yml
+++ b/dev-env/docker-compose.yml
@@ -28,10 +28,13 @@ services:
     container_name: tournesol-dev-api
     environment:
       - SETTINGS_FILE=/backend/dev-env/settings-tournesol.yaml
-    command: 
-        - bash
-        - -c
-        - "python manage.py migrate && python manage.py createcachetable && python manage.py runserver 0.0.0.0:8000"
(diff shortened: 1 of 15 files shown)


dev-env/docker-compose.yml
```diff
@@ -28,10 +28,13 @@ services:
...
-    command: 
-        - bash
-        - -c
-        - "python manage.py migrate && python manage.py createcachetable && python manage.py runserver 0.0.0.0:8000"
+    entrypoint: ["/bin/bash", "-c"]
+    command:
+      - |
+        python manage.py migrate
+        python manage.py createcachetable
+        django-admin compilemessages
```
GresilleSiffle, line 36: Is `django-admin compilemessages` different than `python manage.py compilemessages`.

GresilleSiffle (approved):

PR #470, opened 2022-01-17
[back] ci: add `isort` as a check

GresilleSiffle:
follows the work of the pull request #438 

---

### changed

To improve the back end linters collection, the `./scripts/ci/lint.sh` now also run the `isort --check-only` command. This change affects both the GitHub CI and the local script.

### added

A new script, `./scripts/ci/lint-fix.sh` will help the developpers to automatically fix some errors. Currently, only `isort` is able to fix the target files, hence the " some ". If we decide to use `black`, we will be able to add it to this script.

### future

The scripts`./scripts/ci/lint.sh` and `./scripts/ci/lint-fix.sh` can be integrated as tasks or sub-commands with Python project management tools like `pipenv` or `PDM`.

### DIFF
--- a/backend/.isort.cfg
+++ b/backend/.isort.cfg
@@ -1,0 +1,5 @@
+[settings]
+skip = migrations
+profile = black
+line_length = 99
+
--- a/backend/README.md
+++ b/backend/README.md
@@ -84,7 +84,9 @@ All of them should be automatically triggered by the continuous integration
 system using single script `scripts/ci/lint.sh`.
 
 You are encouraged to run this script locally before creating a new commit.
-Note that the script must be run from the root of the back end folder. 
+Note that the script must be run from the root of the back end folder.
+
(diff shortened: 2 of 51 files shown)


lfaucon:
Love this. Barely looked at the PR

Should we enforce hooks like isort or black from the root of the repository so that it also applies to other sub-directory using python such as the analytics and the soon-TM twitter-bot (ping @aidanjungo) ?

GresilleSiffle:
Thanks!

> Should we enforce hooks like isort or black from the root of the repository so that it also applies to other sub-directory using python such as the analytics and the soon-TM twitter-bot (ping @aidanjungo) ?

I don't know yet. I don't know if those projects could be different enough to require their own configuration, nevertheless I think it would be a very good idea :+1: 

I'll create an issue so that we can investigate how to handle this properly.

PR #475, opened 2022-01-17
[backend] feat: Define minimum number of contributors for a video to appear in recommendations

amatissart:
Note that it may also impact how videos are selected when clicking on "New video" on the Compare page.


GresilleSiffle:
It works :+1: 

I think I need to discuss a little bit to be convinced by how we have implemented  this feature.

One of the objectives of the browser extension is to recommend videos, and having a minimum 2 contributors hard limit will help to make quality recommendations and it sounds right to me. The same goes for the main recommendation stream in the Tournesol front end.

There is still something that worries me a bit. Great videos with only one contributor will be " silenced " by Tournesol, until someone else find those videos on YT and add a comparison. Given the fact that our community is still rather small and growing, a video can have only one contributor for several weeks, or months. It raises the question: Does Tournesol help effectively discovering " mute news " with this new behaviour?

We can say that a video having only one contributor is an edge case, but it's currently the case of 2919 of our videos. Explicitly pushing a video with one contributor in the main stream could help this video to be discovered and rated.

We might want to investigate more sustainable systems to prevent videos to stay with only one contributor:
- having a criteria in the recommendation page, dedicated to find only videos with 1 or 2 contributors, in order to encourage the community to rate them
- or having a dedicated section in the front end, listing rated videos with less than 2 contributors
- or something else

I think I'll be more at peace with this PR if we plan to think about such a system in the future.

GresilleSiffle (approved):

lfaucon:
I create two tickets for future work that would alleviate yor concerns and are very desirable features for us: #485 and #487

lfaucon (approved):

PR #477, opened 2022-01-17
[infra] Enable gzip compression in Nginx and precompress static files

glerzing:
original config file :
[nginx.txt](https://github.com/tournesol-app/tournesol/files/7884622/nginx.txt)

 new config file :
[new_nginx.txt](https://github.com/tournesol-app/tournesol/files/7884624/new_nginx.txt)

I discussed with JST of the BREACH vulnerability. We think it is not a problem in Tournesol.

A further improvement could be to use the compression algorithm Brotli for better performances. Brotli (compared to gzip) is slower for compression, slightly slower for decompression but gives a better compression ratio, for example :
main.45fd59e7.js => 939173 bytes (uncompressed)
main.45fd59e7.js.gz => 289515 bytes (gzip)
main.45fd59e7.js.br => 212442 bytes (brotli)
precompress by default also produces brotli files. But the difficulty is that to use brotli we need to install the nginx module ngx_http_brotli_static_module. Not sure this is possible when installing nginx with apt.
If you think adding Brotli is worth, I will see with JST how this could be done.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -28,11 +28,12 @@
     "redux": "^4.1.2",
     "redux-persist": "^6.0.0",
     "typescript": "~4.5.4",
-    "web-vitals": "^2.1.0"
+    "web-vitals": "^2.1.0",
+    "precompress": "7.0.1"
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "react-scripts build && precompress -i json,js,css,xml,svg,html,ttf build",
     "test": "react-scripts test --testURL=http://localhost:3000",
     "eject": "react-scripts eject",
     "update-schema": "wget -O scripts/openapi.yaml \"${REACT_APP_API_URL:-http://localhost:8000}/schema/\"",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build && precompress -i json,js,css,xml,svg,html,ttf build",
+    "build": "react-scripts build && precompress -t gz -i json,js,css,xml,svg,html,ttf build",
     "test": "react-scripts test --testURL=http://localhost:3000",
     "eject": "react-scripts eject",
     "update-schema": "wget -O scripts/openapi.yaml \"${REACT_APP_API_URL:-http://localhost:8000}/schema/\"",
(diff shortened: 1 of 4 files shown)


infra/ansible/roles/nginx/tasks/main.yml
```diff
@@ -54,25 +54,11 @@
...
-      ssl_ciphers 'ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256';
-      ssl_session_cache shared:SSL:10m;
-      ssl_session_tickets off;
-  notify:
-    - Reload Nginx
+- name: Set Nginx configuration
+  template:
+    src: nginx.conf.j2
+    dest: /etc/nginx/nginx.conf
+  notify: Reload Nginx
```
glerzing, line 61: instead of modifying the default configuration, I set a static configuration that is just copied.

jstainer-kleis (commented):
The trade-off here is the following:
- only modifying some parts of the configuration file makes easier applying them to a new version of the default configuration provided with the distribution package
- freezing the whole configuration file is much simpler and easier to read

In any case tracking upstream modifications of the package default configuration requires manual intervention, both approaches are fine with me.

frontend/package.json
```diff
@@ -28,11 +28,12 @@
...
     "redux-persist": "^6.0.0",
     "typescript": "~4.5.4",
-    "web-vitals": "^2.1.0"
+    "web-vitals": "^2.1.0",
+    "precompress": "7.0.1"
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "react-scripts build && precompress -i json,js,css,xml,svg,html,ttf build",
```
amatissart, line 36: `precompress` generates both `.gz` and `.br` by default. Let's keep `.gz` only:
```suggestion
    "build": "react-scripts build && precompress -t gz -i json,js,css,xml,svg,html,ttf build",
```

amatissart (commented):
Looks very good to me. And I agree the effort to set up brotli is not worth it.

Only a minor comment about the "precompress" usage.

amatissart (approved):

Issue #446, opened 2022-01-11
Disable creating a username that is an email address


aidanjungo:
As username is public (#444) we should not allow to use an email as username.

Also contact users who have a mail as username to ask them to change that.

Options: 
1. reject "@" in usernames
2. reject based on a regexp for emails: `r'.+@.+\..+'`

Include: 
- Adapt login component and routes to enable login with username OR email
- Password reset from email

PR #483, opened 2022-01-18
[backend] feat: Reject username with '@' on register and profile update

amatissart:
First step to address #446


GresilleSiffle (approved):
:+1:

Issue #426, opened 2022-01-07
[extension] fix: Improve authentication from extension


lfaucon:
Currently user's are often logged out if they have not open the website and the extension only display an alert and does not help the user logging back in.

We should investigate how to better handle the login process from the extension

GresilleSiffle:
I've been able to reproduce this behaviour by following these steps. I don't know if I'm exploring the right path, can anyone confirm?

- log in to Tournesol
- from the API administration interface, I manually delete my access_token
- my access_token is still available in my browser local storage, but can't be used to make any request
- try to add a YT video to my rate later list -> this triggers an alert inviting me to log in

lfaucon:
I think you are on the right path. It seems to me that you are simulating your token expiring and that this is one of the main cause of extension not logged in for me that is easy to fix. I'd say this is the priority

I'm not sure about the following second issue:
- From a fully not logged-in browser with extension installed
- Try to rate later
- View alert (normal)
- open new tab tournesol.app
- user is not logged in (normal)
- login, then immediately switch to the youtube tab and try to add to rate later again

I think this scenario fails to be logged in in the extension on the final try and requires to refresh Tournesol's tab after having logged in (75% sure of this)

Third possible improvement is to show a more useful alert, maybe opening the tab automatically (a bit weird) or providing a 1-click button to the login page (probably best)

Note also that we should be careful not to mess-up the website's auth with the extension's auth ^^

GresilleSiffle:
Alright, if the pull request #481 is merged, both cases will trigger the same error message.

Note that I was not able to reproduce the case 2, everything went well 10 times in row.

> Third possible improvement is to show a more useful alert, maybe opening the tab automatically (a bit weird) or providing a 1-click button to the login page (probably best)

Sure. I'm currently following this path: (1) display a really useful alert in a first PR (2) investigate a sustainable fix in second PR.

> Note also that we should be careful not to mess-up the website's auth with the extension's auth ^^

I'll be careful : )

PR #492, opened 2022-01-19
[extension] investigate about iframe and login

GresilleSiffle:
**related to**  #426 

---

This pull request allows the user to login directly from a YouTube video page, and make the extension able to automatically refreshes any expired access token in the background when a logged user arrived on a YouTube video page.

The new `iframe` currently available only on a YouTube video page could be used globally on YouTube, on the home page for instance, if we want the user to be able to connect from everywhere.

This PR also makes it possible to use the Tournesol comparison UI directly in an `iframe`, thanks to the `?embed=1` feature added by @amatissart . However it will still require a bit of code factorization, as I currently don't know how we want to design the comparison flow on YouTube.

**other additions**

New permissions are now required by the extension: `webRequest` and `webRequestBlocking`. They allow the extension to use the `webRequest` API needed by `background.js` to display the Tournesol application in an `iframe`.

There is now a `config/const.js` containing settings that can be shared by content scripts. Useful to define HTML id or CSS selectors without hardcoding them in different files.

The `addRateLaterButton.js` now uses `const` instead of `var`.

## Objectives

When a user click on the "Rate Later" button below a YouTube video, the extension:

**[case 1]** displays the login form in an `iframe` when the user is not logged
- [x] the login form is displayed, using the new front end embed feature 
- [x] `iframe` must be hidden by the extension after the login
- [x] shouldn't be hidden after a login failure
- [x] `iframe` has 100% width
- [x] `iframe` has a sufficient height to display the login form and the browser configuration instructions ( 490px? )
- [ ] automatically retry the rate later action after the login?

**[case 2]** automatically refreshes the access token without requiring any action from the user
- [x] add the `iframe` hidden in each video page
- [x] this must trigger an access token refresh in the background when needed 

**[case 0]** should detect, if possible, if the browser currently allows the cookies from Tournesol to bypass any blocking third-party cookies strategy. If such a configuration is detected the extension must inform the user that its required to accept cookies from Tournesol ( even if they are considered third-party on the YouTube page ). Else the extension won't be able to read the Tournesol `localStorage` in the `iframe`.
- [x] check if it's possible on Chrome/Chromium
- [x] check if it's possible on Firefox

The two previous features have been added in the front end by #493 

The **[case 2]** is still WIP as the user have to click two times on the "Rate Later" button. The first time allows the extension to refresh its token, the second time allows the user to effectively add the video to its rate-later list.
- [x] only one click should be needed

### Edge cases

**(solved)** When a user is connected to Tournesol, and have a YouTube video in an open tab. If the user logout from the Tournesol tab then switch to the YouTube video tab and click on `Rate Later`, the `iframe` actually display the Tournesol home page instead of the login form.

- [x] fix it

## Use cases non handled by this PR

Using the `Rate Later` button in the Tournesol extension menu doesn't display the login `iframe` yet. If we want to keep this button in addition to contextual `Rate Later` below the YouTube player, I'd prefer integrating the login flow in a second PR.

Using the `right click + Rate later on Tournesol` on a YouTube video doesn't display the login `iframe` neither. Adding this feature globally on YouTube will make this PR even bigger, and I'd prefer integrating the `iframe` in another PR.

Currently we provide three ways to add a video to a rate-later list, and this PR handles only one case:
- using the Tournesol menu **[todo]**
- using the Tournesol contextual button :heavy_check_mark: 
- using the right click on a video **[todo]**

The description "This action requires a connection to Tournesol" isn't actually translated, and appears always in English.

This is no close button at the moment. It's possible to close the `iframe` by pressing escape or clicking on the modal background.

## Notes

Using the extesion with a Tournesol integration in an `iframe` requires either:
- having a browser configured to accept third-party cookies ( not safe )
- having a browser blocking third-party cookies BUT explicitly accepting cookies from tournesol.app ( safer )

See the screenshots below to configure your browser.

### Firefox

Go to Settings > Privacy & Security [tab] > Cookies and Site Data [title] > Manage Exceptions... [button], then add `tournesol.app`. Don't forget to remove the auto-generated `http://` exception, only keep the secure `https://`.

capture
   
### Chrome/Chromium

See the settings `Block third-party cookies` and the customized section at the bottom of the screenshot.

capture

### login form and warning message

capture

capture2

### current login flow (the refresh button doesn't exist anymore)

https://user-images.githubusercontent.com/39056254/151147528-ff40a1ed-444a-4d4a-b601-a2e866daae42.mp4

### first WIP screenshot

capture

### old login flow, without modal

https://user-images.githubusercontent.com/39056254/150802806-c976f5bc-3897-4510-9e3f-c5c3bdb9eefa.mp4

### design of the close button (outdated)

capture


browser-extension/background.js
```diff
@@ -26,6 +32,35 @@ chrome.contextMenus.onClicked.addListener(function (e, tab) {
...
+ */
+chrome.webRequest.onHeadersReceived.addListener(
+  function(info) {
+    let headers = info.responseHeaders;
+    for (let i=headers.length - 1; i >= 0; --i) {
+      let header = headers[i].name.toLowerCase();
+      if (header == 'x-frame-options' || header == 'frame-options') {
+        headers.splice(i, 1);
+      }
+    }
```
amatissart, line 48: Can be made slightly more readable:
```suggestion
    const headers = info.responseHeaders.filter(
      h => !['x-frame-options', 'frame-options'].includes(h.name.toLowerCase())
    )
```

browser-extension/addRateLaterButton.css
```diff
@@ -20,3 +20,10 @@ img#tournesol-button-image {
   margin-right: 6px;
   vertical-align: middle;
 }
+
+iframe#x-tournesol-iframe {
+  display: none;
+
+  width: 100%;
+  height: 490px;
```
amatissart, line 28: To make the iframe a bit more esthetically pleasing, by removing the default border:
```suggestion
  height: 490px;
  border: none;
  margin: 0;
```

browser-extension/html/tournesol-iframe.html
```diff
@@ -0,0 +1,4 @@
+<iframe id="x-tournesol-iframe-inner"
+        src="https://tournesol.app/login?embed=1"
+        style="width: 100%; height: 99%;">
+</iframe>
```
amatissart, line 4: By default the `<iframe>` is wrapped in a `<body>` with a non-zero margin.  
Can be simplified with an explicit `<body>` to remove the arbitrary "99%": 

```suggestion
<body style="margin: 0;">
<iframe id="x-tournesol-iframe-inner"
        src="https://tournesol.app/login?embed=1"
        style="width: 100%; height: 100%; border: none; margin: 0;">
</iframe>
</body>
```

amatissart (commented):
A few minor comments, mostly about the iframe appearance.

Otherwise, these changes look very good to me, in order to make the authentication sync more robust between the website and the extension.

In some cases, the authentication error may be less visible than before (explicit alerts are no longer used). But I think it can improved separately, for example by adding the authentication status in the extension menu.

browser-extension/html/tournesol-iframe.html
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/492/commits/4709ab192ca7da1e39a317e5b19f2cc0c46261b9

browser-extension/addRateLaterButton.css
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/492/commits/b267e4d736e984a30339a947ee32e1cf353ab468

browser-extension/addModal.css
```diff
@@ -0,0 +1,24 @@
+#x-tournesol-modal {
+    top: 0;
+    left: 0;
+    position: fixed;
+
+    width: 100%;
+    height: 100%;
+
+    z-index: 1001;
```
amatissart, line 9: The YouTube top bar (`#masthead-container`) uses `z-index: 2020`, so we'll need to go higher :)
```suggestion
    z-index: 9999;
```

browser-extension/html/tournesol-iframe.css
```diff
@@ -0,0 +1,34 @@
...
+    margin: 0;
+    overflow: hidden;
+}
+
+.main {
+    width: 100%;
+}
+
+.heading {
+    height: 22px;
```
amatissart, line 15: Is it necessary to define an explicit "height" here?  
It breaks the layout in smaller screens:
image

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/492/commits/09919b277c2edf4c51654004715159c8c9f955e6

amatissart (approved):

PR #499, opened 2022-01-23
duration added to videorecommandation, score and num of contributors …

sintaz1992:
related to issue #388


browser-extension/addTournesolRecommendations.js
```diff
@@ -139,20 +139,26 @@ const getTournesolComponent = () => {
...
+        var formatted_video_duration = new Date(video.duration * 1000).toISOString().substring(11, 19)
+      }
+      
+      //var formatted_video_duration = video.duration;
+       
       // if (formatted_video_duration.startsWith('00:'))
       //   formatted_video_duration = formatted_video_duration.substring(
       //     3,
       //     formatted_video_duration.length,
       //   );
```
lfaucon, line 158: I think these comments are not useful to keep anymore

browser-extension/addTournesolRecommendations.js
```diff
@@ -139,20 +139,26 @@ const getTournesolComponent = () => {
...
-      // var formatted_video_duration = video.duration;
+       const video_duration = document.createElement('p');
+       video_duration.setAttribute('class', 'time_span');
+
+      // Convert SECONDS to hh:mm:ss or mm:ss format depending on the duration
+      if (video.duration<3600) {
+      var formatted_video_duration = new Date(video.duration * 1000).toISOString().substring(14, 19)
+      } else {
+        var formatted_video_duration = new Date(video.duration * 1000).toISOString().substring(11, 19)
+      }
```
lfaucon, line 150: You can greatly simplify the code with:

```suggestion
      const formatted_video_duration = new Date(video.duration * 1000).toISOString().substring(
          video.duration<3600 ? 14 : 11, 19
      )
```

browser-extension/addTournesolRecommendations.js
```diff
@@ -170,17 +176,18 @@ const getTournesolComponent = () => {
...
+       
+       const video_score = document.createElement('p');
+       video_score.className = 'video_text';
+       console.log(video)
+       video_score.append(
+         'Rated ' +
+           Number(video.rating_n_ratings).toFixed(0) +
+           ' points by ' +
+           video.rating_n_contributors +
+           ' users',
```
lfaucon, line 188: We now prefer to use the word "contributors"

browser-extension/addTournesolRecommendations.js
```diff
@@ -170,17 +176,18 @@ const getTournesolComponent = () => {
...
-      //     Number(video.tournesol_score).toFixed(0) +
-      //     ' points by ' +
-      //     video.rating_n_experts +
-      //     ' users',
-      // );
-      // details_div.append(video_score);
+       
+       const video_score = document.createElement('p');
+       video_score.className = 'video_text';
+       console.log(video)
```
lfaucon, line 182: Unnecessary console.log

lfaucon (changes requested):
Thanks! Great work.

Two extra things:
1. could you add a few pixels between the contributor information and the channel name?
Screenshot 2022-01-23 at 21 51 06

2. You can increment the patch version in manifest.json

browser-extension/addTournesolRecommendations.js
```diff
@@ -170,17 +176,18 @@ const getTournesolComponent = () => {
...
+       video_score.append(
+         'Rated ' +
+           Number(video.rating_n_ratings).toFixed(0) +
+           ' points by ' +
```
amatissart, line 186: `rating_n_ratings` is the number of comparisons the video received. So a text like "X comparisons by N users" would be more appropriate.

And if we want to display the Tournesol score, it would need to be computed from the "criteria_scores".

amatissart (dismissed):

browser-extension/addTournesolRecommendations.js
sintaz1992: @lfaucon  do we prefer to report Tournesol score like 0.45 points by 23 contributors or 14 comparions by 23 contributors? number of comaprisons is understandable BUT
my suggestion is that reporting score would not be very helpful in numbers but then we can think about a descriptiove way like among the top ten percent of scores, rated 50% more than others, somethign lie that but needs backend analysis from my understaning. what do you think?

browser-extension/addTournesolRecommendations.js
sintaz1992: what does it mean to incrase the patch version ?

browser-extension/addTournesolRecommendations.js
lfaucon: 1. I would prefer that we stick with "Rated X points by Y contributors". @sintaz1992 you use the wrong field in your implementation right now. You used the number of comparisons. If we choose to display "X comparisons by N users", then it is important that we implement a way to show the score too
2. About using percentiles, @sintaz1992  you may talk with @aidanjungo  about it on Discord, then make a proposal of how it would be computed , when , and how it would be displayed. if the team approves, we'll implement it.

browser-extension/addTournesolRecommendations.js
lfaucon: That probably scores well on the "relaxing" criteria :)
Nearly 3M views. How is it possible?

sintaz1992:
@amatissart Thanks for the fucntion of calcualtign time , I put it in the js file and use it..is there a better way to do so?

sintaz1992:
@lfaucon all your comments addressed except the patch version .. I didn't understand that one..
about rates, I suggest at the moment we show the "largly_recommanded" (if i;m not mistaken) but i will discuss about the idea of percentiles

browser-extension/addTournesolRecommendations.js
```diff
@@ -182,7 +182,9 @@ const getTournesolComponent = () => {
        video_score.className = 'video_text';
        video_score.append(
          'Rated ' +
-           Number(video.criteria_scores[0].score) +
+           Number(video.criteria_scores[video.criteria_scores.findIndex(element=>{
+           return  element.criteria==='largely_recommended'? true:false
+           })].score) +
```
lfaucon: You don't need to use `findIndex` then retrieve this index from the array. The `find` method does it for you: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

`element.criteria==='largely_recommended'? true:false` is a beginner mistake, simply `element.criteria==='largely_recommended'` is exactly the same thing

browser-extension/addTournesolRecommendations.js
```diff
@@ -19,6 +19,48 @@ document.addEventListener('yt-navigate-finish', process);
...
+      }
+      if (
+        criteria.score != undefined &&
+        criteria.score < min_score &&
+        criteria.criteria != 'largely_recommended'
+      ) {
+        min_score = criteria.score;
+        min_criteria = criteria.criteria;
+      }
+    });
```
lfaucon, line 59: This whole part of the method is not useful in our case, isn't it?

lfaucon (commented):
Starting to look really good :) 

> all your comments addressed except the patch version 

You should update the field `version` in the file `browser-extension/manifest.json`. But first you should update your branch with main. The latest version of the extension is `2.1.0`, you may increase it to `2.1.1` with your latest changes.

> about rates, I suggest at the moment we show the "largly_recommanded" (if i;m not mistaken) but i will discuss about the idea of percentiles

I think it is good that you use the calculateScore method for now so scores are the same as on the website!

GresilleSiffle:
Hello,

> You should update the field version in the file browser-extension/manifest.json. But first you should update your branch with main. The latest version of the extension is 2.1.0, you may increase it to 2.1.1 with your latest changes.

The current version of the extension has increased and is now **2.1.1**. For the moment this version has not been released, so it's not required to bump the version in this pull request.

Handling the extension version can be tricky. To avoid adding this complexity in pull requests, we could only update the version when a release is scheduled. This will avoid conflicts between pull requests.

lfaucon:
> The current version of the extension has increased and is now 2.1.1. For the moment this version has not been released, so it's not required to bump the version in this pull request.

> Handling the extension version can be tricky. To avoid adding this complexity in pull requests, we could only update the version when a release is scheduled. This will avoid conflicts between pull requests.

I disagree. With what you propose we might see version 2.x.y and not now exactly which version of the code was published. In the otherhand, if 1 change on teh extension => version change + conflict resolution we have better guarantee to trace code & version

GresilleSiffle:
Good point, I didn't think about this :+1:

sintaz1992:
Ready to merge!!

lfaucon (approved):

PR #505, opened 2022-01-25
[front] fix: Login form could be stuck in loading state

amatissart:
After a login request has been interrupted (e.g the tab has been closed), the token status can be persisted with the value "loading". 

In this situation, the login form should not be disabled, if no request has been submitted on the current page yet.

### DIFF
--- a/frontend/src/features/login/Login.tsx
+++ b/frontend/src/features/login/Login.tsx
@@ -17,6 +17,7 @@ const Login = () => {
   const { showErrorAlert } = useNotifications();
   const login: LoginState = useAppSelector(selectLogin);
   const dispatch = useAppDispatch();
+  const [formIsDirty, setFormIsDirty] = useState(false);
   const [username, setUsername] = useState('');
   const [password, setPassword] = useState('');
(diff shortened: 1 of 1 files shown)


frontend/src/features/login/Login.tsx
```diff
@@ -17,6 +17,7 @@ const Login = () => {
   const { showErrorAlert } = useNotifications();
   const login: LoginState = useAppSelector(selectLogin);
   const dispatch = useAppDispatch();
+  const [formIsDirty, setFormIsDirty] = useState(false);
```
lfaucon, line 20: Curious about the name choice. It doesn't tell much. Why not `formHasBeenSubmitted`?

lfaucon (approved):

frontend/src/features/login/Login.tsx
amatissart: A form is considered "dirty" when a user has modified any of the inputs, as opposed to a "clean" form. It's a common terminology in UI libraries. But in this case I agree it's unclear. I will follow your suggestion.

PR #508, opened 2022-01-27
[extension] feat: also display login iframe from browser action

GresilleSiffle:
**related to** #426 

**complete the work of** #492 

---

The extension is now able to display the Tournesol login form from the browser action popup.

### additional changes

Make the `addModal.js` content script the only script responsible of displaying and hiding the modal.

Make it more explicit that the login `iframe` is by default integrated in the modal ( if we decide in the future to make the modal able to display anything ).

Make some message listeners able to return a response, instead of just silently succeed or fail.

### DIFF
--- a/browser-extension/addModal.js
+++ b/browser-extension/addModal.js
@@ -27,6 +27,17 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
     if (modal) {
       modal.style.display = 'none';
     }
+    return;
+  }
+
+  if (request.message === "displayModal") {
+    const modal = document.getElementById(EXT_MODAL_ID);
+    modal.style.display = EXT_MODAL_VISIBLE_STATE;
+    if (modal) {
+      sendResponse({success: true});
+    } else {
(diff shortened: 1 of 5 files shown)


browser-extension/addModal.js
```diff
@@ -85,9 +94,49 @@ function addTournesolModal() {
...
+    return false;
+  }
+
+  const displayModal = function displayModal() {
+    modal.style.display = EXT_MODAL_VISIBLE_STATE;
+    iframe.removeEventListener('load', displayModal);
+  }
+
+  // prevent visual blink while refreshing the iframe
+  iframe.addEventListener('load', displayModal);
```
amatissart, line 122: Multiple functions are named `displayModal`. It's probably fine as the inner one uses `const`, but for readability a different name would help.

GresilleSiffle: Thanks!

Fixed by https://github.com/tournesol-app/tournesol/pull/508/commits/0beb7769e92478dc6c157e9111a660cebf7bedbc

amatissart (approved):

Issue #510, opened 2022-01-27
Actions that trigger a modal should be executed after the modal is closed (if auth is ok)


amatissart:
**Current flow:**
 1. Click on "Rate Later" (without being logged in)
 2. A modal appears with a login form
 3. Credentials are accepted, the modal is closed
 4. The user need to click on "Rate later" a second time to execute the action

**Expected flow:**

  The "rate later" action is executed automatically after the used logged in and the modal is closed

PR #511, opened 2022-01-27
[extension] feat: also display login iframe from context menu

GresilleSiffle:
**related to** #426 

**complete the work of** #508 

---

This pull request makes the login `iframe` appear when the users use the context action `Rate later on Tournesol` on a YouTube video.

It also make the automatic access token refresh process available on all YouTube pages. 

After the merge, and if I'm not forgetting anything, all rate later actions of the extension will trigger the login `iframe`.

The following steps are:
- enhanced the UX by adding a close button on top of the `iframe`
- https://github.com/tournesol-app/tournesol/issues/510
- add tests!

### DIFF
--- a/browser-extension/config/const.js
+++ b/browser-extension/config/const.js
@@ -15,7 +15,7 @@ const EXT_MODAL_VISIBLE_STATE = 'flex';
 const EXT_MODAL_INVISIBLE_STATE = 'none';
 // CSS selector identifying the element in the YT page the Tournesol modal
 // will wait for before being added to the DOM
-const EXT_MODAL_WAIT_FOR = 'div#info.ytd-watch-flexy';
+const EXT_MODAL_WAIT_FOR = 'div#content.ytd-app';
 
 // unique HTML id of the Tournesol login iframe
 const IFRAME_TOURNESOL_LOGIN_ID = 'x-tournesol-iframe-login';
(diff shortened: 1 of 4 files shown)


browser-extension/config/const.js
```diff
@@ -15,7 +15,7 @@ const EXT_MODAL_VISIBLE_STATE = 'flex';
 const EXT_MODAL_INVISIBLE_STATE = 'none';
 // CSS selector identifying the element in the YT page the Tournesol modal
 // will wait for before being added to the DOM
-const EXT_MODAL_WAIT_FOR = 'div#info.ytd-watch-flexy';
+const EXT_MODAL_WAIT_FOR = 'div#content.ytd-app';
```
amatissart, line 18: Any reason to wait for this specific element ? 
Now that the iframe is displayed as a modal, maybe it could be appended to the body without waiting for this element ?

amatissart (approved):
Looks good to me :+1:

browser-extension/config/const.js
GresilleSiffle: I currently don't know, as I've always tested the modal with this behaviour.

I'm going to test.

GresilleSiffle: It seems to work without waiting for `div#content`.

I just removed the waiting statement.

Issue #347, opened 2021-12-07
[backend] feat: Create new public API endpoint to access Tournesol statistics


aidanjungo:
Create a new public API endpoint `/stats/` which returns JSON Response containing:

- Number of users
- Number of new users since last 30 days
- Number of video
- Number of new videos since last 30 days
- Number of comparisons
- Number of new comparisons since last 30 days

Issue #509, opened 2022-01-27
[meta] discuss and plan the Tournesol models refactor


GresilleSiffle:
In order to make Tournesol able to work with different kinds of content, and not only YouTube videos, we would like with @amatissart to submit to discussion the following technical specifications.

## new database design

We propose three new models to replace the current `Video` model.

**(1) tournesol.models.Entity**

> other names we have thought of: candidate, alternative, content, scorable, comparable, ratable...

(I personally like `Candidate` more than `Entity` and more than the rest.)

This model will represent an entity that can be compared with other ones, using a specific set of criteria.

We forecast this new model will have the following attributes:
- `id` a unique identifier managed by the database
- `uid` a unique identifier that will be used by the API to reference the entity
- `type` a foreign key to `EntityType`
- `metadata` a JSONB PostgreSQL field holding all metadata related to the entity

The new `uid` field will replace the current `video_id` field of the `Video` model. It could contain a YouTube video ID, or any similar relevant string identifying the entity. To avoid collision between id coming from different sources (YouTube, Wikidata, etc.), we could use namespaces.

Examples:
- `youtube:videoid1abc` or `ytu:videoid1abc` could be the `uid` of the YT video `videoid1abc`
- `wikidata:Q0000000` or `wda:Q0000000` could be the `uid` of the Wikidata resource `Q0000000`

Such namespaces also allow to easily filter the entities by their source. The `type` will only indicate if the document is a video, a document, etc. and won't make the difference between videos coming from different sources.

As PostgreSQL don't perform validation for JSONB fields, we will need to validate the incoming JSON at the serializer level, before the entity is saved in the database, to avoid having incoherent JSON structure for entities of the same type.

**(2) tournesol.models.EntityType**

A type identifying an entity. It could be a video, a document, a podcast, etc.

**(3) tournesol.models.Scrutin**

> other possible names: Ballot, Election, ...

This new model will have a **n to n** relationship with the `Entity` model. A "scrutin" will make entities available for comparisons within its own scope.

For instance, the current tournesol.app application contains only one scrutin, allowing to compare YT videos. If we want Tournesol to not mix in the same ranking videos and presidents, we will be able to create two scrutins, that will make any API consumer able to know that a video can't be compared with a president.

It's still possible to mix entities with different `type` inside the same scrutin.  With this design we will cover all use cases:
- comparisons between a subset of YT videos
- comparisons between YT videos + PeerTube videos
[...]

PR #514, opened 2022-01-28
Public statistics API

farnots:
Linked to https://github.com/tournesol-app/tournesol/issues/347


aidanjungo:
Three small modif:
- count only active user
- count only video which have been rated at least once
- date should be the one from addition to Tournesol and not Youtube

backend/tournesol/tests/test_api_statistics.py
```diff
@@ -0,0 +1,133 @@
+
```
GresilleSiffle, line 1: We can remove the extra blank line here.

```suggestion
```

backend/tournesol/tests/test_api_statistics.py
```diff
@@ -0,0 +1,133 @@
...
+                duration_ms=302
+        )
+
+        Comparison.objects.filter(pk=comparison_1.pk).update(datetime_lastedit=timezone.now() - timedelta(days=5))
+        Comparison.objects.filter(pk=comparison_2.pk).update(datetime_lastedit=timezone.now() - timedelta(days=29))
+        Comparison.objects.filter(pk=comparison_3.pk).update(datetime_lastedit=timezone.now() - timedelta(days=60))
+            
+        self._list_of_comparisons = [comparison_1, comparison_2, comparison_3]
+       
+
```
GresilleSiffle, line 73: We can also remove the extra blank line here.

```suggestion
```

backend/tournesol/tests/test_api_statistics.py
```diff
@@ -0,0 +1,133 @@
...
+    def test_video_count_is_right(self):
+        """
+        Test if the number of video is right
+        """
+
```
GresilleSiffle, line 78: I'd like to propose to reword this a bit. The formulation `is_right` doesn't seem to be needed here, as there no test that will ensure something `is_wrong`.

How about something like this?

```suggestion
    def test_video_stats(self):
    """"
    An anonymous user can get statistics about videos.
    """"
```

backend/tournesol/tests/test_api_statistics.py
```diff
@@ -0,0 +1,133 @@
...
+        last_month_video_count = response.data["last_month_video_count"]
+
+        self.assertEqual(video_count, len(self._list_of_videos))
+        self.assertEqual(last_month_video_count, 2)
+
+    def test_user_count_is_right(self):
+        """
+        Test if the number of user is right
+        """
+
```
GresilleSiffle, line 98:
```suggestion
    def test_user_stats(self):
    """"
    An anonymous user can get statistics about users.
    """"
```

backend/tournesol/tests/test_api_statistics.py
```diff
@@ -0,0 +1,133 @@
...
+        last_month_user_count = response.data["last_month_user_count"]
+
+        self.assertEqual(user_count, len(self._list_of_users))
+        self.assertEqual(last_month_user_count, 1)
+
+    def test_video_comparison_is_right(self):
+        """
+        Test if the number of comparison is right
+        """
+
```
GresilleSiffle, line 118: I'd like to propose to reword this a bit. The formulation `is_right` doesn't seem to be needed here, as there no test that will ensure something `is_wrong`.

How about something like this?

```suggestion
    def test_comparison_stats(self):
    """"
    An anonymous user can get statistics about comparisons.
    """"
```

backend/tournesol/serializers.py
```diff
@@ -324,3 +324,12 @@ def validate(self, attrs):
...
     is_public = BooleanField()
+
+
+class StatisticsSerializer(Serializer):
+    user_count = IntegerField()
+    last_month_user_count = IntegerField()
+    video_count = IntegerField()
+    last_month_video_count = IntegerField()
+    comparison_count = IntegerField()
+    last_month_comparison_count = IntegerField()
```
GresilleSiffle, line 335: This serializer looks great, but I think it can look ever better : ]

If we want to add more stats in the future, for instance, the total number of users in addition to the total number of active users, we may want to reword some fields.

Also, the field `user_count` can be confusing as we are not returning the number of users, but the number of active users.

In order to keep the API as stable as possible, I'd suggest reword it now to keep " doors open ".

What do you think about this suggestion @farnots , @lfaucon and @amatissart ?

```python
class StatisticsSerializer(Serializer):
    users_active = IntegerField()
    users_active_last_month = IntegerField()
    videos_compared = IntegerField()
    videos_compared_last_month = IntegerField()
    comparisons = IntegerField()
    comparisons_last_month = IntegerField()
```

I like to use the model name as prefix (`users_`, `videos_`), as it allows to visually regroup statistics and allows to quickly identify which object is counted. The drawback is field names that are not correct English nouns.

As it's a personal preference I didn't use the GitHub suggestion feature.

backend/tournesol/views/stats.py
```diff
@@ -0,0 +1,76 @@
...
+from rest_framework.permissions import AllowAny
+from rest_framework.response import Response
+
+from core.models import User
+
+from ..models import Comparison, Video
+from ..serializers import StatisticsSerializer
+
+
+class Statistics:
```
GresilleSiffle, line 19: I'm wondering, should we move this into the models folder or somewhere else?

It's not an entity persisted in the database as is, so it's not a Django model, but it's still a representation of an object that needs to be serialized.

Leaving the class here is very OK for me, because it's small and it's the first of its kind.

backend/tournesol/views/stats.py
```diff
@@ -0,0 +1,76 @@
...
+        self.last_month_video_count = last_month_video_count
+
+    def add_comparison_statistics(self, comparison_count, last_month_comparison_count):
+        self.comparison_count = comparison_count
+        self.last_month_comparison_count = last_month_comparison_count
+
+
+@extend_schema_view(
+    get=extend_schema(
+        description="Retrieve statistics"
```
GresilleSiffle, line 38: We can use a dot at the end.

```suggestion
        description="Retrieve statistics."
```

GresilleSiffle (changes requested):
Thanks @farnots , very good work. 

Thanks for taking the time to add more tests, the code is really clean!

I allowed myself to be a bit picky on some method names, and made few suggestions for the future.

Do not hesitate to give feedback too : )

backend/tournesol/serializers.py
amatissart: Another solution would be to split the statistics into 3 objects: `active_users`, `compared_videos` and `comparisons`.
But this API is not critical and can be extended or changed with little difficulty, so any solution is reasonable in the short term.

In the longer term, we could wait for the specifications in #509, and define a more generic schema with statistics per ballot.  
Something like:  
```json
{
  "active_users": {
    "total": 123,
    "joined_last_month": 42
  },
  "ballots": [
    {
      "name": "youtube",
      "compared_entities": {
        "total": 1234,
        "added_last_month": 234
      },
      "comparisons": {
        "total": 12345,
        "edited_last_month": 2345
      }
    }
  ]
}
```

GresilleSiffle: I like this a lot!

As we are going to rework this endpoint after the " ballot " update, I propose to keep the serializer as is for now, and merge it to avoid having many blocked pull requests.

backend/tournesol/tests/test_api_statistics.py
farnots: Ok I made the change.

backend/tournesol/tests/test_api_statistics.py
farnots: Ok I made the change.

backend/tournesol/tests/test_api_statistics.py
farnots: Ok I made the change.

backend/tournesol/tests/test_api_statistics.py
farnots: Ok I made the change.

backend/tournesol/tests/test_api_statistics.py
farnots: Ok I made the change.

backend/tournesol/serializers.py
farnots: I like your suggestion @GresilleSiffle about the wording. Do you want me to update it now or I wait for a complete refactor when the ballot update will be merged ?

backend/tournesol/views/stats.py
GresilleSiffle: Let's keep it here for now then.

backend/tournesol/serializers.py
GresilleSiffle: Let's keep it as it for now :+1:

GresilleSiffle (approved):

Issue #306, opened 2021-11-18
Clean-up layout of Settings>Account and Profile


lfaucon:
Some screen sizes do not look good. See example for example: https://user-images.githubusercontent.com/4726554/142005959-f848abe8-ecff-438c-a508-721e22711d94.png

PR #516, opened 2022-01-28
Clean Account Settings Layout

farnots:
Linked to https://github.com/tournesol-app/tournesol/issues/306

https://user-images.githubusercontent.com/3254286/151596898-709aa87b-590e-4826-bae3-cd2333e36106.mp4

### DIFF
--- a/frontend/src/features/settings/account/EmailAddressForm.tsx
+++ b/frontend/src/features/settings/account/EmailAddressForm.tsx
@@ -10,7 +10,7 @@ import {
 } from '@mui/material';
 import { Link as RouterLink } from 'react-router-dom';
 
-import { FormTextField } from 'src/components';
+import TextField from '@mui/material/TextField';
 import { AccountsService, ApiError, UserProfile } from 'src/services/openapi';
 import { Lens as LensIcon, HelpOutline as HelpIcon } from '@mui/icons-material';
 import { useTheme } from '@mui/styles';
@@ -119,7 +119,7 @@ const EmailAddressForm = () => {
       <Grid container spacing={2} direction="column" alignItems="stretch">
         {isLoading && <CircularProgress />}
         {/* "display" is used here to keep the form state during loading. */}
-        <Grid item md={6} style={{ display: isLoading ? 'none' : undefined }}>
+        <Grid item style={{ display: isLoading ? 'none' : undefined }}>
           {profileData && (
             <Box
               marginBottom={2}
@@ -137,13 +137,13 @@ const EmailAddressForm = () => {
             </Box>
           )}
           <form onSubmit={handleSubmit}>
-            <FormTextField
+            <TextField
               required
               fullWidth
               label={t('settings.newEmailAddress')}
               name="email"
               type="email"
-              formError={formError}
+              error={formError}
               style={{ marginBottom: 16 }}
             />
             <Button
--- a/frontend/src/features/settings/account/EmailAddressForm.tsx
+++ b/frontend/src/features/settings/account/EmailAddressForm.tsx
@@ -143,7 +143,7 @@ const EmailAddressForm = () => {
               label={t('settings.newEmailAddress')}
               name="email"
               type="email"
-              error={formError}
+              formError={formError}
               style={{ marginBottom: 16 }}
             />
             <Button
--- a/frontend/src/features/settings/account/EmailAddressForm.tsx
+++ b/frontend/src/features/settings/account/EmailAddressForm.tsx
@@ -10,7 +10,7 @@ import {
 } from '@mui/material';
 import { Link as RouterLink } from 'react-router-dom';
 
-import TextField from '@mui/material/TextField';
+import { FormTextField } from 'src/components';
 import { AccountsService, ApiError, UserProfile } from 'src/services/openapi';
 import { Lens as LensIcon, HelpOutline as HelpIcon } from '@mui/icons-material';
 import { useTheme } from '@mui/styles';
@@ -137,7 +137,7 @@ const EmailAddressForm = () => {
             </Box>
           )}
           <form onSubmit={handleSubmit}>
-            <TextField
+            <FormTextField
               required
               fullWidth
               label={t('settings.newEmailAddress')}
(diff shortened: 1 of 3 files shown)


GresilleSiffle:
Hello @farnots 

Thanks for adding a preview video :+1: I'll review this pull request tomorrow.

frontend/src/pages/settings/account/Account.tsx
```diff
@@ -36,20 +36,39 @@ export const AccountPage = () => {
...
             sm={12}
             md={9}
+            lg={9}
+            xl={10}
             spacing={3}
           >
-            <SettingsSection title={t('settings.changeEmailAddress')} xs={12}>
+            <SettingsSection
+              title={t('settings.changeEmailAddress')}
+              xs={12}
```
GresilleSiffle, line 45: Since we re-use `xs={12}`, `sm={12}`, etc. several times in this file, how about using a variable to store these values?

In the `AcountPage` component body we could have:
```javascript
const breakP = { xs: 12, sm: 12, md: 9, lg: 9, xl: 10 };
```
and in the returned JSX:
```jsx
xs={breakP.xs}
sm={breakP.sm}
md={breakP.md}
lg={breakP.lg}
xl={breakP.xl}
```

GresilleSiffle (dismissed):
The new breakpoints looks very good, thank you.

In the following review, I've requested to revert one change about the `EmailAddressForm` which introduces a new bug.

I'm not sure about the relevance of my comments about the breakpoints. In fact I don't know if there is a way to not repeat the breakpoints definition in all `<SettingsSection>` components, and just tell them to fit the parent's breakpoints instead.

Maybe @lfaucon and @amatissart you'll have better ideas?

frontend/src/pages/settings/account/Account.tsx
amatissart: > and in the returned JSX:
> 
> ```js
> xs={breakP.xs}
> sm={breakP.sm}
> md={breakP.md}
> lg={breakP.lg}
> xl={breakP.xl}
> ```

can also be written using the [spread syntax](https://reactjs.org/docs/jsx-in-depth.html#spread-attributes):
```jsx
<SettingsSection title="Title" {...breakP}>
```

frontend/src/features/settings/account/EmailAddressForm.tsx
```diff
@@ -137,13 +137,13 @@ const EmailAddressForm = () => {
             </Box>
           )}
           <form onSubmit={handleSubmit}>
-            <FormTextField
```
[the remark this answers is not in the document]
amatissart: Yes, keeping the `FormTextField` seems preferable.  
It is basically a wrapper for `TextField`, so most of the props such as `fullWidth` or `size` can be passed to override the default values. And its implementation can be adjusted if necessary.

frontend/src/features/settings/account/PasswordForm.tsx
```diff
@@ -92,7 +90,6 @@ const PasswordForm = () => {
             label={t('settings.confirmNewPassword')}
             name="password_confirm"
             color="secondary"
-            size="small"
```
amatissart, line 95: Why removing `size="small"` specifically on this page ?  
This introduces some inconsistencies between pages, as all other form fields in the app use `size="small"`. 

And I tend to prefer small text fields, to make forms easier to fit on a single screen, and with a more native look-and-feel; although it's mostly a matter of taste as the design of these forms is not currently specified,

amatissart:
> I'm not sure about the relevance of my comments about the breakpoints. In fact I don't know if there is a way to not repeat the breakpoints definition in all `<SettingsSection>` components, and just tell them to fit the parent's breakpoints instead.
> 
> Maybe @lfaucon and @amatissart you'll have better ideas?

The breakpoints defined on the parent `Grid` and on the `<SettingsSection>` don't represent the same thing; the first ones define the relative width of the settings _section_, besides the settings menu. The second ones define the relative width of the settings _content_ below the section title (as the title always fits the parent width to display an underline). So using the same breakpoints on these 2 components is not necessarily what we want.

It could make sense indeed to change the `SettingsSection` implementation to fit the parent width (equivalent to set `xs={12}`). Nevertheless, all children would then be responsible for configuring their breakpoints, so not sure it would solve anything.

Another approach could be to remove `fullWidth` prop from form fields and buttons or to set a `max-width` on the section content. It would make explicit breakpoints unnecessary, and would possibly simplify the layout.

frontend/src/features/settings/account/EmailAddressForm.tsx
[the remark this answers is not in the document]
farnots: You right, I didn't saw any modification during my tests so I changed it to stay consistent.

farnots:
I made the changes, I'm not sure about your last comment @amatissart and your suggestion

GresilleSiffle (commented):
Thanks @farnots , several points have been resolved :+1: 

About the break points: it currently works. Using the same break points for the container and the different sections makes sense visually. Using the same variable in addition to using the same value might lead to unexpected behaviour the day we want to change the section layout without changing the container layout.

I got a proposition, how about using different variables with the same values? To make it explicit that we can change their values independently, and to prevent developers to consider them as coupled.

Like this, in the `<AccountPage>` component.

```javascript
  // sectionBreakP can be changed independently of breakP
  const breakP = { xs: 12, sm: 12, md: 9, lg: 9, xl: 10 };
  const sectionBreakP = breakP;
  ``` 

and in the `<SettingsSection>`

```javascript
<SettingsSection title={t('something')} {...sectionBreakP}>
```

amatissart (approved):

amatissart:
> ```js
>   // sectionBreakP can be changed independently of breakP
>   const breakP = { xs: 12, sm: 12, md: 9, lg: 9, xl: 10 };
>   const sectionBreakP = breakP;
> ```
> 
> and in the `<SettingsSection>`
> 
> ```js
> <SettingsSection title={t('something')} {...sectionBreakP}>
> ```
> 
> What do you think @amatissart ?

Why not. Either solution is fine. We can change it easily if necessary.  
I approved as it is now :heavy_check_mark:

Issue #461, opened 2022-01-15
[frontend] feat: filter My Comparison page


lfaucon:
On a video card, "{N} comparisons" (see screenshot below) should be a clickable link that takes the user to the "My Comparisons" page filtered for this video. 

Notes:
- The API endpoint for fetching the comparisons already exist.
- For the filter we should use a similar style as for the filter on the Uploader/Channel from the Recommendations page

![Screenshot 2022-01-15 at 16 48 46](https://user-images.githubusercontent.com/8818081/149628092-14e0b6dd-7a55-4862-97bf-95d40ef10660.png)

PR #518, opened 2022-01-30
hyperlink added to compariosn page, ON MYCOMPARIOSN, useEFFECT update…

sintaz1992:
realted to issue #461 

Similar to the filter on channel on recommendation page, I added a filter on comparioslist component

The main change is how useEffect works: before it was just updating if offest changes, now its senestive to changes in location.search for video and offset

if video has no comparions, it will not diret the user to this page from comparion page

### DIFF
--- a/frontend/src/features/videos/PublicStatusAction.tsx
+++ b/frontend/src/features/videos/PublicStatusAction.tsx
@@ -40,9 +40,11 @@ export const UserRatingPublicToggle = ({
     <Box display="flex" alignItems="center" flexWrap="wrap" width="100%" px={1}>
       <Typography variant="caption" style={{ fontSize: '11px' }}>
         {nComparisons > 0 ? (
-          <Trans t={t} i18nKey="video.nComparisonsByYou" count={nComparisons}>
-            {{ count: nComparisons }} comparison by you
-          </Trans>
+          <a href={`/comparisons/?video=${videoId}`}>
+            <Trans t={t} i18nKey="video.nComparisonsByYou" count={nComparisons}>
+              {{ count: nComparisons }} comparison by you
+            </Trans>
+          </a>
         ) : (
           t('video.notYetComparedByYou')
         )}
--- a/frontend/src/features/videos/PublicStatusAction.tsx
+++ b/frontend/src/features/videos/PublicStatusAction.tsx
@@ -1,7 +1,7 @@
 import React, { useContext } from 'react';
 import { useTranslation, Trans } from 'react-i18next';
-import { Tooltip, Typography, Box, Switch } from '@mui/material';
-
+import { Tooltip, Typography, Box, Switch, Link } from '@mui/material';
+import { Link as RouterLink } from 'react-router-dom';
 import { UsersService, ContributorRating } from 'src/services/openapi';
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/comparisons/ComparisonList.tsx
```diff
@@ -18,19 +18,31 @@ function ComparisonsPage() {
...
+          limit: limit,
+          offset: offset,
+        }).then((data) => {
+          setComparisons(data.results);
+          setCount(data.count || 0);
+        })
+      : UsersService.usersMeComparisonsList({ limit, offset }).then((data) => {
+          setComparisons(data.results);
+          setCount(data.count || 0);
+        });
```
lfaucon, line 42: This code could be slightly refactored to avoid repeating the piece:
```TypeScript
      .then((data) => {
          setComparisons(data.results);
          setCount(data.count || 0);
        });
```

Also we prefer using `await` for asynchronous calls, so we are slowly removing the `.then` when possible (@tournesol-app/maintainers do you confirm?)


```suggestion
    const process = async () => {
      const comparisonsRequest = await (filteredVideo
        ? UsersService.usersMeComparisonsListFiltered({
            videoId: filteredVideo,
            limit,
            offset,
          })
        : UsersService.usersMeComparisonsList({ limit, offset }));
      setComparisons(comparisonsRequest.results);
      setCount(comparisonsRequest.count || 0);
    };
    process();
```

frontend/src/features/videos/PublicStatusAction.tsx
```diff
@@ -40,9 +40,11 @@ export const UserRatingPublicToggle = ({
...
       <Typography variant="caption" style={{ fontSize: '11px' }}>
         {nComparisons > 0 ? (
-          <Trans t={t} i18nKey="video.nComparisonsByYou" count={nComparisons}>
-            {{ count: nComparisons }} comparison by you
-          </Trans>
+          <a href={`/comparisons/?video=${videoId}`}>
+            <Trans t={t} i18nKey="video.nComparisonsByYou" count={nComparisons}>
+              {{ count: nComparisons }} comparison by you
+            </Trans>
+          </a>
```
lfaucon, line 47: Screenshot 2022-01-30 at 22 02 24

Please avoid the blue & purple styles for the text. 

The best would be to have a similar style as for the channel name and possibly a similar Tooltip which may say "See my comparisons".

Screenshot 2022-01-30 at 22 03 00

lfaucon (changes requested):
💯  Really awesome feature. Thanks! Only two small comments

frontend/src/features/videos/PublicStatusAction.tsx
amatissart: The implemention of this link is visible here:
https://github.com/tournesol-app/tournesol/blob/1fcb04e2f889e4a9204bdf98bdf2e505acb17b4a/frontend/src/features/videos/VideoCard.tsx#L286-L294

`<Link>` is provided by MUI to get consistent styling.
`<RouterLink>` is another component provided by react-router to navigate internally in the single-page app (whereas clicking on`<a>` would trigger an unnecessary page reload).

sintaz1992:
Ready to merge

lfaucon (approved):
LGTM

Issue #487, opened 2022-01-18
feat: Unsafe recommendations


lfaucon:
Two types of videos are now filtered out of the list of recommended videos: scores < 0 and video compared by only one contributor.

We may enable specifying `unsafe=True` (or similar) in the Tournesol recommendation query and display such videos with a warning on the Video Card with appropriate explanations.

score < 0: may be good but below average of the videos added to Tournesol

only one contributor: may be good but we have low confidence as this has been rated by only one contributor

- [x] Update Backend route
- [x] Feature for video card to display warning signs
- [x] Add Filtering option in frontend recommendations to turn on/off the unsafe option (safe by default)

PR #531, opened 2022-02-02
[#487] Back structure for unsafe videos

farnots:
Linked to #487

### DIFF
--- a/backend/tournesol/views/video.py
+++ b/backend/tournesol/views/video.py
@@ -55,6 +55,11 @@ from ..serializers import VideoSerializer, VideoSerializerWithCriteria
                 "Accepted formats: ISO 8601 datetime (e.g `2021-12-01T12:45:00`) "
                 "or legacy: `dd-mm-yy-hh-mm-ss`."
             ),
+            OpenApiParameter(
+                "unsafe",
+                OpenApiTypes.BOOL,
+                description="Return videos mark as unsafe"
+            ),
             *[
                 OpenApiParameter(
                     crit,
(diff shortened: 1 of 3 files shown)


backend/tournesol/views/video.py
```diff
@@ -154,13 +159,24 @@ def get_queryset(self):
...
-            queryset.annotate(
-                total_score=Sum(F("criteria_scores__score") * criteria_weight)
-            )
-            .filter(total_score__gt=0)
-            .order_by("-total_score")
+
+        unsafe = request.query_params.get('unsafe')
+        show_unsafe = False
+        if unsafe:
+            show_unsafe = unsafe == 'true'
```
amatissart, line 166: (minor) Can be simplified for readibility:
```suggestion
        show_unsafe = request.query_params.get('unsafe') == 'true'
```

backend/tournesol/views/video.py
```diff
@@ -154,13 +159,24 @@ def get_queryset(self):
...
         )
+
+        if show_unsafe is True:
+            queryset = queryset.order_by("-total_score")
+        else:
+            queryset = (
+                queryset
+                .filter(total_score__gt=0)
+                .order_by("-total_score")
+            )
```
amatissart, line 179: When `show_unsafe` is True, we would also expect to include videos without a minimum of contributions. So the filter applied on line 112 needs to be moved too.

farnots: I miss it thanks

backend/tournesol/views/video.py
```diff
@@ -154,13 +162,19 @@ def get_queryset(self):
...
-            )
-            .filter(total_score__gt=0)
-            .order_by("-total_score")
+
+        queryset = queryset.annotate(
+            total_score=Sum(F("criteria_scores__score") * criteria_weight)
         )
+
+        if show_unsafe is True:
+            queryset = queryset.order_by("-total_score")
```
amatissart, line 171: Now that no minimal number of contributions is required, this queryset would also include videos without any rating (e.g videos that has been added to a rate-later list), and with score being null. So an additional filter is necessary:
```suggestion
            queryset = (
                queryset
                .filter(total_score__isnull=False)
                .order_by("-total_score")
            )
```

backend/tournesol/views/video.py
amatissart: I would suggest to move the filter on `rating_n_contributors` here in the same condition. I think that it would help the reader to grasp all what is going on with this safe/unsafe filters.

farnots: Thanks I updated

backend/tournesol/views/video.py
farnots: Done

amatissart (approved):

Issue #525, opened 2022-02-02
refactor: transform the `Video` model into `Entity`


GresilleSiffle:
also see the issue #509 

---

Transform the current model `Video` into the generic model `Entity`.

Keep the current `Video` fields, and add the following:
- `uid` CharField, UNIQUE NOT NULL (with an index) max_length 144  
- `type` CharField, choice, only one choice for the moment -> `video`
- `metadata` JSONB
 
Delete the legacy fields coming from Tournesol v1.

Keep the current `Video` methods in the model `Entity`.

Do not fill the `metadata` field yet, to avoid data duplication between the table columns and this field.

Replace all mentions of `Video` by `Entity`.

PR #536, opened 2022-02-03
[back] refactor: rename `Video` model in `Entity`

GresilleSiffle:
**related to** #525 

---

This pull request replaces the models `Video` and `VideoCriteriaScore` by `Entity` and `EntityCriteriaScore`.

All fields specific to YT videos has been kept in the model `Entity` and will be deleted when the method responsible of handling `metadata` saving is implemented.

The back end and the front end are still compatible according to the CI.

There are still models using a field named `video` to reference an `Entity`. I suggest to refactor them just after this pull request, and not a the same time to keep the diff small. We could also run `black` just after.

## todo

**step 1** - rename
- [x] rename the `Video` model in `Entity`
- [x] rename the `VideoCriteriaScore` model in `EntityCriteriaScore`
- [x] the migrations can easily be run backward
- [x] adapt the `import` statements and the tests

**step 2** - remove the legacy `Entity` fields, not used in Tournesol 2.0
- [x] remove the fields
- [x] remove legacy models too

**step 3** - add the new `Entity` generic fields
- [x] add `uid`
- [x] add `type`
- [x] add `metadata`
- [x] auto-fill `uid` and `type` with a migration
- [x] `uid` and `type` shouldn't have default value enforced by the database

**step 4** - ensure newly created entities are considered videos coming from YT
- [x] add a test

**step 5** - update the admin interface
- [x] add the possibility to filter entities by `type`
- [x] replaced the old `video_id` by the new `uid` column 

## questions

Concerning the **step 2**, we've identified with Adrien three fields that supposedly can be deleted:
- `Entity.embedding`, this field is always null in production
- `Entity.info`, this field is sometimes null sometimes an empty string in production
- `Entity.pareto_optimal`, 69 rows with true in database, 6900 rows with false

If some of them are legacy fields from Tournesol 1.0, it could be the good moment to delete them and remove the associated dead code.


backend/tournesol/models/entity.py
```diff
@@ -25,51 +25,97 @@
...
         YOUTUBE_VIDEO_ID_REGEX, f"Video ID must match {YOUTUBE_VIDEO_ID_REGEX}"
     )
 
+    # TODO: will be replaced by the `uid` field
     video_id = models.CharField(
         max_length=20,
         unique=True,
         help_text=f"Video ID from YouTube URL, matches {YOUTUBE_VIDEO_ID_REGEX}",
         validators=[video_id_regex],
     )
```
amatissart, line 72: After this PR is deployed, this `video_id` is now redundant with `uid`.  

It does not seem that the migration has been specified yet. Should we remove it at the same time as other fields specific to videos, or can it be replaced with a `@property` now ?

backend/tournesol/admin.py
```diff
@@ -9,15 +9,15 @@
...
-    Video,
-    VideoCriteriaScore,
+    Entity,
+    EntityCriteriaScore,
 )
 
 
-@admin.register(Video)
+@admin.register(Entity)
 class VideoAdmin(admin.ModelAdmin):
```
lfaucon, line 18: Looks like we did not anticipate the need to rename the `VideoAdmin` and `VideoCriteriaScoreAdmin` class? 

I'd recommend renaming it right away, because it is easy and it would be very confusing for someone unaware of the current refactor.

backend/tournesol/admin.py
```diff
@@ -38,7 +39,7 @@ def update_metadata(self, request, queryset):
             video.refresh_youtube_metadata(force=True)
```
lfaucon, line 39: Could filter entities by `type="video"`

```Python
    videos = [entity for entity in queryset if entity.type == Entity.TYPE_VIDEO]
    for video in videos:
        video.refresh_youtube_metadata(force=True)
```

backend/tournesol/migrations/0023_make_new_entity_fields_notnull.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.11 on 2022-02-03 14:29
```
lfaucon, line 1: Any reason to create 5 migration files? If not, you could simply concatenate the list of operations.

backend/tournesol/models/entity.py
```diff
@@ -25,51 +25,97 @@
...
 
-class Video(models.Model, WithFeatures, WithEmbedding):
-    """One video."""
+class Entity(models.Model, WithFeatures, WithEmbedding):
+    """
+    A generic entity that can be compared with another one.
 
+    The current model still contains fields from the previous `Video` model.
+    These fields are kept as-is for now to ease the refactor of the Tournesol
+    app, and will be replaced in the future by the `metadata` JSON field.
```
lfaucon, line 34: How about leaving `TODO` formatted comments to simplify coming back for clean-up where needed?

Edited: héhé, I see you did this a lot below. 👌

backend/tournesol/models/entity.py
```diff
@@ -86,29 +132,34 @@ class Video(models.Model, WithFeatures, WithEmbedding):
...
+    # TODO: specific, will be moved to the metadata
     tags = models.ManyToManyField(Tag, blank=True)
 
-    # computed in the Video.recompute_pareto(),
+    # computed in the Entity.recompute_pareto(),
     #  called via the manage.py compute_quantile_pareto command
     # should be computed after every ml_train command (see the devops script)
     # SEE also: {feature}_quantile fields (defined below)
 
+    # XXX is that used?
```
lfaucon, line 147: Not anymore, but could be again part of a ML feature for Entities. Let's delete for now.

lfaucon (commented):
Some of my comments are probably things planned for later (I don't have the content of all next tickets in mind). Feel free to do things in the order you intend 👍 

All I found marked for deletion seemed good to delete indeed.

backend/tournesol/models/entity.py
GresilleSiffle: Yup, I've also regrouped all fields inside to-do blocks to avoid repeating the to-do comment.

backend/tournesol/migrations/0023_make_new_entity_fields_notnull.py
GresilleSiffle: There is indeed no reason to keep all migrations files. I'll regroup them in coherent operations.

backend/tournesol/models/entity.py
GresilleSiffle: Deleted.

backend/tournesol/models/entity.py
```diff
@@ -25,51 +25,97 @@
...
     name = models.CharField(max_length=1000, help_text="Video Title", blank=True)
+    # TODO: specific, will be moved to the metadata
     description = models.TextField(
         null=True, help_text="Video Description from the web page", blank=True
     )
+
+    # TODO: specific, will be moved to the metadata
     caption_text = models.TextField(
         null=True, help_text="Processed video caption (subtitle) text", blank=True
     )
```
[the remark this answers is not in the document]
GresilleSiffle: Deleted.

backend/tournesol/admin.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/536/commits/ad6d4de42589b04634fc98feaabc7b7dba87d8f7.

backend/tournesol/admin.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/536/commits/ad6d4de42589b04634fc98feaabc7b7dba87d8f7.

backend/tournesol/migrations/0021_fill_new_entity_fields.py
```diff
@@ -0,0 +1,28 @@
...
+def migrate_forward(apps, schema_editor):
+    """
+    Set the `uid` and the `type` field of all entities present in the
+    database (supposedly only YT videos are present).
+    """
+    Entity = apps.get_model("tournesol", "Entity")
+    for entity in Entity.objects.iterator():
+        entity.uid = "{}:{}".format(ENTITY_UID_YT_NAMESPACE, entity.video_id)
+        entity.type = ENTITY_TYPE_VIDEO
+        entity.save()
```
amatissart, line 19: "update_fields" can be used here, to avoid potential conflicting updates during the migration (and slightly better performance).
```suggestion
        entity.save(update_fields=["uid", "type"])
```

amatissart (approved):

PR #538, opened 2022-02-04
#487 Unsafe recommandations Front Card Style

farnots:
Linked to #487

I try different style but the simpler one with a tooltips seams the most space convenient.

CleanShot 2022-02-04 at 16 26 00@2x
CleanShot 2022-02-04 at 16 25 51@2x
 

I set to merge on my backend PR, does it's seams right for you ?

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -163,7 +163,10 @@
     "nbContributors_other": "{{count}} contributors",
     "criteriaRatedHigh": "Rated high:",
     "criteriaRatedLow": "Rated low:",
-    "labelShowSettings": "Show settings related to this video"
+    "labelShowSettings": "Show settings related to this video",
+    "unsafeNotEnoughContributor": "May be good but we have low confidence as this has been rated by only one contributor",
+    "unsafeNegativeRating" : "May be good but below average of the videos added to Tournesol"
+
   },
   "notifications": {
     "errorOccurred": "Sorry an error has occurred.",
--- a/frontend/src/features/videos/VideoCard.tsx
+++ b/frontend/src/features/videos/VideoCard.tsx
@@ -23,6 +23,7 @@ import {
   ExpandMore as ExpandMoreIcon,
   ExpandLess as ExpandLessIcon,
 } from '@mui/icons-material';
+import WarningIcon from '@mui/icons-material/Warning';
 import { convertDurationToClockDuration } from 'src/utils/video';
 
 const useStyles = makeStyles((theme) => ({
@@ -83,6 +84,12 @@ const useStyles = makeStyles((theme) => ({
     fontWeight: 'normal',
     fontSize: '0.9em',
     color: theme.palette.neutral.main,
+  },
+  safe_warning: {
+    fontSize: '32px',
+    height: '32px',
+    width: '32px',
+    marginLeft: '5px',
   },
   contributors: {
     fontFamily: 'Poppins',
@@ -193,6 +200,8 @@ function VideoCard({
   let min_score = Infinity;
   let max_criteria = '';
   let min_criteria = '';
+  let safe = false;
+  let safeCause = '';
 
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('sm'), {
     noSsr: true,
@@ -223,6 +232,15 @@ function VideoCard({
 
   const nbRatings = video.rating_n_ratings;
   const nbContributors = video.rating_n_contributors;
+  if (nbContributors != null && nbContributors > 3) {
+    if (total_score > 0) {
+      safe = true;
+    } else {
+      safeCause = t('video.unsafeNegativeRating');
+    }
+  } else {
+    safeCause = t('video.unsafeNotEnoughContributor');
+  }
 
   return (
     <Grid container spacing={1} className={classes.main}>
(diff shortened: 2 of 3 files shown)


frontend/public/locales/en/translation.json
```diff
@@ -163,7 +163,10 @@
     "nbContributors_other": "{{count}} contributors",
     "criteriaRatedHigh": "Rated high:",
     "criteriaRatedLow": "Rated low:",
-    "labelShowSettings": "Show settings related to this video"
+    "labelShowSettings": "Show settings related to this video",
+    "unsafeNotEnoughContributor": "May be good but we have low confidence as this has been rated by only one contributor",
```
GresilleSiffle, line 167: How about not hard coding the number `one` in the translation?

The unsafe characteristic of a video can evolve if the Tournesol community grows, and in the future an unsafe video could be a video with two distinct contributors, three, or even more.

**edit** I now think the original text is better than my following proposition, as I found "low confidence" better than "uncertain", because there is always a bit of uncertainty in a video relevance, even if its score is high. But I'll leave the comment as is, for this history.

Here is an idea:

```javascript
"The relevance of this score is uncertain, due to a low number of distinct contributors."
```

or even

```javascript
"The relevance of this video's score is uncertain, due to a low number of distinct contributors."
```

It also avoid using the pronoun `we`. Who is `we`? The Tournesol community? The Tournesol agorithm? Even if we can agree on an answer, I think we ( the Tournesol team : ) can be more clear by avoiding `we`.

What do you think @amatissart @lfaucon @farnots ?

GresilleSiffle (commented):
Thanks for this visual proposition @farnots 

I didn't review the code yet, I just made few comments.

frontend/public/locales/en/translation.json
lfaucon: I strongly like avoiding "we". I slightly prefer "uncertain" than "low confidence" but like both options

farnots: Done I updated to avoid `we`

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -319,6 +337,13 @@ function VideoCard({
                 <span className={classes.nb_tournesol}>
                   {total_score.toFixed(0)}
                 </span>
+                {safe != null && safe == false && (
+                  <Tooltip title={safeCause} placement="bottom">
+                    <span className={classes.safe_warning}>
+                      <WarningIcon color="warning" fontSize="inherit" />
+                    </span>
+                  </Tooltip>
+                )}
```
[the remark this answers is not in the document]
farnots: I really like your suggestion, it's subtle yet visible.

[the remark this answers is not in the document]
farnots: How the `{...(safe == false` part is working ? Specially the `{...` in this case ?

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -193,6 +200,8 @@ function VideoCard({
   let min_score = Infinity;
   let max_criteria = '';
   let min_criteria = '';
+  let safe = false;
+  let safeCause = '';
```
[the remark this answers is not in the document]
farnots: I switch the condition, the default one is `unsafe = false`

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -223,6 +232,15 @@ function VideoCard({
 
   const nbRatings = video.rating_n_ratings;
   const nbContributors = video.rating_n_contributors;
+  if (nbContributors != null && nbContributors > 0) {
+    if (total_score > 0) {
+      safe = true;
```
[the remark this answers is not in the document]
farnots: You're right

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -223,6 +245,14 @@ function VideoCard({
 
   const nbRatings = video.rating_n_ratings;
   const nbContributors = video.rating_n_contributors;
+  if (nbContributors != null && nbContributors <= 1) {
+    unsafe = true;
+    unsafe_cause = t('video.unsafeNotEnoughContributor');
+    if (total_score < 0) {
+      unsafe = true;
+      unsafe_cause = t('video.unsafeNegativeRating');
+    }
```
amatissart, line 254: This condition should not be nested: a video with a negative score and 2 contributors should be considered "unsafe" too.

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -308,17 +338,28 @@ function VideoCard({
...
-                  className="tournesol"
-                  src={'/svg/tournesol.svg'}
-                  alt="logo"
-                  title="Overall score"
-                  width={32}
-                />
-                <span className={classes.nb_tournesol}>
-                  {total_score.toFixed(0)}
-                </span>
+                <SafeTournesolScoreWrapper
```
amatissart, line 348: Using this wrapper inside the <Box display="flex"> breaks the alignment of the content, as the wrapper consists of a single child.  
I think that it would make sense to use the wrapper above the `<Box>`.

frontend/src/features/videos/VideoCard.tsx
farnots: Done too

amatissart (approved):
We made small adjustments in the layout of the tooltip and in the translations:
image

PR #539, opened 2022-02-04
[#487] Unsafe Recommendations - User filter

farnots:
Linked to #487. Like the PR about the Card view I based this PR on the back.

I'm not sure about the working I use for both the french and the english translation.

CleanShot 2022-02-04 at 17 04 26@2x


farnots:
> Here are propositions we discussed with @amatissart and @lfaucon for the tooltip text linked to the new filter:

@GresilleSiffle @amatissart  There is no Tooltips on the filter itself, should I add one ?

farnots:
I add a Tooltip for the filter component. I choice the simpliest way with a binding to all the choice as we don't yet need for each choice. Would like to know if it's ok for you ?

Each component have always a Tooltip but it's not displayed with and empty String.
CleanShot 2022-02-11 at 11 36 10@2x
CleanShot 2022-02-11 at 11 36 31@2x

GresilleSiffle:
> I add a Tooltip for the filter component. I choice the simpliest way with a binding to all the choice as we don't yet need for each choice. Would like to know if it's ok for you ?

I think it's ok, we could do a refactor only the day we will need this complexity.

farnots:
> Alright, there is currently one conflict wit the file `features/recommendation/SearchFilter.tsx`, after resolution we will be ready to merge.


I fix the conflict 👌🏻

GresilleSiffle (approved):

farnots:
Thanks for fixing the PR @GresilleSiffle

GresilleSiffle:
You're welcome : )))

Thanks again for your work! Will be available on staging soon.

PR #542, opened 2022-02-04
Wiki auth

jstainer-kleis:


lfaucon:
We decided not to do more changes to the wiki authentication for now

Issue #348, opened 2021-12-07
Upgrade to Django 4.0


lfaucon:
https://docs.djangoproject.com/en/4.0/releases/4.0/

glerzing:
I self-assigned this issue. There isn't a lot of modifications to apply. The main issue I had is that :
In video.py, there is: "from languages.languages import LANGUAGES" to import a very long list of languages (around 8000)
but this file (from django-language-field which is not maintained anymore) uses ugettext_lazy, deprecated in Django 4.0.
so, importing this languages list causes an error. It was used for the choices attribute of videos' language

3 possible ways to solve it:
1 - copy the content of languages/languages.py and replace ugettext_lazy by gettext_lazy. django-language-field is on MIT license, so it shouldn't be a problem, I guess.
2 - use settings.LANGUAGES from django.conf. But it only contains around 80 languages so it is probably far from exhaustive
3 - use another library like iso-639, and a formula for formatting, like:
LANGUAGES = [(getattr(lang, 'part1', None) or
                      getattr(lang, 'part2t', None) or
                      getattr(lang, 'part3'),
                      gettext_lazy(lang.name)) for lang in iso639.languages]

lfaucon:
Difficult for me to make a wise choice here. @amatissart or @GresilleSiffle do you have recommendations about the library to use for languages on the backend?

GresilleSiffle:
Hello,

We discussed with Adrien this morning.

Our current proposition is to use the Django settings `LANGUAGES`, even if it contains a small subset of all languages.

We would like to discard the proposition **(1)**, because the package `django-language-field` is in a very slow maintenance mode (or even not maintained anymore?) and do not want to maintain ourselves a fork of it. Still, opening a pull request to make it compatible with Django 4.x might interest the package maintainer.

We would like to discard the proposition **(3)** for now. It seems to be our best alternative for have an inclusive list of languages, but the `language` field of the `Video` model is going to be replaced in the near future by metadata JSON field. As this part of the application is going to be rewritten, our validation process will change too.

Let's use the Django settings `LANGUAGES` then.

PR #544, opened 2022-02-04
[backend] Upgrade to Django 4.0 (#348)

glerzing:
I also updated a few other libraries, even that was not required.

The reason why django-language-field was replaced is discussed in #348 (use of the obsolete function ugettext_lazy)

tested with python 3.8 and 3.9

You may have to delete the tournesol-dev-api image first for it to work.


lfaucon (approved):
LGTM. Thanks!

PR #548, opened 2022-02-05
pytest warnings

glerzing:
In the previous PR (#544), I missed those pytest warnings because I was using `python manage.py test` instead of pytest

The first warning is because USE_L10N is getting deprecated. It is set to true by default anyway.

The second warning (RemovedInDjango41Warning) comes from django-rest-registration. It is not solved yet in the latest version of django-rest-registration, so I just muted the warning.


amatissart (approved):
Thanks

PR #549, opened 2022-02-05
[analytics] add Tournesol scores page

aidanjungo:
For now, only videos with positive Tournesol score are included, will be improved after #531 is merged.

### DIFF
--- a/analytics/analytics.py
+++ b/analytics/analytics.py
@@ -1,12 +1,13 @@
 import streamlit as st
 from PIL import Image
 
-from pages import page_public_dataset
+from pages import page_public_dataset, page_tournesol_scores
 
 st.session_state.data = None
 
 PAGES = {
     "Public Dataset": page_public_dataset,
+    "Tournesol Scores": page_tournesol_scores,
 }
 
 # === Side bar ==================================
--- a/analytics/pages/page_tournesol_scores.py
+++ b/analytics/pages/page_tournesol_scores.py
@@ -1,0 +1,186 @@
+import matplotlib.pyplot as plt
+import pandas as pd
+import plotly.graph_objects as go
+import seaborn as sns
+import streamlit as st
+
+from utils import CRITERIA, CRITERI_EXT, MSG_NO_DATA, TCOLOR, api_get_tournesol_scores
+
+
+def add_expander_select_filters():
+
+    with st.expander("Select channel(s)"):
+
+        st.markdown(
+            "You can select one or several YouTube channels."
+            "If you select none, all channels will be use."
+        )
+
+        if not isinstance(st.session_state.df_scores, pd.DataFrame):
+            st.warning(MSG_NO_DATA)
+            return
+
+        df = st.session_state.df_scores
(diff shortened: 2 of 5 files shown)


analytics/pages/page_tournesol_scores.py
```diff
@@ -0,0 +1,186 @@
...
+from utils import CRITERIA, CRITERI_EXT, MSG_NO_DATA, TCOLOR, api_get_tournesol_scores
+
+
+def add_expander_select_filters():
+
+    with st.expander("Select channel(s)"):
+
+        st.markdown(
+            "You can select one or several YouTube channels."
+            "If you select none, all channels will be use."
```
GresilleSiffle, line 16: We can add a space at the beginning of this string to avoid concatenating sentences.

```suggestion
            " If you select none, all channels will be use."
```

analytics/pages/page_tournesol_scores.py
```diff
@@ -0,0 +1,186 @@
...
+    with st.expander("Correlation coefficients"):
+
+        if not isinstance(st.session_state.df_scores, pd.DataFrame):
+            st.warning(MSG_NO_DATA)
+            return
+
+        df = st.session_state.df_scores
+        st.markdown(
+            "The table bellow shows the pearson correlation"
+            "coefficient between the criteria scores."
```
GresilleSiffle, line 89:
```suggestion
            " coefficient between the criteria scores."
```

analytics/analytics.py
```diff
@@ -1,12 +1,13 @@
...
 from PIL import Image
 
-from pages import page_public_dataset
+from pages import page_public_dataset, page_tournesol_scores
 
 st.session_state.data = None
 
 PAGES = {
     "Public Dataset": page_public_dataset,
+    "Tournesol Scores": page_tournesol_scores,
```
GresilleSiffle, line 10: Very minor comment, feel free to ignore : )

I think we can be more explicit in the name of the page as they not only display different data, but also use different data source ( the first page load a file, the second directly connect to Tournesol with HTTPS ).

How about this:

```python
    "Load Public Dataset": page_public_dataset,
    "View Live Tournesol Scores": page_tournesol_scores,
```

or simply

```python
    "Load Public Dataset": page_public_dataset,
    "Live Tournesol Scores": page_tournesol_scores,
```

GresilleSiffle (commented):
Looks good for me :+1: 

I've only added few minor comments.

analytics/analytics.py
aidanjungo: I think I will change the Public Dataset page to download the dataset automatically  as it is done for the Tournesol score page.

GresilleSiffle (approved):
LGTM :+1:

analytics/analytics.py
aidanjungo: Now it is loaded automatically (and cached) as the score and the page is called "Comparisons Public Dataset"

amatissart (approved):
@aidanjungo Feel free to merge if this branch is ready.

Issue #526, opened 2022-02-02
refactor: create the model `Poll`


GresilleSiffle:
**requires** #525 

also see the issue #509 

---

**(1)**

Create the new model `Poll`.

Fields
- `name` SlugField + unique
- `entity_type`, CharField with the same choices as the one in `Entity` 
- (question) in the future do we want an `open` field to allow / prevent comparisons to be made?
- (question) in the future do we want a boolean flag `ml` telling the ML to run on this poll or not?

Create the first poll in a migration, named `videos`.

Create a FK (can be null for now) from `Comparison` to `Poll`.

Create a FK (can be null for now) from `ContributorRating` to `Poll`.

**(2)**

**n-n** relation between `Poll` and `Criteria`
- with an additional field `rank`, used to order the field on the front end ( values could be 10, 20, 30, 40, etc. to have space for re-ordering ) 

Create the new model `Criteria`.
- `name` SlugField + unique
- `optional` boolean 

Create the new model `CriteriaLocale`.
- FK to `Criteria`
- lang CharField with choices from Django settings LANGUAGES
- label

Plan to add a validation in the API, rejecting requests that contain criteria not available in the current poll.

**(3)**

Configure the admin interface, to be able to add polls and criteria.

lfaucon:
> (question) in the future do we want an open field to allow / prevent comparisons to be made?

Probably yes. Let's not have this field today and create it when we are ready to implement a feature using it.

> (question) in the future do we want a boolean flag ml telling the ML to run on this poll or not?

I think that would be the same thing as the `open` field. 

> Create the first poll in a migration, named videos.

I want a more fancy name to honor the original vision of Tournesol ! 🌻 

> Create a FK (can be null for now) from Comparison to Poll.
> Create a FK (can be null for now) from ContributorRating to Poll.

Is it much easier to allow null for now, instead of setting all comparisons and ratings to the "videos" Poll ? What is your reasonning for proposing this way?

> with an additional field rank, used to order the field on the front end ( values could be 10, 20, 30, 40, etc. to have space for re-ordering )
>
> Create the new model Criteria.
>
> name SlugField + unique
> optional boolean

Do I understand that `rank` is an attribute of the `Criteria` model ?

> Configure the admin interface, to be able to add polls and criteria.

The minimum that works, because it will be a very rare operation.

PR #554, opened 2022-02-07
Create model `Poll`, add foreign keys and rename related fields into `entity`

amatissart:
First step, described in #526 

Note that contrary to what was described initially, the foreign keys do not need to be created with `null=True`. Their default value is specified with a method to find the default poll.

The `name` of this default poll is to be confirmed: `videos` for now.


backend/tournesol/admin.py
```diff
@@ -49,7 +49,7 @@ class EntityCriteriaScoreAdmin(admin.ModelAdmin):
 class ContributorRatingAdmin(admin.ModelAdmin):
```
GresilleSiffle, line 49: future considerations

we could add a filter on entity type and and filter on poll in this admin

backend/tournesol/models/ratings.py
```diff
@@ -25,12 +26,18 @@ class ContributorRating(models.Model, WithFeatures):
...
     is_public = models.BooleanField(
         default=False, null=False, help_text="Should the rating be public?"
     )
 
     class Meta:
-        unique_together = ["user", "video"]
+        unique_together = ["user", "entity", "poll"]
 
     def __str__(self):
         return "%s on %s" % (self.user, self.video)
```
GresilleSiffle, line 43:
```suggestion
        return "%s on %s" % (self.user, self.entity)
```

backend/tournesol/models/entity_score.py
```diff
@@ -0,0 +1,57 @@
...
+        validators=[MinValueValidator(0.0), MaxValueValidator(1.0)],
+        help_text="Top quantile for all rated videos for aggregated scores"
+        "for the given criteria. 0.0=best, 1.0=worst",
+    )
+
+    class Meta:
+        unique_together = ["video", "poll", "criteria"]
+
+    def __str__(self):
+        return f"{self.entity}/{self.criteria}/{self.score}"
```
GresilleSiffle, line 57: We refer to `self.entity` here but the model has still a `video` attribute. Maybe a migration is missing?

backend/tournesol/models/ratings.py
amatissart: Should be fixed in ffdeff7

backend/tournesol/models/entity_score.py
amatissart: Should be fixed in ffdeff7b

GresilleSiffle (approved):

Issue #327, opened 2021-11-27
[frontend] feat: Comparisons on HomePage when not logged in.


lfaucon:
- Videos from the top 10 of recently rated are automatically loaded (fallback to all time if there are not enough)
- Comparison on Home page: If not logged in button "Submit" says "Login to submit"
- Comparison on Home page: If not logged in Random from last month and automatically detect browser's language

GresilleSiffle:
- [ ] display the comparison feature on the home page, for non logged user
  - [ ] videos are from the top 10 recently rated
  - [ ] fall back to video of all time if there are less than 2 videos in the recent top 10
- [ ] display the `Login to submit` button
- [ ] save somewhere the comparison data when the button is clicked
- [ ] after the user has activated its account, send the comparison to the API

## Questions

**(a)** Do we want to also display the comparison feature in the Home page for logged users?

Display this feature will act as an invitation to rate the recent videos, to collect more judgment. Could this be a virtuous and engaging feature? I personally like it.

**(b)** What does "recently" mean? Rated last week? Rated last month?

There may be more inertia in the videos proposed if we use the month, but the risk of displaying bad quality videos is lower.

## Technical specifications

A comparison made by an anonymous user can either be saved in the browser local storage or in the back end's database in a pending state or a dedicated table. The two options have a different level complexity, let's explore what is best. Currently I think the local storage is better, because less complex ( no new database table, no new API endpoints, no concept of anonymous comparison ).

lfaucon:
> after the user has activated its account, send the comparison to the API

Maybe "send the comparison to the API & redirect the user to the comparison page for editing the comparison if she/he feels the need"

> (a) Do we want to also display the comparison feature in the Home page for logged users?

Why not. I find your justification convincing

> (b) What does "recently" mean? Rated last week? Rated last month?

Let's have the same default as for the Recommendations page! Maybe let's implement it in a way such that when the recommendation page is changed so will the home page comparison page.

> A comparison made by an anonymous user can either be saved in the browser local storage or in the back end's database in a pending state or a dedicated table. The two options have a different level complexity, let's explore what is best. Currently I think the local storage is better, because less complex ( no new database table, no new API endpoints, no concept of anonymous comparison ).

I recommend frontend side for simplicity and also for not collecting non-logged-in user data.

> The <ComparisonPage> component will need to be refactored so that two different pages can include the comparison interface without duplicating all code related to the usage of the <VideoSelector> components.

Yes, fully agreed

[...]

PR #555, opened 2022-02-08
[front] feat: comparison on Home page for logged users

GresilleSiffle:
**related to** #327 

---

This pull request aims to create the foundations of the comparison feature on the Home page.

It partially addresses the issue #327 as currently only logged users are able to make comparisons.

It adds a "light" comparison feature on the Home page, with two extra secondary menu for logged users allowing to:
- go to the comparison page
- copy the comparison URL to the clipboard

I didn't update the Home page layout yet as we didn't agree on a precise order for the different sections. It looks fine for a first pull request, but should be enhanced by moving the comparison section higher in the page before making the comparison available for anonymous users.

See a layout proposition in the PR #564 of @amatissart . Could be merged in this branch. 

**new component** `<LightComparison>`
- [x] the comparison is displayed on the Home page, logged users only for now
- [x] no option criteria
- [x] no video selector
- [x] no public / private toggle
- [x] no mention of the number of comparison
- [ ] submitting should redirect the logged user to the comparison page?

The redirection to the comparison page doesn't seem to be required. There is already a button allowing the logged users to go directly to the comparison page, and no automatic redirection allows users to edit they comparison.

**how videos are selected**

- [x] two random videos are picked from the top 20 of the last month
- [x] missing videos are completed by videos from the top 20 of the last year

The top 20 of monthly recommendations corresponds to the recommendations displayed by default in the recommendations page of the front end. This top 20 can be reduced to top 10 if needed.

**to-do**
- [x] add an error message when there is not enough video to compare
- [x] add end-to-end tests

**questions**
- [x] should we add a button allowing **logged users** to copy the URL of the comparison in order to share it with others **logged users**?
- [x] should we add a way for logged users to directly go to the comparison page, to have access to all optional criteria?

I used the `text` variant of the mui `<Button>` component to make a very light secondary menu containing these actions.

## preview

screencapture-localhost-3000-2022-02-09-14_55_24

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -259,7 +259,10 @@
     "extensionNotAvailableOnYourBrowser": "The extension is not available on your web browser. You may use it on <2>Firefox</2>, <4>Google Chrome</4> or <6>Chromium</6>.",
     "collaborativeContentRecommendations": "Collaborative Content Recommendations",
     "whatIsTournesol": "What is Tournesol?",
-    "tournesolPlatformDescription": "Tournesol is an <1>open source</1> platform which aims to <3>collaboratively</3> identify top videos of public utility by eliciting contributors' judgements on content quality. We hope to contribute to making today's and tomorrow's large-scale algorithms <6>robustly beneficial</6> for all of humanity."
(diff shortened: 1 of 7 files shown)


aidanjungo:
Two small comments/questions:
- Should we filter videos with a maximum duration of 10-12min to not afraid users with 2h video?
- Could we propose videos in the same language as the language of the interface?

and for `should we add a button allowing logged users to copy the URL of the comparison in order to share it?`
Yes it would be cool! 😀 Good way to share comparisons page with non register users

`should we add a way for logged users to directly go to the comparison page, to have access to all optional criteria?`
I think yes

lfaucon:
See comment on ticket #327

Issue #540, opened 2022-02-04
Run ML for both trusted and not trusted users


lfaucon:
Context:
The ML algorithm computes both criteria scores per contributor (ContributorRatingCriteriaScore) and for videos (VideoCriteriaScore <- now being renamed EntityCriteriaScore). Currently the algorithm only runs on comparisons from trusted users, so non-trusted users are do not have ContributorRatingCriteriaScore.

Acceptance:
We should run the algorithm on all users too but only for saving the ContributorRatingCriteriaScore of non-trusted users.

PR #556, opened 2022-02-08
Add models Criteria, CriteriaRank and CriteriaLocale

amatissart:
Covers most of points **(2)** and **(3)** described in #526

- [x] New models `Criteria`, `CriteriaLocale`
- [x] New model `CriteriaRank` used as many-to-many between `Poll` and `Criteria`
  Note that `optional` has been moved to this model, as a criteria could be optional in one poll, and required in another.
- [x] Admin configuration, to create/edit polls, criterias, ranks, labels, etc.
- [ ] Validation in the API, rejecting requests that contain criteria not available in the current poll (TODO separately in a new PR)

PR #561, opened 2022-02-09
Run ML on every user

farnots:
Linked to #540

### DIFF
--- a/backend/ml/management/commands/ml_train.py
+++ b/backend/ml/management/commands/ml_train.py
@@ -68,7 +68,6 @@ def fetch_data():
             ccs.weight,
         ]
         for ccs in ComparisonCriteriaScore.objects
-            .filter(comparison__user__in=User.trusted_users())
             .prefetch_related("comparison")
     ]
     return comparison_data


farnots:
@lfaucon Any idea of to tests if my implemetation is working ? I cannot open to review until I test it

amatissart:
> @lfaucon Any idea how to tests if my implemetation is working ? I cannot open to review until I test it

We could probably write a test case, similar to the one introduced recently in #556:
https://github.com/tournesol-app/tournesol/blob/main/backend/tournesol/tests/test_ml_run.py

backend/ml/management/commands/ml_train.py
```diff
@@ -50,46 +50,63 @@
...
+                ccs.comparison.user_id,
+                ccs.comparison.entity_1_id,
+                ccs.comparison.entity_2_id,
+                ccs.criteria,
+                ccs.score,
+                ccs.weight,
+            ]
+            for ccs in ComparisonCriteriaScore.objects
+                .exclude(comparison__user__in=User.trusted_users())
+                .prefetch_related("comparison")
```
lfaucon, line 87: As discussed during the stand-up: We should run the ML on trusted users, and also on everyone.

Let's avoid the code duplication here:

```suggestion
    comparisons_queryset = ComparisonCriteriaScore.objects.all().prefetch_related("comparison")
    if trusted:
        comparison_queryset = comparison_queryset.filter(comparison__user__in=User.trusted_users())
    comparison_data = [
            [
                ccs.comparison.user_id,
                ccs.comparison.entity_1_id,
                ccs.comparison.entity_2_id,
                ccs.criteria,
                ccs.score,
                ccs.weight,
            ]
            for ccs in comparison_queryset
    ]
```

Maybe this piece above can be done with [django's Queryset.values_list](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#values-list)

backend/ml/management/commands/ml_train.py
```diff
@@ -133,11 +150,20 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         criterias_list = Poll.default_poll().criterias_list
-        comparison_data = fetch_data()
+        comparison_data_trusted = fetch_data()
+        compraison_data_not_trusted = fetch_data(trusted=False)
```
lfaucon, line 154: typo `compraison`

backend/ml/management/commands/ml_train.py
```diff
@@ -133,11 +150,20 @@ class Command(BaseCommand):
...
+                comparison_data_trusted, criterias=criterias_list, save=True, verb=-1
             )
-            save_data(glob_scores, loc_scores)
+            save_data(glob_scores_trusted, loc_scores_trusted)
+
+            # Run for all users including non trusted users
+            glob_scores_not_trusted, loc_scores_not_trusted = ml_run(
+                compraison_data_not_trusted, criterias=criterias_list, save=True, verb=-1
+            )
+            save_data(glob_scores_not_trusted, loc_scores_not_trusted, trusted=False)
```
lfaucon, line 168: Currently you have implemented the logic with

Pseudo code:
```
def handle():
    fetch_trusted
    fetch_all
    run_ml_trusted
    save_trusted
    run_ml_all
    save_all
```

What do you think of the refactor:

```
def process():
    fetch
    run_ml
    save

def handle():
    process_all
    process_trusted
```

lfaucon (changes requested):
I agree with @amatissart about writing tests following the example here:

https://github.com/tournesol-app/tournesol/blob/main/backend/tournesol/tests/test_ml_run.py

# Possible tests
Setup:
2 trusted users who agree on some comparisons
10 non-trusted users with oposite opinions on the comparisons

## Test 1
run management command
asserts that everyone has ContributorCriteriaScores
asserts that everyone's ContributorCriteriaScores are aligned with contributors individual opinions
asserts that the EntityCriteriaScores are aligned with trusted contributors opinions

## Test 2
Run only process with `trusted=False`, asserts that no EntityCriteriaScore is created/modified

farnots:
I set the tests  has suggested by @lfaucon with different opinion from the trusted user group and the non trusted group. I check that each video entity have score based only on the trusted user group but that all user have comparison score

backend/tournesol/tests/test_ml_run.py
```diff
@@ -14,10 +14,46 @@ def setUp(self) -> None:
...
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user4, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user5, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user6, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user7, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user8, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user9, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user10, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+
     def test_ml_train(self):
```
GresilleSiffle, line 52: **questions / suggestions**

@farnots @amatissart @lfaucon 

Do you think it's relevant here to split this method into two tests instead of one?

Also the current name, `test_ml_train` is very generic and doesn't tell what the test is going to check. I don't know if it's relevant because I'm not familiar enough with this part of the code yet, but I would suggest to change it to make the code more readable and welcoming for new developers. The test case class is already named `TestMlTrain` so we know we are going to test the machine learning algorithm, and choosing a different name may encourage developers to add different tests in the future instead of adding all assertions into a single big test.

What about `test_computed_scores` and `test_computed_scores_untrusted_users`? Or something else.

farnots: > Do you think it's relevant here to split this method into two tests instead of one?

How do you want to split this method ? The test is to check information before and after the `ml_train` commands. If we split it the tests will be less interesting.

backend/ml/management/commands/ml_train.py
```diff
@@ -50,48 +50,52 @@
 """
 
 
-def fetch_data():
+def fetch_data(trusted=True):
```
amatissart, line 53: Maybe a parameter such as `trusted_only=True` would be more explicit.
(`trusted=False` also includes comparisons from trusted users)

amatissart (commented):
Thanks @farnots 
Except a comment about how `ContributorRatingCriteriaScore` from trusted users are preserved between the 2 passes, it looks good to me.

With the current algorithm, running the score computation a second time on the whole dataset (including non-trusted users) may be expensive.
We are currently working on a new algorithm (see some preliminary work on https://github.com/tournesol-app/tournesol/compare/mehestan), so we might want to wait for its complete implementation before applying this change.

backend/tournesol/tests/test_ml_run.py
GresilleSiffle: > How do you want to split this method ?

Actually I don't know. My question was motivated by the fact we have here one test case with only one test, and I wanted to know if this was intended and relevant or if this test was actually checking two different and independent use cases that could be split in two different tests.

If it's not relevant to split, it's good as is then :+1:

backend/ml/management/commands/ml_train.py
```diff
@@ -130,17 +134,24 @@ def save_data(video_scores, contributor_rating_scores):
         ]
     )
```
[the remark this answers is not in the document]
lfaucon: Correct.

`save_data` with `trusted=False` should only update the `ContributorRating` and `ContributorRatginCriteriaScore` for non-trusted users.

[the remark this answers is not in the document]
farnots: I didn't have that in mind. Thanks for the info. I add some filter on the `contributor_rating_scores` to make sure we only save the right one

backend/tournesol/tests/test_ml_run.py
```diff
@@ -14,10 +14,46 @@ def setUp(self) -> None:
...
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user4, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user5, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user6, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user7, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user8, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user9, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
+        ComparisonCriteriaScoreFactory(comparison__user=not_trusted_user10, comparison__entity_1=self.video1, comparison__entity_2=self.video2, score=-10)
```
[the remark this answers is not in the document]
farnots: Thanks for your suggestion

backend/ml/management/commands/ml_train.py
farnots: Yes great idea thanks

backend/ml/management/commands/ml_train.py
```diff
@@ -117,7 +141,12 @@ def save_data(video_scores, contributor_rating_scores):
     rating_ids.update(
         {(rating.user_id, rating.entity_id): rating.id for rating in created_ratings}
     )
-    ContributorRatingCriteriaScore.objects.all().delete()
+
+    if trusted_only:
+        ContributorRatingCriteriaScore.objects.filter(contributor_rating__user__in=User.trusted_users()).delete()
+    else:
+        ContributorRatingCriteriaScore.objects.exclude(contributor_rating__user__in=User.trusted_users()).delete()
+
```
farnots, line 149: Only delete the one that will be re-generated. All trusted user in trusted mode and only non trusted user in all mode

backend/ml/management/commands/ml_train.py
```diff
@@ -50,48 +50,72 @@
...
+            ]
+        )
+
+    if trusted_only:
+        trusted_contributor_rating_scores = set(
+            (contributor_id, video_id, criteria, score, uncertainty)
+            for (contributor_id, video_id, criteria, score, uncertainty) in contributor_rating_scores
+            if (contributor_id,) in User.trusted_users().values_list (
+                "id"
             )
```
farnots, line 106: Filter to get only the `ContributorRatingScore` we whan to update.

backend/ml/management/commands/ml_train.py
```diff
@@ -130,17 +159,36 @@ def save_data(video_scores, contributor_rating_scores):
...
 
 class Command(BaseCommand):
     help = "Runs the ml"
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--skip_untrusted',
+            nargs='?',
+            help='Skip ML run on untrusted users',
+        )
```
farnots, line 180: This optional argument is only added for test purpose. Do you think it's relevent @amatissart ?

backend/ml/management/commands/ml_train.py
amatissart: I took the liberty to make some adjustments, especially here to avoid redundant SQL queries. 
```python
            for (contributor_id, video_id, criteria, score, uncertainty) in contributor_rating_scores
            if (contributor_id,) in User.trusted_users().values_list (
                "id"
            )
```
would fetch the list trusted users for every score.

amatissart (approved):

Issue #482, opened 2022-01-18
[front] feat: the log in page displays link to sign up


GresilleSiffle:
Here is the UX/UI suggested on Discord

![compare-quick-wins](https://user-images.githubusercontent.com/39056254/149972432-a0f24fc5-98a2-4c22-b702-0efc437d09b6.png)

PR #562, opened 2022-02-09
[#482] Add link to /signup on the login page

sigmike:
Linked issue: #482 

I chose the first alternative from the suggested UI. The link to the demo doesn't seem very useful here but I don't know (and I don't know where it should link to).

In English I used the same label as the header button, but the suggested French label is different from the header so you may want something else.

I used the outlined variant to differentiate it from the submit button, but the confusion can still happen I guess. It looks like that:
image

### DIFF
--- a/frontend/src/features/login/Login.tsx
+++ b/frontend/src/features/login/Login.tsx
@@ -111,6 +111,17 @@ const Login = () => {
             {t('login.forgotYourPassword')}
           </Link>
         </Box>
+        <Box my={2}>
+          <Button
+            color="secondary"
+            fullWidth
+            variant="outlined"
+            component={RouterLink}
+            to="/signup"
+          >
+            {t('login.signUp')}
+          </Button>
+        </Box>
       </ContentBox>
     </>
   );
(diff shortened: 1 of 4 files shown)


frontend/src/features/login/Login.tsx
```diff
@@ -111,6 +111,17 @@ const Login = () => {
             {t('login.forgotYourPassword')}
           </Link>
         </Box>
+        <Box my={2}>
```
amatissart, line 114: To make clearer the distinction between login and signup, what about adding a divider and message addressed to users who don't have an account yet?

```suggestion
        <Divider />
        <Box my={2}>
          <Typography variant="subtitle1" gutterBottom>
            No account on Tournesol yet?
          </Typography>
```

(with translation of course)

amatissart (commented):
Thanks for this PR! 
That will definitely help to make the experience smoother for newcomers.

I just made a suggestion to hopefully avoid confusions between the two buttons.

frontend/src/features/login/Login.tsx
sigmike: I added your suggestion.
image

amatissart (approved):

PR #564, opened 2022-02-10
More compact layout for home page, and display light comparison above the fold

amatissart:
This is a suggestion, based on #327 
||Before|After| 
|-|-|-|
|Not logged in|image| image|
|Logged in|image |image|

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -27,7 +27,7 @@
   },
   "submit": "submit",
   "lightComparison": {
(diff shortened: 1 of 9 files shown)


GresilleSiffle:
I like new layout :+1: 

We might be walking away a bit from the UX recommendations we had so far by adding a rather big paragraph as a subtext of the main title. I don't know if it can be overwhelming for the users.

What do you think @lfaucon @aidanjungo @amatissart about removing the second sentence starting with **we hope to...**? 

It will reduce the amount of text present on the Home page and the removed sentence will still be present in the About page.

PR #574, opened 2022-02-13
[front] fix: Clips scores between 0 and 1 on RadarChart

### LINKED ISSUE #570: [front] fix: Video analysis page radarchart should clip scores between 0 and 1

lfaucon:
Context:
Scores were scaled automatically when displayed such that if the max score was very small it appeared that all scores are large. This PR fixes this issue by forcing a fixed sclae between 0 and 1.

Closes #570

### DIFF
--- a/frontend/src/components/CriteriaRadarChart.tsx
+++ b/frontend/src/components/CriteriaRadarChart.tsx
@@ -10,6 +10,9 @@ import { VideoSerializerWithCriteria } from 'src/services/openapi';
 import { getCriteriaName } from 'src/utils/constants';
 import { useTranslation } from 'react-i18next';
 
+const RADER_CHART_CRITERIA_SCORE_MIN = -1;
+const RADER_CHART_CRITERIA_SCORE_MAX = 1;
+
 interface Props {
   video: VideoSerializerWithCriteria;
 }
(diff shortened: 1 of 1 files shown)


frontend/src/components/CriteriaRadarChart.tsx
```diff
@@ -43,25 +55,28 @@ const CriteriaRadarChart = ({ video }: Props) => {
...
-            fillOpacity={0.6}
-          />
-        </RadarChart>
-      )}
-    </div>
+    <RadarChart width={300} height={300} outerRadius="80%" data={data}>
+      <PolarGrid />
+      <PolarAngleAxis dataKey="criteria" tick={renderCustomAxisTick} />
+      {/* An invisible PolarRadiusAxis used to enforce the axis between 0 and 1 */}
+      <PolarRadiusAxis domain={[0, 1]} axisLine={false} tick={false} />
```
amatissart, line 72: Why not using a scale between -1 and 1, to visualize negative scores too?

amatissart (approved):

sintaz1992:
I agree that it does not look good if the max score is very small. It's a good solution but how about normalizaing them, such that max score is 1? the differene will be that we will show more precise compariosn between scores of one video , at the moment the relation between scores will not be kept

Issue #527, opened 2022-02-02
refactor: comparisons API now handles `Poll`


GresilleSiffle:
**requires** #526 

also see the issue #509

---

Refactor the comparisons API to handles `Poll`.

### back end

**to-do**

- use uid instead of video_id in JSON payload of all comparisons endpoints
- update the front end accordingly

**done**

Create a migration to set the value of the comparison poll FK to the previously created Poll with name `videos`.

Create a migration to remove the NULLABLE constraint on this FK.

`GET /users/me/comparisons/{pollname}/` - get all my comparisons made in this poll

`POST /users/me/comparisons/{pollname}/` create a new comparison in this poll
 
`GET|PUT /users/me/comparisons/{pollname}/{uid_a}/{uid_b}/` get or partial update of a comparison in this poll 

`GET /users/me/comparisons/{pollname}/{uid_a}/` get all comparisons that include the entity with `uid_a` in this poll 

Add a validation ensuring the submitted criteria name match those in the requested poll ( a match in the name will be sufficient for now ).

Replace `video_id_a` & `video_id_b` in URLs by `uid_a` & `uid_b`

Replace `video_a` & `video_b` in JSON representation of comparisons by `entity_a` & `entity_b`

Adapt the tests.

### front end

**done**

Update the API schema and re-generate the services.

Update the components to use the new endpoints.

Hardcode the new `pollname` in the front end ( as a constant ).

lfaucon:
How do we handle the current calls to `GET /users/me/comparisons/`. Will this be a breaking change? Or do we maintain complete backward compatibility? In case of breaking change, it should just be a matter of refreshing the frontend web page for users, so not too bad, but better to avoid.

PR #576, opened 2022-02-14
refactor: use `uid` in comparisons URL instead of `video_id`

GresilleSiffle:
**related to** #527 

**follows the work of pull request** https://github.com/tournesol-app/tournesol/pull/563

---

This pull request refactors the comparisons API, to use entity UID in its URLs and make the front end compatible. It also changes the JSON representation of a comparison by replacing `video_a` and `video_b` keys by `entity_a` and `entity_b`.

This pull request doesn't replace the usage of `video_id` in the JSON representation of a comparison yet, I'd prefer doing this in a distinct PR as it might require few tweaks of the video serializers and maybe other part of the application. 

**back end**
- [x] in URL replace occurrences of `video_id_a` / `video_id_b` by `uid_a` / `uid_b`
- [x] in URL replace occurrences of `video_id` by `uid`
- [x] in the serializers, replace fields `video_a`  / `video_b` by `entity_a` / `entity_b`
- [x] `uid` URL parameters now contain the entity UID value instead of the previous video ID
- [ ] the video serializers still use `video_id` for now...

**front end**
- [x] in the schema URLs replace `video_id_a` / `video_id_b` by `uid_a` / `uid_b`
- [x] in the schema URLs replace `video_id` by `uid`
- [x] update services usage, replace fields `videoIdA`  / `videoIdB` by `uidA` / `uidB`
- [x] update services usage, replace fields `video_id_a`  / `video_id_b` by `entity_a` / `entity_b`
- [x] update services usage, replace fields `videoId` by `uid`
- [x] now use "namespaced" UID in `/comparisons/` URLs
- [ ] the video serializers still use `video_id` for now...

### new API design

capture

### DIFF
--- a/frontend/src/utils/constants.ts
+++ b/frontend/src/utils/constants.ts
@@ -44,7 +44,9 @@ export const getCriteriaName = (t: TFunction, criteria: string) => {
 };
 
 export const YOUTUBE_POLL_NAME = 'videos';
-export const YT_UID_PREFIX = 'yt%3A';
+
+const UID_DELIMITER = '%3A';
+export const UID_YT_NAMESPACE = 'yt' + UID_DELIMITER;
 
 export const recommendationFilters = {
   date: 'date',
--- a/frontend/src/utils/constants.ts
+++ b/frontend/src/utils/constants.ts
@@ -44,6 +44,7 @@ export const getCriteriaName = (t: TFunction, criteria: string) => {
 };
 
 export const YOUTUBE_POLL_NAME = 'videos';
+export const YT_UID_PREFIX = 'yt%3A';
 
 export const recommendationFilters = {
   date: 'date',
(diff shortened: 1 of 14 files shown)


frontend/src/utils/constants.ts
```diff
@@ -45,6 +45,9 @@ export const getCriteriaName = (t: TFunction, criteria: string) => {
 
 export const YOUTUBE_POLL_NAME = 'videos';
 
+const UID_DELIMITER = '%3A';
+export const UID_YT_NAMESPACE = 'yt' + UID_DELIMITER;
```
amatissart, line 49: I would recommend to define the delimiter as `:`, and to URL-encode it strictly when necessary (i.e in URL paths, as the OpenAPI client seems to handle `:` as a special case, although it is a valid character). 
For example it would be incorrect to use the encoded version in a request body.

amatissart (approved):

Issue #529, opened 2022-02-02
refactor: create the `Poll` API


GresilleSiffle:
**requires** #527

also see the issue #509

---

**backend**

Create new endpoints for the polls.

All those endpoints are public.

**(1)** `GET /polls/{pollname}` to fetch all poll's information PLUS the the list of its criteria.

**(2)** `GET /polls/{pollname}/recommendations` to fetch recommendations ordered by the " Tournesol score " DESC

**(3)** `GET /polls/{pollname}/entities` to fetch all entities, regardless of the score, without any default filter (except entities are still related to the poll scope)

Add tests.

lfaucon:
This ticket was closed without implementing the third endpoint.

Either:
1. Deemed redundant with the unsafe recommendations? In which case, let's move the ticket to DONE on the kanban
2. Planned to be implemented? in which case, let's re-open the ticket and proceed with the implementation.

I'm slightly in favor of option 2 because it is a more efficient endpoint when the extra fetching, summing and sorting of scores is not needed.

amatissart:
When implementing the routes `GET /entities/` routes, it seemed indeed redundant to have  `GET /polls/{pollname}/entities/`. That's why it was closed in #585.

What would be a use case for this endpoint which would not be addressed by either `GET /polls/{pollname}/recommendations/?unsafe=true` (to fetch all entities scored in a poll), or `GET /entities/?type={entityType}` (to fetch all entities present in the database, including those without comparison)?

lfaucon:
Ok 👌 moving to DONE

PR #577, opened 2022-02-14
Implement minimal version of poll recommendations API

amatissart:
Implements item **(2)** in https://github.com/tournesol-app/tournesol/issues/529

Similar to `GET /video/` to fetch a list of recommendation, with a few differences

In the **response**: 
  * fields specific to a video are not present. Instead a `metadata` dict is returned (empty for now)
  * `total_score` is present in the output (as a `FloatField`) and computed with a default weight of 10 for each criteria (instead of 50 in the /video/ route, to be consistent with the score currently computed by the frontend).

In the **query parameters**:
  * `?criteria=50` is replaced with `?weights[criteria]=10`, as the name of the criterias are now dynamic
  * filters based on video metadata (language, uploader, etc.) are not implemented.  
  Maybe a generic `?metadata[key]=value` could be used instead in the future?

### DIFF
--- a/backend/tournesol/views/polls.py
+++ b/backend/tournesol/views/polls.py
@@ -1,9 +1,17 @@
 import logging
 
-from rest_framework.generics import RetrieveAPIView
+from django.conf import settings
+from django.db.models import Case, F, Prefetch, Q, Sum, When
+from django.shortcuts import get_object_or_404
+from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
+from rest_framework import serializers
+from rest_framework.generics import ListAPIView, RetrieveAPIView
+from rest_framework.pagination import LimitOffsetPagination
 
-from tournesol.models import Poll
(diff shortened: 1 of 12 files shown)


backend/tournesol/views/polls.py
```diff
@@ -12,7 +25,100 @@ class PollsView(RetrieveAPIView):
...
+                        }
+                    )
+                ]
+            ),
+        ],
+    )
+)
+class PollsRecommendationsView(ListAPIView):
+    permission_classes = []
+    pagination_class = LimitOffsetPagination
```
GresilleSiffle, line 62: I think the `pagination_class` can be removed here, as the current value `LimitOffsetPagination` is our default value configured in the settings. The `ListAPIView` being a child of `GenericAPIView`, `PollsRecommendationsView` should automatically inherit of this default value.

See:
https://github.com/encode/django-rest-framework/blob/efc7c1d664e5909f5f1f4d07a7bb70daef1c396e/rest_framework/generics.py#L46

```suggestion
```

GresilleSiffle (approved):

PR #583, opened 2022-02-15
[#463][front] : adds language icon in the collapsed navigation bar

### LINKED ISSUE #463: The language button should be on the landing page
Currently, the language icon is not in the collapsed navigation bar
image

If find it weird that it is hidden inside the navigation bar.

The "language icon" could at least be visible in the collapsed bar, like that

image

Or be added on the top left of the page.

ramprakashram:
Context:
Language Icon is missing on the sidepanel when it is collapsed.
This PR fixes it by adding the LanguageSelector component when it is collapsed and hides the text when collapsed.

Fixes #463

### DIFF
--- a/frontend/src/components/LanguageSelector.tsx
+++ b/frontend/src/components/LanguageSelector.tsx
@@ -7,7 +7,7 @@ import MenuButton from './MenuButton';
 import makeStyles from '@mui/styles/makeStyles';
 
 interface LanguageSelectorProps {
-  hideLanguageName?: boolean;
+  languageName?: boolean;
 }
 
 const useStyles = makeStyles({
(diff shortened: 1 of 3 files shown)


GresilleSiffle (changes requested):
Thanks @ramprakashram for your contribution :+1: 

The current implementation works fine but can be further improved by using only one `<LanguageSelector>` component instead of two (one in the footer, and one below).

Less code, less bugs ; )

I'd suggest to simply remove the `<LanguageSelector>` from the `<Footer>` to avoid JSX duplication, and toggle its `hideLanguageName` prop according to the value of `drawOpen`.

We can even rename `hideLanguageName` into simply `languageName` to be able to use the `drawOpen` value directly as a prop, without using the negation with `!` (this will be easier to read).


Like this:
```JSX
{drawerOpen && <Footer />}
<LanguageSelector languageName={drawerOpen} />
```

Thank you.

ramprakashram:
Sure @GresilleSiffle , will update as you suggested.
Thanks

ramprakashram:
@GresilleSiffle I have updated the PR with the following changes as you suggested:

- Moved LanguageSelector Component from the Footer Component to SideBar Component.
- Renamed the prop _hideLanguageName_ in LanguageSelector Component to _languageName_

Thanks for your review comments :)

amatissart (approved):
Looks good to me. Thanks!

I took the liberty of applying small adjustements in the SideBar style to align the language selector icon with the other icons.

ramprakashram:
@amatissart My bad, I missed that alignment somehow. Thanks.

GresilleSiffle (approved):

PR #584, opened 2022-02-15
[front] refactor: `Comparison` now uses UIDs (and more)

GresilleSiffle:
**related to** #527 

---

This pull request make some front end components compatible with the new entity `uid` instead of the previous `video_id`. The main change is in the `<Comparison>` component which is now able to handle URLs containing UIDs (it's still compatible with the old `videoA` / `videoB` parameters, this keeps all previously shared comparison links valid).

To keep the back end backward compatible, its API responses include the `uid` field **in addition** to `video_id`.

To keep the front end backward compatible, the components **fallback** to `video_id` if there is no `uid` field in a video object.

### details

`features.videos.VideoCard.VideoCard`
- [x] now uses `uid`
- [x] is backward compatible

`features.comparisons.Comparison.Comparison`
- [x] now compatible with URL parameters `uidA` / `uidB`
- [x] still compatible with legacy URL parameters `videoA` / `videoB`
- [x] also works with mixed `uidA` / `videoB` and  `videoA` / `uidB`
- [x] when conflicting parameters co-exist, the couple `uidA` / `uidB` prevail 

`utils.action`
- [x] `AddToRateLaterList` is now compatible with `uid`
- [x] `RemoveFromRateLater` is now compatible with `uid`
- [x] `CompareNowAction` is now compatible with `uid`

`features.videos.PublicStatusAction.PublicStatusAction`
- [x] now compatible with `uid`

### DIFF
--- a/frontend/src/features/comparisons/Comparison.tsx
+++ b/frontend/src/features/comparisons/Comparison.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect, useCallback, useMemo } from 'react';
-import { useHistory, useLocation } from 'react-router-dom';
+import { Redirect, useHistory, useLocation } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 
 import makeStyles from '@mui/styles/makeStyles';
@@ -39,6 +39,29 @@ const useStyles = makeStyles((theme: Theme) => ({
     color: theme.palette.text.secondary,
   },
 }));
+
+const rewriteLegacyParameters = (
+  uidA: string,
+  uidB: string,
+  legacyA: string | null,
+  legacyB: string | null,
+  paramVidA: string,
+  paramVidB: string
+) => {
+  const searchParams = new URLSearchParams();
+  searchParams.append(paramVidA, uidA);
+  searchParams.append(paramVidB, uidB);
+
+  if (legacyA && uidA === '') {
+    searchParams.append(paramVidA, UID_YT_NAMESPACE + legacyA);
+  }
+
+  if (legacyB && uidB === '') {
+    searchParams.append(paramVidB, UID_YT_NAMESPACE + legacyB);
+  }
+
+  return searchParams;
+};
 
 /**
  * The comparison UI.
@@ -88,24 +111,18 @@ const Comparison = () => {
     : searchParams.get(legacyParams.vidB) || '';
 
   // step 3: Clean the URL by replacing legacy parameters by UIDs.
-  useEffect(() => {
-    const legacyA = searchParams.get(legacyParams.vidA);
-    const legacyB = searchParams.get(legacyParams.vidB);
-
-    searchParams.delete(legacyParams.vidA);
-    searchParams.delete(legacyParams.vidB);
-
-    if (legacyA && uidA === '') {
-      searchParams.append(uidParams.vidA, UID_YT_NAMESPACE + legacyA);
-    }
-
-    if (legacyB && uidB === '') {
-      searchParams.append(uidParams.vidB, UID_YT_NAMESPACE + legacyB);
-    }
-
-    history.push('?' + searchParams.toString());
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  const legacyA = searchParams.get(legacyParams.vidA);
+  const legacyB = searchParams.get(legacyParams.vidB);
+  searchParams.delete(legacyParams.vidA);
+  searchParams.delete(legacyParams.vidB);
+  const newSearchParams = rewriteLegacyParameters(
+    uidA,
+    uidB,
+    legacyA,
+    legacyB,
+    uidParams.vidA,
+    uidParams.vidB
+  );
 
   const [selectorA, setSelectorA] = useState<VideoSelectorValue>({
     videoId: videoA,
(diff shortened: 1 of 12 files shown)


frontend/src/features/comparisons/Comparison.tsx
```diff
@@ -80,21 +103,25 @@ const Comparison = () => {
       if (searchParams.get(videoKey) !== videoId) {
         searchParams.delete(videoKey);
         if (videoId) {
-          searchParams.append(videoKey, videoId);
+          if (Object.values(uidParams).includes(videoKey)) {
+            searchParams.append(videoKey, UID_YT_NAMESPACE + videoId);
+          } else {
+            searchParams.append(videoKey, videoId);
+          }
```
amatissart, line 110: I think that we could find a simpler way.

What about considering `videoA` and `videoB` as deprecated, and replacing them with the new `uidA` and `uidB` in a separate `useEffect` executed on the initial page load. Then, the rest of the implementation could assume that the new uid params are used.

amatissart (commented):
There is a weird behavior on my local environment: the URL is rewritten multiple times when accessing a comparison page with `uidA` and `uidB` params, possibly due to a double-decoding issue.

frontend/src/features/comparisons/Comparison.tsx
```diff
@@ -61,8 +62,30 @@ const Comparison = () => {
...
+      uidA: 'uidA',
+      uidB: 'uidB',
+    };
+  }, []);
+
+  // try to read UIDs from the URL...
+  const encodedUidA: string = searchParams.get(uidParams.uidA) || '';
+  const encodedUidB: string = searchParams.get(uidParams.uidB) || '';
+  const uidA: string = encodedUidA ? decodeURIComponent(encodedUidA) : '';
+  const uidB: string = encodedUidB ? decodeURIComponent(encodedUidB) : '';
```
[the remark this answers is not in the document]
GresilleSiffle: That's true, thanks!

Fixed by https://github.com/tournesol-app/tournesol/pull/584/commits/cf16d7b9c465b63f7beca8b7bb42ac62940836cd.

GresilleSiffle:
> There is a weird behavior on my local environment: the URL is rewritten multiple times when accessing a comparison page with `uidA` and `uidB` params, possibly due to a double-decoding issue.

Thanks for spotting this. 

After few tests, it's not related to a double decoding issue. I found the bug in the `onChange` method, it's caused by a uid treated like an id. Gonna fix it.

frontend/src/utils/video.ts
```diff
@@ -16,6 +17,44 @@ export function isVideoIdValid(videoId: string) {
...
+ *
+ * The `uid` is expected to have the form `{namespace}:{id}`. The potential
+ * occurrences of the delimiter `:` in the identifier are considered part of it.
+ *
+ * Example with different `uid`:
+ *
+ *     yt:videoABCDEF -> videoABCDEF
+ *     yt:video:ABCDE -> video:ABCDE
+ */
+export function idFromEntity(entity: VideoObject): string {
```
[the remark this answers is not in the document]
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/584/commits/efbc55e1e6973f392b6584b40d25cfafdc405cdd.

frontend/src/features/comparisons/Comparison.tsx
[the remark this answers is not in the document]
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/584/commits/efbc55e1e6973f392b6584b40d25cfafdc405cdd.

frontend/src/features/comparisons/Comparison.tsx
```diff
@@ -61,8 +62,50 @@ const Comparison = () => {
...
+      searchParams.append(uidParams.vidA, UID_YT_NAMESPACE + legacyA);
+    }
+
+    if (legacyB && uidB === '') {
+      searchParams.append(uidParams.vidB, UID_YT_NAMESPACE + legacyB);
+    }
+
+    history.push('?' + searchParams.toString());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
amatissart, line 108: Using `useEffect` still leads to unexpected side-effects. This effect runs after the first render, and `<VideoSelector>` components may trigger URLs rewrite themselves. 

To avoid this complexity, I would recommend to handle this redirection as soon as possible, for example by returning a react-router `<Redirect>` on the first render, if legacy params need to be rewritten.

Something like:
```
if (hasLegacyParams) {
  const newParams = rewriteParams()
  return (
    <Redirect
      to={{ pathname: location.pathname, search: newParams.toString() }}
    />
  );
}
```

GresilleSiffle: Sure, I switched from `useEffect` to a redirection and removed unnecessary code in the last 4 commits.

frontend/src/features/comparisons/Comparison.tsx
```diff
@@ -61,8 +90,35 @@ const Comparison = () => {
...
+      vidA: 'uidA',
+      vidB: 'uidB',
+    };
+  }, []);
+  const legacyParams: { vidA: string; vidB: string } = useMemo(() => {
+    return {
+      vidA: 'videoA',
+      vidB: 'videoB',
+    };
+  }, []);
```
amatissart, line 104: Instead of using `useMemo`, couldn't these key names be defined as constants outside of the components ?

amatissart (approved):

PR #586, opened 2022-02-16
Fix empty metadata being rejected in Entity admin form

amatissart:
The bug was observed when saving an Entity in the admin form, even if the value of "metadata" was unchanged.


GresilleSiffle (approved):

lfaucon (commented):
Awesome! Thanks 😊

PR #587, opened 2022-02-16
[front] fix: PUT comparison request encoding chars twice

GresilleSiffle:


amatissart (approved):
Looks good to me.

I confirm that the bug is related to the OpenAPI client upgrade in #580.
In the recent releases, the generated client is in charge of encoding the URL path component, using `encodeURI` by default:
https://github.com/ferdikoomen/openapi-typescript-codegen/blob/v0.19.0/docs/openapi-object.md#openapiencode_path

PR #590, opened 2022-02-17
Update README.md

### LINKED ISSUE #573: Making README in backend more clear for new developers
When I was going through the README In backend [README](https://github.com/tournesol-app/tournesol/blob/main/backend/README.md) , I found it confusing why should I add a google API key. it would be nice to explain when it useful to do so.

sintaz1992:
closes #573

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -57,6 +57,8 @@ Both `dev-env/run-docker-compose.sh` and `dev-env/run-db-and-local-django.sh` de
 
 ## Setup Google Api Key
 
+* This step is important in development, especially if you add a new video to the database and the relevant information should be fetched from Youtube API.
+
 * Go to https://console.cloud.google.com/apis/ and create a new project named `tournesol`
 
 * Setup you credentials by getting a API key. Do not restrict its use for development purpose.
--- a/backend/README.md
+++ b/backend/README.md
@@ -57,7 +57,7 @@ Both `dev-env/run-docker-compose.sh` and `dev-env/run-db-and-local-django.sh` de
 
 ## Setup Google Api Key
 
-* This step is important in development, especially if you add a new video to the database and the relevant information should be fetched from Youtube API.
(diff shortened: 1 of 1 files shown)


GresilleSiffle:
Hello @sintaz1992 

Thanks for taking this issue.

I think the current additions are still a bit confusing for two reasons:

**(1)** The title `Setup Google Api Key` is at the same level of the `Install` title. It's not that bad, but by doing this we are telling the readers these two steps are not that related, and one could think its installation is complete without reading the API key section. I'd like propose to include this as the last subtitles of the `Install` section.

Maybe like this:

```markdown
### Setup Google API Key (recommended)
```

**(2)** I think the usage of words such a `important`, `good` and so on, are not always adapted in a documentation file. In our case, the word `important` doesn't explicitly state if the API key set-up is mandatory or optional. 

Maybe something like this:

```markdown
Configuring a Google API key allows the application to fetch video metadata directly from YouTube. The application should still start without it, but some features will be missing. It's recommended to get and configure this API key if you plan to contribute regularly to the project, to make your environment closer to the production one.
```

GresilleSiffle (changes requested):
I'd like to request changes if there is no objection : )

See my previous comment for the reasons and the requested changes.

GresilleSiffle:
> @GresilleSiffle 's comment point (1) also would require to modify the title for the Dependencies section? I'd be happy merging this PR with addressing only suggestino (2)

I think yes, but it's not necessary to address this in this pull request. Mainly because the top level `dev-env` sub-directory already mentions the `expect` dependency, it's not as "missable" as the API key set-up. We could leave it as is for now. A good place to move this section is inside the `Automatic installation (recommended)` title, as its first sub-section.

sintaz1992:
@GresilleSiffle Thanks for the review ...I used your suggestion for addressing point 2 and lets leave point 1 to another pull request. So it should be ready to merge?

GresilleSiffle (approved):

PR #595, opened 2022-02-17
Reject comparison when non-optional criteria is missing

amatissart:
A validation that should have been implemented in #560

### DIFF
--- a/backend/tournesol/tests/test_api_comparison.py
+++ b/backend/tournesol/tests/test_api_comparison.py
@@ -43,7 +43,7 @@ class ComparisonApiTestCase(TestCase):
     non_existing_comparison = {
         "entity_a": {"video_id": _uid_01.split(":")[1]},
         "entity_b": {"video_id": _uid_03.split(":")[1]},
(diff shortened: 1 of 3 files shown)


lfaucon (approved):
LGTM

backend/tournesol/tests/test_api_comparison.py
```diff
@@ -763,3 +763,13 @@ def test_invalid_criteria_in_comparison(self):
         self.assertContains(
             response, "not a valid criteria", status_code=status.HTTP_400_BAD_REQUEST
         )
+
+    def test_missing_non_optional_criteria_in_comparison(self):
```
GresilleSiffle, line 767: How about moving this test near to the other tests using the same POST view? 

Let's say just after the test `test_authenticated_cant_create_criteria_scores_without_mandatory`?

GresilleSiffle (approved):
Thanks!

Just added a minor comment that might help the maintenance of the `test_api_comparison` file.

Issue #455, opened 2022-01-14
[front] Default for language of videos based on the selected language for video recommendations


lfaucon:
Probably, select by default the list of language available in the user's `navigator.languages`.

Persist modifications to the language filter by the user in localStorage

Open questions:
- How to handle URL sharing?
- Include language option when requesting video for reco

amatissart:
* when recommendation is displayed: the "language" filter is added in the URL dynamically if it's not already present
* to select this language dynamically:
  * use the value stored in localStorage
  * or select a list of languages derived from `navigator.languages` 
* when user clicks on the language filter: the filter value is persisted in localStorage

PR #602, opened 2022-02-18
Set default languages on recommendations (#455)

sigmike:
For issue #455.

I changed the language translation keywords because I moved the available recommendations languages to a constant using the ISO language codes.

I made the change to keep a `language=` in the URL when all the languages are unselected and to do that I added an optional parameter to `useListFilter` that makes it add items to the URL even when the value is falsy. That means the `date` param is also kept in the URL when the user unselects all the date filters. It doesn't change the behavior so I kept it like that, but we could also make `useListFilter` only apply this behavior on specific params instead.

### DIFF
--- a/frontend/src/features/recommendation/LanguageFilter.tsx
+++ b/frontend/src/features/recommendation/LanguageFilter.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { ChoicesFilterSection } from 'src/components';
+import { availableRecommendationsLanguages } from 'src/utils/constants';
 
 interface Props {
   value: string;
@@ -10,11 +11,16 @@ interface Props {
 function LanguageFilter(props: Props) {
   const { t } = useTranslation();
 
-  const languageChoices = {
-    en: t('language.english'),
-    fr: t('language.french'),
-    de: t('language.german'),
-  };
+  const languageChoices: Record<string, string> = useMemo(
+    () =>
+      Object.fromEntries(
+        availableRecommendationsLanguages.map((language) => [
+          language,
+          t(`language.${language}`),
+        ])
+      ),
+    [t]
+  );
 
   return (
     <ChoicesFilterSection
--- a/frontend/src/features/recommendation/LanguageFilter.tsx
+++ b/frontend/src/features/recommendation/LanguageFilter.tsx
@@ -1,7 +1,10 @@
 import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { ChoicesFilterSection } from 'src/components';
-import { availableRecommendationsLanguages } from 'src/utils/constants';
+import {
+  availableRecommendationsLanguages,
+  getLanguageName,
+} from 'src/utils/constants';
 
 interface Props {
   value: string;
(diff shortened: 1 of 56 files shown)


frontend/src/features/recommendation/LanguageFilter.tsx
```diff
@@ -10,11 +11,16 @@ interface Props {
...
-    en: t('language.english'),
-    fr: t('language.french'),
-    de: t('language.german'),
-  };
+  const languageChoices: Record<string, string> = useMemo(
+    () =>
+      Object.fromEntries(
+        availableRecommendationsLanguages.map((language) => [
+          language,
+          t(`language.${language}`),
```
amatissart, line 19: Unfortunately, i18next-parser is not able to understand that 3 different translations keys are used here. When running `yarn run i18n:parse` the existing translation keys will be deleted.

I think that we need to keep an explicit object with all translation keys, similarly to how the criteria names are currently translated:
https://github.com/tournesol-app/tournesol/blob/0b6718530ea7db21ee5f9a9fcc5d2be8b156c829/frontend/src/utils/constants.ts#L30-L44

frontend/src/pages/videos/VideoRecommendation.tsx
```diff
@@ -34,13 +43,29 @@ function VideoRecommendationPage() {
   }
 
   useEffect(() => {
+    if (location.search === locationSearchRef.current) return;
+    locationSearchRef.current = location.search;
```
amatissart, line 47: Could you explain why this is necessary? Is it to avoid duplicate requests to the server? Is that not redundant with the effect dependencies?

frontend/src/utils/recommendationsLanguages.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    [[], ''],
+  ];
+
+  testCases.forEach(([input, expected]) =>
+    it(`converts ${JSON.stringify(input)}`, () => {
+      const languagesGetter = jest.spyOn(window.navigator, 'languages', 'get');
+      languagesGetter.mockReturnValue(input as string[]);
+      expect(recommendationsLanguagesFromNavigator()).toEqual(expected);
+    })
+  );
```
amatissart, line 21: Thanks for the unit tests! They are greatly appreciated, and very relevant in this case :+1:

amatissart (commented):
It looks very good to me, except a few minor comments/questions. Thanks again @sigmike!

frontend/src/features/recommendation/LanguageFilter.tsx
sigmike: Ok, it should be fixed now.

frontend/src/pages/videos/VideoRecommendation.tsx
sigmike: I had to add new dependencies[[1]](https://github.com/tournesol-app/tournesol/pull/602/files#diff-216c121f61c465f9624a7b2c17723cdbaea037aa9a085d990423439036c4ff28L43-R68) to the effect hook, but we don't want to trigger a reload when these new dependencies are changed (but we still want the function to be updated).

In this case `searchParams` will change at the same time as the location so it's not a problem, but `history` may change at any time. It probably doesn't, but to be safe and to handle future changes where new dependencies may really change, I added this check to make sure we trigger a reload only when the location changes.

sigmike: > (but we still want the function to be updated)

Actually we don't really care in this case, but removing the dependencies would raise a warning in the linter.

amatissart (approved):

frontend/src/pages/videos/VideoRecommendation.tsx
amatissart: Ok I've just added a comment to explain this.

PR #607, opened 2022-02-20
Fixed issue 571 - Display statistics on Home page and About page

iamnkc:
Please find the attached image for statistics on Home page.

Full Screen
image

Small screen
image

Please review and let me know if I need to change any piece of code.

### DIFF
--- a/frontend/src/components/Statssection.tsx
+++ b/frontend/src/components/Statssection.tsx
@@ -1,0 +1,91 @@
+import { Grid } from '@mui/material';
+import { StatsService } from 'src/services/openapi';
+import type { Statistics } from 'src/services/openapi';
+import React, { useEffect, useState } from 'react';
+import useMediaQuery from '@mui/material/useMediaQuery';
+
+interface statsType {
+  text: string;
+  count: number;
+  last_month_count: number;
+}
+
+const StatsUi = (props: statsType) => {
+  return (
+    <Grid item xs={4}>
+      <span style={{ fontSize: '12px' }}>{props.text}</span> <br />
+      <span style={{ fontSize: '50px', lineHeight: '1em' }}>
+        {' '}
+        {props.count}{' '}  
+      </span>{' '}
+      <br />
+      {
+        <span
+          style={
+            props.last_month_count >= 0 ? { color: 'green' } : { color: 'red' }
+          }
+        >
+          {(props.last_month_count >= 0 ? '+ ' : '') + props.last_month_count}
+        </span>
+      }
+    </Grid>
+  );
+};
+
+const StatsSection = () => {
+  const [data, setData] = useState({
+    user_count: 0,
+    last_month_user_count: 0,
+    video_count: 0,
+    last_month_video_count: 0,
+    comparison_count: 0,
+    last_month_comparison_count: 0,
+  });
+
+  const matches = useMediaQuery('(min-width:600px)');
+
+  const updateState = (value: Statistics) => {
+    setData({
+      user_count: value.user_count,
+      video_count: value.video_count,
+      comparison_count: value.comparison_count,
+      last_month_user_count: value.last_month_user_count,
+      last_month_video_count: value.last_month_video_count,
+      last_month_comparison_count: value.last_month_video_count,
+    });
+  };
+
+  useEffect(() => {
+    StatsService.statsRetrieve()
+      .then(updateState)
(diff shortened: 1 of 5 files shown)


frontend/src/components/Statssection.tsx
```diff
@@ -0,0 +1,91 @@
...
+        >
+          {(props.last_month_count >= 0 ? '+ ' : '') + props.last_month_count}
+        </span>
+      }
+    </Grid>
+  );
+};
+
+const StatsSection = () => {
+  const [data, setData] = useState({
```
sandre35, line 36: You should precise the type of `data` to `Statistics` like `useState<Statistics>`

frontend/src/components/Statssection.tsx
```diff
@@ -0,0 +1,91 @@
...
+      last_month_comparison_count: value.last_month_video_count,
+    });
+  };
+
+  useEffect(() => {
+    StatsService.statsRetrieve()
+      .then(updateState)
+      .catch((error) => {
+        console.log(error);
+      });
```
sandre35, line 63: There is no need to rewrite `setData` with `updateState` call directly `setData`

frontend/src/components/Statssection.tsx
```diff
@@ -0,0 +1,91 @@
...
+      last_month_video_count: value.last_month_video_count,
+      last_month_comparison_count: value.last_month_video_count,
+    });
+  };
+
+  useEffect(() => {
+    StatsService.statsRetrieve()
+      .then(updateState)
+      .catch((error) => {
+        console.log(error);
```
sandre35, line 62: use `console.error` here as it is an error that may happen here

frontend/src/pages/about/PublicDownloadSection.tsx
```diff
@@ -1,12 +1,32 @@
...
+    setUserCount(value.user_count);
+    setVideoCount(value.video_count);
+    setComparisonCount(value.comparison_count);
+  };
+
+  useEffect(() => {
+    StatsService.statsRetrieve()
+      .then(updateState)
+      .catch((error) => {
+        console.log(error);
```
sandre35, line 26: `console.error`

sandre35 (changes requested):
Hi @iamnkc,

Your contribution looks pretty cool. Please adress remarks below :)
Also, you seemed to duplicate the tournesol project into yours. There is no need to do that. You can just create a branch from the tournesol repository and then ask a PR to merge into main!

frontend/src/components/Statssection.tsx
```diff
@@ -0,0 +1,91 @@
...
+interface statsType {
+  text: string;
+  count: number;
+  last_month_count: number;
+}
+
+const StatsUi = (props: statsType) => {
+  return (
+    <Grid item xs={4}>
+      <span style={{ fontSize: '12px' }}>{props.text}</span> <br />
```
sandre35, line 16: Also, try to use the props `sx`instead of `style`
There is another active PR that dealing with this issue. So we try our best not to introduce anomaly into the code

GresilleSiffle:
Hello @iamnkc 

Thanks for your contribution :+1: 

Here are some feedback:

**(1)** we don't use the black background anywhere else in the application, it seems a bit rough compared to the rest of the page. to keep the colors balanced, I'd suggest to use either the blue already available of the white of the `Contribute!` section.

**(2)** I'd suggest to simplify the labels by using:
- active users instead of user count ( or even just users )
- comparisons instead of comparisons count
- rated videos instead of video count

**(3)** there is no indication about the meaning of the `+ {number}` under each stat. I'd suggest to at least use a caption always visible under the section, like `Tournesol statistics with their evolution during the last 30 days`. It could also be a paragraph before, or even a combination of title + subtitle. What are your preferences @aidanjungo @lfaucon @amatissart ? any other idea ? Adrien also proposed to use a tooltip.

iamnkc:
Hi @sandre35 and @GresilleSiffle , 
Thanks for the valuable feedback, I have pushed all the changes that @sandre35 as suggested. From the next issue onwards, I'll create a branch from main repo and raise a PR.  For display component, Please suggest the changes so that I can implement them.

GresilleSiffle:
Sure, we will review the UI with the rest of the team to give you clear guidelines.

lfaucon:
Rebased and replaced by #652

Issue #565, opened 2022-02-10
[front] bug: Sidepanel does not appear selected for page Recommendations


lfaucon:
Home appears selected when the Home page is open
<img width="398" alt="Screenshot 2022-02-10 at 17 57 56" src="https://user-images.githubusercontent.com/8818081/153457727-c1cd9348-ffb3-4f87-99a7-9d1be8db8f1c.png">

Recommendations does not
<img width="390" alt="Screenshot 2022-02-10 at 17 58 03" src="https://user-images.githubusercontent.com/8818081/153457735-7b45a831-0cd4-4a65-90c3-35fa7ef2614e.png">

PR #609, opened 2022-02-20
Correct logic for highlighted component in navbar

sandre35:
Related issue: https://github.com/tournesol-app/tournesol/issues/565

The problem was that the target url beneath selecting "Recommendations" has the url parameter `date=Month`
I choose to correct the logic by deleting all possibles url params when checking for the component to highlight


lfaucon (approved):
Thanks for the fix!

Issue #528, opened 2022-02-02
refactor: contributor_ratings API now handles `Poll`


GresilleSiffle:
**requires** #527

also see the issue #509

---

Refactor the contributor_ratings API to handles `Poll`.

**back end**

Create a migration to set the value of the contributor_ratings poll FK to the previously created Poll with name `videos`.

Create a migration to remove the NULLABLE constraint on this FK.

`GET /users/me/contributor_ratings/{pollname}` get all my ratings per entity for this poll

`POST /users/me/contributor_ratings/{pollname}` (e.g to initialize an entity rating as public/private for this poll)

`GET|PUT|PATCH /users/me/contributor_ratings/{pollname}/{uid}` get rating, or update the private/public status

`PATCH /users/me/contributor_ratings/{pollname}/_all` to update all ratings statuses in a poll

Adapt the tests.

**front end**

Update the API schema and re-generate the services.

Update the components to use the new endpoints.

PR #611, opened 2022-02-21
[back] refactor: contributor ratings API now compatible with `Poll`

GresilleSiffle:
**related to** #528 

---

This pull request addresses a subset of tasks described in #528 .

**(1)** It adds the poll name as a new URL path parameter in the ratings API, and **(2)** make the front end compatible.

**(3)** It also removes the default poll value of the `ContributorRating.poll` field. Now the poll must be set before saving any `ContributorRating` instance. According to this I've updated the `ml_train` command to explicitly set the default poll before saving ratings.

I created a new mixin view able to work with the `{poll_name}` URL path parameter. It retrieves the parameter from the URL and make if available in the view and its serializer. It's the same behaviour we already have in the comparisons API, but factorized to avoid code duplication and related bugs. I didn't update the comparisons API to use this new mixin to keep this PR focused on ratings, but I'll do right after this PR is merged.

**back end to-do**

`GET users/me/contributor_ratings/`
- [x] now becomes `users/me/contributor_ratings/{poll_name}`
- [x] update the tests

`POST users/me/contributor_ratings/`
- [x] now becomes `users/me/contributor_ratings/{poll_name}`
- [x] update the tests

`GET users/me/contributor_ratings/{video_id}`
- [x] now becomes `users/me/contributor_ratings/{poll_name}/{video_id}`
- [x] update the tests

`PATCH users/me/contributor_ratings/{video_id}`
- [x] now becomes `users/me/contributor_ratings/{poll_name}/{video_id}`
- [x] update the tests

`PUT users/me/contributor_ratings/{video_id}`
- [x] now becomes `users/me/contributor_ratings/{poll_name}/{video_id}`
- [x]  **there is current no test**, no need to adapt them

**front end to-do**
 - [x] re-generate the OpenAPI schema
 - [x] update the requests requiring the new `pollName` parameter


backend/tournesol/views/mixins/poll.py
```diff
@@ -0,0 +1,46 @@
...
+        except ObjectDoesNotExist:
+            return self.response_404_poll_doesnt_exist(request_kwargs[self.poll_parameter])
+
+    def response_404_poll_doesnt_exist(self, poll_name: str):
+        return Response(
+            {
+                "detail": "The requested poll {0} doesn't exist.".format(poll_name),
+            },
+            status=status.HTTP_404_NOT_FOUND,
+        )
```
amatissart, line 32: Returning a 404 response is not correct here, as `.initial()` is unable to handle it.

An exception should be used instead, possibly with `get_object_or_404` shortcut provided by Django:

```suggestion
    def poll_from_kwargs_or_404(self, request_kwargs):
        return get_object_or_404(Poll, name=request_kwargs[self.poll_parameter])
```

amatissart (approved):

PR #614, opened 2022-02-22
Adds tournesol_score field to entity model, updates frontend and extension to read from it

### LINKED ISSUE #519: [back] feat: Include aggregated scores in API response containing criteria scores
Context:

Currently the frontend and the extensions both compute an aggregated score. It makes more sense for this score to be computed only in one place on the backend. 

Acceptance:
- [ ] Update necessary backend api routes to include the Tournesol's scores
- [ ] Update the frontend to use the score provided by the backend
- [ ] Update the extension to use the code provided by the backend

sintaz1992:
closes #519 
By calculating the tournesol_score when ml_train is happening, the front-end and extesnion can directly read the tournesol_score from backend.

- I needed to change the url address in browser_extension to test the update locally (from api.tounesol to localhost:8000)
which should be modified before merging. However in general is there a way the browser understands getting from localhost or api on its own? 
- I had to modify the test samples for video Card (by adding tournesol_score)
-


backend/ml/management/commands/ml_train.py
```diff
@@ -79,6 +80,10 @@ def save_data(video_scores, contributor_rating_scores):
     Saves in the scores for Entities and ContributorRatings
     """
     EntityCriteriaScore.objects.all().delete()
+    for video_id, criteria, score, uncertainty in video_scores:
+       video=EntityCriteriaScore(entity_id=video_id).entity
+       video.tournesol_score+=10*score
```
sandre35, line 85: You should use some plugin within your IDE to highlight syntax issues. Install at least `isort` and `pylint` or `flake8`.
Details of unsuccessful checks cqn be seen in logs of Gihub CI.
Here you must insert spaces around arithmetic operators.

backend/tournesol/migrations/0033_alter_entity_tournesol_score.py
```diff
@@ -0,0 +1,18 @@
...
+        ('tournesol', '0032_entity_tournesol_score'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='entity',
+            name='tournesol_score',
+            field=models.FloatField(default=0, help_text='Tournesol ScoreCalculated from the criteria_score from certified contributors'),
+        ),
+    ]
```
sandre35, line 18: You have created to much migrations for this issue. You can combine them into one squash.
The more migrations you create, the longer the database will take to update its columns

browser-extension/addTournesolRecommendations.js
```diff
@@ -32,6 +32,7 @@ const convertDurationToClockDuration = (duration) => {
 
 
 const calculateScore= (video) => {
+  console.log(video)
```
sandre35, line 35: A little oversight ;) 
This line of code can be deleted

browser-extension/utils.js
```diff
@@ -42,7 +42,7 @@ export const fetchTournesolApi = async (url, method, data) => {
   if (data) {
     body["body"]= JSON.stringify(data)
   }
-  return fetch(`https://api.tournesol.app/${url}`, body).then(r => {
+  return fetch(`http://localhost:8000/${url}`, body).then(r => {
```
sandre35, line 45: This change looks weird to me.

sandre35 (changes requested):
Hi @sintaz1992,
Really good contribution !
Please look at my remarks, especially concerning your IDE :)

sandre35:
Hi @amatissart, @GresilleSiffle,
I think a little rebase is required on the PR to solve conflicts and update the code to the latest refractor of the back :)

GresilleSiffle:
Hello @sintaz1992 @sandre35 

Thanks again for the work and the review, it's really appreciated.

Yes, the pull request is based on an old version of Tournesol and must be rebased or updated in some way. The diff currently shows old code that is going to be re-introduced after the merge, rebasing will fix this.

I'm currently working on the contributor ratings API refactor, which will change some files also changed by this pull request. In order to ease the work of everyone, I recommend to wait for the PR https://github.com/tournesol-app/tournesol/pull/611 to be merged, before rebasing this one.

@sintaz1992 if you need help for rebasing do no hesitate to ask, we will help. 

Here is one advise about migrations:

You can create only one migration of you are just altering the models without directly touching to the data. It's a good rule of thumb to separate the data migration and the schema migration to be able as a developer to run the data migration back and forth easily.

How to create only one migration? The automatic way to do this, is to revert the migration one by one, by running `python manage.py migrate tournesol 0026`, then deleting the migration files from 0027 to the last, then running `python manage.py migrate makemigrations` again.

backend/ml/management/commands/ml_train.py
GresilleSiffle: If it helps, there is a section in the back end's `README.md` describing how to run locally the code quality checks.

Here:
https://github.com/tournesol-app/tournesol/blob/main/backend/README.md#code-quality

With docker the command is:
```shell
docker exec tournesol-dev-api scripts/ci/lint.sh
```

The output can be impressive at first, because there are a lot of warnings ( not errors ) we should take care of, or just disable. You can leave those warnings. What's relevant for you is to check if you see `chk1=1`, `chk2=1` (these are in the top 5 lines of the output), or `chk3=1` (this one is at the end), the digit 1 means one check has failed and the output will describe the problem.

Here is an example of error, viewed directly from GitHub, you need to click on the Python Code Quality Checks section, then navigate to the line 10:
https://github.com/tournesol-app/tournesol/runs/5289368490#step:6:10

lfaucon (dismissed):

amatissart (commented):
One thing to keep in mind: with the ongoing "core refactor" to handle multiple polls in the database, this "tournesol_score" will need to be associated to a pair (poll, entity) instead of being defined directly on the `Entity` model.

@sintaz1992 Let us know about your availability in the next few days, to address the requested changes in this PR, and possibly assign yourself on the follow-up task #631.

GresilleSiffle:
I just squashed and rebased the branch. The current diff should be identical as the previous one, minus one comment.

It should be easier to work on it now : )

Note for the future, let's avoid merging `main` into this branch to avoid adding conflicts.

backend/tournesol/models/entity.py
```diff
@@ -72,6 +72,13 @@ class Meta:
         "from certified contributors",
     )
 
+    tournesol_score = models.FloatField(
+        null=False,
+        default=0,
+        help_text="Tournesol Score"
+        "Calculated from the criteria_score from certified contributors",
```
GresilleSiffle, line 79: Note for self:
- should be nullable
- null by default
- "le score de l'entité agrégée sur l'ensemble des critères et l'ensemble des utilisateurs pour un scrutin donnée"

GresilleSiffle:
### to-do

**field Entity.tournesol_score**
- [x] should accept null value
- [x] should allow blank value in forms validation 
- [x] should have null has a default value

backend/ml/management/commands/ml_train.py
```diff
@@ -98,15 +103,39 @@ def save_data(video_scores, contributor_rating_scores, poll, trusted_only=True):
             ]
         )
 
+        entities = []
+        for entity in Entity.objects.prefetch_related("criteria_scores").iterator():
+            entity.tournesol_score = 10 * sum(
+                [
+                    criterion.score
+                    for criterion in entity.criteria_scores.iterator()
+                    if criterion.score
```
GresilleSiffle, line 112: not useful

backend/ml/management/commands/ml_train.py
```diff
@@ -98,15 +103,39 @@ def save_data(video_scores, contributor_rating_scores, poll, trusted_only=True):
             ]
         )
 
+        entities = []
+        for entity in Entity.objects.prefetch_related("criteria_scores").iterator():
+            entity.tournesol_score = 10 * sum(
+                [
```
GresilleSiffle, line 109: not useful

frontend/src/utils/entity.ts
```diff
@@ -10,6 +10,7 @@ export const videoFromRelatedEntity = (entity: RelatedEntityObject): Video => {
     views: entity.metadata.views,
     uploader: entity.metadata.uploader,
     language: entity.metadata.language,
+    tournesol_score: null,
```
GresilleSiffle, line 13: the `tournesol_score` can be used in the `VideoSerializerWithCriteria` intead of `VideoSerializer` to avoid this addition

backend/ml/management/commands/ml_train.py
```diff
@@ -100,6 +101,14 @@ def save_data(video_scores, contributor_rating_scores, poll, trusted_only=True):
             for video_id, criteria, score, uncertainty in video_scores
         )
 
+        entities = []
+        for entity in Entity.objects.prefetch_related("criteria_scores").iterator():
+            entity.tournesol_score = 10 * sum(
+                criterion.score for criterion in entity.criteria_scores.iterator()
+            )
+            entities.append(entity)
+        Entity.objects.bulk_update(entities, ["tournesol_score"])
```
amatissart, line 110: I see some issues with this implementation: 
 * all entities will be updated, including entities not related to the current poll, as well as entities without score (where `tournesol_score` should remain null)
 * in Django 4.0 [.prefech_related() is ignored when used with `.iterator()`](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#prefetch-related`.iterator)
 Although it may change in future releases, see https://github.com/django/django/pull/15334)

My suggestion would be to remove `.iterator()` and filter on the score poll;
```python
        for entity in (
            Entity.objects.filter(criteria_scores__poll=poll)
            .distinct()
        ):
```

amatissart (changes requested):

frontend/src/utils/entity.ts
amatissart: Is there an obstacle that prevents this change?

backend/ml/management/commands/ml_train.py
lfaucon: I find it weird to fetch the criteria scores from the db in order to compute the tournesol_score at this moment because the ML algorithm just finished computing them and we are in the saving phase. Why not computing the tournesol_scores directly from the list `video_scores` returned by the ML algorithm?

Also we may in the fuure make the tournesol_score more complicated thant just `10 x sum(criteria_scores)`. It would probably be beneficial to already separate the computation in its own method.

sintaz1992:
> One thing to keep in mind: with the ongoing "core refactor" to handle multiple polls in the database, this "tournesol_score" will need to be associated to a pair (poll, entity) instead of being defined directly on the `Entity` model.
> 
> @sintaz1992 Let us know about your availability in the next few days, to address the requested changes in this PR, and possibly assign yourself on the follow-up task #631.

Thanks to @GresilleSiffle for taking care of the issues ...I learned a lot just by looking at the commits

backend/ml/management/commands/ml_train.py
GresilleSiffle: > I find it weird to fetch the criteria scores from the db in order to compute the tournesol_score at this moment because the ML algorithm just finished computing them and we are in the saving phase. Why not computing the tournesol_scores directly from the list video_scores returned by the ML algorithm?

I've considered it at first, but the loop would have been less readable and more bug-prone, mainly because the items in `video_scores` are not entities, but criteria with their score ordered by criteria name. It would have required either a a copy with a transformation of the data structure, or a loop with several conditions inside. That's why I've opted for an extra SQL query, which simplify all of this with a rather low cost.

> Also we may in the future make the tournesol_score more complicated thant just 10 x sum(criteria_scores). It would probably be beneficial to already separate the computation in its own method.

That's a good idea. As the algorithm is going to change I don't know if it's still relevant but yes, it's a good idea.

frontend/src/utils/entity.ts
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/614/commits/0b4be828aaa764bebbccf09b7474ce6e220fb559.

backend/ml/management/commands/ml_train.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/614/commits/a48fd5d4ca5e65b8714b5314da3ee41e0080eaa1.

backend/ml/management/commands/ml_train.py
```diff
@@ -100,6 +101,18 @@ def save_data(video_scores, contributor_rating_scores, poll, trusted_only=True):
...
         )
 
+        entities = []
+        for entity in (
+            Entity.objects.filter(criteria_scores__poll=poll)
+            .distinct()
+            .prefetch_related("criteria_scores")
+        ):
+            entity.tournesol_score = 10 * sum(
+                criterion.score for criterion in entity.criteria_scores.iterator()
```
amatissart, line 111: After running "ml_train" with debug messages about SQL queries, I can see that this `.iterator()` too would cancel the effect of `prefetch_related()`:
```suggestion
                criterion.score for criterion in entity.criteria_scores.all()
```

frontend/src/features/videos/VideoCard.tsx
```diff
@@ -192,7 +192,9 @@ function VideoCard({
   const { getCriteriaLabel } = useCurrentPoll();
 
   const videoId = videoIdFromEntity(video);
-  let total_score = 0;
+  const tournesolScore =
+    'tournesol_score' in video ? video.tournesol_score : null;
```
amatissart, line 196: As a consequence, the score would not be visible after deployment, until a new ml_train has run.

amatissart (approved):

PR #617, opened 2022-02-23
[back] refactor: comparisons API now uses `PollScopedViewMixin`

GresilleSiffle:
Following the work of #611 , the comparisons API now uses the `PollScopedViewMixin` class.

As we are using several level of inheritance in the comparisons API class based views, I took the time to ensure the `PollScopedViewMixin` is correctly used by adding a new test for each HTTP methods.

### DIFF
--- a/backend/tournesol/tests/test_api_ratings.py
+++ b/backend/tournesol/tests/test_api_ratings.py
@@ -83,7 +83,7 @@ class RatingApi(TestCase):
         # the default poll must not contain the rating
         with self.assertRaises(ObjectDoesNotExist):
             ContributorRating.objects.select_related("poll", "user", "entity").get(
-                poll__name=self.poll_videos,
+                poll=self.poll_videos,
                 user=self.user1,
                 entity__video_id=self.video3.video_id,
             )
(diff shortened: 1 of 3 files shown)


backend/tournesol/tests/test_api_ratings.py
```diff
@@ -62,17 +62,28 @@ def test_authenticated_cant_create_non_existing_poll(self):
...
+                poll__name=non_existing_poll,
+                user=self.user1,
+                entity__video_id=self.video3.video_id,
+            )
+
+        # the default poll must not contain the rating
         with self.assertRaises(ObjectDoesNotExist):
             ContributorRating.objects.select_related("poll", "user", "entity").get(
-                poll=self.poll_videos,
+                poll__name=self.poll_videos,
```
amatissart, line 86: `self.poll_videos` is an instance of Poll, so this lookup would not match anyway:

```suggestion
                poll=self.poll_videos,
```

GresilleSiffle: Sorry, bad copy / paste.

amatissart (approved):

Issue #572, opened 2022-02-11
refactor: Run ML for each poll


amatissart:
Instances of `Comparison`, `EnityCriteriaScore` and `ContributorRatingCriteriaScore` are now associated with a `poll`.

So the ML should run for each poll separately.

Pseudo code: 
```
for each poll
   for each criteria in this poll 
      run "ml_train"
```

amatissart:
@farnots Don't hesitate to ping me if the description is unclear

PR #620, opened 2022-02-23
Run ML on each Poll

farnots:
Linked to #572 

⚠️ If the PR https://github.com/tournesol-app/tournesol/pull/561 is merged before, some conflict should be fixed.


amatissart:
@farnots Even if "video" is currently the only type that can be used in a Poll, it should still be possible to set up a test case where a second poll about videos would be created, possibly with some new criterias on videos. Then, a test could check that `EntityCriteriaScore` instances are correctly created to cover all pairs of (poll, criteria).

lfaucon (approved):
Apart from the lack of test, this works for me :)

farnots:
> @farnots Even if "video" is currently the only type that can be used in a Poll, it should still be possible to set up a test case where a second poll about videos would be created, possibly with some new criterias. Then, a test could check that `EntityCriteriaScore` instances are correctly created to cover all pairs of (poll, criteria).

I added a factory to create a new poll and test the code against it

amatissart:
@farnots Now that #561 has been merged and deployed, you can solve the conflicts on this branch.

farnots:
@amatissart @lfaucon OK for review

PR #622, opened 2022-02-24
Use frontend recommendations languages in extension

sigmike:
Related to #349.

It requires either [!602](https://github.com/tournesol-app/tournesol/pull/602) to be deployed in production or a manual change of the local storage on tournsol.app.

For now this PR only includes the code that loads the recommendations languages from the website's localStorage and uses them in the recommendations request.

If the recommendations languages change on the website and you refresh the youtube home page the recommendations are still in the previous language because the iframe that refreshes the data hasn't finished yet when the recommendations are loaded. A second refresh will load recommendations in the updated languages. A solution may be to wait for the iframe to complete but it may be tricky and may add a delay. A better solution may be to trigger the extension's storage change immediatly when the website changes its own storage (with a custom event captured by the extension for example).

Another solution would be to store the recommendations languages in the backend and not care about it in the extension at all. The extension would omit the `language` parameter in the request and the backend would use the stored languages of the user.

Remaining to do:

- [x] Remove the language selection UI from the recommendations
- [x] Use the navigator languages when there's no stored languages

### DIFF
--- a/browser-extension/manifest.json
+++ b/browser-extension/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "Tournesol Extension",
-  "version": "2.1.3",
+  "version": "2.1.4",
   "description": "Open Tournesol directly from Youtube",
   "permissions": [
     "https://tournesol.app/",
--- a/browser-extension/manifest.json
+++ b/browser-extension/manifest.json
@@ -58,13 +58,10 @@
     },
     {
       "matches": ["https://tournesol.app/*"],
-      "js": ["fetchTournesolToken.js"],
-      "run_at": "document_end",
-      "all_frames": true
-    },
-    {
-      "matches": ["https://tournesol.app/*"],
-      "js": ["fetchTournesolRecommendationsLanguages.js"],
+      "js": [
+        "fetchTournesolToken.js",
+        "fetchTournesolRecommendationsLanguages.js"
+      ],
       "run_at": "document_end",
       "all_frames": true
     }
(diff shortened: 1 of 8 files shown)


browser-extension/manifest.json
```diff
@@ -61,6 +61,12 @@
       "js": ["fetchTournesolToken.js"],
       "run_at": "document_end",
       "all_frames": true
+    },
+    {
+      "matches": ["https://tournesol.app/*"],
+      "js": ["fetchTournesolRecommendationsLanguages.js"],
```
GresilleSiffle, line 67: Is there a reason to not put this script in the same array as `fetchTournesolToken.js`? They are both triggered at `document_end` with the same `all_frames` parameter.

GresilleSiffle (commented):
Hello @sigmike 

>  A better solution may be to trigger the extension's storage change immediatly when the website changes its own storage (with a custom event captured by the extension for example).

This indeed looks very elegant. Do you think it's possible to follow this path?

browser-extension/manifest.json
sigmike: No, I just forgot it was possible. I merged them.

Actually we could run it at `document_start` but it probably wouldn't be enough to solve the timing issue mentioned above.

sigmike:
> Do you think it's possible to follow this path?

I implemented that in d3d12071d36220570c67c3b8118a2f49a564172a

> Deprecate language choice from the extension (from #349)

I'm not sure what you meant by "deprecate". I removed the links completely in 4ee1a4ef6047d9570069e028c0313dd8d9345573.

lfaucon (commented):
Please increase the extension version in `manifest.json`.

lfaucon (approved):
Thanks for the changes

PR #630, opened 2022-02-27
feat: Display language names instead of codes in EntityAdmin

lfaucon:
As an admin, using the admin interface, I find more readable to see names of languages ("Spanish" and "Estonian") rather than codes ("es" and "et") so that I can identify and manually correct language assignments of video entities

### DIFF
--- a/backend/tournesol/admin.py
+++ b/backend/tournesol/admin.py
@@ -98,7 +98,8 @@ class EntityAdmin(admin.ModelAdmin):
     @staticmethod
     @admin.display(description="language", ordering="metadata__language")
     def get_language(obj):
-        return LANGUAGE_CODE_TO_NAME_MATCHING.get(obj.metadata.get("language"))
+        language_code = obj.metadata.get("language")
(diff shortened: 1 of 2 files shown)


backend/tournesol/admin.py
```diff
@@ -98,7 +98,7 @@ def get_publication_date(obj):
     @staticmethod
     @admin.display(description="language", ordering="metadata__language")
     def get_language(obj):
-        return obj.metadata.get("language")
+        return LANGUAGE_CODE_TO_NAME_MATCHING.get(obj.metadata.get("language"))
```
amatissart, line 101: As a precaution, the language code can still be used as a fallback.  
It would avoid confusion in case when an unknown value is present in the database, even though it should not happen given the current implementation.
```suggestion
        language_code = obj.metadata.get("language")
        return LANGUAGE_CODE_TO_NAME_MATCHING.get(language_code, language_code)
```

amatissart (approved):

PR #623, opened 2022-02-25
WIP / [back] refactor: comparisons API does not use `video_id` in body anymore

GresilleSiffle:

Issue #597, opened 2022-02-17
refactor: `RelatedVideoSerializer` should accept the `uid` in addition to `video_id`


GresilleSiffle:
Finally I choose to create a new `RelatedEntitySerializer` instead of making `RelatedVideoSerializer` able to work with both `uid` and `video_id`.

PR #635, opened 2022-02-28
[back][front] refactor: ratings API uses `RelatedEntitySerializer`

GresilleSiffle:
**related to** #597

**follows the work of the pull request** #623

### DIFF
--- a/backend/tournesol/serializers/rating.py
+++ b/backend/tournesol/serializers/rating.py
@@ -2,7 +2,7 @@ from django.db.models import Q
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import extend_schema_field
 from rest_framework.exceptions import ValidationError
-from rest_framework.fields import BooleanField, RegexField, SerializerMethodField
+from rest_framework.fields import BooleanField, CharField, SerializerMethodField
 from rest_framework.serializers import ModelSerializer, Serializer
 
 from core.utils.constants import YOUTUBE_VIDEO_ID_REGEX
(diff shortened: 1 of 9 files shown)


backend/tournesol/serializers/rating.py
```diff
@@ -51,24 +51,24 @@ def to_internal_value(self, data):
...
+        entity_serializer.is_valid(raise_exception=True)
+        entity = Entity.objects.get(uid=uid)
         user = self.context["request"].user
-        if user.contributorvideoratings.filter(entity=video).exists():
+        if user.contributorvideoratings.filter(entity=entity).exists():
             raise ValidationError(
-                "A ContributorRating already exists for this (user, video)",
+                "A ContributorRating already exists for this (user, entity)",
                 code="unique",
             )
```
amatissart, line 70: The uniqueness should be checked for (user, entity, poll) instead of (user, entity)

backend/tournesol/serializers/rating.py
```diff
@@ -51,24 +51,24 @@ def to_internal_value(self, data):
 
 
 class ContributorRatingCreateSerializer(ContributorRatingSerializer):
-    video_id = RegexField(YOUTUBE_VIDEO_ID_REGEX, write_only=True)
+    uid = RegexField(rf"yt:{YOUTUBE_VIDEO_ID_REGEX[1:]}", write_only=True)
```
amatissart, line 54: Tweaking the regex here does not seem very robust, and very specific to videos. Maybe we could remove this regex validation, and rely on the `RelatedEntitySerializer` to do the necessary checks (depending on the current poll, etc.)

backend/tournesol/serializers/rating.py
GresilleSiffle: That's true, it also means a test is missing here.

backend/tournesol/serializers/rating.py
GresilleSiffle: Good, I've used a `CharField` instead.

Will rework how `uid` validation works in the next PR.

amatissart (approved):

PR #636, opened 2022-03-01
WIP / [back] feat: `uid` are validated based on the entity type

GresilleSiffle:
still very WIP

### DIFF
--- a/backend/tournesol/serializers/entity.py
+++ b/backend/tournesol/serializers/entity.py
@@ -9,6 +9,7 @@ from rest_framework.fields import CharField, RegexField
 from rest_framework.serializers import ModelSerializer
 
 from core.utils.constants import YOUTUBE_VIDEO_ID_REGEX
+from tournesol.entities.video import YOUTUBE_UID_NAMESPACE
 from tournesol.models import Entity, EntityCriteriaScore
 from tournesol.utils.api_youtube import VideoNotFound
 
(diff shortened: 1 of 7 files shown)


backend/tournesol/serializers/entity.py
```diff
@@ -216,16 +215,28 @@ class Meta:
...
+        """
+        Validate the `uid` against the regex provided by the entity class.
+        """
+        split_uid = value.split(UID_DELIMITER)
 
         if len(split_uid) <= 1 or not split_uid[1]:
             raise ValidationError("Malformed `uid`.")
 
+        poll = self.context.get("poll")
+        if poll:
```
GresilleSiffle, line 227: Currently all views using a `RelatedEntitySerializer` inherit from `PollScopedViewMixin` and thus the `poll` variable is never `None`.

I'm not sure about the relevance of this condition here. Are there future cases in which we will use the `RelatedEntitySerializer` in a view not related to a poll? I've opted for a safe approach but if we are confident that all operations on a related entity will be executed in the scope a of poll, we could remove this condition.

amatissart: Yes, if we consider that a poll is always defined in this context, it seems reasonable to write `self.context["poll"]`, and fail with a `KeyError` as soon as possible.

amatissart (approved):

PR #637, opened 2022-03-02
[tests] tests: add end-to-end tests related to the comparison page

GresilleSiffle:
Here are few end-to-end tests ensuring the comparison UI is working properly.

This test case is not exhaustive yet, but is a good starting point. It should fail easily if the comparison UI is broken during the development. 

capture

**use cases not covered by these tests**
- pasting video URLs directly in the video selectors input fields (more details in the code)
- requesting new video with the `new video` circular button (note: this feature is subject to change)

### DIFF
--- a/tests/cypress/integration/frontend/comparisonPage.ts
+++ b/tests/cypress/integration/frontend/comparisonPage.ts
@@ -1,4 +1,29 @@
 describe('Comparison page', () => {
+
+  const deleteComparison = (idA, idB) => {
+    cy.sql(`
+        DELETE FROM tournesol_comparisoncriteriascore
+        WHERE comparison_id = (
+            SELECT id
+            FROM tournesol_comparison
+            WHERE entity_1_id = (
+                SELECT id FROM tournesol_entity WHERE metadata->>'video_id' = '${idA}'
+            ) AND entity_2_id = (
+                SELECT id FROM tournesol_entity WHERE metadata->>'video_id' = '${idB}'
+            )
+        );
+      `);
+
+    cy.sql(`
+        DELETE FROM tournesol_comparison
+            WHERE entity_1_id = (
+                SELECT id FROM tournesol_entity WHERE metadata->>'video_id' = '${idA}'
+            ) AND entity_2_id = (
+                SELECT id FROM tournesol_entity WHERE metadata->>'video_id' = '${idB}'
+            );
+      `);
+  };
+
   describe('authorization', () => {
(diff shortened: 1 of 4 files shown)


amatissart:
@GresilleSiffle the test about "my rated videos" is failing because the count of rated videos may have changed after a comparison is submitted in the new test case. I would recommend to relax the condition in this test:

```diff
diff --git a/tests/cypress/integration/frontend/myRatedVideosPage.spec.ts b/tests/cypress/integration/frontend/myRatedVideosPage.spec.ts
index 5cc7f99f..bfb16596 100644
--- a/tests/cypress/integration/frontend/myRatedVideosPage.spec.ts
+++ b/tests/cypress/integration/frontend/myRatedVideosPage.spec.ts
@@ -5,7 +5,7 @@ describe('"My rated videos" page', () => {
     cy.get('input[name="password"]').click().type("tournesol").type('{enter}');
 
     // All rated videos are listed.
-    cy.contains('Showing videos 1 to 20 of 263').should('be.visible');
+    cy.contains(/Showing videos 1 to 20 of \d+/).should('be.visible');
 
     // Mark all videos as public.
     cy.contains('button', 'Options', {matchCase: false}).click();
@@ -25,7 +25,7 @@ describe('"My rated videos" page', () => {
 
     // Mark all videos as public, the filter is reset and all videos are listed
     cy.contains('button', 'Mark all as public').click();
-    cy.contains('Showing videos 1 to 20 of 263').should('be.visible');
+    cy.contains(/Showing videos 1 to 20 of \d+/).should('be.visible');
   })
 
   it('visit ratings page with filter in URL', () => {
```

tests/cypress/integration/frontend/comparisonPage.ts
```diff
@@ -0,0 +1,160 @@
...
+      cy.visit('/comparison');
+
+      cy.focused().type('user1');
+      cy.get('input[name="password"]').click()
+        .type('tournesol').type('{enter}');
+
+      // add one video, and ask for a second one
+      cy.get('input[placeholder="Paste URL or Video ID"]').first()
+        .type(videoAUrl.split('?v=')[1], {delay: 0});
+      cy.get('button[data-testid=new-video').last().click()
```
amatissart, line 102: Selecting a second video at random may lead to an existing comparison, which would cause this test to fail.  

Maybe it would be cleaner to select the 2 videos deterministically, and remove the comparison from the database in `before()`  if it's already present.

amatissart (approved):

Issue #627, opened 2022-02-27
[front] feat: Language filter for recommendation language should propose all languages.


lfaucon:
Currently only languages French, English and German are available for selection, but there is no good reason and it makes Tournesol unwelcoming for contributors with a different language.

We should enable more languages by using a multi-select component https://mui.com/api/multi-select-unstyled/#main-content

sigmike:
What language list should we use? The [ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)?

GresilleSiffle:
Good question : )

In the back end we currently use a subset of languages which represent all languages that Django provides a translation for.

See the current list ( directly from the Django source code ):
https://github.com/django/django/blob/e0442a628eb480eac6a7888aed5a86f83499e299/django/conf/global_settings.py#L56

When the back end fetch the video metadata, it will store the language in the database only if it matches one language of this list. 
See:
https://github.com/tournesol-app/tournesol/blob/8f86cecca830d4109757096a7e74aaeda9a14226/backend/tournesol/utils/video_language.py#L51

Do you think it's a good idea to use theses languages @amatissart @lfaucon ?

lfaucon:
Yes, it's good to use Django's `settings.LANGUAGES` with a good way to get this list to the frontend.

sigmike:
We also need the list in the browser extension to filter the languages from the browser[[1]](https://github.com/tournesol-app/tournesol/blob/1028d98feea29f841e552ad79748afaa112087cc/browser-extension/background.js#L102). But if we accept all the languages then maybe we can just remove this filter. How would the backend behave if the recommendations request includes languages it doesn't know?

> with a good way to get this list to the frontend

What is a good way to do that? A new API call?

amatissart:
After our discussion last Friday, it seems that 2 approaches may be used to provide a list of available languages both in the frontend and the extension:

1. Generate a static JSON file, using the list internally by `langdetect`. Here is a way to build this list programmatically:

```python
from langdetect import detector_factory
detector_factory.init_factory()
detector_factory._factory.get_lang_list()
```

2. Implement a new API endpoint to expose the list of languages for which videos actually exist in the database.
Advantage: the list would be shorter and the options would be more likely to return actual recommendations
Drawback: one more asynchronous call may be required in frontend and extension before recommendations can be loaded.

Either approach seems reasonable to me.

[...]

PR #642, opened 2022-03-04
Propose all languages in recommendations

sigmike:
Related issue: #627. See the comments there.


frontend/src/features/recommendation/SearchFilter.spec.tsx
```diff
@@ -122,20 +123,35 @@ describe('Filters feature', () => {
...
-    fireEvent.click(languageCheckbox);
+    const languageFilter = queryByTestId(document, 'search-language-filter');
+    const autocomplete = queryByTestId(languageFilter, 'autocomplete');
+    expect(autocomplete).not.toBeNull();
+    const input = within(autocomplete).getByRole('textbox');
+    expect(autocomplete).not.toBeNull();
+
+    if (action === 'add') {
+      autocomplete.focus();
+      userEvent.type(input, language);
```
sigmike, line 144: I had to use `userEvent` here because triggering a change with `fireEvent` as suggested in many answers on https://stackoverflow.com/questions/60882089/how-to-test-material-ui-autocomplete-with-react-testing-library didn't work here. I don't know why. It looks like `userEvent` is recommended with mui anyway: https://stackoverflow.com/a/71242000/188760.

backend/scripts/generate_languages_json.py
```diff
@@ -0,0 +1,18 @@
+from django.utils import translation
```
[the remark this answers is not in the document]
sigmike: I removed this script (see https://github.com/tournesol-app/tournesol/issues/627#issuecomment-1063242392) but the suggestions are noted.

frontend/src/features/recommendation/LanguageFilter.tsx
```diff
@@ -1,37 +1,50 @@
...
+    <TitledSection title={t('filter.language')}>
+      <Autocomplete
+        multiple
+        options={availableRecommendationsLanguages}
+        getOptionLabel={getOptionLabel}
+        renderInput={(params) => <TextField {...params} variant="standard" />}
+        value={arrayValue}
+        onChange={handleChange}
+        data-testid="autocomplete"
+      />
```
lfaucon, line 46: I find the multi-select component slightly nicer with the below options. Especially the `"outlined"` variant.

```suggestion
      <Autocomplete
        multiple
        options={availableRecommendationsLanguages}
        getOptionLabel={getOptionLabel}
        renderInput={(params) => <TextField {...params} variant="outlined" />}
        value={arrayValue}
        onChange={handleChange}
        filterSelectedOptions
        limitTags={3}
        data-testid="autocomplete"
      />
```

lfaucon (approved):
Thanks for working on this feature! Works and looks good for me, but maybe @GresilleSiffle  or @amatissart would like to give it a quick look to before merging and shipping this change.

frontend/src/features/recommendation/LanguageFilter.tsx
amatissart: In addition to these suggestions, I would also add `disableClearable` to remove the "clear" button. To gain some space, and it's not so useful as the selection option can be removed individually, and using the backspace key.

The `TextField` should also use `color="secondary"` to be consistent with other filters.

frontend/src/features/recommendation/LanguageFilter.tsx
```diff
@@ -1,37 +1,50 @@
...
-    <ChoicesFilterSection
-      title={t('filter.language')}
-      multipleChoice
-      choices={languageChoices}
-      {...props}
-    ></ChoicesFilterSection>
+    <TitledSection title={t('filter.language')}>
+      <Autocomplete
+        multiple
+        options={availableRecommendationsLanguages}
```
amatissart, line 40: The order of these options may look random and not be very practical. 
Maybe we could sort the labels by alphabetical order?
image

frontend/src/features/recommendation/LanguageFilter.tsx
sigmike: I applied all these suggestions.

amatissart (approved):
Thanks @sigmike. It looks good to me, I just pushed an additional commit to align the arrow at the bottom of the field.

Issue #543, opened 2022-02-04
API endpoint for recommendations from a specific user `/users/:username:/recommendations`. Best would be to reuse the same code as for aggregated recommendations.


glerzing:

glerzing:
Ok, thanks for the info

Issue #363, opened 2021-12-14
feat: personnal Recommendations from Tournesol's users


lfaucon:
A page to receive recommendations based on the public ContributorRatingsCriteriaScores of a user rather than the VideoCriteriaScores

- [x] #540
- [x] #543
- [x] #1003
- [x] #1056
- [x] safe/unsafe reco. differ between collective and public personal page
  - see issue: https://github.com/tournesol-app/tournesol/issues/1036
  - see PR: https://github.com/tournesol-app/tournesol/pull/1037
- [ ] (optional) By default this page should only show public data, but if a user is looking at the recommendation page of him/herself, then an option could appear to also show videos rated privately.

lfaucon:
@glerzing I proposed this ticket to you, but as I see you probably have not started, @farnots will start to work on the first step #540

glerzing:
ok, he can start to work on the first step.

glerzing:
I don't think I would be efficient in the frontend part, I would prefer to specialize in the backend, if there is enough work in the backend.

lfaucon:
During discussion on Friday 25 March we agreed that this page should reuse the layout of the "Compared items" page

sigmike:
> A page in user's profile "Your public recommendations" with a link to be shared.

The user's profile looks like that:
![image](https://user-images.githubusercontent.com/28259/160675674-a241e43b-366d-417a-a933-069e4ef2a94c.png)

You want a link after the username input that leads to the user's personal recommendations page?

lfaucon:
In our discussion last Friday we suggested that the personnal  page would instead be incorporated into the already existing "Compared Items" page

A few simple first steps may be:
- [ ] Enable sorting by personnal scores the list of compared videos
- [ ] Display the personnal scores on this page next to the VideoCards

sigmike:
I did that in https://github.com/tournesol-app/tournesol/pull/732.

sigmike:
The changes were made in #732 and disabled in f600f19c7 saying that it should be discussed first.

PR #644, opened 2022-03-07
user-specific recommendations backend (#543)

glerzing:
**related to** #543 

---

It took me more than one month, I lost a lot of time on a few tricky queries.

I tried to avoid duplicating code using inheritance. The function get_scores though still contains some almost-duplicated code.

I kept using the "total_score". But should the ContributorRating model have a tournesol_score, that would be used for these user recommendations? When would it be recomputed?

Don't hesitate to suggest improvements.


backend/tournesol/views/polls.py
```diff
@@ -22,6 +22,24 @@
 
 logger = logging.getLogger(__name__)
 
+WEIGHTS_PARAMETER = OpenApiParameter(
+    "weights",
+    OpenApiTypes.OBJECT,
+    style="deepObject",
+    description="Weights for criteria in this poll."
+                " The default weight is 10 for each criteria.",
```
lfaucon, line 30: Isn't the default weight 50 ?

lfaucon (commented):
Don'tworry about taking time on this task. And thanks for the work. It will be a very valuable feature to have for User Experience :) I'll give a thourough review tomorrow.

backend/tournesol/views/polls.py
glerzing: It looks like it changed recently. Maybe to allow for more arbitrary weights than just percentages.

backend/tournesol/views/user_recommendations.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            self.queryset = Entity.objects.filter(
+                contributorvideoratings__user=self.user,
+                contributorvideoratings__is_public=True
+            )
+
+        return super().get_queryset()
+
+
+class ExternalUserRecommendations(ContributorRecommendations):
```
lfaucon, line 106: Discussing the naming with @GresilleSiffle : 
We'd recommend `PublicContributorRecommendations` for the `/{username}/` and `PrivateContributorRecommendations` for the `/me/` endpoint

backend/tournesol/views/user_recommendations.py
```diff
@@ -0,0 +1,120 @@
...
+        self.kwargs["name"] = self.kwargs["poll_name"]  # parent class calls it 'name'
+
+        if self.request.query_params.get("include_private") == "true":
+            self.queryset = Entity.objects.filter(
+                contributorvideoratings__user=self.user
+            )
+        else:
+            self.queryset = Entity.objects.filter(
+                contributorvideoratings__user=self.user,
+                contributorvideoratings__is_public=True
```
lfaucon, line 100: For the `/me/` endpoint, we should show all contributor ratings private or not private.

lfaucon (dismissed):

backend/tournesol/views/user_recommendations.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    View class to access another user's recommendations
+    """
+
+    def get_queryset(self):
+        self.user = get_object_or_404(User, username=self.kwargs["username"])
+        self.kwargs["name"] = self.kwargs["poll_name"]  # parent class calls it 'name'
+
+        self.queryset = Entity.objects.filter(
+            contributorvideoratings__user__username=self.user,
```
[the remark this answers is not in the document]
lfaucon: Also, we should avoid the word "video" and opt for "entity" instead

backend/tournesol/views/user_recommendations.py
glerzing: So, no "include_private" URL parameter? In the issue #363, the 4th point is:
"(optional) By default this page should only show public data, but if a user is looking at the recommendation page of him/herself, then **an option** could appear **to also show videos rated privately**."

Can you confirm if the parameter "include_private" should be remove?

backend/tournesol/views/user_recommendations.py
glerzing: ok

backend/tournesol/views/user_recommendations.py
lfaucon: According to the ticket, we need one way to see private information and one to see public information. This could be done with an optional parameter "include_private" but I find nice that we would use `/me/` endpoints for everything private and `/{username}/` for everything public. It for example bring that we can easily implement caching on `/{username}/` endpoints and that it is easier to differentiate when we deal with private "me" information and when not

backend/tournesol/serializers/user_recommendations.py
```diff
@@ -0,0 +1,38 @@
...
+    class Meta:
+        model = Entity
+        fields = [
+            "uid",
+            "metadata",
+            "n_comparisons",
+            "n_contributors",
+            "total_score",
+            "contributorvideoratings",
+        ]
```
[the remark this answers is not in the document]
glerzing: Do you mean that the tournesol_score of the Entity should be reused, or that a user-specific "tournesol_score" should be added to ContributorRating? And if it is added to ContributorRating, when should it be recomputed? (when there is a
 request that needs it? When there is a new rating? periodically?).
I don't know how to interface my work with what was done in the PR #561

backend/tournesol/views/user_recommendations.py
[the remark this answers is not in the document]
glerzing: The filter looks dubious indeed, well seen. I'll investigate tomorrow why the unit tests work anyway, there are also some weird results with the PersonalRecommendations.

backend/tournesol/views/user_recommendations.py
glerzing: ok, you're right

backend/tournesol/views/user_recommendations.py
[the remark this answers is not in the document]
glerzing: Actually, this was already working. Not sure why, maybe because username is a unique field.

backend/tournesol/tests/factories/poll.py
```diff
@@ -8,7 +8,7 @@ class PollFactory(DjangoModelFactory):
     class Meta:
         model = Poll
 
-    name = factory.Sequence(lambda n: "Poll #%s" % n)
+    name = factory.Sequence(lambda n: "Poll_nb%s" % n)
```
glerzing, line 11: the name was not URL-friendly

backend/tournesol/serializers/contributor_recommendations.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class ContributorRecommendationsSerializer(RecommendationSerializer):
+    """ Format the output of the contributor_recommendations views """
+    contributor_ratings = ContributorRecommendationsRating(many=True,
+                                                           source="contributorvideoratings")
+
+    class Meta(RecommendationSerializer.Meta):
+        fields = list((set(RecommendationSerializer.Meta.fields)
+                      - {"criteria_scores"}) | {"contributor_ratings"})
```
glerzing, line 29: Do you think we should keep "criteria_scores" (potentially with another name)? For example, if you think it would be nice to have a direct comparison between the contributor's criteria score and the global criteria scores, without having to call the /entities/{uid} API.

backend/tournesol/views/user_recommendations.py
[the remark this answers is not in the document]
glerzing: (but I changed it to contributorvideoratings__user anyway, it is cleaner)

backend/tournesol/serializers/contributor_recommendations.py
```diff
@@ -0,0 +1,29 @@
...
+class ContributorRecommendationsRating(ModelSerializer):
+    criteria_scores = ContributorRecommendationsScore(many=True)
+
+    class Meta:
+        model = ContributorRating
+        fields = ["is_public", "criteria_scores"]
+
+
+class ContributorRecommendationsSerializer(RecommendationSerializer):
+    """ Format the output of the contributor_recommendations views """
```
GresilleSiffle, line 23: As the name of the class is concise and explicit, you can choose to not add a docstring here.

GresilleSiffle (dismissed):
Hello @glerzing 

Thanks again for all your work.

I was adding comments and suggestions and we realized with @amatissart that the introduction of a direct inheritance of `PollsRecommendationView` by `ContributorRecommendations` could lead to undesirable and opaque side effects. The tight coupling between those views will also make their maintenance more complex.

Nevertheless inheriting the recommendation features, like the filters, is still a good idea. To simplify the review, I'm gonna refactor the current code by adding a new class, that will be directly inherited by `PollsRecommendationView` and `ContributorRecommendations`. This way, both classes will share the same features, while being as less coupled as possible.

If there is no objection, I'll continue this pull request.

**to-do**
- [x] uncoupling of `PollsRecommendationView` and `ContributorRecommendations`
- [x] re-document the API
- [x] refactor the serializers when possible
- [x] check the tests

backend/tournesol/views/contributor_recommendations.py
```diff
@@ -0,0 +1,119 @@
...
+                ),
+            )
+        )
+
+        return queryset.annotate(
+            total_score=Sum(
+                F("contributorvideoratings__criteria_scores__score") * criteria_weight,
+            ),
+            filter=Q(contributorvideoratings__poll=poll)
+            & Q(contributorvideoratings__user=user),
```
GresilleSiffle, line 67: This filter doesn't seem necessary to me, as the contributor ratings are already filtered by the prefetch on `contributorvideoratings`.

By inspecting the generated SQL, adding this filter only adds `tournesol_contributorrating.poll_id` to the SQL GROUP BY clause.

The tests still pass without this filter.

Generated SQL:
- with the filter: https://pastebin.com/raw/qA24VxX1
- without the filter: https://pastebin.com/raw/CLxHaXJF

backend/tournesol/tests/test_api_contributor_recommendations.py
```diff
@@ -0,0 +1,335 @@
...
+            format="json",
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data["count"], 2)
+
+        for entity in response.data["results"]:
+            self.assertEqual(entity["is_public"], True)
+
+    def test_recommendations_are_filtered_by_pole(self):
```
glerzing, line 136: pole => poll

backend/tournesol/tests/test_api_contributor_recommendations.py
```diff
@@ -0,0 +1,335 @@
...
+        non-existing poll.
+        """
+        response = self.client.get(
+            f"/users/{self.user1.username}/recommendations/non-existing", format="json"
+        )
+        self.assertEqual(response.status_code, 404)
+
+        self.client.force_authenticate(self.user1)
+        response = self.client.get(
+            f"/users/{self.user1.username}/recommendations/non-existing", format="json"
```
glerzing, line 198: did you mean "users/me/recommendations/non-existing" ?

backend/tournesol/tests/test_api_contributor_recommendations.py
```diff
@@ -0,0 +1,335 @@
...
+            f"/users/{user.username}/recommendations/{self.poll.name}?unsafe=true",
+            format="json",
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data["count"], 1)
+
+        # authenticated checks
+        self.client.force_authenticate(self.user1)
+        response = self.client.get(
+            f"/users/{user.username}/recommendations/{self.poll.name}", format="json"
```
glerzing, line 233: here too, is it rather "users/me/recommendations/{self.poll.name}" ?

backend/tournesol/tests/test_api_contributor_recommendations.py
GresilleSiffle: I'm using the contributor public recommendations endpoint here, `/users/{self.user1.username}/` is the URL I need.

I'm making two requests in a row because I'm checking the behaviour with an anonymous user then a logged user.

backend/tournesol/tests/test_api_contributor_recommendations.py
GresilleSiffle: Thanks, fixed.

backend/tournesol/tests/test_api_contributor_recommendations.py
GresilleSiffle: Here too, I'm performing checks with both anonymous and authenticated users, that's making I'm making several requests to the same endpoint.

backend/tournesol/views/contributor_recommendations.py
```diff
@@ -0,0 +1,111 @@
...
+                    ) from error_value
+            else:
+                used_weight = 10
+            criteria_cases.append(
+                When(
+                    contributorvideoratings__criteria_scores__criteria=crit,
+                    then=used_weight,
+                )
+            )
+        criteria_weight = Case(*criteria_cases, default=0)
```
GresilleSiffle, line 44: This part is duplicated is both poll views and contributor recommendations views.

Could be factorized in a method.

backend/tournesol/serializers/contributor_recommendations.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+
+    @extend_schema_field(ContributorCriteriaScore(many=True))
+    def get_criteria_scores(self, obj):
+        return ContributorCriteriaScore(
+            obj.contributorvideoratings.first().criteria_scores.all(), many=True
+        ).data
+
+    def get_is_public(self, obj) -> bool:
+        return obj.contributorvideoratings.first().is_public
```
amatissart, line 32: Calling `.first()` does not take advantage from the "prefetched" querysets, which are cached as `.all()`:

```suggestion
    def get_criteria_scores(self, obj):
        return ContributorCriteriaScore(
            obj.contributorvideoratings.all()[0].criteria_scores.all(), many=True
        ).data

    def get_is_public(self, obj) -> bool:
        return obj.contributorvideoratings.all()[0].is_public
```

backend/tournesol/views/contributor_recommendations.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+    def filter_unsafe(self, queryset, filters):
+        show_unsafe = filters["unsafe"]
+        if not show_unsafe:
+            # Ignore RECOMMENDATIONS_MIN_CONTRIBUTORS, only filter on the
+            # total score
+            queryset = queryset.filter(total_score__gt=0)
+
+        return queryset.distinct()
```
amatissart, line 50: Calling `.distinct()` is not related to the "unsafe" filter, right?  
If that's still necessary, it should be moved to the definition of the base queryset in `get_queryset()`.

GresilleSiffle: That's true, I'm going to check if it's required.

GresilleSiffle: This `DISTINCT` doesn't seem to be required in my opinion. 

The generated SQL uses a `GROUP BY "tournesol_entity"."id"` clause which must not generate duplicated rows.

I've manually ran the query locally and on staging by adding and removing a `DISTINCT` clause, each time the number of returned rows is the same.

Here the generated SQL query without `DISTINCT`:

```sql
SELECT 
  "tournesol_entity"."id", 
  "tournesol_entity"."uid", 
  "tournesol_entity"."type", 
  "tournesol_entity"."tournesol_score", 
  "tournesol_entity"."rating_n_ratings", 
  "tournesol_entity"."rating_n_contributors", 
  SUM(
    (
      "tournesol_contributorratingcriteriascore"."score" * CASE WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'largely_recommended' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'reliability' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'pedagogy' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'importance' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'layman_friendly' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'entertaining_relaxing' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'engaging' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'diversity_inclusion' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'better_habits' THEN 10 WHEN "tournesol_contributorratingcriteriascore"."criteria" = 'backfire_risk' THEN 10 ELSE 0 END
    )
  ) AS "total_score" 
FROM 
  "tournesol_entity" 
  LEFT OUTER JOIN "tournesol_contributorrating" ON (
    "tournesol_entity"."id" = "tournesol_contributorrating"."entity_id"
  ) 
  LEFT OUTER JOIN "tournesol_contributorratingcriteriascore" ON (
    "tournesol_contributorrating"."id" = "tournesol_contributorratingcriteriascore"."contributor_rating_id"
  ) 
WHERE 
  (
    "tournesol_contributorrating"."is_public" 
    AND "tournesol_contributorrating"."poll_id" = 1 
    AND "tournesol_contributorrating"."user_id" = 1
  ) 
GROUP BY 
  "tournesol_entity"."id" 
HAVING 
  SUM(
    (
      "tournesol_contributorratingcriteriascore"."score" * CASE WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'largely_recommended'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'reliability'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'pedagogy'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'importance'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'layman_friendly'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'entertaining_relaxing'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'engaging'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'diversity_inclusion'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'better_habits'
      ) THEN 10 WHEN (
        "tournesol_contributorratingcriteriascore"."criteria" = 'backfire_risk'
      ) THEN 10 ELSE 0 END
    )
  ) > 0.0 
ORDER BY 
  "total_score" DESC, 
  "tournesol_entity"."id" DESC;
```

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/644/commits/bb9dbebaede6f6aedbc990bba9c3d71439cbde2d.

amatissart (approved):

PR #645, opened 2022-03-07
reduce pytest's uninteresting logs

glerzing:
These logs are retrieved using "python manage.py test", but it is the same logs for pytest, except that pytest only displays the logs for the tests that failed and not all the logs.

Initially, when running "python manage.py test", I had 1690 lines of log, mostly useless:
[log1.txt](https://github.com/tournesol-app/tournesol/files/8008430/log1.txt)

The use of "logging.getLogger('factory').setLevel(logging.WARN)" reduces the amount of logs to 157 lines:
[log2.txt](https://github.com/tournesol-app/tournesol/files/8008431/log2.txt)

The use of "logging.getLogger('faker').setLevel(logging.INFO)" reduces the amount of logs to 118 lines:
[log3.txt](https://github.com/tournesol-app/tournesol/files/8008439/log3.txt)

To further remove the last useless logs, I had to set the Django log level to "INFO". This is more controversial, maybe there is a better way to do it. I could also have modified the settings-tournesol.yaml instead. The only logs that remain (45 lines) look pretty meaningful:
[log4.txt](https://github.com/tournesol-app/tournesol/files/8008449/log4.txt)

### DIFF
--- a/backend/settings/settings.py
+++ b/backend/settings/settings.py
@@ -358,7 +358,15 @@ LOGGING = {
     },
     "root": {
         "handlers": ["console"],
-        "level": os.environ.get("DJANGO_LOG_LEVEL", "INFO"),
+        "level": os.environ.get("DJANGO_LOG_LEVEL", "DEBUG"),
+    },
+    "loggers": {
+        "factory": {
+            "level": "WARN"
+        },
+        "faker": {
+            "level": "INFO"
+        }
     },
 }
 
--- a/backend/settings/settings.py
+++ b/backend/settings/settings.py
@@ -358,7 +358,7 @@ LOGGING = {
     },
     "root": {
         "handlers": ["console"],
-        "level": os.environ.get("DJANGO_LOG_LEVEL", "DEBUG"),
+        "level": os.environ.get("DJANGO_LOG_LEVEL", "INFO"),
     },
 }
 
--- a/backend/tournesol/tests/factories/video.py
+++ b/backend/tournesol/tests/factories/video.py
@@ -1,5 +1,4 @@
 import datetime
-import logging
 import random
 import string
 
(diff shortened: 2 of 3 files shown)


backend/tournesol/tests/factories/video.py
```diff
@@ -7,6 +8,7 @@
 from core.tests.factories.user import UserFactory
 from tournesol.models import Entity, EntityCriteriaScore, VideoRateLater
 
+logging.getLogger('factory').setLevel(logging.WARN)
```
amatissart, line 11: It seems reasonable to disable log messages related to `factory` and `faker`. However it would be cleaner to set these levels directly in the `LOGGING` configuration in "settings.py":

```py
    "loggers": {
        "factory": {
            "level": "WARN"
        },
        "faker": {
            "level": "INFO"
        }
    },
```

Otherwise, we depend on implicit side-effects when this module is loaded, and this change may not be applied when running tests on other files only.

backend/settings/settings.py
```diff
@@ -358,7 +358,7 @@
     },
     "root": {
         "handlers": ["console"],
-        "level": os.environ.get("DJANGO_LOG_LEVEL", "DEBUG"),
+        "level": os.environ.get("DJANGO_LOG_LEVEL", "INFO"),
```
amatissart, line 361: I think I found prefer to keep the "DEBUG` level by default here, especially for local environments and CI. It does not seem unreasonable to provide as many details as possible when debugging a failing test case. 

And even in situations where this would be too verbose for developers, `pytest` provides some options to customize this behavior, such as `pytest --show-capture=no` (and variants such as `--show-capture=stdout`) to reduce the amount of visible logs, or `pytest -x` to stop on the first failure.

glerzing: I agree, it's more cautious to keep the DEBUG logs

amatissart (approved):

PR #651, opened 2022-03-08
First step to generalize `VideoSelector` into EntitySelector

amatissart:
* Use `uid` instead of `videoId` in comparison page and selector
* Split `VideoSelector` and `VideoCard` into small components to enable reusability
* Adjustments in `VideoCard` style to make it slightly more compact


GresilleSiffle:
I'm currently testing the comparison interface, it looks like the compatibility with old URL parameters `videoA` and `videoB` don't work anymore.

Do you see the same behaviour @amatissart ?

GresilleSiffle (approved):

amatissart:
> I'm currently testing the comparison interface, it looks like the compatibility with old URL parameters `videoA` and `videoB` don't work anymore.
> 
> Do you see the same behaviour @amatissart ?

Yes, it's a regression introduced in #632. My bad. I will propose a fix separately.

PR #652, opened 2022-03-08
Display statistics on home page and about page

lfaucon:


lfaucon:
@iamnkc I rebased your work here and improved the style. Thanks a lot for your work.

Screenshot 2022-03-08 at 18 42 51

TODO
- [ ] add French translations

GresilleSiffle:
I've added translation for the metrics name, and I saw I forgot to also translate the tooltips' text.

One thing that could be improved, is the `<StatsUi>`. The fact that this component is a grid item makes it very specific, and less re-usable. For instance, if I try to add `<Divider />` between three successive `<StatsUi>`, the last one will be displayed on a second row because each of these component are flex item with hard-coded `sm={4}`. I don't know if this is a good idea but I tend to think leaving the parent component responsible of the layout make it easier to re-use `<StatsUi>` everywhere in the front end.

What are your opinion about that @amatissart @lfaucon ?

```jsx
      <Grid container sx={{ maxWidth: 900 }}>
        <StatsUi
          text={t('stats.activeUsers')}
          count={data.userCount}
          lastMonthCount={data.lastMonthUserCount}
        />
        {/* ... */}
      </Grid>
```

would become 

```jsx
      <Grid container sx={{ maxWidth: 900 }}>
        <Grid item xs={12} sm={4}>
          <StatsUi
            text={t('stats.activeUsers')}
            count={data.userCount}
            lastMonthCount={data.lastMonthUserCount}
          />
          {/* ... */}
        </Grid>
      </Grid>
```

lfaucon:
Indeed, I think that things are more easily reusable when the Grid container and Grid items are in the same component!

GresilleSiffle:
I've added the missing translations :ok_hand: the branch is ready to be merged for me.

amatissart (approved):

PR #661, opened 2022-03-10
Implement `EntityCard` and `EntityInput`

amatissart:
* `EntityCard` can be used to represent a `RelatedEntityObject` returned by the API. 
`VideoCard` is still present, to be used on pages which are still specific to videos (rate-later list, recommendations, etc.)

* A generic `EntityInput` is used in the new `EntitySelector` (as a replacement for `VideoSelector`) and can handle both entity types: videos (a text field where id or URL can be pasted), and candidate (autocomplete with a list of options fetched from the API).


frontend/src/components/entity/EntityImagery.tsx
```diff
@@ -0,0 +1,82 @@
...
+        maxHeight="280px"
+        justifyContent="center"
+        sx={{
+          '& img': {
+            flex: 1,
+            objectFit: 'contain',
+          },
+        }}
+      >
+        <img src={entity.metadata.image_url} />
```
GresilleSiffle, line 75: Suggestion: we could add the attribute `alt` here for a better accessibility.

```suggestion
        <img src={entity.metadata.image_url} alt={entity.metadata.name} />
```

GresilleSiffle (approved):
Great! :+1: :+1: :+1:

Issue #604, opened 2022-02-18
[back] feat: Endpoint for non-connected videos


lfaucon:
For a user the comparisons connect entities into a graph. For  the ratings to be more accurate it can be useful to help users have a fully connected graph of comparisons (more info on wikipedia https://en.wikipedia.org/wiki/Connectivity_(graph_theory).

To help users we want to provide an endpoint `users/me/unconnected_entities/{entity_ui}` that will return the list of entities not connected to `entity_uid`. Note that if the user has fully connected graph of comparison then this endpoint returns an empty list

Acceptance criteria:
- The endpoint has been implemented with an efficient (linear time) algorithm to extract the list of videos
- The endpoint returns similar paginated results as other endpoints for entities
- The endpoint is tested for edge cases

PR #663, opened 2022-03-11
API for non connected entities

farnots:
Linked to #604 

## Implementation

This is an endpoint for showing all non-related entity for an user based on a source video. The relation is based on comparison set by the user. 

## Tests 

This is graphics representation of the tests made in order of their own declaration.

Drawing-1

Drawing-2

Drawing-3

Drawing-4


lfaucon (changes requested):

Vikka (changes requested):
Hi,
Suggestion to improve pythonicness of the code.

backend/tournesol/tests/test_api_unconnected_entities.py
```diff
@@ -0,0 +1,248 @@
...
+        response = self.client.get(
+            "{}/{}/".format(self.user_base_url, self.video_source.id),
+            format="json",
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["count"], 2)
+        self.assertEqual(
+                list(map(lambda x: x["uid"], response.data["results"])),
+                list(map(lambda x: x.uid, self.unrelated_video))
```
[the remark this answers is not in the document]
GresilleSiffle: Yes, I also recommend using the full word `entity` if the resulting statement fits a single line.

- `ent` is acceptable I think if it makes the statement fit on a single line
- `e` is not pythonic
- `x` is confusing

In case of doubt, we can always refer to the Zen of Python (and PEP8): https://peps.python.org/pep-0020/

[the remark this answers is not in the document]
Vikka: with entity, the line is 82, which is black compliant and acceptable if it's for readability

Vikka:
To fix flake8 issue, merge main into your branch

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def get_queryset(self):
+        # Get related entities from source
+        source_node = Entity.objects.none()
+
+        if self.kwargs.get("uid"):
+            entity_uid = self.kwargs.get("uid")
+            source_node = Entity.objects.get(uid=entity_uid)
+        else:
+            return Entity.objects.none()
```
Vikka, line 68:
```suggestion
        if not self.kwargs.get("uid"):
            return Entity.objects.none()

        entity_uid = self.kwargs.get("uid")
        source_node = Entity.objects.get(uid=entity_uid)
```

Vikka (changes requested):
If you like suggested changes, consider accepting it.

http://haacked.com/archive/2019/06/03/suggested-changes/

> One nice aspect of suggesting changes in this way is how GitHub handles credit for the change. If the maintainer accepts the change, GitHub creates a commit with the change with the commenter as a [co-committer](https://help.github.com/en/articles/creating-a-commit-with-multiple-authors). The commenter sees their avatar in the history of the file they helped improve.

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,86 @@
...
+        # Get all comparison for the user
+        user_comparisons = Comparison.objects.filter(
+            poll=self.poll_from_url, user=self.request.user
+        )
+
+        for comparison in user_comparisons.iterator():
+            if comparison.entity_1 not in user_all_entities:
+                user_all_entities.add(comparison.entity_1)
+            if comparison.entity_2 not in user_all_entities:
+                user_all_entities.add(comparison.entity_2)
```
Vikka, line 82: Why check if entity_1 and 2 is in the set? Just add it, if it's already in it, it won't add a duplicate. 
See https://docs.python.org/3/tutorial/datastructures.html#sets

```suggestion
            user_all_entities.add(comparison.entity_1)
            user_all_entities.add(comparison.entity_2)
```

Vikka (changes requested):

farnots:
I resolved the flake8 but fucked up my branch with the rebase.... Trying to resolve the issue. I hate rebase :/

farnots:
> I resolved the flake8 but fucked up my branch with the rebase.... Trying to resolve the issue. I hate rebase :/

Fixed, sorry for the force push

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,84 @@
...
+
+        local_entities.update(new_set_to_push)
+
+        return local_entities
+
+    def get_queryset(self):
+        # Get related entities from source
+        source_node = Entity.objects.none()
+
+        if not self.kwargs.get("uid"):
```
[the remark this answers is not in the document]
farnots: Done

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,88 @@
...
+                    ))
+                )
+            )
+
+        self.__already_visited_node__.add(entity)
+
+        new_set_to_push = set()
+        for element in local_entities:
+            if element not in self.__already_visited_node__:
+                new_set_to_push.update(self.get_related_entities(element))
```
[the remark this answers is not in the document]
farnots: Done, thanks for the tips

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def get_queryset(self):
+        # Get related entities from source
+        source_node = Entity.objects.none()
+
+        try:
+            entity_uid = self.kwargs.get("uid")
+            source_node = Entity.objects.get(uid=entity_uid)
+        except ObjectDoesNotExist:
+            raise Http404
```
lfaucon, line 55: I think this could be simplified into: 

```suggestion
        source_node = get_object_or_404(Entity, uid=self.kwargs.get("uid"))
```

See: https://docs.djangoproject.com/en/4.0/topics/http/shortcuts/#get-object-or-404

⚠️ needs the import too if you accept the suggestion

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,81 @@
...
+    permission_classes = [IsAuthenticated]
+
+    __already_visited_node__ = set()
+    __all_entites__ = set()
+    __all_connections__ = defaultdict(set)
+
+    def get_related_entities(self, entity_id):
+
+        self.__already_visited_node__.add(entity_id)
+        self.__all_entites__.remove(entity_id)
```
lfaucon, line 38: This line raises a 500 error in the case where a user calls the endpoint with the uid of an Entity that exists but they have not yet compared.

I think it would be easily solved and the appropriate behaviour would be to return `__all_entites__` in the case of a not compared entity uid ?

I would recommend adding this in the test suite too :)

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,81 @@
...
+            poll=self.poll_from_url, user=self.request.user
+        )
+
+        for comparison in user_comparisons.iterator():
+            user_all_entities.add(comparison.entity_1)
+            user_all_entities.add(comparison.entity_2)
+
+        return Entity.objects \
+            .filter(id__in=(entity.id for entity in user_all_entities)) \
+            .exclude(id__in=(entity_id for entity_id in user_related_entities))
```
lfaucon, line 81: Can simplify the request. This will be useful when the number of non-connected is small.

```suggestion
        return Entity.objects \
            .filter(id__in=user_all_entities - user_related_entities)
```

⚠️ Needs to also accept suggestion that mosuggestion

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,81 @@
...
+        user_all_entities = set()
+
+        # Get all comparison for the user
+        user_comparisons = Comparison.objects.filter(
+            poll=self.poll_from_url, user=self.request.user
+        )
+
+        for comparison in user_comparisons.iterator():
+            user_all_entities.add(comparison.entity_1)
+            user_all_entities.add(comparison.entity_2)
```
lfaucon, line 77: When you access `comparison.entity_1` Django will make an additional db request because you have not `prefetch` the related objects.

In this case you actually only need the ids so I suggest the following:

```suggestion
            user_all_entities.add(comparison.entity_1_id)
            user_all_entities.add(comparison.entity_2_id)
```

lfaucon (changes requested):
Requesting changes again, but this is getting much better! Thanks for making changes and sorry to have been very slow art reviewing again 👍

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,81 @@
...
+    generics.ListAPIView
+):
+    """
+    API view for showing unconnected entities
+    """
+    serializer_class = EntitySerializer
+    permission_classes = [IsAuthenticated]
+
+    __already_visited_node__ = set()
+    __all_entites__ = set()
```
lfaucon, line 32: Typo: "entities" (sorry the typo comes from one of my comments ^^)

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,81 @@
...
+            entity_uid = self.kwargs.get("uid")
+            source_node = Entity.objects.get(uid=entity_uid)
+        except ObjectDoesNotExist:
+            raise Http404
+
+        comparisons = list(Comparison.objects.filter(
+            poll=self.poll_from_url, user=self.request.user
+        ))
+
+        self.__all_entites__ = set(c.entity_1_id for c in comparisons).union(
```
lfaucon, line 61: It looks like this set is never used, except for when elements are removed from it (line 38). 

One possible refactor would be to completely remove it? 

However I would probably prefer to make use of it. For example `get_related_entities` could not return anything but instead simply empty the `__all_entites__` as it is currently doing (could be renamed in this case `remove_connected_entities`). Then the final `return` statement could be simplified to use `self.__all_entities__`.

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,81 @@
...
+):
+    """
+    API view for showing unconnected entities
+    """
+    serializer_class = EntitySerializer
+    permission_classes = [IsAuthenticated]
+
+    __already_visited_node__ = set()
+    __all_entites__ = set()
+    __all_connections__ = defaultdict(set)
```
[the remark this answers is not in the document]
GresilleSiffle: Yes, one single underscore should be enough.

With two starting and trailing underscores, these variables really look like special python variable like `__name__` which is unexpected.

backend/tournesol/views/unconnected_entities.py
farnots: Thanks

backend/tournesol/views/unconnected_entities.py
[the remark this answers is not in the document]
farnots: Ok thanks for your review. I fix it

backend/tournesol/views/unconnected_entities.py
farnots: Nice, I should have think about this the first time

backend/tournesol/views/unconnected_entities.py
farnots: Yes good catch. I switch the purpose of the `__all_entities__` as `_entities_without_comparison_` to fix this issue (I added a test case). Every entities without comparison with the same type are now also return from the API

backend/tournesol/views/unconnected_entities.py
farnots: As I updated the request using `Q`, I think the separation of include and exclude make the code more readable. What do you thing about it ?

backend/tournesol/views/unconnected_entities.py
farnots: Thanks

backend/tournesol/views/unconnected_entities.py
farnots: Yeah thanks a lot, you can see this comment https://github.com/tournesol-app/tournesol/pull/663#discussion_r856302769 with the switch of purpose of this element

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,83 @@
...
+        user_comparisons = Comparison.objects.filter(
+            poll=self.poll_from_url, user=self.request.user
+        )
+
+        for comparison in user_comparisons.iterator():
+            user_all_entities.add(comparison.entity_1_id)
+            user_all_entities.add(comparison.entity_2_id)
+
+        return Entity.objects \
+            .filter(Q(id__in=user_all_entities) | Q(id__in=self._entities_without_comparison_)) \
```
lfaucon, line 82: I see that `_entities_without_comparison_` can potentially contain a lot of Entities completely unrelated with the user as it is defined here:

```Python
self._entities_without_comparison_ = set(
            Entity.objects.filter(
                type=self.poll_from_url.entity_type,
                comparisons_entity_1=None,
                comparisons_entity_2=None).values_list('id', flat=True)
            .exclude(id=source_node.id))
```

I don't think that we want to include `_entities_without_comparison_` in non-connected.

backend/tournesol/views/unconnected_entities.py
```diff
@@ -0,0 +1,74 @@
...
+    generics.ListAPIView
+):
+    """
+    API view for showing unconnected entities
+    """
+    serializer_class = EntitySerializer
+    permission_classes = [IsAuthenticated]
+
+    _already_visited_node_ = set()
+    _entities_without_comparison_ = set()
```
lfaucon, line 31: TODO clean_up

lfaucon (approved):
LGTM. Thanks for addressing all comments! I'm looking forward to see how many unconnected entities I have XD

PR #664, opened 2022-03-11
Suggestion to adjust style of the comparison page

amatissart:
* The "magic wand" button (used to select a new video automatically) is replaced with a more explicit button with text "Auto"

* The style of the criteria sliders is updated to be closer to the initial mockup.  
The name and logo of each criteria is aligned on the left.  
The checkbox used on each optional criteria is aligned on the right.  
It also makes the layout more consistent when criterias names are of different lengths.

|Before|After|
|-|-|
|image|image

Feedbacks are welcome :pray:

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -19,7 +19,7 @@
     "comparisonInPublicDataset": "This comparison is included in the public dataset.",
     "comparisonInPublicDatasetAfterSubmission": "After submission, this comparison will be included in the public dataset.",
     "editComparison": "Edit comparison",
-    "criteriaSkipped": "Skipped",
+    "criteriaSkipped": "skipped",
     "submitAComparison": "Submit a comparison"
   },
   "comparisons": {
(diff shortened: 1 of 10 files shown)


frontend/public/locales/en/translation.json
```diff
@@ -153,27 +145,36 @@
...
   "username": "Username",
   "profile": "Profile",
+  "metrics": {
+    "evolutionDuringTheLast30Days": "Evolution during the last 30 days."
+  },
+  "stats": {
+    "activeUsers": "Active users",
+    "comparisons": "Comparisons",
+    "ratedVideos": "Rated videos"
+  },
```
lfaucon, line 155: I see you moved some fields around. What is the order that you try to have ?

lfaucon (approved):
Looks awesome 👍

GresilleSiffle (approved):
This is great :+1: 

Personal opinion: I'd tend to avoid using a divider between the criteria sliders, as they are already visually well separated for me, with their title and the slider itself. I think the UI will be more lightweight without them.

PR #673, opened 2022-03-13
refactor: Rename TweetedVideo model and related name to entity model

lfaucon:
I found that the related name `"video"` was not well chosen as it is not intuitive `my_entity.video` returns the twitter bot information. I renamed both the model and the related name


backend/twitterbot/models/tweeted.py
```diff
@@ -12,13 +12,13 @@
...
 
-class TweetedVideo(models.Model):
+class TweetInfo(models.Model):
     """One tweeted video."""
 
     video = models.ForeignKey(
         Entity,
         on_delete=models.CASCADE,
-        related_name="video",
+        related_name="tweet_info",
```
amatissart, line 21: "video" was indeed misleading. However, this `related_name` represents a queryset for all "tweet info" instances that are related to this video. 

So `tweet_info_set` may be more appropriate, or simply `tweets` ?

lfaucon: Good catch. Renamed to `tweets` in 42162947

amatissart (approved):

PR #678, opened 2022-03-15
Prettify extension

sigmike:
Related issue: #625

The `browser_extension` directory now contains many files that should not be packaged so I moved all the source code into a `src` subdirectory to avoid having to exclude all these files.

I copied eslint and prettier configs from the frontend, removed all react and typescript options and added the `webextensions` env to eslint.

I ran prettier on all the source files (html, css, etc.) in 2bdb28b596e762ce1fc7b4ce71c4173549c0dd70, but eslint only checks the js files.

I fixed all the errors reported by eslint in 6259f58e458f6c6c742d78d1f7ae7bdc7fe72269. A few comments:

- I removed `const.js`. I didn't find a way to make eslint know that this file is prepended to all other files so it complains about the constants being undefined. Since `const.js`' constants are actually only used in a single file, I just moved them there. If we really want a separate `const.js` we can probably tell eslint to ignore these undefined variables, or we could add it in `web_accessible_resources` and load it asynchronously (https://stackoverflow.com/questions/48104433/how-to-import-es6-modules-in-content-script-for-chrome-extension).
- Some variable assignments are removed because they are never used.
- I added `const` in front of variable assignments that didn't declare the variable, assuming they are not global variables.
- I would change all the `var` to `let` or `const`, but since this rule is not enforced by eslint config I didn't touch them.

I tested the extension briefly and it seems to work, but I don't know everything the extension can do.

I think squashing the commits of this PR would make the history more difficult to read if we ever need to. Having separate commits for the global move, prettier, and the eslint fixes seems better to me.

Remaining to do:
- [ ] Run the linter in the CI


GresilleSiffle (approved):
Thanks a lot @sigmike :+1: 

I installed the extension on Chromium 99 and Firefox 91 on Debian. Everything looks fine, so far I've tested:
- refreshing the recommendations
- expanding the recommendations
- manual login through the iframe
- automatic login after a token refresh on https://tournesol.app 

I suggest to update the patch version to `2.1.6` because the code has changed since 2.1.5, even if it adds no new feature for the user.

amatissart (approved):

Issue #680, opened 2022-03-17
[front] feat: make it possible to have a home page by poll


GresilleSiffle:
... to describe ...

PR #686, opened 2022-03-17
[front] feat: each poll can have its own Home page

GresilleSiffle:
**related to** #680 

---

This poll request adds the home route to the `PollRoutes`, allowing different polls to have different home page.

I added a dummy home page for the poll `presidentielle2022` to ensure everything is working.


frontend/src/app/PollRoutes.tsx
```diff
@@ -81,7 +88,7 @@ const PollRoutes = ({ pollName }: Props) => {
           return;
         }
         return (
-          <route.type key={route.id} path={`${basePath}/${route.url}`}>
+          <route.type key={route.id} path={`${basePath}/${route.url}`} exact>
```
amatissart, line 91: At first, I was concerned that using `exact` could break some existing URLs, but it seems to be fine. And react-router will route URLs with a trailing slash to the same route. :+1:

amatissart (approved):

frontend/src/app/PollRoutes.tsx
GresilleSiffle: Si, I added exact to make the routing independent of the routes declaration order. For instance, it prevents the page `/presidentielle2022/comparison` to match the home page `/presidentielle2022/`.

If at some point we experience unexpected behaviours we could always declare this prop conditionally.

Issue #626, opened 2022-02-25
[front] feat: Criteria scores BarChart


lfaucon:
Create a component similar to the CriteriaRadarChart.jsx, but which displays scores as a BarChart instead.

sigmike:
Design that was suggested in chat:

![tournesol_bar_chart](https://user-images.githubusercontent.com/28259/159001376-fcff2ebb-539e-450f-bdd8-d88e1379e958.png)

PR #688, opened 2022-03-18
Add Bar chart on video analysis page

sigmike:
Related issue: #626

Initial attempt to add the Bar chart:

image

Moving the ticks in the middle doesn't seem possible with recharts. It may be possible to solve that by making 2 aligned charts.

Most of the code is copied from the Radar chart. Some of it should be refactored if both charts are going to stay.

### DIFF
--- a/frontend/src/components/Card.tsx
+++ b/frontend/src/components/Card.tsx
@@ -3,7 +3,7 @@ import Card from '@mui/material/Card';
 
 interface Props {
   children: React.ReactNode;
-  sx: object | undefined;
+  sx?: object;
   [props: string]: unknown;
 }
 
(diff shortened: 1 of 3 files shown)


lfaucon:
Screenshot 2022-03-18 at 22 27 53

Pushed a commit to improve the layout, including:
- centering the icons
- removeing the margin created by the left Y-Axis
- a tooltip that displays the exact values

lfaucon (approved):
LGTM!

The design of the Video Analysis page will be in any case improved as part of #694

sigmike:
> centering the icons

But this hides small bars. We don't care? I can try the aligned dual chart.

lfaucon:
I discussed yesterday evening with @aidanjungo and @amatissart and it did not seem to important that very small scores are hidden. The tooltip helps to see the exact value.

Unless you think it wll be very easy to do the double bar chart, I'd recommend keeping this visualization as it is now and revisiting it only later if we feel the need.

PR #692, opened 2022-03-18
Fix a race condition when the 2 entity selectors are updated concurrently

amatissart:
Due to how the location was memoized by the callback `onChange`, the URL change could fall in a race condition when the 2 "auto" buttons were activated simultaneously. Only one of the 2 changes would be persisted in the URL.

Reading `window.location` just in time should solve the issue.


lfaucon (approved):
Could not reproduce the issue with the new implementation 💯

PR #693, opened 2022-03-18
feat: display and editing for EntityCriteriaScore Admin

lfaucon:


backend/tournesol/admin.py
```diff
@@ -104,7 +104,17 @@ def get_language(obj):
...
 class EntityCriteriaScoreAdmin(admin.ModelAdmin):
-    pass
+    list_display = (
+        'entity',
+        'criteria',
+        'score'
+    )
+    list_editable = (
+        'score',
+    )
```
amatissart, line 114: I am not convinced that make this `score` editable is useful, as they well be overridden on the next ml run.

amatissart (approved):

Issue #681, opened 2022-03-17
[front] feat: design and implement a comparison series


GresilleSiffle:
## feature specifications

As a Tournesol user, on the `presidentielle2022` poll, I can start a comparison series, behaving like a tutorial.

**From where the comparison series is accessible?**

... describe where the starting point is in the app, and what are the conditions ... 

The implementation of the conditions should be done with another issue.

**Should the user be logged before starting a comparison series?**

Yes, to ease the development process.

**Should this component be usable by any poll?**

Yes we would like to, but this could be done in a later development.

**Key points**

- each step consists of a comparison page, with a specific message related to the current step
- (these messages should be easily configurable)
- the first two candidates are proposed randomly
- a breadcrumb indicates the current position (step) in the series and the remaining comparisons to do
- clicking on submit saves the comparison and goes to the next comparison page, and increases the step value by 1
- submitting a comparison already made doesn't increase the step value

**How to refresh  the candidates between each submission ?**

... to decide with the team ...

## technical specifications

Create a `<ComparisonSeries />` component.

This component should have two props:
- `length`, the length of the series
- `messages`, an object containing translated messages, each of them associated with a step

## remaining to-do

- [x] #706
- [x] review them with the team (reviewed with Aidan, Adrien and partially by Louis) // can be updated at any time
- [x] #707
- [x] #726
- (distinct topic) add a link in the future `presidentielle2022` home page
- (distinct topic) add a result page

PR #696, opened 2022-03-21
[front] feat: add the possibility to create series of comparison

GresilleSiffle:
**related to** #681 

---

This pull request gives the possibility to developers to create one or more comparison series per poll, anywhere in the application.

A comparison series consists of successive comparisons, with a fixed length either defined by the poll, or customized. Each step in the series (even the first one) can trigger a dialog box to give useful information to the users. This new dialog box is a standalone component and is not specific to comparison series. 

After each comparison, one of the two compared entities is refreshed. For now the refresh strategy is pretty straightforward, but in the future we could use different suggestion strategies by simply passing a selection function to the `<ComparisonSeries>` component.

After the final step of the series, the user can continue its comparisons indefinitely until he / she arrives on an already made comparison.

**next steps**

Make the suggested comparisons aware of the potential comparisons already made by the user ( check the back end ). Also make the starting point dynamic, according to the comparisons already made.

### to-do

**new component** `<ComparisonSeries>`
- [x] create a `<ComparisonSeries>`
- [x] with prop `length`
- [x] with prop `messages`
- [x] display the progression
  - [x] create a dummy UI
  - [x] make it works with `length`
  - [x] display the message of the current step
  - [x] fix the forbidden non-null assertion 
- [ ] after the last comparison the user should be redirected to somewhere

**new component** `<DialogBox>`
- [x] can display a title
- [x] can display a undetermined number of paragraphs
- [x] can be closed by clicking on a button

**Comparison**
- [x] create a process that automatically refresh one of the two compared entities after each comparison
  - [x] make the comparison series able to suggest new UIDs
  - [x] make the comparison series able to suggest which entity ( left / right ) should be refreshed
  - [x] avoid suggesting the same comparison in the same series

**page Comparison**
- [x] do not hard-code dialogs

**new entities shortcuts**
- [x] move `selectNextEntity` outside of `<ComparisonSeries>`
- [x] move `alreadyComparedWith` outside of `<ComparisonSeries>`

**URL**
- [x] should be available at URL `/comparison?series=true`
- [x] if the value of `series` is not `true`, the standard comparison UI is displayed

**poll specificity**
- [x] should the length of the series be stored in the polls constant?

### preview

An example of comparison series. The algorithm doesn't suggest comparisons that have already been made.

https://user-images.githubusercontent.com/39056254/159707088-7f15c44c-a357-4774-b129-2c230a73c9d3.mp4

### UX/UI exploration

with a mui `<Stepper />`

This component shows the number of comparison made and the number of remaining comparisons. It adds several new visual information, one chip per step, which are not mandatory (and in my opinion adds non necessary complexity in the UI). This could be simplify by displaying only three steps, the first one, the last one, and the current one with a label like " 4 / 7 ". 

capture

with a mui `<LinearProgress />`

capture2

alternative

capture3


amatissart (approved):
The wording of the tutorial dialogs will need translations and possibly other adjustments in a separate PR. Apart from it looks good to me :+1:

GresilleSiffle:
Fine,  I'm going to merge this first step and document what's missing in the original issue #696 

- add translation for the dialog
- review them with the team
- make the component aware of the existing comparison
- add a link in the future `presidentielle2022` home page
- add a result page

Issue #689, opened 2022-03-18
[front] feat: Language and date selectors should be explicit that everything is selected when nothing is selected


lfaucon:
# Date filter
Either:
- Small "x" next to the filtering options to show the selected filter can be clicked to be removed
- An "All time" option

# Language filter
- Display "All languages" text when no languages is filtered

PR #700, opened 2022-03-23
Explicit "everything" filter

sigmike:
Related issue: #689 

I tried the small "x" but it doesn't look good when it's warped on a new line on some window widths, so I added an "All time" option instead.

For the languages I shortened the French version to "Toutes" because the input is not long enough for "Toutes les langues".

image
image

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -118,6 +118,7 @@
     "thisWeek": "This week",
     "thisMonth": "This month",
     "thisYear": "This year",
+    "allTime": "All time",
     "uploadDate": "Upload date",
     "language": "Language",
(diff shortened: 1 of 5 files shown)


GresilleSiffle:
LGTM :+1: 

Works fine. Good for you @amatissart ?

amatissart (approved):
Looks good to me :+1: 

Perhaps the `CheckCircle` icon could be replaced with a `RadioButtonUncheckedIcon` to be more consistent with native radio buttons where a choice is selected at all time. Or maybe even use the `<Radio>` component provided by MUI. But I'm ok with addressing that separately.

GresilleSiffle:
I create a new issue to potentially rework the upload date filter, so that we can merge these new additions.

Issue #512, opened 2022-01-27
[twitter-bot] Integration of the Twitter bot in the main repo


aidanjungo:
[Tournesol-twitter-bot](https://github.com/tournesol-app/tournesol-twitter-bot) was until now in a separate repo, it should be integrated into the main repo to interact with the database an be automatize from the server.

The integration will be breakdown into a few steps:

- [ ] Add new Django models to save video which have been tweeted
- [ ] Write test for those models
- [ ] Add the Tournesol-twitter-bot functions (with a bit of refactor) to the main repo (/tournesol/twitter-bot/)
- [ ] Write test for those functions
- [ ] Deploy on the server and use cron job to automatize

aidanjungo:
@amatissart don't hesitate to comment and/or add things that I could missed 🙂

PR #702, opened 2022-03-23
Modification for Twitterbot deployement

aidanjungo:
Linked to #512 

Must be merge only after #567


amatissart:
I updated "get-vm-secrets.sh" and defined a function to read the settings values more safely. The previous implementation would break if minor changes were applied in "settings.yaml" (e.g order of keys, presence of quotes or commas, etc.)

As a consequence #916 needs to be merged and deployed first, to install the new dependency on the server.

amatissart (approved):

PR #703, opened 2022-03-23
feat: Home Page for presidentielle2022

lfaucon:


aidanjungo:
> I changed the title to "Beneficial Collaborative Decisions". It is probably less easy to understand as previously "Collaborative Content Recommendations". I like that it targets more the heart of what Tournesol is, especially in the context of multiple Polls. I would be ok if others prefer to keep the original title on the video home page. what do you think @aidanjungo ?

For the videos I prefer "Collaborative Content Recommendations" but for the presidents (and possibly other topic) I like the new one.

PR #704, opened 2022-03-23
backend/tournesol/entities annotations, code layout and type homogenization

Vikka:
First simple pr.
Almost nothing, I read the code and saw some stuff I can improve.

The only "big" change is the returned type of `get_uid_regex` that now return `str` instead of `str | None`.
AFAIK Only used in backend/tournesol/serializers/entity.py:L241, type truthiness is used.
I like simpler signature. PEP20, simple is better than complex. Optional[str] is more complex than str.

### DIFF
--- a/backend/tournesol/entities/video.py
+++ b/backend/tournesol/entities/video.py
@@ -42,10 +42,10 @@ class VideoEntity(EntityType):
         )
 
     @classmethod
-    def get_uid_regex(cls, namespace):
+    def get_uid_regex(cls, namespace) -> str:
         if namespace == YOUTUBE_UID_NAMESPACE:
             return YOUTUBE_UID_REGEX
-        return None
+        return ''
 
     def update_metadata_field(self):
         from tournesol.utils.api_youtube import VideoNotFound, get_video_metadata
--- a/backend/tournesol/entities/video.py
+++ b/backend/tournesol/entities/video.py
@@ -70,7 +70,6 @@ class VideoEntity(EntityType):
         are older than `VIDEO_METADATA_EXPIRE_SECONDS`.
         The request can be forced with `.refresh_metadata(force=True)`.
         """
-        return self.instance.last_metadata_request_at is None or (
-            timezone.now() - self.instance.last_metadata_request_at
-            >= timedelta(seconds=settings.VIDEO_METADATA_EXPIRE_SECONDS)
-        )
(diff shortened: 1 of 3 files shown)


backend/tournesol/entities/video.py
```diff
@@ -64,13 +64,12 @@ def update_metadata_field(self):
...
         are older than `VIDEO_METADATA_EXPIRE_SECONDS`.
         The request can be forced with `.refresh_metadata(force=True)`.
         """
-        return self.instance.last_metadata_request_at is None or (
-            timezone.now() - self.instance.last_metadata_request_at
-            >= timedelta(seconds=settings.VIDEO_METADATA_EXPIRE_SECONDS)
-        )
+        return self.instance.last_metadata_request_at is None \
+               or (timezone.now()-self.instance.last_metadata_request_at
+                   >= timedelta(seconds=settings.VIDEO_METADATA_EXPIRE_SECONDS))
```
GresilleSiffle, line 75: This change makes the CI to fail.

It's caused by `.github/workflows/backend-ci.yml` which runs `./scripts/ci/lint.sh`.

https://github.com/tournesol-app/tournesol/blob/665e2100e4bce4d969b9fee822a264571f2ad86e/.github/workflows/backend-ci.yml#L52

This script run several code quality tools, and will return 1 if one of the checks fails.

You can check the code quality by running this script from the `tournesol/backend` folder before submitting a PR. 

Let's keep the code as it was before the change, as it's " black " formatted*, so everyone can reproduce it by simply running `black` on the file (see https://github.com/psf/black).

```suggestion
        return self.instance.last_metadata_request_at is None or (
            timezone.now() - self.instance.last_metadata_request_at
            >= timedelta(seconds=settings.VIDEO_METADATA_EXPIRE_SECONDS)
        )
```

* note to self: add `black` to the CI : )

GresilleSiffle (changes requested):
**edit:** I clicked on the commit suggestion button just to see if it was possible to directly commit on a fork from a pull request, and it is. So I'm going to merge as soon as the CI is green.

Hello @Vikka 

Thanks for your contribution. All changes are OK except one that make the CI to fail, so I made a suggestion to quickly fix the problem.

After that, we will be ready to merge.

Thank you again :+1: :snake:

GresilleSiffle (approved):

PR #705, opened 2022-03-24
[front] feat: add translations for the `presidentielle2022` tuto

### LINKED ISSUE #706: [front] feat: make it possible to translate the tutorials

GresilleSiffle:
**is subtask of** #681 

close #706 and #707

---

This PR add the possibility to translate the tutorial of each poll. It comes with a dialog proposition that could be updated at any moment. It also makes the `<ComparisonSeries>` component aware of the existing comparisons made by the user (see the commit c2b5e97, I accidentally pushed two features in the same branch, tell me if you prefer two PR @amatissart). 

I use several sub-keys in the translation so that we can add dialog for more than one series in a given poll.

The keys are " generic " and not talkative, but they help to add and remove paragraph rather than easily. Contrary to even more generic keys containing several paragraphs inside on string, which makes working on the translations uncomfortable.

**known limitations**

**(1)** updating an already made comparison by manually selecting a candidate results in a " soft lock "

I developed a fix on a local branch, I'll create a second pull request to make the reviews easier (see #721). 

**(2)** refreshing the page with make the comparison series to restart at 1

This is a missing feature, I'll update the component when **(1)** will be merged.

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -415,8 +415,8 @@
         },
         "title7": "Thanks a lot!",
         "message7": {
-          "p10": "One last comparison and you will be able to see your results.",
-          "p20": "We sincerely hope you've appreciated this new voting system."
+          "p10": "One last comparison and you will be able to see your current preferences.",
+          "p20": "We sincerely hope you appreciated this new voting system, which may have allowed you to better think about why you support some candidates."
         }
       }
     }
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -381,15 +381,15 @@
       "tutorial": {
         "title1": "Welcome on Tournesol 🌻",
         "message1": {
-          "p10": "In order to familiarize yourself with the comparison system, small messages like this one will be displayed.",
-          "p20": "Let's start with a first comparison between two candidates.",
-          "p30": "Move the blue handle below the candidates, then submit to select the person who should be president.",
-          "p40": "The more the handle is in the center, the more the candidates are considered similar."
+          "p10": "This tutorial will guide you step by step through a comparison series.",
+          "p20": "First, move the blue handles and adjust your preferences by moving them more or less far. Save when your choice is made.",
+          "p30": "The more your opinion is clear-but on a criterion, the more the handle should be close to the slider extremity.",
+          "p40": "If you find candidates similar, the handle should remain close to the center."
         },
         "title2": "Congratulations !",
         "message2": {
-          "p10": "Let's continue with other candidates.",
-          "p20": "Note that you can manually choose the candidates by using the drop down menu located just above the candidates' photos."
+          "p10": "You have made your first comparison \uD83E\uDD73",
+          "p20": "Note that you can manually choose the candidates by using the drop down menus located just above the candidates' photos."
         },
         "title3": "Tip",
         "message3": {
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -383,7 +383,7 @@
         "message1": {
           "p10": "This tutorial will guide you step by step through a comparison series.",
           "p20": "First, move the blue handles and adjust your preferences by moving them more or less far. Save when your choice is made.",
-          "p30": "The more your opinion is clear-but on a criterion, the more the handle should be close to the slider extremity.",
+          "p30": "The more your opinion is clear-cut on a criterion, the more the handle should be close to the slider extremity.",
           "p40": "If you find candidates similar, the handle should remain close to the center."
         },
         "title2": "Congratulations !",
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -393,14 +393,14 @@
         },
         "title3": "Tip",
         "message3": {
-          "p10": "Don't worry if you are not confident. You will be able to update all comparisons later, at your own pace.",
-          "p20": ""
+          "p10": "Don't worry if the handles' position doesn't feel exact.",
+          "p20": "You will be able to update all comparisons later, at your own pace"
         },
         "title4": "Did you see?",
         "message4": {
           "p10": "There are also optional criteria under the main ones.",
           "p20": "Try to unfold and use the optional criteria to give a more complete opinion about the candidates.",
-          "p30": ""
+          "p30": "It's also possible to uncheck an optional criterion if you don't want to give your opinion on it."
         },
         "title5": "Is it better than voting for a single person?",
         "message5": {
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -2,6 +2,9 @@
   "components": {
     "filtersButton": "filters"
   },
+  "dialogBox": {
+    "continue": "continue"
+  },
   "video": {
     "labelShowSettings": "Show settings related to this video",
     "nbViews": "{{nbViews}} views",
@@ -44,9 +47,6 @@
     "goToComparison": "Go to comparison"
   },
   "submit": "submit",
-  "dialogBox": {
-    "continue": "continue"
-  },
   "videoSelector": {
     "newVideo": "Select a new video automatically",
     "autoEntityButton": "Auto",
@@ -375,5 +375,49 @@
     "videos": "videos",
     "entityCandidate": "candidate",
     "entityVideo": "video"
+  },
+  "presidentielle2022": {
+    "dialogs": {
+      "tutorial": {
+        "title1": "",
+        "message1": {
+          "p10": "",
+          "p20": "",
+          "p30": "",
+          "p40": ""
+        },
+        "title2": "",
+        "message2": {
+          "p10": "",
+          "p20": ""
+        },
+        "title3": "",
+        "message3": {
+          "p10": ""
+        },
+        "title4": "",
+        "message4": {
+          "p10": "",
+          "p20": ""
+        },
+        "title5": "",
+        "message5": {
+          "p10": "",
+          "p20": "",
+          "p30": ""
+        },
+        "title6": "",
+        "message6": {
+          "p10": "",
+          "p20": "",
+          "p30": ""
+        },
+        "title7": "",
+        "message7": {
+          "p10": "",
+          "p20": ""
+        }
+      }
+    }
   }
 }
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -123,8 +123,8 @@
     "thisYear": "This year",
     "allTime": "All time",
     "uploadDate": "Upload date",
-    "language": "Language",
     "allLanguages": "All languages",
+    "language": "Language",
     "uploader": "Uploader"
   },
   "ratings": {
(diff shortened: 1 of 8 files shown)


frontend/src/pages/comparisons/Comparison.tsx
```diff
@@ -15,10 +29,36 @@ const ComparisonPage = () => {
...
+      const comparisons = await getUserComparisons(pName);
+      setAlreadyMadeComparisons(
+        comparisons.map((c) => c.entity_a.uid + '/' + c.entity_b.uid)
+      );
+    }
+
+    if (series === 'true') {
+      getUserComparisonsAsync(pollName);
+    }
+  }, [pollName, series]);
```
amatissart, line 61: As it's used in a series only, would not it make sense to move this effect into `ComparisonSeries`?

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
```diff
@@ -50,14 +52,20 @@ const ComparisonSeries = ({
...
 
   const searchParams = new URLSearchParams(location.search);
   const uidA: string = searchParams.get(UID_PARAMS.vidA) || '';
   const uidB: string = searchParams.get(UID_PARAMS.vidB) || '';
 
+  useEffect(() => {
+    if (alreadyMadeComparisons !== undefined) {
+      setComparisonsMade(alreadyMadeComparisons);
+    }
+  }, [alreadyMadeComparisons, setComparisonsMade]);
```
amatissart, line 67: If I am not mistaken the "alreadyMadeComparisons" may be empty on the first render. So the random entities could be selected before the comparisons have been fetched.

Maybe this component should render a loader in this case, and wait for the comparisons list before initializing the comparison?

frontend/src/pages/comparisons/Comparison.tsx
```diff
@@ -15,10 +29,36 @@ const ComparisonPage = () => {
...
   const tutorialAlternatives = options?.tutorialAlternatives ?? undefined;
   const tutorialDialogs = options?.tutorialDialogs ?? undefined;
+  const [alreadyMadeComparisons, setAlreadyMadeComparisons] = useState<
+    Array<string>
+  >([]);
+
+  let dialogs;
+  if (tutorialDialogs) {
+    dialogs = tutorialDialogs(t);
+  }
```
amatissart, line 43: (minor) Can be rewritten as a `const`:
```suggestion
  const dialogs = tutorialDialogs ? tutorialDialogs(t) : undefined;
```

frontend/src/pages/comparisons/Comparison.tsx
GresilleSiffle: At first I thought it was a good idea to receive the existing comparisons as a prop, so that the parent component can control which comparisons will be avoided by the series, but I don't see which use cases would require such a system.

I'm going to move this effect in `<ComparisonSeries>`.

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
GresilleSiffle: > If I am not mistaken the "alreadyMadeComparisons" may be empty on the first render. So the random entities could be selected before the comparisons have been fetched.

That's true, most of the time is empty during the first render. Good idea to add a loader.

frontend/src/pages/comparisons/Comparison.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/705/commits/aa5edac2e780ebbb16249a46fce15689afdccf6b.

The code is a bit more complex now, mainly because I needed to call `genInitialComparisonParams` after `getAlternativesAsync` and `getUserComparisonsAsync`. I chained three promises in the `useEffect`, and it works, but I don't know if this is elegant or the React way to do it.

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/705/commits/08ca381787b24919047781ace366b182b13d898f.

frontend/public/locales/en/translation.json
```diff
@@ -375,5 +375,49 @@
...
     "entityVideo": "video"
+  },
+  "presidentielle2022": {
+    "dialogs": {
+      "tutorial": {
+        "title1": "Welcome on Tournesol \uD83C\uDF3B",
+        "message1": {
+          "p10": "In order to familiarize yourself with the comparison system, small messages like this one will be displayed.",
+          "p20": "Let's start with a first comparison between two candidates.",
+          "p30": "Move the blue handle below the candidates, then submit to select the person who should be president.",
```
lfaucon, line 386: The French version specifies "d'après vous". I would prefer to also add here "who should be presnident according to you".

frontend/public/locales/en/translation.json
```diff
@@ -375,5 +375,49 @@
...
+          "p40": "The more the handle is in the center, the more the candidates are considered similar."
+        },
+        "title2": "Congratulations !",
+        "message2": {
+          "p10": "Let's continue with other candidates.",
+          "p20": "Note that you can manually choose the candidates by using the drop down menu located just above the candidates' photos."
```
lfaucon, line 392: Should "menu" be plural?

frontend/public/locales/en/translation.json
```diff
@@ -375,5 +375,49 @@
...
+        },
+        "title5": "Is it better than voting for a single person?",
+        "message5": {
+          "p10": "Contrary to the current voting system in France, here you are invited to give your opinion on several candidates.",
+          "p20": "Thus it's possible to express your favourite choice, your second choice, the third one, etc.",
+          "p30": "This system avoids problem like « tactical vote », and doesn't encourage the polarization of political opinions."
+        },
+        "title6": "How does it work?",
+        "message6": {
+          "p10": "It's by aggregating comparisons of all users that Tournesol is able to understand their preferences.",
```
lfaucon, line 411: Could be simplified into: "By aggregating comparisons of all users Tournesol is able to understand their preferences"

```suggestion
          "p10": "By aggregating comparisons of all users Tournesol is able to understand their preferences.",
```

frontend/public/locales/en/translation.json
```diff
@@ -375,5 +375,49 @@
...
+          "p20": "It's not the candidate who received the most voices who is elected, but the one with the most favourable sum of opinions.",
+          "p30": "The objective is to get the results that fits the best the users' global opinion."
+        },
+        "title7": "Thanks a lot!",
+        "message7": {
+          "p10": "One last comparison and you will be able to see your results.",
+          "p20": "We sincerely hope you've appreciated this new voting system."
```
lfaucon, line 418:
```suggestion
          "p20": "We sincerely hope you appreciated this new voting system."
```

I'm 90% sure of this :)

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
```diff
@@ -16,6 +26,14 @@ interface Props {
   length: number;
 }
 
+export function getUserComparisons(
+  pollName: string
+): Promise<ComparisonModel[]> {
+  return UsersService.usersMeComparisonsList({
+    pollName,
+  }).then((data) => data.results ?? []);
```
lfaucon, line 34: I am quite convince that it makes the code simpler and less error prone to use `async` and `await` whenever needed while avoiding `Promises` and `then` as much as possible. As of now we have a bit of a mix of both styles which is probably worse than either one?

lfaucon (commented):
Not approving, because I have not tested nor reviewed the code changes to the ComparisonSeries component

frontend/public/locales/en/translation.json
GresilleSiffle: I think so.

Fixed by https://github.com/tournesol-app/tournesol/pull/705/commits/f963a07da5157a207f4a7579ecd252cbed9fc5b3.

frontend/public/locales/en/translation.json
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/705/commits/91597d1560abff84e939dd37bf267525fe74d387.

The sentence is closer to the french one now.

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
GresilleSiffle: Yes you're right. I also remember you made this suggestion a long time ago in another pull request.

I'll see what I can do.

frontend/public/locales/en/translation.json
GresilleSiffle: Thanks!

Fixed by https://github.com/tournesol-app/tournesol/pull/705/commits/923e6b9239bc62ae3b4ab29bd00b13e4b8ba94d6.

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/705/commits/6ce3adc9dc6d6e8fce008dcf1a988cec1778393c.

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
```diff
@@ -16,6 +26,16 @@ interface Props {
   length: number;
 }
 
+async function getUserComparisons(
+  pollName: string
+): Promise<ComparisonModel[]> {
+  const comparisons = await UsersService.usersMeComparisonsList({
+    pollName,
+  });
```
amatissart, line 34: By default the number of comparisons is limited to 30 per page. 
100 should be enough, (66 pairs of candidates in this poll).
```suggestion
  const comparisons = await UsersService.usersMeComparisonsList({
    pollName,
    limit: 100,
  });
```

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
```diff
@@ -50,32 +74,72 @@ const ComparisonSeries = ({
...
+          });
+        } else {
+          // stop loading if no alternatives have been provided
+          setIsLoading(false);
+        }
+      });
+    } else {
+      // stop loading if no series is going to be rendered
+      setIsLoading(false);
+    }
```
amatissart, line 140: One way to make that a bit simpler maybe (and `Promise.all()` makes the 2 requests concurrent instead of sequential).

Feel free to apply it or not.

```suggestion
    if (length >= MIN_LENGTH) {
      const comparisonsPromise = getUserComparisonsAsync(pollName);
      const alternativesPromise = getAlternatives
        ? getAlternativesAsync(getAlternatives)
        : Promise.resolve();
      Promise.all([comparisonsPromise, alternativesPromise])
        .then(([comparisons, entities]) => {
          if (entities && initialize.current && (uidA === '' || uidB === '')) {
            setFirstComparisonParams(
              genInitialComparisonParams(entities, comparisons, uidA, uidB)
            );
          }
        })
        .then(() => {
          setIsLoading(false);
        });
    } else {
      // stop loading if no series is going to be rendered
      setIsLoading(false);
    }
```

amatissart (approved):

PR #709, opened 2022-03-24
add jinja2 maximum version until flake8-html next release

Vikka:
The issue of jinja2 new version is already fixed but not released: https://github.com/lordmauve/flake8-html/commit/edef580c6d5e0ccb3ab54aa31b42b4359f7b9235

PR #710, opened 2022-03-24
use the os.environ.get shortcut with default arg

Vikka:


Vikka:
Flake8 issue is solved by this PR: https://github.com/tournesol-app/tournesol/pull/709

backend/settings/settings.py
```diff
@@ -16,15 +16,14 @@
 import yaml
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
+from dotenv import load_dotenv
```
amatissart, line 19: This import could be moved above the comment about `BASE_DIR`.

backend/settings/settings.py
```diff
@@ -16,15 +16,14 @@
...
 
+load_dotenv()
 
 server_settings = {}
-SETTINGS_FILE = (
-    "SETTINGS_FILE" in os.environ
-    and os.environ["SETTINGS_FILE"]
-    or "/etc/django/settings-tournesol.yaml"
-)
+SETTINGS_FILE = os.getenv("SETTINGS_FILE", "/etc/django/settings-tournesol.yaml")
```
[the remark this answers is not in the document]
Vikka: Exactly. This is a very practical way to handle env variable.

backend/settings/settings.py
Vikka: thanks, updated.

amatissart (approved):

PR #711, opened 2022-03-25
[front] security: update `minimist` to 1.2.6 from 1.2.5

GresilleSiffle:
see: https://security.snyk.io/vuln/SNYK-JS-MINIMIST-2429795


GresilleSiffle:
This won't fix the problem I think, as `minimist` is a sub-dependency of other packages.

PR #723, opened 2022-03-28
[front] refactor: snackbars are now displayed in the top right corner

GresilleSiffle:
**Context** 

The snackbars were displayed at the bottom center of the screen, which could hide the submit button in the `<Comparison>` component and other UI components depending on the page.

As there is already the side bar on the left, I excluded the anchors bottom left, center left and top left.

As on mobile the bottom center is very similar to the bottom right I chose the top right position. I also excluded center top to avoid hiding potential important UI components as most of our pages display their content in a centered container

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -17,3 +17,8 @@ code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
     monospace;
 }
+
+.SnackbarItem-wrappedRoot.belowTopBar {
+  /* display alerts below the top bar, to avoid hiding buttons */
+  top: 66px;
+}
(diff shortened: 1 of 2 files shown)


amatissart:
Some e2e tests seem to fail because the snackbar messages are displayed on top of the login/logout button, making it unaccessible. Beside tests, it may also be an issue for real users.  

Do you think it would be possible / preferable to display these message in the top-right corner, just below the topbar?

GresilleSiffle:
Indeed, I just saw the tests.

I'm going to set this pull request to draft for now, to clearly see the screens of the failing tests.

GresilleSiffle:
As you suggested, I think it's a good idea to display the notifications just below the top bar.

Updating the `top` CSS property works well.

amatissart (approved):

Issue #721, opened 2022-03-28
[front] fix: `ComparisonSeries` must not block when an existing comparison is updated


GresilleSiffle:
Instead the `<ComparisonSeries>` component must:
- not increment the `step` state
- not updated its state `comparisonsMade`
- not display several times the same dialog
- suggest a new comparison

PR #724, opened 2022-03-28
[front] feat: make `ComparisonSeries` "resumable"

GresilleSiffle:
must be rebased on main when #721 is merged

---

If a user that has already finished the tutorial manually add the `?series` in the presidentielle 2022 comparison page, he/she will arrive on the comparison series page, with the stepper showing everything is finished. From here, it will be possible to continue adding more comparisons, the UI automatically suggestion new a candidate after each submit.

I think it's a good feature to be able to follow an infinite chain of suggested comparisons. This allows users to not be forced to click on the drop down menus to select new candidates. But this feature might be removed when we will add the result page.


amatissart (approved):
Nice!

Maybe we would expect the dialogs from an intermediary step to not show up when a series is resumed, or to replace it with a dedicated message. But that can be discussed separately, if relevant.

Issue #726, opened 2022-03-29
[front] feat: all comparisons must be private on `presidentielle2022`


GresilleSiffle:
**context**

(1) the primary goal of `presidentielle2022` is not to collect data, the poll will have a short life compared to the poll `videos`

(2) at the moment no use cases need users to set their comparisons public

(3) political opinion are sensitive personal data, better to keep them private

PR #727, opened 2022-03-29
[front] feat: polls must be configured to allow public comparisons

GresilleSiffle:
**related to** #726 

---

Following our "private by default" policy, polls must be explicitly configured to allow users to set their comparisons public.

A poll with the `comparisonsCanBePublic` set to true, will have its entity card display the public / private checkbox. The front end page `/ratings/` will also display the filter by visibility option and update visibility buttons.

For now it's a front end only configuration, but it will be easy to make it a back configuration (a field in the Poll model for instance).

| a poll configured with `comparisonsCanBePublic` = true | a poll without `comparisonsCanBePublic`  |
|---|---|
| capture2 | capture |
| cap | compa |

### DIFF
--- a/frontend/src/pages/videos/VideoRatings.tsx
+++ b/frontend/src/pages/videos/VideoRatings.tsx
@@ -48,7 +48,7 @@ const NoRatingMessage = ({ hasFilter }: { hasFilter: boolean }) => {
 };
 
 const VideoRatingsPage = () => {
-  const { name: pollName } = useCurrentPoll();
+  const { name: pollName, options } = useCurrentPoll();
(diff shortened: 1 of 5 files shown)


frontend/src/pages/videos/VideoRatings.tsx
```diff
@@ -114,6 +114,10 @@ const VideoRatingsPage = () => {
     }
   };
 
+  const comparisonsCanBePublic = () => {
+    return options?.comparisonsCanBePublic === true;
+  };
```
amatissart, line 119: same here

amatissart (approved):
a few functions can be changed into simple `const`.

Otherwise, it looks good to me.

PR #731, opened 2022-03-29
Define feature flag to enable poll presidentielle2022 per environment

amatissart:
A custom environment variable is used to add the poll `presidentielle2022` to the list of selectable polls in frontend.
This mechanism relies on variables that are embedded during the build time: 
see https://create-react-app.dev/docs/adding-custom-environment-variables/ for more details.

This new poll will be enabled in development environment, in tournesol-vm, and on staging. (Not on production for the moment).

This way, the "main" branch remains deployable on all environments at any time. And the feature can easily be disabled at deploy time if necessary.


lfaucon (approved):
Awesome

PR #733, opened 2022-03-29
feat[back]: api endpoint for correlations between criterias for contributor ratings

lfaucon:


backend/tournesol/views/criteria_correlations.py
```diff
@@ -0,0 +1,83 @@
...
+
+class ContributorCriteriaCorrelationsView(GenericAPIView):
+    
+    @extend_schema(
+        responses={
+            200: ContributorCriteriaCorrelationsSerializer
+        }
+    )
+    def get(self, request, poll_name):
+        poll = Poll.objects.get(name=poll_name)
```
GresilleSiffle, line 60: Using an non-existing poll raises an error 500.

As we use the `poll_name` URL path parameter in a lot of views, we have created a `PollScopedViewMixin` class that you can use to handle everything for you.

It will retrieve the poll from the URL, and will make it available in the view and the serializer context. It also takes care of returning a proper 404 when the requested poll name doesn't exist.

I suggest:
1. to make the view inherit from `PollScopedViewMixin` and use `self.poll_from_url` 
2. add a test ensuring 404 is returned instead of 500

You can check `ComparisonDetailApi` to see a usage of `PollScopedViewMixin` and `ComparisonApiTestCase.test_authenticated_cant_read_non_existing_poll` for an example of test.

backend/tournesol/views/criteria_correlations.py
```diff
@@ -0,0 +1,83 @@
...
+        return None
+
+
+class ContributorCriteriaCorrelationsSerializer(serializers.Serializer):
+    criterias = serializers.ListField(child=serializers.CharField(), default=list)
+    correlations = serializers.ListField(child=serializers.ListField(child=serializers.FloatField(), default=list), default=list)
+    slopes = serializers.ListField(child=serializers.ListField(child=serializers.FloatField(), default=list), default=list)
+    
+
+class ContributorCriteriaCorrelationsView(GenericAPIView):
```
GresilleSiffle, line 52: **(1)**

We can add a docstring here to display the description of the endpoint in the generated documentation.

For instance:

```python
"""Retrieve the correlation between... of the logged user, for the given poll."""
```

**(2)**

The endpoint is currently public and accessible as an anonymous user. Should it be private?

backend/tournesol/views/criteria_correlations.py
```diff
@@ -0,0 +1,83 @@
+
```
GresilleSiffle, line 1:
```suggestion
```

backend/tournesol/tests/test_api_criteria_correlation.py
```diff
@@ -0,0 +1,110 @@
...
+            CriteriaRankFactory(criteria__name="hello", poll=self.poll, rank=1),
+            CriteriaRankFactory(criteria__name="world", poll=self.poll, rank=2),
+            CriteriaRankFactory(criteria__name="!", poll=self.poll, rank=3)
+        ]
+        self.entities = EntityFactory.create_batch(10)
+        self.ratings = [ContributorRatingFactory(user=self.user, entity=entity) for entity in self.entities]
+
+    def test_correlations_no_data(self):
+        response = self.client.get(f"/users/me/criteria_correlations/{self.poll.name}/")
+        assert response.status_code == status.HTTP_200_OK
```
GresilleSiffle, line 73: Most of our tests use `self.assertEqual` from Django.

This new test case "breaks" the coherence with the rest of the application, and contributors may be confused.

If this test case is not a proposition to switch from `self.assertEqual` to `assert`, I propose to stick to the current style for now (see the test_api_comparison.py file for instance).

backend/tournesol/tests/test_api_criteria_correlation.py
```diff
@@ -0,0 +1,110 @@
...
+    PollWithCriteriasFactory,
+)
+from tournesol.tests.factories.ratings import (
+    ContributorRatingCriteriaScoreFactory,
+    ContributorRatingFactory,
+)
+from tournesol.views.criteria_correlations import compute_correlation, compute_slope
+
+
+def test_compute_correlation():
```
GresilleSiffle, line 23: `test_compute_correlation` and `test_compute_slope` are outside the test case for a reason?

GresilleSiffle (changes requested):
Hello @lfaucon 

I took the time to review and I hit the request changes button because I saw some bugs that we can easily fix.

I'll probably be working on something else until the end of the week but I'll be available to help in case of need.

Thanks : )

backend/tournesol/tests/test_api_criteria_correlation.py
lfaucon: They don't use the setup and don't touch the database so it felt natural to keep them out of the test case class.

backend/tournesol/tests/test_api_criteria_correlation.py
```diff
@@ -0,0 +1,110 @@
...
+        {e: 24 + 3 * i for i, e in enumerate(entities)}, 
+    ), -3/2)
+    assert np.allclose(compute_slope(
+        {e: i for i, e in enumerate(entities)}, 
+        {e: 24 + 3 * i for i, e in enumerate(entities)}, 
+    ), 3)
+
+
+
+class CorrelationAPI(TestCase):
```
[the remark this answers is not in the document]
lfaucon: Fixed in 94f1d5d0

backend/tournesol/views/criteria_correlations.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy as np
+from django.http import JsonResponse
+from drf_spectacular.utils import OpenApiResponse, OpenApiTypes, extend_schema
```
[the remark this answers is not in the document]
lfaucon: Removed in 56ba5ddb

backend/tournesol/views/criteria_correlations.py
lfaucon: Done 1) in fb09eb7d

I don't see the problem with 2) ... Non-authenticated seem to receive back a 401 response, isn't it? Base on your other comment, I added the test that makes sure of it.

backend/tournesol/views/criteria_correlations.py
```diff
@@ -0,0 +1,83 @@
...
+    
+    @extend_schema(
+        responses={
+            200: ContributorCriteriaCorrelationsSerializer
+        }
+    )
+    def get(self, request, poll_name):
+        poll = Poll.objects.get(name=poll_name)
+        criterias = poll.criterias_list
+        n = len(criterias)
```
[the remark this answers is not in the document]
lfaucon: Fully agree! :)

backend/tournesol/views/criteria_correlations.py
lfaucon: Thanks! Done in a0435a02

backend/tournesol/tests/test_api_criteria_correlation.py
lfaucon: Done in e2ba66d3

amatissart (approved):

amatissart:
@GresilleSiffle I rebased and fixed a small bug on this branch, to ensure that only ratings related to the current poll are considered. It looks good to me now. Do you want to have a look as you had requested some changes?

GresilleSiffle:
Good for me :+1:

Issue #740, opened 2022-03-31
[front] feat: design the personal result page for the `presidentielle2022`


GresilleSiffle:
### functional specifications

After having finished the tutorial, the user must be automatically redirected to its personal and private result page (sometimes also called feedback page in our previous discussions).

It displays **(a)** a title and a clear description (see the mock-up here:  https://miro.com/app/board/uXjVOMy5GR4=/ )

This page can be accessible **(b)** at any moment directly from {**suggestion:** the new sidebar item My results } .

This page must display two widgets:
- **(c)** the three most important criteria according to the user's comparisons
- **(d)** the three best candidates according to the user's comparisons (based on the should be president criterion right?)

See the mock-up here:

The widget **(e)** has one more feature: a share on Twitter button. // describe the expected behaviour of this button //

The widget **(f)** has two more features:
- a button allowing to see the previously made comparisons
- a button allowing to go directly to the comparison UI to add new comparisons

Moreover **(g)** this page also invite the user to check the Tournesol original `videos` poll.

Acceptance criteria: **(a) (b) (c) (d) (e) (f) (g)**

The issue can be made in several pull requests.

PR #741, opened 2022-03-31
Recompute individual scores on comparison submit, for Mehestan only

amatissart:
For polls using the new "Mehestan" algorithm, individual scores are quite cheap to compute and can be updated every time a comparison is created, updated or deleted.

Moreover, it's a prerequisite to implement the "results" page described in #740

### DIFF
--- a/backend/tournesol/tests/test_api_comparison.py
+++ b/backend/tournesol/tests/test_api_comparison.py
@@ -2,6 +2,7 @@ import datetime
 from copy import deepcopy
 from unittest.mock import patch
 
+from django.core.management import call_command
 from django.db.models import ObjectDoesNotExist, Q
 from django.test import TestCase
 from django.utils import timezone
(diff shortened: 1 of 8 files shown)


backend/tournesol/tests/test_api_comparison.py
```diff
@@ -923,3 +932,76 @@ def test_invalid_criteria_in_comparison(self):
...
+                score=score,
+            )
+
+        self.client = APIClient()
+
+    def test_update_individual_scores_after_new_comparison(self):
+        call_command("ml_train")
+
+        self.assertEqual(ContributorRatingCriteriaScore.objects.count(), 4)
+        self.assertEqual(EntityCriteriaScore.objects.count(), 4)
```
GresilleSiffle, line 966: Should we also check the individual scores are not present before making the post?

```python
        self.assertEqual(
            ContributorRatingCriteriaScore.objects
                .filter(contributor_rating__user=self.user2)
                .count(),
            0
        )
```

GresilleSiffle (approved):

PR #760, opened 2022-04-02
Implement 'share on Twitter' button about most important criteria

amatissart:
The content of the tweet was suggested by @aidanjungo 
The translation may need reviewing.

For each criterion, the number of emojis is between 1 and 6.

### DIFF
--- a/frontend/src/features/feedback/StackedCriteriaPaper.tsx
+++ b/frontend/src/features/feedback/StackedCriteriaPaper.tsx
@@ -11,6 +11,7 @@ import {
   ListItemText,
   Typography,
 } from '@mui/material';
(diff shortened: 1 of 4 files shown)


frontend/src/features/feedback/StackedCriteriaPaper.tsx
```diff
@@ -22,6 +23,14 @@ interface Props {
   };
 }
 
+const openTwitterPopup = (text: string) => {
```
lfaucon, line 26: Could be somewhere in a `utils.ts` file to be more discoverable and reuseable?

lfaucon (approved):
Approved with a small suggestion

lfaucon (approved):

PR #794, opened 2022-04-05
feat: Multiple small improvements to Django admin

lfaucon:
* Display meaningful columns for ComparisonCS and ContributorRatingCS admins
* Enable filtering ComparisonCS and ContributorRatingCS by Poll
* Show total number of Comparison and ComparisonCriteriaScores on PollAdmin

Screenshot:
Screenshot 2022-04-05 at 21 33 48

### DIFF
--- a/backend/tournesol/admin.py
+++ b/backend/tournesol/admin.py
@@ -188,15 +188,21 @@ class ComparisonAdmin(admin.ModelAdmin):
 @admin.register(ComparisonCriteriaScore)
 class ComparisonCriteriaScoreAdmin(admin.ModelAdmin):
     list_filter = (
-        "comparison__poll__name",
+        'comparison__poll__name',
     )
     list_display = (
         'id',
+        'comparison',
         'criteria',
         'score'
     )
     readonly_fields = (
         'comparison',
+    )
+    search_fields = (
+        'criteria',
+        'comparison__entity_1__uid',
+        'comparison__entity_2__uid',
     )
 
 
--- a/backend/tournesol/admin.py
+++ b/backend/tournesol/admin.py
@@ -148,11 +148,15 @@ class ContributorRatingCriteriaScoreAdmin(admin.ModelAdmin):
     )
     list_display = (
         'id',
+        'contributor_rating',
         'criteria',
         'score'
     )
     readonly_fields = (
         'contributor_rating',
+    )
+    search_fields = (
+        'contributor_rating__entity__uid',
     )
 
 
--- a/backend/tournesol/admin.py
+++ b/backend/tournesol/admin.py
@@ -4,7 +4,7 @@ Defines Tournesol's backend admin interface
 
 from django.contrib import admin
 from django.contrib.admin.filters import SimpleListFilter
-from django.db.models import Q, QuerySet
+from django.db.models import Count, Q, QuerySet
 
 from .models import (
     Comparison,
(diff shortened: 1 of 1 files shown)


backend/tournesol/admin.py
```diff
@@ -197,17 +217,27 @@ class PollAdmin(admin.ModelAdmin):
...
     inlines = (CriteriasInline,)
 
-    @admin.display(description="Nb of criterias")
+    @admin.display(description="# criterias")
     def get_n_criterias(self, obj):
         return obj.criterias.count()
 
+    @admin.display(description="# comparisons")
+    def get_n_comparisons(self, obj):
+        return obj.comparisons.count()
```
GresilleSiffle, line 235: Not sure, because it has been a long time since I've worked with the administration interface, but this line may generate an SQL query for each object displayed in the list.

I'm going to check.

GresilleSiffle (commented):
Thanks Louis :ok_hand: 

I'm going to check if those new additions generate a correct amount of SQL queries.

backend/tournesol/admin.py
```diff
@@ -177,7 +187,17 @@ def get_poll_name(self, obj):
 
 @admin.register(ComparisonCriteriaScore)
 class ComparisonCriteriaScoreAdmin(admin.ModelAdmin):
-    pass
+    list_filter = (
+        "comparison__poll__name",
+    )
+    list_display = (
```
GresilleSiffle, line 193: In the previous interface we were also able to see the related comparison (user + entity A + entity B). This new `list_display` doesn't provide these information.

It's not a problem of course, but it may be more meaningful if we are able to see the comparison of each criterion. We can display those information by adding `comparison` to the `list_display` for instance. I still don't know if this will generate an extra SQL query by criterion, adding `comparison` might require to update the query set.

**edit** adding `comparison` to `list_display` is safe according to my PostgreSQL logs

What do you think @lfaucon ?

We also might want to display the state of the comparison public, or not, but this is clearly a new feature that doesn't need to be in this PR.

backend/tournesol/admin.py
GresilleSiffle, line 235: These new additions are OK for now, but we should keep in mind we are generating extra 3 extra SQL queries for each poll.

This is clearly manageable for now because we have only 2 polls in production. I'll try to fix this in a next PR.

backend/tournesol/admin.py
```diff
@@ -143,7 +143,17 @@ def get_poll_name(self, obj):
 
 @admin.register(ContributorRatingCriteriaScore)
 class ContributorRatingCriteriaScoreAdmin(admin.ModelAdmin):
-    pass
+    list_filter = (
+        "contributor_rating__poll__name",
+    )
+    list_display = (
```
GresilleSiffle, line 149: Here too we are now not displaying the related entity.

See my comment about the `list_display` of `ComparisonCriteriaScoreAdmin`.

backend/tournesol/admin.py
```diff
@@ -177,7 +187,17 @@ def get_poll_name(self, obj):
 
 @admin.register(ComparisonCriteriaScore)
 class ComparisonCriteriaScoreAdmin(admin.ModelAdmin):
-    pass
+    list_filter = (
+        "comparison__poll__name",
+    )
+    list_display = (
+        'id',
+        'criteria',
```
GresilleSiffle, line 195: If we want to display the related comparison for each criterion. Feel free to drop this suggestion if not relevant.

```suggestion
        'comparison',
        'criteria',
```

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/794/commits/1e87ea98145b9dbcc20e4ac51b469749b0e208de.

backend/tournesol/admin.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/794/commits/d226c0df727c390b8145c0e20af618d376e62c5d.

GresilleSiffle (approved):

PR #799, opened 2022-04-07
[back] feat: add a mgmt cmd to delete inactive users

GresilleSiffle:
This pull request adds a new management command to delete old inactive users.

Inactive users are user with the `is_active` field set to `False`. Old inactive are inactive users who have created their account a certain amount a days ago.

This amount is define by the new setting `APP_CORE["MGMT_DELETE_INACTIVE_USERS_PERIO"']`. This setting is automatically configured to use by default the `REST_REGISTRATION.REGISTER_VERIFICATION_PERIOD` value, so that it equals the lifetime of an activation email. The objective of this configuration is to avoid having valid activation emails linked to deleted users.

The PR also adds a new systemd unit responsible of running daily at 2:20am the management command.

Bonus: when ran in a terminal supporting ANSI-colored output, the command will use fancy colors.

**to-do**
- [x] create the command
- [x] make the time delta configurable
- [x] add tests
- [x] create the scheduled task + a failure alert

### DIFF
--- a/backend/core/management/commands/delete_inactive_users.py
+++ b/backend/core/management/commands/delete_inactive_users.py
@@ -39,6 +39,8 @@ class Command(BaseCommand):
                     f"{n_users_to_delete - deleted[0]} users were NOT deleted"
                 )
             )
+            self.stdout.write(self.style.WARNING("warning partial success"))
         else:
             self.stdout.write(self.style.SUCCESS(f"{deleted[0]} users deleted"))
+            self.stdout.write(self.style.SUCCESS("success"))
         self.stdout.write("end")
--- a/backend/core/management/commands/delete_inactive_users.py
+++ b/backend/core/management/commands/delete_inactive_users.py
@@ -5,19 +5,30 @@ import datetime
 
 from django.core.management.base import BaseCommand, CommandError
 from django.utils import timezone
+
 from core.models.user import User
+from settings import settings
+
 
 class Command(BaseCommand):
-    help = 'Delete users that have not activated their account.'
+    help = "Delete users that have not activated their account."
 
     def handle(self, *args, **options):
-        self.stdout.write(self.style.SUCCESS(f'start command: {__name__}'))
+        self.stdout.write(self.style.SUCCESS(f"start command: {__name__}"))
+        days_to_keep = settings.APP_CORE["MGMT_DELETE_INACTIVE_USERS_PERIOD"]
 
-        delta = timezone.now() - datetime.timedelta(days=1)
-        n_users = User.objects.filter(is_active=False, date_joined__lt=delta).count()
-        self.stdout.write(self.style.SUCCESS(f'{n_users} inactive users found'))
+        # display the configuration if more verbosity is asked
+        if options.get("verbosity", 1) > 1:
+            self.stdout.write(
+                self.style.SUCCESS(f"MGMT_DELETE_INACTIVE_USERS_PERIOD: {days_to_keep}")
+            )
+
+        delete_before = timezone.now() - datetime.timedelta(days=days_to_keep)
+        n_users = User.objects.filter(
+            is_active=False, date_joined__lt=delete_before
+        ).count()
+        self.stdout.write(self.style.SUCCESS(f"{n_users} inactive users found"))
 
         deleted = User.objects.filter(is_active=False).delete()
-        self.stdout.write(self.style.SUCCESS(f'{deleted[0]} users deleted'))
-        self.stdout.write(self.style.SUCCESS(f'{n_users - deleted[0]} were NOT deleted'))
-        self.stdout.write(self.style.SUCCESS(f'end'))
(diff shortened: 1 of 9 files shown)


backend/core/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,46 @@
...
+        ).count()
+
+        # display the configuration if more verbosity is asked
+        if options.get("verbosity", 1) > 1:
+            self.stdout.write(f"MGMT_DELETE_INACTIVE_USERS_PERIOD: {days_to_keep}")
+
+        self.stdout.write(f"{n_users_to_delete} old inactive users found")
+        deleted = User.objects.filter(
+            is_active=False, date_joined__lt=delete_before
+        ).delete()
```
amatissart, line 33: Any reason to count and delete use users separately? Technically, the count may change between the 2 queries. And I think any problem during the delete would cause an exception, so the condition `n_users_to_delete != deleted[0]` can only be True when the count has changed in the meantime, right?

backend/settings/settings.py
```diff
@@ -374,3 +374,18 @@
...
+APP_CORE = {
+    "MGMT_DELETE_INACTIVE_USERS_PERIOD":
+        server_settings.get(
+            "CORE_DELETE_INACTIVE_USERS_PERIOD",
+            # if no configuration has been provided, align the value with the
+            # validity of the validation emails.
+            REST_REGISTRATION.get(
+                "REGISTER_VERIFICATION_PERIOD",
+                7
+            )
```
amatissart, line 389: `REST_REGISTRATION.REGISTER_VERIFICATION_PERIOD` accepts a `timedelta` value:
https://django-rest-registration.readthedocs.io/en/latest/detailed_configuration/register.html#register-verification-period

So overriding the default value in `REST_REGISTRATION` would break this configuration.

backend/core/tests/management/test_delete_inactive_users.py
```diff
@@ -0,0 +1,70 @@
...
+        ).count()
+
+        call_command("delete_inactive_users", stdout=out)
+        output = out.getvalue()
+        n_users_after = User.objects.count()
+
+        self.assertIn("2 old inactive users found", output)
+        self.assertIn("2 users deleted", output)
+        self.assertIn("success", output)
+        self.assertNotIn("were NOTE deleted", output)
```
amatissart, line 45: typo?
```suggestion
        self.assertNotIn("were NOT deleted", output)
```

backend/core/management/commands/delete_inactive_users.py
GresilleSiffle: My original idea was to show in the logs that among this number of users, only this number have been deleted.

I think this is not that relevant if indeed the `.delete()` method fails completely in case of error on an item, as an alert will be raised on Discord. I don't know exactly how `.delete()` works here, the documentation says that its implementation may be subject to change.

In any case, if `.delete()` fails completely in case of error, it could be useful to display in the logs number of users to delete. Maybe I can read the code of `.delete()` and see if there are any obvious exceptions that are raised.

backend/settings/settings.py
GresilleSiffle: Thanks, good catch.

backend/core/management/commands/delete_inactive_users.py
GresilleSiffle: It's not clear what exceptions could be raised by `.delete()`.

According to the Django source code, it could be `TypeError` if the query set is misconfigured.
[See here](https://github.com/django/django/blob/06ebaa9e287137d4496a2736b055f1c6aed95839/django/db/models/query.py#L937)

Or it could be any exceptions raised by the `SQLCompiler.connection.cursor.execute` method.
[See here](https://github.com/django/django/blob/06ebaa9e287137d4496a2736b055f1c6aed95839/django/db/models/sql/compiler.py#L1381)

To not add unnecessary complexity, I think we could wait for an exception to be raised to see exactly what we should expect. In fact, it's even not necessary to catch an exception here, but I wanted to log the remaining number of users to delete still in the database, which require to catch the exception.

backend/settings/settings.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/799/commits/c3ffa21b16e6e562503fc2f73f6886711b0c14ed.

backend/core/management/commands/delete_inactive_users.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/799/commits/a7cf57fe84063fe0469ce5f6c231758872c9e26f and https://github.com/tournesol-app/tournesol/pull/799/commits/c6bc698c0e03aba0e0aeba1e1501fe9456d3a96b.

amatissart (approved):

PR #800, opened 2022-04-07
All paginations at the bottom of their pages

aidanjungo:
To have something uniform for the positions of the paginations, I change from the top to the bottom for:

- Comparisons page
- Rate Later page

### DIFF
--- a/frontend/src/components/Pagination.tsx
+++ b/frontend/src/components/Pagination.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Button, Paper, Typography } from '@mui/material';
 import { useTranslation, Trans } from 'react-i18next';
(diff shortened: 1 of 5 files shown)


amatissart (approved):
Seems reasonable indeed. Thanks!

I just added a commit to scroll on top, on every page change. It was already implemented in the pages that used a pagination at the bottom already. It's now directly implemented by the `Pagination` component.

GresilleSiffle:
Thanks @aidanjungo . I'm always a bit lost when I'm browsing the front end, sometimes the pages are at the top, sometimes at the bottom...

For the future, do you think it's a good idea to have the pagination at the top and the bottom ? Because having to scroll down each time we want to go to the next / previous page is rather uncomfortable, as it requires an extra amount of unneeded scrolling.

PR #807, opened 2022-04-13
[back] perf: speed-up Mehestan criteria processing with multiprocesses

GresilleSiffle:
It's just an idea, I haven't tested it yet against a big database.

### unexpected behaviour

**(1)**

There is currently a bug when running `python manage.py test` or `pytest` locally on my laptop, the following exceptions are raised by the test `ComparisonWithMehestanTest.test_update_individual_scores_after_new_comparison`:

```
psycopg2.OperationalError: SSL error: decryption failed or bad record mac
django.db.utils.OperationalError: SSL error: decryption failed or bad record mac
```

They may be related to this: https://code.djangoproject.com/ticket/31637

**(2)**

The `Pyscopg` documentation says the connection to the database must be created after forks.

> **libpq** connections [shouldn’t be used by a forked processes](https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNECT), so when using a module such as [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing) or a forking web deploy method such as FastCGI make sure to create the connections after the fork.

See: https://www.psycopg.org/docs/usage.html#thread-and-process-safety

The `PostgreSQL` also warns about this

> **Warning** On Unix, forking a process with open libpq connections can lead to unpredictable results because the parent and child processes share the same sockets and operating system resources. For this reason, such usage is not recommended, though doing an exec from the child process to load a new executable is safe.

See: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT

Note that according to the Django 4.0 documentation, the framework should automatically create a new connections when needed. Closing the connections before forking might allow Django to automatically re-create connections in the child processes.

See: https://docs.djangoproject.com/en/4.0/ref/databases/#connection-management

**(3)**

I'm currently investigating how to properly create a new database connection in each child processes.

It seems the `ml_train` works well if all connections are closed before forking and processing individually each criterion.

Still, the `ml_train` fails when it is run by `python manage.py test` or `pytest`, with the following exceptions:

```
psycopg2.InterfaceError: connection already closed
django.db.utils.InterfaceError: connection already closed
```

As Django should automatically create new connections when needed, I think the problem might come from the way the tests are run, i.e. with a single transaction for each test, which makes the multiprocessing code to complain.

**important notes** without closing the database before forking, the tests involving Mehestan failed only on my desktop, now they are failing even on the CI ( which is coherent ).

**(4)**

...

### DIFF
--- a/backend/ml/outputs.py
+++ b/backend/ml/outputs.py
@@ -84,16 +84,27 @@ def save_contributor_scores(
         for contributor_id, video_id, _, _, _ in scores_list
         if (contributor_id, video_id) not in rating_ids
     )
-    created_ratings = ContributorRating.objects.bulk_create(
-        ContributorRating(
-            poll_id=poll.pk,
-            entity_id=video_id,
-            user_id=contributor_id,
-        )
-        for contributor_id, video_id in ratings_to_create
+    ContributorRating.objects.bulk_create(
+        (
+            ContributorRating(
+                poll_id=poll.pk,
+                entity_id=video_id,
+                user_id=contributor_id,
+            )
+            for contributor_id, video_id in ratings_to_create
+        ),
+        ignore_conflicts=True,
     )
+    # Refresh the `ratings_id` with the newly created `ContributorRating`s.
     rating_ids.update(
-        {(rating.user_id, rating.entity_id): rating.id for rating in created_ratings}
+        {
+            (contributor_id, entity_id): rating_id
+            for rating_id, contributor_id, entity_id in ContributorRating.objects.filter(
+                poll_id=poll.pk,
+            ).values_list(
+                "id", "user_id", "entity_id"
+            )
+        }
     )
 
     scores_to_delete = ContributorRatingCriteriaScore.objects.filter(
--- a/backend/ml/outputs.py
+++ b/backend/ml/outputs.py
@@ -88,10 +88,10 @@ def save_contributor_scores(
         (
             ContributorRating(
                 poll_id=poll.pk,
-                entity_id=video_id,
+                entity_id=entity_id,
                 user_id=contributor_id,
             )
-            for contributor_id, video_id in ratings_to_create
+            for contributor_id, entity_id in ratings_to_create
         ),
         ignore_conflicts=True,
     )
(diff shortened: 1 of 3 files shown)


amatissart:
I fear that conflicts may occur when new `ContributorRating` are created (in parallel for each criterion). In practice, it should be unlikely, as the `ContributorRating` are supposed to be created before a comparison is submitted.

Still, I think that it would be safer and simple enough to ignore the conflict when these `ContributorRating` are created in bulk (in "outputs.py"):

```diff
     created_ratings = ContributorRating.objects.bulk_create(
-        ContributorRating(
-            poll_id=poll.pk,
-            entity_id=video_id,
-            user_id=contributor_id,
-        )
-        for contributor_id, video_id in ratings_to_create
+        (
+            ContributorRating(
+                poll_id=poll.pk,
+                entity_id=video_id,
+                user_id=contributor_id,
+            )
+            for contributor_id, video_id in ratings_to_create
+        ),
+        ignore_conflicts=True,
     )
 ```

File: backend/ml/outputs.py [hunk 1]
```diff
@@ -84,16 +84,27 @@ def save_contributor_scores(
...
+    # Refresh the `ratings_id` with the newly created `ContributorRating`s.
     rating_ids.update(
-        {(rating.user_id, rating.entity_id): rating.id for rating in created_ratings}
+        {
+            (contributor_id, entity_id): rating_id
+            for rating_id, contributor_id, entity_id in ContributorRating.objects.filter(
+                poll_id=poll.pk,
+            ).values_list(
+                "id", "user_id", "entity_id"
+            )
```
amatissart, line 106: This does not cover the case when `single_user_id` is defined.   
Why not simply reusing the `ratings` that was defined previously?

```suggestion
            for rating_id, contributor_id, video_id in ratings.values_list(
                "id", "user_id", "entity_id"
            )
```

GresilleSiffle: Sorry I didn't see there was already a `ratings` defined.

amatissart (approved):

Issue #628, opened 2022-02-27
[extension] bug: Home page recommendations are missing when navigating to Youtube Home from a video


lfaucon:
When navigating to Youtube Home either using the Youtube Logo or the sidebar option (see screenshot), I find that Tournesol recommendations are not anymore included on the Home page as they should be.

<img width="985" alt="Screenshot 2022-02-27 at 10 39 19" src="https://user-images.githubusercontent.com/8818081/155877151-a9d1fb6c-a192-4737-a302-2137d00b4a80.png">

PR #809, opened 2022-04-13
Fix youtube recommendations not always displayed

sigmike:
Related issue: #628

The content script is not always injected when there's a page change on youtube. So if it was injected on a /watch URL and then not injected when returning to /, then the recommendations are empty and therefore not displayed.

The fix always loads the recommendations even on non root URL, so that the videos are available when returning to /.

This may increase the number of unnecessary requests, but it depends on how often there's a page load without a script injection. It seems to be the case on the majority of page change, so it should not be a problem.

Other possible solutions:

- revert d4ab2e2c1965b5219027d3a27a66af09ffd0f43e and possibly find a better solution to speed up loading
- in `process` load the recommendations if they have never been loaded (but that will reintroduce a delay before they're displayed)


amatissart (changes requested):
I see 2 problems with this change:

-  it does not fix all user flows:
   - Browsing to [a video page](https://www.youtube.com/watch?v=dQw4w9WgXcQ), then clicking on the Youtube logo:
   Tournesol recommendations are visible :heavy_check_mark: 
   - Browsing to [my subscriptions feed](https://www.youtube.com/feed/subscriptions), then clicking on the Youtube logo.
   Tournesol recommendations are not visible on the home page :x:
   There is probably a deeper issue with how the Tournesol recommendations are injected into the page.

- loading the Tournesol recommendations on every visit on Youtube seems wasteful:  
The requests triggered by the extension already represent a large part of the load on Tournesol API (our users seem to visit youtube.com more frequently than tournesol.app, what a shame :smile:).  
So adding more requests, including on visits where users are unlikely to see Tournesol recommendations anyway (e.g direct visit on a video page, or even embedded videos on other websites) does not sound like a good idea. Fetching the recommendations lazily, only when the home page is actually loaded seems more reasonable, even though it will introduce a small delay. After all Youtube navigation itself is not instant.

sigmike:
Ok I fixed the recommendations when coming from the subscriptions and made the change to request the recommendations only on the youtube root URL (with a small delay).

browser-extension/src/addTournesolRecommendations.js
```diff
@@ -225,11 +226,17 @@ function process() {
...
+    // If the content script is loaded on a non-root URL the recommendations
+    // are not loaded. So if the user then goes to the root URL and the content
+    // script is not reloaded, we need to load the recommendations.
+    loadRecommandations();
   }
 }
 
 function handleResponse({ data: videosReponse }) {
   areRecommandationsLoading = false;
+  areRecommendationsLoaded = true;
```
lfaucon, line 239: Optionally, a nice refactor here may be to keep track of `recommendationsRequestStatus` first equal to `"idle"`, `"loading"`, then `"loaded"`. So that we don't have the risk to forget to modify both `areRecommandationsLoading` and `areRecommendationsLoaded`

lfaucon (approved):
LGTM

amatissart (approved):

PR #812, opened 2022-04-14
[back] fix: possible email duplication

GresilleSiffle:
This pull request prevents users to create several account by using lower/upper variant of the same email address.

The front end doesn't enforce using only lower case email addresses as some providers might use case sensitive ones. 

I also added a `clean()` method on the `User` model so that we cannot create inconsistencies from the admin interface.

**to-do**
- [x] make it work for `/accounts/register/`
- [x] make it work for `/accounts/register-email/` 
- [x] translate the message 
- [x] add tests
- [x] make the admin UI aware of this

### DIFF
--- a/backend/core/locale/fr/LC_MESSAGES/django.po
+++ b/backend/core/locale/fr/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-04-05 15:08+0000\n"
+"POT-Creation-Date: 2022-04-14 09:39+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -22,14 +22,18 @@ msgstr ""
 msgid "email address"
 msgstr "adresse e-mail"
 
-#: core/serializers/user.py:18
+#: core/serializers/user.py:19
 #, python-format
 msgid "'%(name)s' is a reserved username"
 msgstr "'%(name)s' est un nom d'utilisateur reservé"
 
-#: core/serializers/user.py:20
+#: core/serializers/user.py:21
 msgid "'@' is not allowed in username"
 msgstr "Un nom d'utilisateur ne peut pas contenir '@'"
+
+#: core/serializers/user.py:39 core/serializers/user.py:59
+msgid "A user with this email address already exists."
+msgstr "Un utilisateur avec cet e-mail existe déjà."
 
 #: core/templates/accounts/register/body.html:5
 msgid "Please verify your Tournesol account by clicking on this link:"
--- a/backend/core/models/user.py
+++ b/backend/core/models/user.py
@@ -5,6 +5,7 @@ Defines Tournesol's User model and user preferences
 import logging
 
 from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
 from django.db.models import CheckConstraint, F, Func, Q, Value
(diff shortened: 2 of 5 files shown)


backend/core/serializers/user.py
```diff
@@ -21,26 +22,53 @@ def _validate_username(value):
...
+def validate_email_uniqueness_iexact(value: str) -> str:
+    """Raise an error if an email is considered already in use.
+
+    Even if emails are considered case-sensitive by the RFC 5321, several
+    email providers don't follow this specification. This validator ensures
+    users won't be able to create several accounts with the same address
+    by changing only the case.
+    """
+    try:
+        User.objects.get(email__iexact=value.lower())
```
amatissart, line 35: `.lower()` is redundant with "iexact", right ?
```suggestion
        User.objects.get(email__iexact=value)
```

backend/core/serializers/user.py
```diff
@@ -21,26 +22,53 @@ def _validate_username(value):
...
-    ])
+    email = EmailField(
+        validators=[
+            UniqueValidator(
+                queryset=User.objects.all(),
+                message=_("A user with this email address already exists."),
+                lookup="iexact",
+            ),
+        ]
+    )
```
amatissart, line 63: Cannot the same field definition be used in the `RegisterUserSerializer`?  
It would make the maintenance easier to use the same mechanism in both.

backend/core/locale/fr/LC_MESSAGES/django.po
```diff
@@ -22,15 +22,19 @@ msgstr ""
...
 msgstr "'%(name)s' est un nom d'utilisateur reservé"
 
-#: core/serializers/user.py:20
+#: core/serializers/user.py:21
 msgid "'@' is not allowed in username"
 msgstr "Un nom d'utilisateur ne peut pas contenir '@'"
 
+#: core/serializers/user.py:39 core/serializers/user.py:59
+msgid "A user with this email address already exists."
+msgstr "Un utilisateur avec cet e-mail existe déjà."
```
amatissart, line 36: Thanks for creating the missing translation :pray:

amatissart (commented):
The approach looks good to me as a first step. 
I only have minor comments.

backend/core/serializers/user.py
GresilleSiffle: I just realized it too, it's indeed useless.

backend/core/serializers/user.py
GresilleSiffle: Sure!

backend/core/serializers/user.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/812/commits/518bb64042d9a76db4f4e128cdf0146fe415484f.

backend/core/models/user.py
```diff
@@ -244,6 +245,17 @@ def ensure_email_domain_exists(self):
...
 
+    def clean(self):
+        value = self.email
+
+        similar = User.objects.filter(email__iexact=value)
+        n_similar = similar.count()
+
+        if n_similar > 1:
+            raise ValidationError(_("A user with this email address already exists."))
+        elif n_similar == 1 and similar[0].pk != self.pk:
```
amatissart, line 256: (minor)
This can be written in an slightly simpler way:
```py
similar_emails = User.objects.filter(email__iexact=self.email).exclude(pk=self.pk)
if similar_emails.exists():
    ...
```

In addition, this will trigger a single `.exists()` query, instead of 1 exact `.count()` + 1 fetch for instance `[0]`.

backend/core/models/user.py
```diff
@@ -244,6 +245,17 @@ def ensure_email_domain_exists(self):
...
+            raise ValidationError(_("A user with this email address already exists."))
```
amatissart, line 257: `ValidationError` accepts a dict to associate the error with a specific field in the form:
```suggestion
            raise ValidationError(
                {"email": _("A user with this email address already exists.")}
            )
```

amatissart (approved):

backend/core/models/user.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/812/commits/06d499b7747f8b3c93db04346e98525f6ca74f21.

backend/core/models/user.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/812/commits/06d499b7747f8b3c93db04346e98525f6ca74f21.

PR #813, opened 2022-04-17
chore: Upgrade minimist to >=1.2.6 in frontend and tests

lfaucon:
This was recommended as a critical security upgrade by dependabot

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -77,6 +77,7 @@
     "fetch-mock-jest": "^1.5.1",
     "husky": "^7.0.1",
     "i18next-parser": "^5.3.0",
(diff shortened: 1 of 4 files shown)


amatissart:
`minimist` has been upgraded to v1.2.6 in #713. This is already the version appearing in yarn.lock files. So this PR would have no effect, and `minimist` should not be listed as a direct dependency in package.json as it's only used by other packages.

The warning triggered by dependabot seems to be a false positive in this case. The message refers to "frontend/api/package-lock.json" which does not exist on "main".

lfaucon:
I was indeed confused that the security warning poped up, when we seem to have the target version already. Any idea why?

PR #814, opened 2022-04-17
[front] Improve uploaded date filter

aidanjungo:


frontend/public/locales/en/translation.json
```diff
@@ -494,4 +494,4 @@
       }
     }
   }
-}
+}
```
lfaucon, line 497: Not sure what happened here. It is usually prefered to leave an empty line at the end of files.

frontend/public/locales/en/translation.json
```diff
@@ -148,12 +148,12 @@
...
-    "thisMonth": "This month",
-    "thisYear": "This year",
-    "allTime": "All time",
-    "uploadDate": "Upload date",
+    "today": "A day",
+    "thisWeek": "A week",
+    "thisMonth": "A month",
+    "thisYear": "A year",
+    "allTime": "Always",
+    "uploadDate": "Upload since",
```
lfaucon, line 156: I think it will be clearer this way:

```suggestion
    "uploadDate": "Uploaded since",
```

frontend/public/locales/fr/translation.json
```diff
@@ -501,4 +501,4 @@
       }
     }
   }
-}
+}
```
lfaucon, line 504: Same comment

lfaucon (approved):
Approved with comments :) Thanks

frontend/public/locales/en/translation.json
```diff
@@ -148,12 +148,12 @@
...
-    "thisMonth": "This month",
-    "thisYear": "This year",
-    "allTime": "All time",
-    "uploadDate": "Upload date",
+    "today": "A day",
+    "thisWeek": "A week",
+    "thisMonth": "A month",
+    "thisYear": "A year",
+    "allTime": "Always",
+    "uploadDate": "Uploaded since",
```
amatissart, line 156: Probably not very important, but videos "uploaded since a week" doesn't sound like correct english to me. Contrary to something like "uploaded over the past week".

frontend/src/features/recommendation/RecommendationApi.tsx
```diff
@@ -9,7 +9,8 @@ export const getRecommendedVideos = async (
   const conversionTime = new Map();
   const params = new URLSearchParams(searchString);
   conversionTime.set('Any', 1);
-  conversionTime.set('Today', dayInMillisecondes);
+  // "Today" is 36 hours to get most of the videos from the previous day
+  conversionTime.set('Today', dayInMillisecondes * 1.5);
```
amatissart, line 13: Note that the publication datetime of a video is currently truncated to keep the date only (i.e the hours/minutes/seconds are not preserved). I don't see any reason for not keeping the exact publication date (maybe legacy reasons I am not aware of?) 

So it may also be relevant in another PR to update the metadata mechanism to save the exact datetime and to make this filter behave more consistently.

lfaucon: I see. I was not aware and I don't think there are any reason for this. That's part of the reason reason why we very rarely saw videos in the "Today" filtered list.

frontend/public/locales/en/translation.json
lfaucon: "Uploaded for a month/week/year" would be the translation proposed by Google translate: https://translate.google.ch/?sl=fr&tl=en&text=Mises%20en%20ligne%20depuis%20un%20mois&op=translate

frontend/src/features/recommendation/RecommendationApi.tsx
aidanjungo: Will be solve by #818

Issue #671, opened 2022-03-13
[backend] feat: endpoint to extract one contributor's cyclic inconsistencies


lfaucon:
It may happen that a contributor compares contents A, B and C on criteria X and submits that A > B, B > C and C > A. This is an impossibility. We should open an endpoint that helps contributors extract such inconsistencies in their comparisons to make it possible for them to correct their ratings.

The endpoint may be:

`/users/me/inconsistencies/{pollname}`

which returns a list of inconsistencies. Each inconsistencies should contain the name of the incorrectly Criteria, the three ComparisonCriteriaScores, the three Entities and their metadata.

glerzing:
I don't agree with the needs.

Contributors probably won't enjoy having to fix inconsistencies. This could feel a bit like a failure, and make them more careful with new comparisons, whereas we want it to be pleasant and effortless. I think this goes against user experience.

And inconsistencies are normal to human reasoning. It can probably even be a valuable information for machine learning, to be able to estimate how reliable are a contributor's ratings. And maybe there is a value in keeping the database "natural", and let any enterprise importing the database make their own pre-processing. But I am not a machine learning expert.

What are others' opinions on this?

Omicron666:
On the detection of inconsistencies themselves, you actually need to check for cycles. Not just 3 elements or you will miss many of them.
Also thinking about cycles is probably easier.

lfaucon:
After a groupd discussion, we agreed that it is important for a contributor to have access to his or her list of inconsistencies in order to correct them because having inconsistencies reduces the contributor's impact on the aggregated scores

lfaucon:
A second interesting type of inconsistencies is the differences between ContributorRatingCriteriaScores and the Comparison

@lenhoanglnh Can you provide the mathy formula to extract this type of inconsistencies?

glerzing:
I am quite lost with this feature.

For the cyclic inconsistencies, the easy, naive approach's complexity for 3-elements-cycles is in O(n³) => not great
The paragraph 3.2 of the paper https://www.researchgate.net/publication/338671588_Inconsistency_indices_for_incomplete_pairwise_comparisons_matrices
suggests a complicated formula for calculating cyclic inconsistencies in incomplete PC. It looks like its time complexity is in O(n²), and it should also work for cyclic inconsistencies of length > 3. Is that what you want me to implement?

[...]

PR #815, opened 2022-04-17
part 1/2 : score inconsistencies (next: cyclic inconsistencies) (#671)

glerzing:
The score inconsistencies measure the difference between the rating scores calculated by the machine learning and the comparisons.

I decided to allow low threshold values (even 0), but to limit the number of results in the response to 100 (even if the count can get higher) to limit the size of the response.

A potential issue us that the machine learning rating scores may not have been updated with the latest comparisons. This is quite problematic that the ContributorRatingCriteriaScores can't be updated directly after each modification, which would also allow for quick feedback, notably in the contributor recommendations.

The next part is on cyclic inconsistencies, which is trickier in terms of algorithms and computation time.

Feel free to suggest corrections or improvements.

### DIFF
--- a/backend/tournesol/serializers/inconsistencies.py
+++ b/backend/tournesol/serializers/inconsistencies.py
@@ -1,0 +1,42 @@
+from rest_framework import serializers
+from rest_framework.serializers import Serializer
+
+
+class ScoreInconsistencySerializer(Serializer):
+    """
+    Serializer for one element
+    """
+
+    inconsistency = serializers.FloatField()
+    entity_1_uid = serializers.CharField()
+    entity_2_uid = serializers.CharField()
+    criteria = serializers.CharField()
+    comparison_score = serializers.FloatField()
+    entity_1_rating = serializers.FloatField()
+    entity_2_rating = serializers.FloatField()
+
+
+class ScoreInconsistenciesSerializer(Serializer):
+    """
+    Returns the score inconsistencies and related data.
+    Used from a dictionary, not a queryset.
+    The count is exhaustive, but the results list is truncated at 100 elements.
+    """
+    mean_inconsistency = serializers.FloatField()
+    count = serializers.IntegerField()
+    results = ScoreInconsistencySerializer(many=True)
+
+
+class ScoreInconsistenciesFilterSerializer(serializers.Serializer):
+
+    inconsistency_threshold = serializers.FloatField(
+        default=5.0,
+        help_text="Comparisons with an inconsistency score above this threshold are "
+                  "considered inconsistent and listed in the response (max. 100 responses)",
+    )
+
+    date_gte = serializers.DateTimeField(
+        default=None,
+        help_text="Restrict the search to entities created or edited after this date"
+                  "Accepted formats: ISO 8601 datetime (e.g `2021-12-01T12:45:00`) ",
+    )
--- a/backend/tournesol/serializers/inconsistencies.py
+++ b/backend/tournesol/serializers/inconsistencies.py
@@ -6,37 +6,45 @@ class ScoreInconsistencySerializer(Serializer):
     """
     Serializer for one element
     """
-
     inconsistency = serializers.FloatField()
+    criterion = serializers.CharField()
     entity_1_uid = serializers.CharField()
     entity_2_uid = serializers.CharField()
-    criteria = serializers.CharField()
     comparison_score = serializers.FloatField()
     entity_1_rating = serializers.FloatField()
     entity_2_rating = serializers.FloatField()
 
 
-class ScoreInconsistenciesSerializer(Serializer):
+class ScoreInconsistenciesStatsSerializer(Serializer):
     """
-    Returns the score inconsistencies and related data.
-    Used from a dictionary, not a queryset.
-    The count is exhaustive, but the results list is truncated at 100 elements.
+    Return some statistics for each criterion,
+    to be able to compare the results for the different criteria
     """
     mean_inconsistency = serializers.FloatField()
+    inconsistent_comparisons_count = serializers.IntegerField()
+    comparisons_count = serializers.IntegerField()
+
+
+class ScoreInconsistenciesSerializer(Serializer):
+    """
+    Returns the score inconsistencies and some statistics.
+    Used from a dictionary, not a queryset.
+    """
     count = serializers.IntegerField()
     results = ScoreInconsistencySerializer(many=True)
+    stats = serializers.DictField(child=ScoreInconsistenciesStatsSerializer(), label="criterion")
 
 
-class ScoreInconsistenciesFilterSerializer(serializers.Serializer):
+class ScoreInconsistenciesFilterSerializer(Serializer):
 
     inconsistency_threshold = serializers.FloatField(
         default=5.0,
         help_text="Comparisons with an inconsistency score above this threshold are "
-                  "considered inconsistent and listed in the response (max. 100 responses)",
+                  "considered inconsistent and listed in the response (max. 100 responses).",
     )
 
     date_gte = serializers.DateTimeField(
         default=None,
-        help_text="Restrict the search to entities created or edited after this date"
-                  "Accepted formats: ISO 8601 datetime (e.g `2021-12-01T12:45:00`) ",
+        help_text="Restrict the search to entities created or edited after this date.\n"
+                  "Accepted formats: ISO 8601 datetime (e.g `2021-12-01T12:45:00`).",
     )
--- a/backend/tournesol/serializers/inconsistencies.py
+++ b/backend/tournesol/serializers/inconsistencies.py
@@ -10,9 +10,10 @@ class ScoreInconsistencySerializer(Serializer):
     criterion = serializers.CharField()
     entity_1_uid = serializers.CharField()
     entity_2_uid = serializers.CharField()
-    comparison_score = serializers.FloatField()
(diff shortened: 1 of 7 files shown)


backend/tournesol/views/inconsistencies.py
```diff
@@ -0,0 +1,213 @@
...
+            entity_1 = comparison_criterion["comparison__entity_1__uid"]
+            entity_2 = comparison_criterion["comparison__entity_2__uid"]
+            criterion = comparison_criterion["criteria"]
+            comparison_score = comparison_criterion["score"]
+
+            entity_1_score = None
+            entity_2_score = None
+            entity_1_uncertainty = None
+            entity_2_uncertainty = None
+            # TODO: after passing to Python 3.10, sort criteria_ratings and optimize with bisect
```
glerzing, line 102: I have let a TODO.  It would be much more efficient to sort the criteria_ratings list first and then use dichotomy, but bisect only offers the "key" parameter since python 3.10 (https://docs.python.org/3/library/bisect.html).

glerzing: This "key" parameter would allow to use dichotomy on both "criteria" and "contributor_rating__entity__uid" values of the dictionary (without the "key" parameter, bisect is not suited for dictionaries).

backend/tournesol/views/inconsistencies.py
```diff
@@ -0,0 +1,213 @@
...
+
+        min_rating_difference = base_rating_difference - uncertainty
+        max_rating_difference = base_rating_difference + uncertainty
+
+        if comparison_score <= -10:
+            min_inconsistency = inconsistency_calculation(min_rating_difference)
+        elif comparison_score >= 10:
+            min_inconsistency = inconsistency_calculation(max_rating_difference)
+        else:
+            root = comparison_score / sqrt(100 - comparison_score**2)
```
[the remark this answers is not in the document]
glerzing: 10 is indeed the max of a comparison, and 100 represents 10²

backend/tournesol/views/inconsistencies.py
```diff
@@ -0,0 +1,213 @@
...
+        3 - always decreasing, if c >= 10
+        So, the inconsistency is either i(R - U) in case 1, inconsistency(R + U)
+        in case 3, or 0 in case 2.
+        The value of R for which i(r) = 0 is C / sqrt(100 - C²), which
+        helps in case 2 to know if the minimum is 0 by checking if this root
+        is in [R - U, R + U].
+
+        There is also an imprecision of 0.5 on the comparison,
+        since the comparisons are made on floats and not integers,
+        which is subtracted on the result of the previous calculation.
```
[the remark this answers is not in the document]
glerzing: If you have a real preference of, say, 3.5001, you can chose the set a comparison score of 4. So, there is at most a 0.5 imprecision in the comparison score if it's rounded to the nearest integer.

The comparison_score has a linear effect on the inconsistency. Adding 0.5 to the comparison_score will either increase the inconsistency by 0.5, or decrease it by 0.5. But since the imprecision could be both ways and we want to minimize the inconsistency, 0.5 is subtracted to the inconsistency.

Maybe this needs to be rephrased.

backend/tournesol/views/inconsistencies.py
glerzing: yes, it's a better idea

backend/tournesol/views/inconsistencies.py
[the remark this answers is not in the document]
glerzing: In case someone is interested by how the inconsistency function varies **with the rating_difference**, here is a curve with uncertainty = 0 and comparison_score = -5
inconsistency
And here is how it would look like if there weren't the absolute value (and the imprecision of 0.5)
inconsistency_no_abs

backend/tournesol/serializers/inconsistencies.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class ScoreInconsistenciesSerializer(Serializer):
+    """
+    Returns the score inconsistencies and related data.
+    Used from a dictionary, not a queryset.
+    The count is exhaustive, but the results list is truncated at 100 elements.
+    """
+    mean_inconsistency = serializers.FloatField()
+    count = serializers.IntegerField()
```
glerzing, line 26: I realise that it may be more interesting to have the mean inconsistency and the count on a per_criterion basis. This way, you could see for which criteria the inconsistency is the highest. What do you think? Do you agree that I should change the API to have the mean_inconsistency and the count **per criterion**?

glerzing: Actually, that's what I am currently implementing with the cycles (the results are split per criterion).

backend/tournesol/serializers/inconsistencies.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class ScoreInconsistenciesSerializer(Serializer):
+    """
+    Returns the score inconsistencies and some statistics.
+    Used from a dictionary, not a queryset.
+    """
+    count = serializers.IntegerField()
+    results = ScoreInconsistencySerializer(many=True)
+    stats = serializers.DictField(child=ScoreInconsistenciesStatsSerializer(), label="criterion")
```
glerzing, line 35: DictFields are not used elsewhere. But it's quite practical for per-criterion stats.

backend/tournesol/views/inconsistencies.py
```diff
@@ -0,0 +1,216 @@
...
+                                 uncertainty) -> float:
+        """
+        Calculate the inconsistency between the comparison
+        criterion score and the general rating of the entity.
+
+        Let's note R the rating difference (rating2 - rating1),
+        r its variable counterpart, U the uncertainty and C the
+        comparison (positive if the 2nd entity is preferred).
+        Consider this function i:
+        i(r) = |C - 10r / sqrt(r² + 1)| (cf issue #671 discussion)
```
lfaucon, line 169: Do you think it may be useful for the users to also include in the API response the value of `10r / sqrt(r² + 1)` and show it to the users as an "expected comparison score"?

lfaucon (approved):
I made a few comments that are worth to address before merging, otherwise I really like it. I look forward to seeing this in action

backend/tournesol/views/inconsistencies.py
glerzing: It would indeed be problematic to tell contributors that their comparison is wrong without indicating an expected score.
But on the other side, it would incentivize them to just set their comparison to the expected value.
Maybe we could, on the frontend, highlight in green on the cursor a range of comparison values that give an acceptable inconsistency (for example, [round(expected score) - 4  , round(expected_score) + 4]).
Another issue is that changing the comparison value will change the rating and thus the expected value. This makes it hard to give an exact range.

in short: I suggest for the backend to indeed return the expected value, and the frontend will convert it to a range of "acceptable" values. It is hard to calculate an exact value for the size of this range, so this can be arbitrary.

backend/tournesol/views/inconsistencies.py
```diff
@@ -0,0 +1,216 @@
...
+        is in [R - U, R + U].
+
+        There is also an imprecision of 0.5 on the comparisons,
+        since the comparisons are made on integers (considering that contributors
+        have to "mentally round" their preferences to the nearest integer).
+        This imprecision is subtracted on the result of the previous calculation.
+        """
+
+        base_rating_difference = entity_2_calculated_rating - entity_1_calculated_rating
+        comparison_max = 10
```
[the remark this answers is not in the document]
glerzing: Ok to modify that.
I also think constants.py should be in tournesol/utils, not in core/utils.
Besides, I think the MAX_VALUE for ratings should be renamed RATING_MAX, put in constants.py and have its comment fixed (it is not between 0 and 100, but between -100 and 100). I can change that in this PR if you agree.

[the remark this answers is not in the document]
glerzing: Besides, the score inconsistencies would be more relevant if, whenever you make a new comparison, the personal and global ratings for this entity are updated immediately. If possible, we should try to enable the machine learning for a single entity, and for a single (user/entity) pair, otherwise we could face numerous small issues.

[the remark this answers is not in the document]
lfaucon: > I think the MAX_VALUE for ratings should be renamed RATING_MAX

This one is never used. I think it is legacy code that instead should be removed.

> I also think constants.py should be in tournesol/utils, not in core/utils. This would be better in a specific PR, as it requires to modify multiple files.

Yes, or we could keep two `constants.py` files in the respective apps with the constants that they use most. But as of now all constants are related to the `tournesol` app

> Besides, the score inconsistencies would be more relevant if, whenever you make a new comparison, the personal and global ratings for this entity were updated immediately. If possible, we should try to enable the machine learning for a single entity, and for a single (user/entity) pair, otherwise we could face numerous small issues.

Very good point to raise this here! We already have implemented a version of the algorithm that updates a single user's ratings when comparisons are submitted. @lenhoanglnh @amatissart what is the timeline and blockers to switch the video poll to Mehestan?

backend/tournesol/views/inconsistencies.py
lfaucon: > I suggest for the backend to indeed return the expected value, and the frontend will convert it to a range of "acceptable" values.

That would be great. I see you have already implemented the change! Thanks

lfaucon (approved):

PR #817, opened 2022-04-18
refactor{frontend} change score display and condition of display in V…

benjaminlebigot:
refactor{frontend} change score display and condition of display in VideoCardScores.tsx

made a function to display score instead of multiply the result by ten each time. You can find this function in utils/criteria.ts.
the ratedHigh is displayed if max_score is higher than 0
the ratedLow is displayed if min_score is lower than 0


lfaucon (approved):
LGTM

amatissart (approved):

PR #825, opened 2022-04-19
WIP / [back] fix: possible email duplication using the `+` symbol

GresilleSiffle:


backend/core/locale/fr/LC_MESSAGES/django.po
```diff
@@ -18,23 +18,32 @@ msgstr ""
...
 
-#: core/models/user.py:28
+#: core/models/user.py:29
 msgid "email address"
 msgstr "adresse e-mail"
 
-#: core/serializers/user.py:19
+#: core/models/user.py:278
+#, fuzzy, python-format
+#| msgid "A user with this email address already exists."
```
amatissart, line 27: this comment about a fuzzy string can be removed
```suggestion
```

backend/core/models/user.py
```diff
@@ -227,21 +227,78 @@ def trusted_users(cls) -> QuerySet["User"]:
...
+            - BOB+HELLO@example.org
+            - etc.
+        """
+        email_split = email.split("@")
+
+        # if there is no `@`, do nothing
+        if len(email_split) == 1:
+            return email
+
+        local_part = email_split[0]
```
amatissart, line 260: Theoretically the local part may contain one or multiple `@`. So `.rsplit("@", 1)` should be used instead to get the local part.

backend/core/models/user.py
```diff
@@ -227,21 +227,78 @@ def trusted_users(cls) -> QuerySet["User"]:
...
+
+        if users.exists():
+            raise ValidationError(
+                {
+                    "email": _(
+                        "A user with an email starting with '%(email)s' already exists"
+                        " in this domain." % {"email": f"{local_part_split[0]}+"}
+                    )
+                }
+            )
```
amatissart, line 282: After verification, using the Django `ValidationError` is fine and is correctlly handled by DRF to format the error, and make it correctly rendered on frontend :+1:

It's also possible to call this validation in `validate_email` instead of `validate`. In this case the ValidationError needs to receive the error string only (without the "email" key).

GresilleSiffle: Great!

Removing the `email` key is fine, but as the email validation is also called in `User.clean`, the error message won't be displayed near the email field in the administration interface.

To keep the message near the email field, it will required either **(1)** to keep the validation in the general `validate` method of the serializers ; or **(2)** to move the exception raising outside of the `validate_email_unique_with_plus` method, so that each parent methods will be able to format the exception correctly ; or **(3)** or keep it as is, remove the key `email` in the exception, and from the `User.email` method catch and re-raise an exception to add the missing email field.

amatissart: I would find option **`(3)`** slightly more explicit, but either (1) or (3) looks good to me.

backend/core/models/user.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/825/commits/d3e35e798b2827a6a37902a4ee837f4a1d4f9ac8.

backend/core/models/user.py
GresilleSiffle: I implemented **(3)**.

Fixed by https://github.com/tournesol-app/tournesol/pull/825/commits/5b07b80a1a8c0b08d2b97e3a498207786bba733e.

backend/core/models/user.py
```diff
@@ -227,21 +227,76 @@ def trusted_users(cls) -> QuerySet["User"]:
...
+        users = User.objects.filter(
+            Q(email__iexact=f"{local_part_split[0]}@{email_split[-1]}")
+            | (
+                Q(email__istartswith=f"{local_part_split[0]}+")
+                & Q(email__iendswith=f"@{email_split[-1]}")
+            ),
+        )
+
+        if username:
+            users.exclude(username=username)
```
amatissart, line 272: I had missed that. The queryset is not updated here. 
```suggestion
        if username:
            users = users.exclude(username=username)
```

amatissart (changes requested):

backend/core/models/user.py
GresilleSiffle: I always do the same mistakes.

amatissart (approved):

Issue #802, opened 2022-04-08
[front] feat: Enable displaying personal criteria scores in criteria scores charts


lfaucon:
Create a wrapper that enables to display `CriteriaBarChart` or `CriteriaRadarChart` with personnal scores instead of the scores from a video. This could be achieved with a simple toggle or checkbox above the chart.

Optionally it would be best if we are able to display double bars or two Radars to makes comparing one's scores with the community score. For example with the below design:

https://recharts.org/en-US/api/RadarChart
<img width="303" alt="Screenshot 2022-04-17 at 14 47 28" src="https://user-images.githubusercontent.com/8818081/163715087-2064cc44-9f24-4915-ad65-14479772f20e.png">

PR #828, opened 2022-04-19
Add checkbox to display personal criteria scores

sigmike:
Related issue: #802 

image

Remaining to do:

- [x] Position checkbox and add label
- [x] Enable checkbox only when logged in
- [x] Refactor common parts of Radar and Bar charts
- [x] Change the color or style of the personal scores on the bar chart?

### DIFF
--- a/frontend/src/components/CriteriaBarChart.tsx
+++ b/frontend/src/components/CriteriaBarChart.tsx
@@ -12,19 +12,13 @@ import {
 import {
   VideoSerializerWithCriteria,
   Recommendation,
-  EntityCriteriaScore,
 } from 'src/services/openapi';
 import { useCurrentPoll } from 'src/hooks/useCurrentPoll';
 import { displayScore, criteriaIcon } from 'src/utils/criteria';
-import { PRESIDENTIELLE_2022_POLL_NAME } from 'src/utils/constants';
-
-const BAR_CHART_CRITERIA_SCORE_MIN = -1;
-const BAR_CHART_CRITERIA_SCORE_MAX = 1;
-
-const between = (a: number, b: number, x: number | undefined): number => {
-  // clips x between a and b
-  return Math.min(b, Math.max(a, x || 0));
-};
+import useCriteriaChartData, {
+  CriteriaChartDatum,
+} from 'src/hooks/useCriteriaChartData';
+import { useTranslation } from 'react-i18next';
 
 const criteriaColors: { [criteria: string]: string } = {
   reliability: '#4F77DD',
@@ -39,17 +33,38 @@ const criteriaColors: { [criteria: string]: string } = {
   default: '#506ad4',
 };
 
+// Copied from https://stackoverflow.com/a/59387478
+const lighten = (color: string, lighten: number) => {
+  const c = color.replace('#', '');
+  const rgb = [
+    parseInt(c.substr(0, 2), 16),
+    parseInt(c.substr(2, 2), 16),
+    parseInt(c.substr(4, 2), 16),
+  ];
+  let returnstatement = '#';
+  rgb.forEach((color) => {
+    returnstatement += Math.round(
+      (255 - color) * (1 - Math.pow(Math.E, -lighten)) + color
+    ).toString(16);
+  });
+  return returnstatement;
+};
+
 const SizedBarChart = ({
-  criteriaScores,
+  data,
+  personalScoresActivated,
+  domain,
   width,
   height,
 }: {
-  criteriaScores: Array<EntityCriteriaScore>;
+  data: CriteriaChartDatum[];
+  personalScoresActivated: boolean;
+  domain: number[];
   width?: number;
   height?: number;
 }) => {
-  const { getCriteriaLabel, options, name: pollName } = useCurrentPoll();
-  const mainCriterionName = options?.mainCriterionName ?? '';
+  const { getCriteriaLabel } = useCurrentPoll();
+  const { t } = useTranslation();
 
   const renderCustomAxisTick = ({
     x,
@@ -73,7 +88,7 @@ const SizedBarChart = ({
         xmlns="http://www.w3.org/2000/svg"
       >
         <style>{'.emoji { font-size: 42px; fill: black; }'}</style>
-        <rect x="0" y="15" width="60" height="60" fill="white" />
+        <rect x="0" y="0" width="60" height="60" fill="white" />
         {emoji ? (
           <text x="9" y="9" dominantBaseline="hanging" className="emoji">
             {emoji}
@@ -86,39 +101,11 @@ const SizedBarChart = ({
     );
   };
 
-  const shouldDisplayChart = criteriaScores && criteriaScores.length > 0;
-
-  if (!shouldDisplayChart) {
-    return <div></div>;
-  }
-
-  const data = criteriaScores
-    .filter((s) => mainCriterionName !== s.criteria)
-    .map((s) => {
-      // TODO: below is a bad hack necessary because scores on multiple
-      // polls are not scaled in the same way. We should have a simpler
-      // manner to address this.
-      const _score = s.score || 0;
-      const clipped_score = between(
-        BAR_CHART_CRITERIA_SCORE_MIN,
-        BAR_CHART_CRITERIA_SCORE_MAX,
-        pollName === PRESIDENTIELLE_2022_POLL_NAME ? 4 * _score : _score
-      );
-      return {
-        ...s,
-        clipped_score,
-      };
-    });
-
   return (
     <BarChart layout="vertical" width={width} height={height} data={data}>
-      <XAxis
-        type="number"
-        domain={[BAR_CHART_CRITERIA_SCORE_MIN, BAR_CHART_CRITERIA_SCORE_MAX]}
-        hide={true}
-      />
+      <XAxis type="number" domain={domain} hide={true} />
 
-      <Bar dataKey="clipped_score" barSize={20}>
+      <Bar dataKey="clippedScore" barSize={20}>
         {data.map((entry) => (
           <Cell
             key={entry.criteria}
@@ -126,6 +113,19 @@ const SizedBarChart = ({
           />
         ))}
       </Bar>
+      {personalScoresActivated && (
+        <Bar dataKey="clippedPersonalScore" barSize={20}>
+          {data.map((entry) => (
+            <Cell
+              key={entry.criteria}
+              fill={lighten(
+                criteriaColors[entry.criteria] || criteriaColors['default'],
+                0.5
+              )}
+            />
+          ))}
+        </Bar>
+      )}
       <YAxis
         type="category"
         dataKey="criteria"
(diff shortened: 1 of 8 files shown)


lfaucon:
@sigmike What do you see in the Django Admin ? You can access it at http://localhost:8000/admin if you run the dev-env. That should answer your question

amatissart:
> @lfaucon, @GresilleSiffle or @amatissart: How can I get data from `UsersService.usersMeContributorRatingsRetrieve` in dev-env? I always get an empty array in the `criteria_scores`. Is there a uid+user that has personal scores?

The user `user` is a superuser created specially for development purposes, but without any comparison or scores by default. You should use one of `user1`, `user2`, ... `user5` to retrieve contributor scores.

lfaucon:
> @lfaucon, @GresilleSiffle or @amatissart: How can I get data from `UsersService.usersMeContributorRatingsRetrieve` in dev-env? I always get an empty array in the `criteria_scores`. Is there a uid+user that has personal scores?

Your question also makes me realize that the toggle button for showing personnal scores should only appear when there exists such personnal scores. Otherwise a text saying "You have not compared this video" should be displayed.

sigmike:
Ok I found videos with personal scores. But some videos have been compared and still have an empty `criteria_scores` in `UsersService.usersMeContributorRatingsRetrieve` (for example in dev-env http://127.0.0.1:3000/video/VKsekCHBuHI with `user1`). Maybe there isn't *enough* comparisons? If that's the reason we should change the message "You have not compared this video".

Besides that I think the PR is complete. Here's what it looks like:
image

We can probably improve the colors and transparency on the radar chart when the personal scores are displayed, but I'm not sure how.

When the user cannot display the personal scores because they are not logged in or have no personal criteria scores, the checkbox is disabled and a tooltip is displayed.

frontend/src/components/CriteriaBarChart.tsx
```diff
@@ -123,10 +129,15 @@ const CriteriaBarChart = ({ video }: Props) => {
         content={(props: TooltipProps<number, string>) => {
           const payload = props?.payload;
           if (payload && payload[0]) {
-            const { criteria, score } = payload[0].payload;
+            const { criteria, clippedScore, personalScore } =
+              payload[0].payload;
             return (
               <pre>
-                {getCriteriaLabel(criteria)}: {displayScore(score)}
+                {getCriteriaLabel(criteria)}: {displayScore(clippedScore)}
```
lfaucon, line 136: I think we want to display the true score here, not the clippedScore. The score is clipped only for display... Now that I think of it, I'm not absolutely sure about what we exactly want. Maybe it would help if we had som indications that bars are clipped for very good scores?

Otherwise to be the most honest we could display something like:

"Displayed as 10.0; Real score is 17.4"

frontend/src/components/PersonalScoreChartWrapper.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const PersonalScoreChartWrapper = ({
+  video,
+  children,
+}: {
+  video: VideoSerializerWithCriteria;
+  children: React.ReactNode;
+}) => {
+  return (
+    <>
```
lfaucon, line 67: Is this `React.Fragment` needed?

frontend/src/hooks/useCriteriaChartData.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    return criteriaScores
+      .filter((s) => s.criteria != 'largely_recommended')
+      .map(({ score, criteria }) => {
+        const clippedScore = between(SCORE_MIN, SCORE_MAX, score);
+        return {
+          criteria,
+          clippedScore,
+          personalScore: personalScoreByCriteria[criteria],
+        };
```
lfaucon, line 43: We either need to:
1. Fix the scale (from -1 to 1) on the BarChart and Radar charts. In this case we could even not have clippedScores.
2. Have `clippedPersonalScore` added in the data for the display

Otherwise when have a personnal score > 1.0 (displayed as >10.0) we obtain for example:

Screenshot 2022-04-22 at 10 30 22


It is important that the scale is fixed such that all the BarCharts and RadarChart from one video to another are easily comparable. It is not a worry that we lose the information when the score if >1

frontend/src/components/CriteriaBarChart.tsx
```diff
@@ -37,8 +31,28 @@ const criteriaColors: { [criteria: string]: string } = {
...
+    parseInt(c.substr(4, 2), 16),
+  ];
+  let returnstatement = '#';
+  rgb.forEach((color) => {
+    returnstatement += Math.round(
+      (255 - color) * (1 - Math.pow(Math.E, -lighten)) + color
+    ).toString(16);
+  });
+  return returnstatement;
+};
```
lfaucon, line 49: I really like the choice of colors here! Great pick

lfaucon (changes requested):
I only did a partial review. I'll need to learn about context providers to do a more complete review.

frontend/src/components/CriteriaBarChart.tsx
sigmike: I changed back to displaying the non clipped score.

frontend/src/components/PersonalScoreChartWrapper.tsx
sigmike: No, I removed it.

frontend/src/hooks/useCriteriaChartData.ts
sigmike: I added a `clippedPersonalScore` and used it as value.

How can I get a video with a personal score > 1.0?

sigmike:
I tried to improve the radar colors:
image

Yellow is Tournesol score. I tried blue for the user score but yellow and blue do not mix well.

And I made the charts work better on smaller screens in 5561f38939c5f5a8b5ebfdfa538570b569cfee68 (though it may be better to do that in another PR).

frontend/src/hooks/useCriteriaChartData.ts
lfaucon: > How can I get a video with a personal score > 1.0?

If you have access to the Django admin, you can manually change the ContributorRatingCriteriaScore in it.

lfaucon:
> And I made the charts work better on smaller screens in https://github.com/tournesol-app/tournesol/commit/5561f38939c5f5a8b5ebfdfa538570b569cfee68 (though it may be better to do that in another PR).

This is indeed a very welcome change! Let's make it in a separated PR, unless you think it will cause conflicts?

sigmike:
There will be some conflicts but they can be solved (and we still have the commit that applies to this PR). I made another PR in https://github.com/tournesol-app/tournesol/pull/847 and reverted the change here.

frontend/src/components/CriteriaBarChart.tsx
```diff
@@ -153,7 +163,7 @@ const CriteriaBarChart = ({ video }: Props) => {
   // ResponsiveContainer adds the width and height props to its child component.
   // We need the width to position the icons.
   return (
-    <ResponsiveContainer width="100%" aspect={1}>
+    <ResponsiveContainer width="100%" height={500}>
```
sigmike, line 166: @lfaucon I switched back to a fixed height because otherwise on some screen size the icon is cropped.

frontend/src/pages/videos/VideoAnalysis.tsx
```diff
@@ -28,11 +33,16 @@ function VideoAnalysis() {
         <Grid item xs={12}>
           <VideoCard video={video} showPlayer={false} />
         </Grid>
-        <Grid item xs={12} md={6}>
-          <Card>
-            <CriteriaBarChart video={video} />
-          </Card>
-        </Grid>
+        {shouldDisplayCharts && (
+          <Grid item xs={12}>
```
sigmike, line 37: I made the chart full width because I found it looked weird with only half the width of the player. But we can change it back if you prefer (`md={6}`).

lfaucon: Looks good to me. We will reorganize the layout of this page as we add a few elements to it soon.

Fyi @adrausch

frontend/src/components/CriteriaBarChart.tsx
sigmike: I changed that in d03d117767202a8c519ae67291ecae602cebe0c0 because the height should be calculated from the criteria count. This change could also be a separate PR though.

frontend/src/pages/videos/VideoAnalysis.tsx
sigmike: Actually I reverted this change in 569d6c27f063863512e2a445648a3f7ee71ba781 since it's not related to the issue addressed here.

frontend/src/components/CriteriaBarChart.tsx
sigmike: These changes have been made in `main` (a50885df52296d75ae1d7a73994891a4c819d860).

GresilleSiffle:
Currently reviewing and rebasing on `main`...

GresilleSiffle (commented):
LGTM

I don't know if @lfaucon you want to have a final look? I'm not that familiar with the codes organized with hooks (I still need to learn : ).

Maybe the thing that would have helped me is more comments in the new hooks, like a documentation string for each exported component. Theses documentations can answer the questions: when to use this hook, what are the limitations, etc.

 For instance,  did the `usePersonalCriteriaScores` work only with the `videos` poll? If yes it should be described.

If it helps, here is a hook with documented functions I found very useful:
https://github.com/tournesol-app/tournesol/blob/main/frontend/src/hooks/useNotifications.ts

sigmike:
The `useCriteriaChartData` hook is really just a way to refactor what's common between the bar and radar criteria charts. If we don't plan to use multiple criteria charts anymore we can probably remove it. It's a hook because it uses hooks internally. I added an explicit interface with some comments.

The `usePersonalCriteriaScores` is just a wrapper around a context. I made a context to separate the state from the component because it had to affect 2 charts and we may want to checkbox to be distant from the charts (we may want it above the titles to also change the distribution chart for example). I made it a custom hook instead of directly using `useContext` so that it would be easy to change the internal behavior without changing its usage. For example we may want to store this state in redux instead of a context (since redux is barely used in the app I actually forgot it was there). The names of the hook's return value are pretty explicit so I'm not sure adding comments on them would be helpful.

I can add more comments about how it should be used, but it's not really a generic hook like the one you mentioned.

> For instance, did the `usePersonalCriteriaScores` work only with the `videos` poll? If yes it should be described.

It should work on any poll as long as `usersMeContributorRatingsRetrieve` also works (its description mentions videos explicitly). It only uses the `video` prop to get the `uid` for this request. I made the change to take a `uid` instead of a `video` so that it matches the API. I tried to add it to the candidate result page and it seems to work (but my dev-env database is not complete enough to be sure, I had to fake the API result).

Some messages talk about video though (for example "You have not compared this video yet"). We could say "entity" instead but that doesn't seem very use friendly. A different message per poll may be better, possibly with a generic fallback.

amatissart (approved):

Issue #833, opened 2022-04-20
epic: Candidates final ranking page


amatissart:
- [ ] Use `SearchFilter` with following options
  - "sort by criteria" dropdown.
  - same filter "?score_mode" as in #830 
- [ ] Use `EntityCard` with scores for each candidate, in a new `EntityList` component
  - Show "deviation" in addition to "Tournesol score"? How to represent it?

GresilleSiffle:
Should we make the recommendation page able to display any type of entity?

GresilleSiffle:
Here is how I see the plan:

**step 1**

Create a `RecommendationPage` that is an exactly a copy of `VideoRecommendationPage`.

Make it work with the poll API instead of the video API, and replace all occurrences of videos by entities in this new file.

**step 2**

Check how this page behaves with the candidate entity type and how much work is required to transform the `VideoList` into an `EntityList`.

An `EntityList` could be a wrapper around `VideoList` and `CandidateList`, selecting the right component according to the currently selected poll.

The list should use `EntityCard`.

**step 3**

Reconfigure the filters.

**step 4**

More?

PR #838, opened 2022-04-21
[front] refactor: reco. page uses polls API instead of the legacy video

GresilleSiffle:
**related to** #833 

---

First step of the issue #833.

The recommendation page now uses the new poll API instead of the legacy video API to retrieve the recommended entities.

It still only works with videos for now, but I'll make it even more generic in the next PR. I also want to update `getRecommendedVideos` to make it work with any type of entity.

**bonus**

There is now a `videoWithCriteriaFromRecommendation` function able to build a `VideoSerializerWithCriteria` object from a `Recommendation` object. Useful to rework the high level components, like the recommendation page, which are using child components that still use legacy video types.  

The limit parameter of the recommendation API is now a parameter of the `getRecommendedVideos` function.

### DIFF
--- a/frontend/src/pages/recommendations/RecommendationPage.tsx
+++ b/frontend/src/pages/recommendations/RecommendationPage.tsx
@@ -34,7 +34,7 @@ function RecommendationsPage() {
     results: [],
   };
 
-  const [entities, setVideos] = useState(prov);
+  const [entities, setEntities] = useState(prov);
   const entitiesCount = entities.count || 0;
 
   const searchParams = useMemo(
@@ -74,7 +74,7 @@ function RecommendationsPage() {
 
     const fetchVideos = async () => {
       setIsLoading(true);
-      setVideos(
+      setEntities(
         (await getRecommendedVideos(limit, location.search, criterias)) || []
       );
       setIsLoading(false);
--- a/frontend/src/pages/recommendations/RecommendationPage.tsx
+++ b/frontend/src/pages/recommendations/RecommendationPage.tsx
@@ -3,7 +3,10 @@ import { useLocation, useHistory } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import { Box } from '@mui/material';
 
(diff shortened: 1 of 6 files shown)


frontend/src/pages/recommendations/RecommendationPage.tsx
```diff
@@ -16,26 +19,32 @@ import {
...
+  const prov: PaginatedRecommendationList = {
     count: 0,
     results: [],
   };
-  const [videos, setVideos] = useState(prov);
-  const [isLoading, setIsLoading] = useState(true);
-  const location = useLocation();
-  const history = useHistory();
+
+  const [entities, setVideos] = useState(prov);
```
lfaucon, line 37: Is it hard to rename `setVideos` into `setEntities` here?

lfaucon (approved):
Works for me and the coode looks good 👍

frontend/src/utils/entity.ts
```diff
@@ -17,6 +22,20 @@ export const videoFromRelatedEntity = (entity: RelatedEntityObject): Video => {
...
 
+export const videoWithCriteriaFromRecommendation = (
+  entity: Recommendation
+): VideoSerializerWithCriteria => {
+  const video = videoFromRelatedEntity(entity);
+
+  return {
+    ...video,
+    criteria_scores: entity.criteria_scores,
+    tournesol_score: entity.tournesol_score || null,
```
amatissart, line 33: 0 could be transformed into null here.
```suggestion
    tournesol_score: entity.tournesol_score ?? null,
```

amatissart (commented):
It looks good to me, except a minor issue in `videoWithCriteriaFromRecommendation`

frontend/src/pages/recommendations/RecommendationPage.tsx
GresilleSiffle: No, I just forgot to update the setter name. Good catch.

PR #839, opened 2022-04-21
[back] feat: recommendations can be filtered by metadata

GresilleSiffle:
**related to** #833 

will unlock #838 

---

The objective is to upgrade the poll API recommendations endpoint so that it provides as much features as the video API recommendations.

This should ease the development of a recommendations page in the front end that works with different entity type.

**to-do**
- [x] make it possible to filter by single value metadata `metadata[name]=hello`
- [x] make it possible to filter by multiple values metadata `metadata[language]=fr,pt` 
- [x] add tests for the new metadata filter

### DIFF
--- a/backend/tournesol/entities/base.py
+++ b/backend/tournesol/entities/base.py
@@ -34,15 +34,9 @@ class EntityType(ABC):
     @classmethod
     def filter_metadata(cls, qst, filters):
         for key, values in filters:
-
             # add several OR expressions if multiple values are provided
             if len(values) > 1:
-                expression = Q(**{"metadata__" + key: values[0]})
-
-                for val in values[1:]:
-                    expression = expression | Q(**{"metadata__" + key: val})
-
-                qst = qst.filter(expression)
+                qst = qst.filter(**{"metadata__" + key + "__in": values})
             else:
                 qst = qst.filter(**{"metadata__" + key: values[0]})
 
--- a/backend/tournesol/entities/base.py
+++ b/backend/tournesol/entities/base.py
@@ -33,7 +33,6 @@ class EntityType(ABC):
     @classmethod
     def filter_metadata(cls, qst, filters):
         for key, values in filters:
-            # add several OR expressions if multiple values are provided
             if len(values) > 1:
                 qst = qst.filter(**{"metadata__" + key + "__in": values})
             else:
--- a/backend/tournesol/entities/base.py
+++ b/backend/tournesol/entities/base.py
@@ -2,6 +2,7 @@ import logging
 from abc import ABC, abstractmethod
 from typing import Type
 
+from django.db.models import Q
 from django.utils import timezone
 from django.utils.functional import cached_property
 from rest_framework.serializers import Serializer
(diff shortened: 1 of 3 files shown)


backend/tournesol/entities/base.py
```diff
@@ -30,6 +31,24 @@ def filter_date_lte(cls, qs, dt):
     def filter_date_gte(cls, qs, dt):
         return qs.filter(add_time__gte=dt)
 
+    @classmethod
+    def filter_metadata(cls, qst, filters):
+        for fil in filters:
+            key, values = fil
```
lfaucon, line 37: Why not:

```suggestion
        for key, values in filters:
```

backend/tournesol/entities/base.py
```diff
@@ -30,6 +31,24 @@ def filter_date_lte(cls, qs, dt):
...
+
+            # add several OR expressions if multiple values are provided
+            if len(values) > 1:
+                expression = Q(**{"metadata__" + key: values[0]})
+
+                for val in values[1:]:
+                    expression = expression | Q(**{"metadata__" + key: val})
+
+                qst = qst.filter(expression)
```
lfaucon, line 46: Would an `__in` operation work here out of the box, instead of the multiple `|` ("or") operations?

backend/tournesol/views/polls.py
```diff
@@ -77,6 +102,15 @@ def filter_by_parameters(self, request, queryset, poll: Poll):
         if date_gte:
             queryset = poll.entity_cls.filter_date_gte(queryset, date_gte)
 
+        metadata_filters = [
+            (self._metadata_from_filter(key), value)
+            for (key, value) in request.query_params.lists()
```
lfaucon, line 107:
```suggestion
            (self._metadata_from_filter(key), values)
            for (key, values) in request.query_params.lists()
```

backend/tournesol/views/polls.py
```diff
@@ -55,6 +74,12 @@ class PollRecommendationsBaseAPIView(PollScopedViewMixin, ListAPIView):
     It doesn't define any serializer, queryset nor permission.
     """
 
+    def _metadata_from_filter(self, filtr: str):
+        """
+        _metadata_from_filter("metadata[language]") -> "language"
+        """
+        return filtr.split("[")[1][:-1]
```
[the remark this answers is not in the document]
GresilleSiffle: No, I didn't think about it.

I used `metadata[uploader]` because in the poll API we are already using this syntax to filter by criteria ( `weights[better_habits]`), and I didn't want to add a new behaviour here.

backend/tournesol/entities/base.py
GresilleSiffle: I'm going to test. It's possible that the generated query runs faster with `IN` rather than with `OR`.

backend/tournesol/views/polls.py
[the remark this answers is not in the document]
lfaucon: Ok! Thanks for the detailed explanation


What do you think of making effort such that query parameters in the frontend urls fit more closely to query parameters of the API ?

backend/tournesol/entities/base.py
```diff
@@ -30,6 +31,17 @@ def filter_date_lte(cls, qs, dt):
...
         return qs.filter(add_time__gte=dt)
 
+    @classmethod
+    def filter_metadata(cls, qst, filters):
+        for key, values in filters:
+            # add several OR expressions if multiple values are provided
+            if len(values) > 1:
+                qst = qst.filter(**{"metadata__" + key + "__in": values})
+            else:
+                qst = qst.filter(**{"metadata__" + key: values[0]})
```
lfaucon, line 41:
```suggestion
            qst = qst.filter(**{"metadata__" + key + "__in": values})
```

backend/tournesol/views/polls.py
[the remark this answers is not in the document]
GresilleSiffle: I would say it's a good idea.

We should check if the syntax `weights[better_habits]` is the default behaviour of DRF or if this syntax is enforced by the OpenAPI specifications. If yes we might want to update the front end instead of the back end.

backend/tournesol/entities/base.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/commit/7d1b3f0f460eda684594092cd2f639b7457a6c3f.

lfaucon (approved):
Works for me

PR #846, opened 2022-04-22
[Dev-Env] Ignore Files from Docker in Dev-Env

adrausch:
Files created by Docker were not ignored by Git in the Dev-Env.
Add in gitignore file their paths.

### DIFF
--- a/dev-env/.gitignore
+++ b/dev-env/.gitignore
@@ -1,1 +1,3 @@
 /db-data
+.dockerignore
+Dockerfile


amatissart (approved):
I am ok with this change. 

Out of curiosity, do you know at what moment exactly are these files created? I find it especially surprising that a `Dockerfile` would be created automatically when running the dev env.

adrausch:
> I am ok with this change.
> 
> Out of curiosity, do you know at what moment exactly are these files created? I find it especially surprising that a `Dockerfile` would be created automatically when starting the dev env.

I never use Docker, so when I had an error on the installation of docker-compose on my macOS laptop, I saw in the documentation (https://docs.docker.com/compose/install/) that Docker recommanded (of course!) their software "Docker Desktop for Mac". Therefore I chose to install this software.

So I think these files were created by "Docker Desktop".

lfaucon (approved):

PR #847, opened 2022-04-22
Improve charts on smaller screens

sigmike:
Related comment: https://github.com/tournesol-app/tournesol/pull/828#issuecomment-1106376425

The charts are now responsive and the same size. They work well down to a window width of 420px.

image

image

### DIFF
--- a/frontend/src/components/CriteriaBarChart.tsx
+++ b/frontend/src/components/CriteriaBarChart.tsx
@@ -16,10 +16,6 @@ import { displayScore } from 'src/utils/criteria';
 const BAR_CHART_CRITERIA_SCORE_MIN = -1;
 const BAR_CHART_CRITERIA_SCORE_MAX = 1;
 
-interface Props {
-  video: VideoSerializerWithCriteria;
-}
-
 const between = (a: number, b: number, x: number | undefined): number => {
   // clips x between a and b
   return Math.min(b, Math.max(a, x || 0));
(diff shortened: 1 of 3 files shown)


frontend/src/components/CriteriaBarChart.tsx
```diff
@@ -94,46 +95,48 @@ const CriteriaBarChart = ({ video }: Props) => {
...
+            <Cell
+              key={entry.criteria}
+              fill={criteriaColors[entry.criteria] || criteriaColors['default']}
+            />
+          ))}
+        </Bar>
+        <YAxis
+          type="category"
+          dataKey="criteria"
+          tick={renderCustomAxisTick}
```
lfaucon, line 117: The method `renderCustomAxisTick` does not work well anymore, because it was using the hardcoded value `250` for the width of the graph. We should find a way to know the width of the compoennt and use this or maybe specify `left: 50%` somehow

Screenshot 2022-04-22 at 16 39 35

lfaucon (changes requested):

frontend/src/components/CriteriaBarChart.tsx
sigmike: It should be fixed (0c4d5d4d2034326ce852d28d5e40a0e1e97afcc8).

frontend/src/components/CriteriaRadarChart.tsx
```diff
@@ -74,25 +75,27 @@ const CriteriaRadarChart = ({ video }: Props) => {
...
-        tick={false}
-      />
-      <Radar
-        dataKey="score"
-        stroke="#8884d8"
-        fill="#8884d8"
-        fillOpacity={0.6}
-      />
-    </RadarChart>
+    <ResponsiveContainer width="100%" height={500}>
```
lfaucon, line 78:
```suggestion
    <ResponsiveContainer width="100%" aspect={1}>
```

I tested this and it removes a lot of the blank space on thinner screens

lfaucon (approved):
Thanks for this contribution! One step closer to having a nice mobile-friendly video analysis page

frontend/src/components/CriteriaRadarChart.tsx
sigmike: Indeed. I applied that to both charts.

PR #851, opened 2022-04-25
Add admin for ContributorScaling

lfaucon:


backend/tournesol/admin.py
```diff
@@ -161,6 +162,23 @@ class ContributorRatingCriteriaScoreAdmin(admin.ModelAdmin):
     )
 
 
+@admin.register(ContributorScaling)
+class ContributorScalingAdmin(admin.ModelAdmin):
```
GresilleSiffle, line 166: The current admin configuration allows to filter by poll, and I think it could be interesting to be able to filter by criteria.

And also by username? I'm curious about my scaling : )

```suggestion
class ContributorScalingAdmin(admin.ModelAdmin):
    search_fields = (
        'criteria',
        'user__username',
    )
```

GresilleSiffle (approved):

PR #852, opened 2022-04-25
Display emoji criteria icons on bar chart

sigmike:
Related issue: #850

For now only the emojis are displayed:

image

Remaining to do:
- [x] Change colors


sigmike:
@lfaucon What color should we use for the bars? For the video criteria I used the color of the icon, but with emojis I'm not sure we can do that since they differ on each system and they contain multiple colors.

lfaucon:
Thanks for this! Let's stick to Tournesol's secondary blue color if we don't find anything better. Otherwise I would have thought to pick the colors bysed of the colors of the emoji, but it is not easy to do and it won't make sense on all devices.

PR #855, opened 2022-04-25
Implement `EntityTabsBox` to use in video selector and analysis page

amatissart:
This adds a new component `EntityTabsBox`
Accepts a list of `EntityTab` in props, to define how to fetch the different lists of entities.

A few limitations:
 * uses `RowVideoCard` (to be made more generic in the future)
 * tabs style is based on MUI `Tabs` (could be customized?)

The first use-case will be to enrich the VideoInput, with the possibility to pick a video from multiple tabs (see #856)

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -82,6 +82,9 @@
     "autoEntityButton": "Auto",
     "pasteUrlOrVideoId": "Paste URL or Video ID"
   },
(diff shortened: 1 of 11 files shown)


amatissart:
I rebased this branch, and replaced the `RowVideoCard` with a `RowEntityCard` to be more generic.

GresilleSiffle (approved):

PR #856, opened 2022-04-25
Use `EntityTabsBox` to pick a video in `VideoInput`

amatissart:
image


frontend/src/features/entity_selector/EntityInput.tsx
```diff
@@ -1,40 +1,163 @@
...
+          const response = await getRecommendations(
+            'videos',
+            20,
+            '?date=Month',
+            []
+          );
+          return response.results ?? [];
+        },
+      },
+    ],
```
lfaucon, line 102: I'd really like that we use here the non-connected entities api endpoint for an additional list. If you think it is too much for this PR, let me know and I'll create a ticket for it

lfaucon (changes requested):
When an error occurs fetching the list, the previous list is not updated. I think it would be better to remove the current list on click, show a loader while the new list is being fetched and display the new list or an error message accordingly

Screenshot 2022-06-11 at 21 34 13

frontend/src/features/entity_selector/EntityInput.tsx
amatissart: The loader and error message have been added.

An additional tab is also present to fetch unconnected entities (the tab is disabled if the other entity has not been selected yet).

I noticed that the "unconnected_entities" API could return incorrect results after successive requests. A fix is implemented in #984

GresilleSiffle:
I tested again, the ok / loading / error states work properly on my laptop :ok_hand: 

One thing we could improve in the future is the wording, or its documentation. I feel "suggestions" / "à comparer" / "à relier" are good names, but also close to each other and can be confounded by users.

As the comparison page is the most important page of the application, and as this new feature will greatly improve the UX, how about adding (in a second PR) a paragraph in each tab, describing with one or two lines the nature of the list?

For "à comparer", the paragraph could be: "Ces vidéos viennent de votre liste à comparer plus tard."

For "à relier", the paragraph could be: "Ces vidéos sont isolées à cause d'un trop faible nombre de comparaison. Aidez-les à sortir de cet état."

It's just an idea. There are a lot of solution to make those terms clear for the users:
- adding a dedicated help page, that can be reached by clicking on a interrogation mark on the comparison page
- adding tooltips
- adding an extra required step in the tutorial
- etc.

I think adding paragraphs is rather quick, and can be easily reverted if one day we decide to change the the comparison UI/UX. We could even have a tutorial/assistance setting, when activated these kind of paragraphs will be displayed everywhere in the application. When the users feel comfortable, they can disable the assistance.

capture

GresilleSiffle (approved):

lfaucon (approved):

PR #860, opened 2022-04-26
[front] Switch Criteria Filter Mode: multiple or single criteria (#843)

### LINKED ISSUE #843: [front] ux: simplify criteria filter
**Context:**
The criteria choices on the recommendation tab is overly complicated and very rarely used. We want to simplify it to be a lot more user-friendly. For this we could use the same selector for criteria as we used in the Result Page of the Presidentielle2022 Poll

**Current state:**
Screenshot 2022-04-22 at 11 33 35

**Target state:**
Screenshot 2022-04-22 at 11 34 17

Note that depending on the decided implementation we will maybe need:
- A small backend change to consider the default weight to 0 instead of 50
- Adapting the extension

adrausch:
Closes #843 

### desktop

capture

full

### mobile

mobile

### DIFF
--- a/frontend/src/components/CriteriaIcon.tsx
+++ b/frontend/src/components/CriteriaIcon.tsx
@@ -15,7 +15,6 @@ const CriteriaIcon = ({
   return (
     <Box
       sx={{
-        display: 'inline',
         ...sx,
       }}
     >
--- a/frontend/src/components/CriteriaIcon.tsx
+++ b/frontend/src/components/CriteriaIcon.tsx
@@ -15,9 +15,7 @@ const CriteriaIcon = ({
   return (
     <Box
       sx={{
-        '& img': {
-          display: 'block',
-        },
+        display: 'inline',
         ...sx,
       }}
     >
--- a/frontend/src/components/CriteriaIcon.tsx
+++ b/frontend/src/components/CriteriaIcon.tsx
@@ -15,6 +15,9 @@ const CriteriaIcon = ({
   return (
     <Box
       sx={{
+        '& img': {
+          display: 'block',
+        },
         ...sx,
       }}
     >
--- a/frontend/src/features/recommendation/CriteriaFilter.tsx
+++ b/frontend/src/features/recommendation/CriteriaFilter.tsx
@@ -214,7 +214,7 @@ function MultipleCriteriaFilter({ setFilter }: FilterProps) {
 }
 
 function CriteriaFilter({ setFilter }: FilterProps) {
-  const [isMultipleFilter, setIsMultipleFilter] = React.useState(false);
+  const [isMultipleFilter, setIsMultipleFilter] = useState(false);
   const { t } = useTranslation();
   return (
     <>
--- a/frontend/src/features/recommendation/CriteriaFilter.tsx
+++ b/frontend/src/features/recommendation/CriteriaFilter.tsx
@@ -186,6 +186,7 @@ function MultipleCriteriaFilter({ setFilter }: FilterProps) {
                   container
                 >
                   <CustomSlider
+                    data-testid={'filter-criterion-slider-' + criteria.name}
                     name={criteria.name}
                     defaultValue={parseInt(
                       searchParams.get(criteria.name) || '50'
--- a/frontend/src/features/recommendation/CriteriaFilter.tsx
+++ b/frontend/src/features/recommendation/CriteriaFilter.tsx
@@ -221,23 +221,29 @@ function CriteriaFilter({ setFilter }: FilterProps) {
   return (
     <>
       <TitledSection title={t('filter.criteria')}>
-        <FormGroup>
-          <FormControlLabel
-            control={
-              <Switch
-                checked={isMultipleFilter}
-                onChange={() => setIsMultipleFilter(!isMultipleFilter)}
+        <Grid container spacing={1}>
+          <Grid item>
+            <FormGroup>
+              <FormControlLabel
+                control={
+                  <Switch
+                    checked={isMultipleFilter}
+                    onChange={() => setIsMultipleFilter(!isMultipleFilter)}
+                  />
+                }
+                label={t('filter.multipleCriteria') as string}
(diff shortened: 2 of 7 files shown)


frontend/src/features/recommendation/CriteriaFilter.tsx
```diff
@@ -49,14 +54,17 @@ const CustomSlider = withStyles({
     backgroundColor: 'currentColor',
   },
 })(Slider);
-
-interface Props {
+interface ValueLabelProps {
```
GresilleSiffle, line 57: minor: we can add a blank line for more clarity

```suggestion

interface ValueLabelProps {
```

GresilleSiffle (commented):
Thanks @adrausch 

This feature will be very useful :+1: 

I added few suggestions you can incorporate in the code easily by simply clicking on the buttons "Commit suggestion".

### bonus

I see error in my JavaScript console when I move the sliders, and I also reproduce on `main`. Do you see the same error @amatissart @lfaucon ?

```
MUI: A component is changing the default value state of an uncontrolled Slider after being initialized. To suppress this warning opt to use a controlled Slider.
```

lfaucon:
> ### bonus
> I see error in my JavaScript console when I move the sliders, and I also reproduce on `main`. Do you see the same error @amatissart @lfaucon ?
> 
> ```
> MUI: A component is changing the default value state of an uncontrolled Slider after being initialized. To suppress this warning opt to use a controlled Slider.
> ```

I see the same. It's a classic

frontend/src/features/recommendation/SearchFilter.spec.tsx
```diff
@@ -99,8 +99,7 @@ describe('Filters feature', () => {
     expect(queryAllByTestId(languageFilter, 'autocomplete')).toHaveLength(1);
 
     // Check criteria filters presence
-    // 1 slider per criteria, "Neutral" position by default
-    expect(screen.queryAllByLabelText(/neutral/i)).toHaveLength(3);
+    expect(screen.queryByLabelText('multiple criteria')).not.toBeNull();
```
GresilleSiffle, line 102: can be simplify with

```suggestion
    expect(screen.queryByLabelText('multiple criteria')).toBeVisible();
```

GresilleSiffle:
I recommend to also add a new isolation test now that we have introduced a new behaviour in this already tested component: the possibility to switch on and off the multiple criteria panel. It should be rather easy and quick. I can handle it tomorrow morning if needed.

amatissart:
I haven't look at the implementation in detail, but I obseved one issue:

By default the selected "criteria" is "largely_recommended", but no parameters in applied in the API request, so the results are actually sorted by "tournesol_score" (which is currently defined as an aggregation of all criteria with the same weight).

However when selecting another criteria and then back to "largely_recommended", the results are sorted correctly. But then it's no longer possible to get the default order.

https://user-images.githubusercontent.com/4726554/165372429-97b591ce-198d-41a4-b57b-db6bc7941dac.mp4

lfaucon:
> Should we add a "default" option in the list, to sort the recommendations by "Tournesol score"?

I'd prefer the simpler option where we sort only on "largely_recommended" from the start because we decide to move away from the way the Tournesol score is currently computed

GresilleSiffle:
> I'd prefer the simpler option where we sort only on "largely_recommended" from the start because we decide to move away from the way the Tournesol score is currently computed

If we sort by the largely recommended criterion, will the Tournesol score appear unordered?

If we it could be confusing for the users, as there is no explanation on what the Tournesol score is.

Few solutions I see:

**(1)** rework how to Tournesol score is computed in this PR, should be based on the largely recommended criterion?

**(2)** get rid of the Tournesol score, and instead display the score related to the selected criterion

frontend/src/features/recommendation/SearchFilter.spec.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/860/commits/486caad8e6d5573cdbc01bc8bf56f8e6579abfb9.

lfaucon:
> If we sort by the largely recommended criterion, will the Tournesol score appear unordered on the recommendations page?
> If yes, it could be confusing for the users, as there is no explanation on what the Tournesol score is, and how it should be understood.
> Few solutions I see:
> (1) rework how to Tournesol score is computed in this PR, should be based on the largely recommended criterion?
> (2) get rid of the Tournesol score, and instead display the score related to the selected criterion

We really want to display a fixed Tournesol score that does not change based on user's search. We had one that changes in the past and it was extremely confusing. (2) is a strong No from me.

We will do (1) but it is ok to do it in a next iteration and have some small level of confusion until then (Note it is by far not the most confusing thing about the whole platform). The tournesol scores will only be slightly un-ordered and selecting specific criterias will make it slightly more un-ordered in an expected way.

I'm also fine with a default that adds the same weight to all criteria just like the current Tournesol scores are computed. @adrausch Can you implement either the default option in the dropdown (equal weight to all criteria) or correctly initialize the filters to sort by `largely_recommended` as it is initially selected? Thanks!

frontend/src/components/CriteriaIcon.tsx
```diff
@@ -15,9 +15,6 @@ const CriteriaIcon = ({
   return (
     <Box
       sx={{
-        '& img': {
-          display: 'block',
-        },
```
amatissart, line 20: This is not critical, but removing this style breaks the alignment when the icon is used besides a text:
Peek 03-05-2022 13-28


It would be cleaner to keep this as "block", and apply "flex" styling on the parent component where this icon is used:
```
display: flex,
align-items: center
```

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/860/commits/34c3cd08abb100f2195dd0f6e1e930f7cb6f770a.

GresilleSiffle (approved):

PR #862, opened 2022-04-28
[front] fix: recommendations page not displaying anything...

GresilleSiffle:
... when the date filter is active.

The `/polls/` API, used by the recommendations page, expects the date to be formatted with the ISO 8601 norm. Contrary to the legacy `/video/` which uses a custom format.

The front end recommendations page now uses only ISO 8601 dates when querying the back end.

It fixes a regression introduced by aa33d80cffefc18d089d70a50e7880ee19644b7f.

I think we can merge this as is, and I'll add an end-to-end test to ensure this bug doesn't come back.

### DIFF
--- a/frontend/src/features/recommendation/RecommendationApi.tsx
+++ b/frontend/src/features/recommendation/RecommendationApi.tsx
@@ -20,46 +20,26 @@ const buildDateURLParameter = (
   pollName: string,
   params: URLSearchParams
 ): void => {
-  function format(str: string) {
-    if (str.length == 1) {
-      return '0'.concat(str);
-    } else if (str.length == 4) {
-      return str.slice(2);
-    } else {
-      return str;
-    }
-  }
-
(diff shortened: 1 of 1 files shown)


frontend/src/features/recommendation/RecommendationApi.tsx
```diff
@@ -20,46 +20,26 @@ const buildDateURLParameter = (
...
+    conversionTime.set('Week', dayInMillisecs * 7);
+    conversionTime.set('Month', dayInMillisecs * 31);
+    conversionTime.set('Year', dayInMillisecs * 365);
 
     if (params.get('date')) {
       const date = params.get('date');
       params.delete('date');
+
       if (date != 'Any') {
         // TODO: figure out why adding 1 month is needed here
```
amatissart, line 40: this comment is now obsolete

amatissart (approved):

PR #869, opened 2022-05-01
Implement 3 different `score_mode` for poll recommendations computed by Mehestan

### LINKED ISSUE #841: back: Dynamically compute collective scores based on different weights for trusted users (`?score_mode`)
Possibly with a new parameter `?score_mode=default/trusted_only/all_equal`

amatissart:
Closes #841 

## Description

Mehestan will compute global scores with 3 different "score modes":
 * `default`: non-trusted users are given a low voting weight to control their influence on the global scores
 * `all_equal`: all users receive the same voting right when computing global scores (note that the individual scores are still scaled according to the scaling process that use a sample of trusted users as a reference)
 * `trusted_only`: non-trusted users are ignored

The poll recommendations view accepts a new parameter `?score_mode` to select the mode to use when computing the `total_score` (weights per criteria can still be applied).

## Implementation

It brings additional complexity in the database models though: `EntityCriteriaScore` now includes a `score_mode` field. As a consequence, it's no longer sufficient to join/prefetch the "criteria_scores" to get scores associated with an entity when building API responses. Up to 3 scores may exist for each entity/poll/criteria, so an extra filter is now required.

To avoid confusion with the previous behavior, the "related_name" on the Entity model has been renamed from `criteria_scores` to `all_criteria_scores`. And to make this migration easier, this PR includes 2 helpers:
 * `queryset.prefetch_scores(poll, score_mode)` will apply a filtered `Prefetch()`, accessible as `entity._prefetched_criteria_scores`
 * `entity.criteria_scores` is a property that will evaluate the prefetched scores as a list. If no prefetch has been used, it will fetch the scores with the default score mode ("default"). Thanks to this property, the existing serializers can remain unchanged.


GresilleSiffle:
Hello,

I'm currently reading this new code : ) What do you think @amatissart of displaying the score mode in the admin interface?

```diff
diff --git a/backend/tournesol/admin.py b/backend/tournesol/admin.py
index aa4160b4..25f95b21 100644
--- a/backend/tournesol/admin.py
+++ b/backend/tournesol/admin.py
@@ -109,8 +109,12 @@ class EntityCriteriaScoreAdmin(admin.ModelAdmin):
     list_display = (
         'entity',
         'criteria',
+        'score_mode',
         'score'
     )
+    list_filter = (
+        'score_mode',
+    )
     search_fields = (
         'entity__uid',
     )

```

GresilleSiffle (approved):
Looks good to me, no error when running the ML training with Mehestan.

Should we run a `yarn update-schema` to add the new `score_more` parameter in the `openapi.yaml` file?

I've run it locally, and there are several unstaged modifications introduced by the distribution of criteria scores feature we should avoid to mix with this PR.

Issue #720, opened 2022-03-28
[EPIC] Add account validation through vouching mechanism to distribute trust to users who do not belong to trusted entities


BerangereColbois:
- [x] [back] create structure to support the vouching mechanism
      * store connections between users, and local trust values between two peers 
      * store list of users trusting and trusted by each user
- [x] [back] compute global trust values per users, based on the [EigenTrust algorithm](ilpubs.stanford.edu:8090/562/1/2002-56.pdf). This value can be used to know how much we trust every user later on.
- [x] [front] modify sign up and profile page accordingly, to get and display all information listed above
- [x] [front] add a vouching system, where each user can vouch for other users, by selecting how much they trust the user,  Modify the profile page by adding the following fields :
      * list of users the user vouched for
      * list of users who vouched for this user
- [x] #1178
- [x] [back] #1071 
- [x] [back] refactor: On the user model "voting-right" -> "trust score". Rename the column in the addtional file added to the public dataset.
- [x] #1176
- [x] #1060
- [ ] UNSURE: Voting rights displayed on user profile page should not be the trust score, but the average voting rights on all their rated videos displayed as a percentage.

PR #870, opened 2022-05-01
[draft] vouching mechanism

BerangereColbois:
[Draft] Vouching mechanism
Back end : 
1. Modification of User model to add fields for diplomas, competencies and trust score
2. Regarding vouching (grouped in the vouch folder) :
    a. Improved Eigentrust algorithm to compute the global trust score of each user given the list of pre-trusted users and the vouching connections
    b. Model for the vouching mechanism
    c. Tests for the algo

Front end : 
    a. Public profile page (not linked to username yet, and pretty ugly)
    b. Draft of modification page


lfaucon:
Really awesome to see a draft for this feature! 🤩 

Let us know how we can help you. We can discuss on our Discord if it helps you!

GresilleSiffle:
According to GitHub there are currently 656 files updated by this feature.

Maybe something went wrong during the commit, or the fork of repository? Let us know @BerangereColbois if you need/want help for somehting.

amatissart:
@BerangereColbois Thanks for the draft. Do you know why the files permissions have been modified from 0644 to 0755 on this branch? It should not be necessary and makes the changes difficult to read for now ^^

Also 5 items had been listed in #720. It may be relevant to focus on just a few of these items to begin with. So don't hesitate to split your changes into multiple small changes if you think it would help with the review.

GresilleSiffle:
> Also 5 items had been listed in https://github.com/tournesol-app/tournesol/issues/720. It may be relevant to focus on just a few of these items to begin with. So don't hesitate to split your changes into multiple small changes if you think it would help with the review.

This is a very good advice, it should even make the development process easier.

If the 5 points listed in the issue #720 can be implemented separately, we should implement them separately. For instance, the diplomas can be added in the back end with the proper migration and tests in a first PR. Then, the front end can be updated afterward with a focus on the look and feel of the UI.

BerangereColbois:
Hi, 
Thank you a lot for all those useful reviews!
First, I'm sorry I changed the permissions, I didn't intend to commit them that way I did something wrong just after initially pulling the repo, sorry for that!
Regarding the separation of the different tasks, it indeed seems easier!
Would you mind if I do new drafts for each of the tasks?

GresilleSiffle:
> First, I'm sorry I changed the permissions, I didn't intend to commit them that way I did something wrong just after initially pulling the repo, sorry for that!

No problem, sometimes it happens. Do you think you can either update this PR or create a new one with only the code updates (without the permissions)?

> Would you mind if I do new drafts for each of the tasks?

Sure, you can create as many drafts as you want. Be careful about something: if the draft 3 is based on the draft 2, updates on the draft 2 might automatically break the draft 3. So yes you can create several drafts, but keep in mind that it's a good practice to ask for review as soon as possible when you think your code is stable, so that it's merged quickly and will allow the following drafts to be based on stable and approved code.

PR #871, opened 2022-05-03
[back] feat: add the `Voucher` model 

BerangereColbois:
[backend] Draft for the first part of issue #720 

A model was added to represent vouching connection, and the user model was updated to store more info that will, later on, be useful for the vouching, and to store the global trust score of the user.

The diplomas and competencies should be displayed on the user's public profile page so that users can vouch either for someone or for a specific field (one could vouch for the diplomas or the competencies only).


backend/vouch/models.py
```diff
@@ -0,0 +1,69 @@
+from django.db import models
+from django.db.models import ObjectDoesNotExist
+from core.models.user import User
+
+# Create your models here.
+class vouch(models.Model):
```
GresilleSiffle, line 6: We can stick to the Python coding style by using a capital letter.

```suggestion
class Vouch(models.Model)
```

backend/vouch/models.py
```diff
@@ -0,0 +1,69 @@
...
+        related_name="vouchinguser",
+    )
+    vouchedfor = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        help_text="The contributor",
+        related_name="vouchedforuser",
+    )
+    is_public = models.BooleanField(
+        default=False, null=False, help_text="Should the vouching be public?"
```
GresilleSiffle, line 21: By default, model fields are not null.

```suggestion
        default=False, help_text="Should the vouching be public?"
```

See: https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.Field.null

backend/vouch/models.py
```diff
@@ -0,0 +1,69 @@
...
+        db_index=True,
+    )
+
+    def save(self, *args, **kwargs):
+        if not self._state.adding and (
+            self.creator_id != self._loaded_values['creator_id']):
+         raise ValueError("Updating the value of creator isn't allowed")
+        super().save(*args, **kwargs)
+
+    def get_vouchedfor(vouching):
```
GresilleSiffle, line 47: This method could be a static.

```suggestion
    @staticmethod
    def get_vouchedfor(vouching):
```

backend/vouch/models.py
```diff
@@ -0,0 +1,69 @@
...
+        except ObjectDoesNotExist:
+            pass
+        else:
+            return vouchedfor, False
+        vouchedfor = vouch.objects.filter(vouchedfor = vouchedfor)
+        return vouchedfor, True
+
+
+    
+
```
GresilleSiffle, line 69: Extra blank lines can be removed.

```suggestion
```

GresilleSiffle (commented):
Hello @BerangereColbois 

Thanks a lot for the submitted work : ) I don't know if you PR is ready for a review or not but I left some comments.

I made few suggestions to improve the code, do not hesitate to tell me I you think they are not relevant.

**(1)**

About the name of some variables, do you think it will be more readable and less confusing to have the attributes `vouch.from` and `vouch.to` instead of `vouch.vouching` and `vouch.vouchedfor` (or even `from_user` / `to_user`)?

The to and from are rather assertive and I like a lot `vouch.from_user` / `to_user` as it explicitly tells the developers we are currently working with `User` instance.  

**(2)**

English is not my mother tongue, so I'm not sure about what I'm going to say but I think voucher is more used than vouch? I don't know.

backend/vouch/models.py
```diff
@@ -0,0 +1,69 @@
...
+            pass
+        else:
+            return vouching, False
+        vouching = vouch.objects.get(vouching = vouching)
+        return vouching, True
+
+    def get_vouching(vouchedfor):
+        try:
+            vouchedfor = vouch.objects.filter(vouchedfor = vouchedfor)
+        except ObjectDoesNotExist:
```
GresilleSiffle, line 60: I think the `filter` method doesn't raise `ObjectDoesNotExist`.

Maybe something like this will be better

```python
vouchedfor = vouch.objects.filter(vouchedfor = vouchedfor)
return vouchedfor, vouchedfor.exists()
```

backend/vouch/models.py
```diff
@@ -0,0 +1,69 @@
...
+        blank=False,
+        help_text="trust value assigned by vouchinguser to vouchedforuser",
+    )
+    uncertainty = models.FloatField(
+        default=0,
+        blank=False,
+        help_text="Uncertainty about the vouching score",
+    )
+    criteria = models.TextField(
+        default=0,
```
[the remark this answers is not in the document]
BerangereColbois: Hi, actually I put this field but don't necessarily intend to use it directly, the idea was to be able to store for which aspect the user vouching has vouched for (for example one could say "I vouch for this user having this diploma, but I don't know anything about their competencies"), but I thought it would be easier to start by just allowing users to vouch for a user in their entirety, and then improve it if time allows. May I ask you for a suggestion on how you would do that properly?

backend/vouch/models.py
BerangereColbois: @GresilleSiffle Thank you a lot for all your remarks, they are really precious!

GresilleSiffle: My pleasure @BerangereColbois 

Tell me when you think this PR is ready so we'll have a final look with the team : )

GresilleSiffle:
Hello @BerangereColbois 

Following our previous work I just created a pull request from a branch or mine to your fork, so we can collaborate more easily : )

https://github.com/BerangereColbois/tournesol/pull/1

After having reviewed and merged this branch, the current PR #871 will be automatically updated by GitHub.

Thank you

backend/vouch/models.py
```diff
@@ -0,0 +1,60 @@
...
+        help_text="The trust value given by the vouching user to receiving user.",
+    )
+    # XXX: What is the vouching score described in the help text? Is it the
+    # `trust_value`? If yes we should use the same vocabulary to avoid any
+    # confusion.
+    uncertainty = models.FloatField(
+        default=0,
+        help_text="Uncertainty about the voucher score.",
+    )
```
amatissart, line 41: Maybe that's because I am not familiar with eigentrust, but I wonder what exactly is the status of these values: will they be defined by the user when creating a voucher, or computed by the algorithm, or something else?

backend/vouch/models.py
```diff
@@ -0,0 +1,60 @@
...
+    attribute = models.TextField(
+        max_length=32,
+        help_text="Name of the attribute corresponding to this voucher.",
+        db_index=True,
+    )
```
amatissart, line 46: To make things more explicit about which attributes are supported, I would recommend to define a list of choices that are accepted in this field. 

For now, maybe we could even define an enum with a single choice:
```py
class VoucherAttribute(models.TextChoices):
    IDENTITY = "identity"
```
and set `choices=VoucherAttribute.choices` on this TextField.
See https://docs.djangoproject.com/en/4.0/ref/models/fields/#enumeration-types for more details.

amatissart (commented):
Hello @BerangereColbois 
I submitted a few comments and suggestions about the data model.  
There are also a few questions about how the fields will interact with the algorithm.

You may also have noticed that the backend lints are failing. This is due to minor code style issues (too many blank lines, line too long, etc.). You can see the details [here](https://github.com/tournesol-app/tournesol/runs/6441938608?check_suite_focus=true#step:6:10). Only flake8 errors (in "vouch/models.py") are blocking, pylint warnings are not related to your changes and can be ignored.

You can also find instructions about how to run the linter locally [in backend/README.md](https://github.com/tournesol-app/tournesol/tree/main/backend#code-quality).

backend/core/models/user.py
```diff
@@ -35,8 +35,14 @@ class User(AbstractUser):
         max_length=100, blank=True, null=True, help_text="Last name"
     )
     title = models.TextField(null=True, blank=True, help_text="Your position")
+    diplomas = models.TextField(
+        null=True, blank=True, help_text="Your diplomas"
+    )
+    competencies = models.TextField(
+        null=True, blank=True, help_text="Your competencies"
+    )
```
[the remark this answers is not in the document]
BerangereColbois: Hello @amatissart, thank you a lot for all those reviews!
Here is the idea I have about vouching mechanism, which answers some of your questions below : 
I want users to be able to tell how much they trust other users on different aspects (I can vouch for someone because I know they are who they pretend to be, or because I know they have the diplomas or competencies they claim). I think we can remove fields of diplomas and competencies for now if we think it is sufficient for users to say how much they vouch for a profile globally, which gives users more freedom on how they assess the importance of the different aspects they can vouch for by giving them a trust value. I was thinking that it would be easier to let users give trust scores to users without constraint (other than the trust value being non-negative, I'll enforce that thank you for the remark!) so that they can choose their metrics more easily, and then scale the result within the algorithm (the algorithm is implemented, I will do a PR for it once it is cleaned and documented so that you can see how it works). Do you think it would be a better option to enforce a certain metric in terms of user experience? 
Regarding the attribute, it was for users to be able to select what they were vouching for about (the PoP, diplomas or competencies mostly) but if we remove them I think we can also remove the attribute field too, at least for now?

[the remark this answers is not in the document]
amatissart: > Do you think it would be a better option to enforce a certain metric in terms of user experience?

The UX will need to be discussed with the team. But if this "trust_value" is given by the user themselves, then I think an explicit scale would be preferable. Otherwise low values could be ambiguous. 

> Regarding the attribute, it was for users to be able to select what they were vouching for about (the PoP, diplomas or competencies mostly) but if we remove them I think we can also remove the attribute field too, at least for now? 

As you wish. If you think we can do without this `attribute`, then it's okay with me.

BerangereColbois:
@amatissart Thank you for your reviews on draft #872, I thought it would make more sense to add the updated version here as the algo calls the new vouch model.
Some of your reviews are fixed in this version, but I still feel like some aspects could be optimized more.
I still have some testing and refactoring to do, so this is not the final version, but is it ok if I close #872 now that I have put together the two parts of the code here?

GresilleSiffle:
Hello @BerangereColbois 

Thanks for addressing all Adrien's comments.

I'm going to push few commits today and tomorrow to make the tests pass.

GresilleSiffle:
I just squashed the new migrations of the `core` application to avoid doing back and forth by adding / removing / adding fields.

I also added tests for the model `Voucher` to cover its static method. It will also encourage us to add more tests if the model evolves.

GresilleSiffle:
I just squashed the new migrations of the `core` application to avoid doing back and forth by adding / removing / adding fields.

I also added tests for the model `Voucher` to cover its static method. It will also encourage us to add more tests if the model evolves.

GresilleSiffle (approved):
LGTM

PR #872, opened 2022-05-03
[backend] Modified eigentrust algo added

BerangereColbois:
Related to second part of issue #720 
The modified version of[ eigentrust algorithm](http://ilpubs.stanford.edu:8090/562/1/2002-56.pdf) that will be used to compute the global trust score of each user based on how other users vouched for them, and the tests of the algorithm.


BerangereColbois:
This draft depends on #871, I will update it once the first draft is done!

backend/vouch/trustalgo.py
```diff
@@ -0,0 +1,76 @@
...
+    Normalize the trust values per user, and set diagonal values at 1
+    This makes C an ergotic stochastic matrix
+    """
+    assert(len(C[0])==len(C))
+    for i in range(len(C)):
+        C[i][i] = 1
+        tmp = np.sum(C[i])
+        for j in range(len(C)):
+            C[i][j] = C[i][j] / float(tmp)
+    return C
```
amatissart, line 23: In most cases when handling numpy objects, it's preferable to use methods provided by numpy. They are implemented in optimized C code and are significantly more performant than python `for` loops.

Here we could write something like
```py
np.fill_diagonal(C, 1)
return C / C.sum(axis=1, keepdims=True)

**Review by amatissart — COMMENTED** [review]
I understand this PR is still a draft. So feel free to ignore my comments about optimization if you want to focus on the algorithm first.

File: backend/vouch/trust_algo.py [hunk 2]
```diff
@@ -0,0 +1,86 @@
...
+    # vouch much to pre-trusted set
+    C = normalize_trust_values(C_, trust_status)
+    # get the a vector of trust given to each users by the network
+    trust_vector = get_trust_vector(C, p)
+    # normalize and rescale it
+    trust_vector = trust_vector/np.sum(trust_vector)
+    voting_weight = rescale(trust_vector, trust_status)
+    for k, u in enumerate(users):
+        u.trust_score = float(voting_weight[k])
+        u.save()
```
amatissart, line 85: As a single field is modified, and the computation may take some time, it's safer to update this field only (to avoid reverting changes that have been written into the database in the meantime).
```suggestion
        u.save(update_fields=["trust_score"])
```

backend/vouch/trust_algo.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    trust_size = np.sum(np.array(trust_status) > 0, axis=0)  # size of the pre-trusted set
+    for i in range(len(C[0])):
+        vouching_size = np.sum(np.array(C[i]) > 0, axis=0)
+        for j in range(len(C[0])):
+            trust_indic = 1 if trust_status[j] > 0 else 0  # whether user j is pre-trusted
+            vouch_indic = 1 if C[i][j] > 0 else 0  # whether user j has been vouched by user i
+            C[i][j] = alpha / (alpha * trust_size + vouching_size) * trust_indic
+            C[i][j] += 1 / (alpha * trust_size + vouching_size) * vouch_indic
+    return C
```
amatissart, line 24: (minor)
Optimizing this function may not be very worthwhile with the current number of users, but most (if not all) this operation could probably be computed without python loop.
For example:
```py
trust_indic = (trust_status > 0).astype(int)
vouch_indic = (C > 0).astype(int)
```

backend/vouch/trust_algo.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+from vouch.models import Voucher
+from core.models import user
+
+# hyper parameter of algorithm - select them wisely
+a = 0.2  # define the minimum ratio of trust that is left to pre-trusted users
+eps = 0.00000001
```
amatissart, line 7: can be written
```py
eps = 0.000_000_01
```
or simply
```py
eps = 1e-8
```

amatissart (commented):
Thanks @BerangereColbois 
I made a few suggestions to optimize the queries used to fetch the input data. But overall it looks good to me.

What would be our strategy to run this algorithm after some vouchers have been created on production? Do you think it would be useful/reasonable to run the algorithm and update the `trust_score` as soon as a new Voucher has been submitted? Or should we consider scheduling a job to run this algorithm periodically at fixed times?

backend/vouch/trust_algo.py
GresilleSiffle: Should we also use upper case name for these parameters, to respect the Python PEP8 style guide?

See: https://peps.python.org/pep-0008/#constants

BerangereColbois:
Hi, thank you for all your precious advice @amatissart @GresilleSiffle!
I couldn't find an optimal way to use an iterator, but I modified the code to query for all vouchers only once. I think it is equivalent in terms of queries but please let me know if I'm wrong.
Regarding the frequency of trust scores updates, I had the same interrogation. I think it would be nice to do it quite often just after deploying the vouching mechanism as current users may use it simultaneously. Still, after a while, maybe the frequency can be lower to match the vouching mechanism utilisation? I was thinking that if people don't use it much it would be possible to update the scores directly, but if it is used often maybe we could schedule it regularly instead. Without knowing how much users will use the mechanism I think it is hard to tell, but I'd be very interested in discussing it with you!

backend/vouch/trust_algo.py
```diff
@@ -0,0 +1,86 @@
...
+    users = list(user.User.objects.all().annotate(_is_trusted=Q(pk__in=user.User.trusted_users())))
+    trust_status = [int(u._is_trusted) for u in users]
+    nb_users = len(users)
+
+    # create a matrix C where entry c_ij is the trust value user i gave to user j
+
+    C_ = np.zeros([nb_users, nb_users], dtype=float)
+    for v in Voucher.objects.iterator():
+        i = users.index(v.by)
+        j = users.index(v.to)
```
amatissart, line 70: Not critical for now, but using `users.index()` would make this loop quadratic, assuming the total the number of Vouchers will be at least O(nb of users).

A mapping between users and indices could be built beforehand.

backend/vouch/tests/test_trust_algo.py
```diff
@@ -0,0 +1,125 @@
...
+                Voucher(by=self.user_9, to=self.user_4, trust_value=94),
+                Voucher(by=self.user_9, to=self.user_5, trust_value=95),
+            ]
+        )
+
+    def test_trust_algo(self):
+        trust_algo()
+        users = list(user.User.objects.all())
+        assert users[1].trust_score >= 0.9999999
+        assert users[9].trust_score == pytest.approx(0, 0.00000001)
```
amatissart, line 125: These assertions depend only on the email trust status and not on the Vouchers themselves.  
It would be useful to add a check about one of the values that are expected to be modified thanks to the vouchers. Otherwise, a regression in the algorithm could easily go unnoticed.

amatissart (commented):
Thanks @BerangereColbois!

An additional test case would be welcome to ensure that the algorithm is actually reading the `Voucher` content (see my comment below).

Apart from that, it looks good to me. If that's okay with you, I think we can merge the algorithm implementation and discuss the exact detail about how to update the trust scores in a separate issue.

backend/vouch/tests/test_trust_algo.py
BerangereColbois: Hi, 
Thank you, I didn't notice that we could obtain the same result with an unworking vouching mechanism. I fixed that and added verification that we can update the vouchers.

GresilleSiffle:
Hello @BerangereColbois 

I'm going to push few commits to make the core more clear (without altering the behaviour of course).

So if you plan to continue working on this branch, do not forget to run `git pull` before adding new commits : )

Thank you

GresilleSiffle:
Currently investigating how to use type hints compatible with NumPy n-dimension arrays.

https://github.com/ramonhagenaars/nptyping

amatissart:
@GresilleSiffle In this specific case I would find the docstring more helpful than using custom typing hints, which may not be fully supported by IDEs, and might be disturbing for contributors.

amatissart (approved):

GresilleSiffle:
Yes you're right, if the IDE is not correctly enforcing the types, the comments will be more clear.

PR #873, opened 2022-05-03
fix: Safe recommendations filter on tournesol_score instead of total_score

lfaucon:


backend/tournesol/views/polls.py
```diff
@@ -161,7 +162,7 @@ def annotate_with_total_score(self, queryset, request, poll: Poll):
                 F("criteria_scores__score") * criteria_weight,
                 filter=Q(criteria_scores__poll=poll),
             )
-        )
+        ).filter(total_score__isnull=False)
```
lfaucon, line 165: I found it made more sense to have this operation right after the annotation. This means that entities which have no EntityCriteriaScores will not be returned.

Should we also keep a filter on `tournesol_score__isnull=False` ?

backend/tournesol/views/contributor_recommendations.py
```diff
@@ -38,18 +38,16 @@ def annotate_with_total_score(self, queryset, request, poll: Poll, user):
             total_score=Sum(
                 F("contributorvideoratings__criteria_scores__score") * criteria_weight,
             ),
-        )
+        ).filter(total_score__isnull=False)
```
amatissart, line 41: a similar filter is missing in `PollsRecommendationsViews` in polls.py, line 215

amatissart (approved):

GresilleSiffle (approved):

PR #874, opened 2022-05-03
Add load_tweetinfo management command

aidanjungo:
Will be use to add to the TweetInfo all the video already tweeted by the old version of the Twitterbot.

### DIFF
--- a/backend/twitterbot/management/commands/load_tweetinfo.py
+++ b/backend/twitterbot/management/commands/load_tweetinfo.py
@@ -16,7 +16,7 @@ class Command(BaseCommand):
             "-f",
             "--file",
             type=str,
-            help="Relative path to the input file.",
+            help="Relative path to the input file. Each line must follow the structure {yt_video_id};{BOT_NAME}.",
         )
 
         # Optional argument
--- a/backend/twitterbot/management/commands/load_tweetinfo.py
+++ b/backend/twitterbot/management/commands/load_tweetinfo.py
@@ -16,7 +16,10 @@ class Command(BaseCommand):
             "-f",
             "--file",
             type=str,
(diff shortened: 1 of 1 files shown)


backend/twitterbot/management/commands/load_tweetinfo.py
```diff
@@ -0,0 +1,70 @@
...
+class Command(BaseCommand):
+    help = "Load already tweeted videos from a file and save data as TweetInfo."
+
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            "-f",
+            "--file",
+            type=str,
+            help="Relative path to the input file.",
```
amatissart, line 19: It could be useful to add some details about the format of this file in this help message.

amatissart (approved):

backend/twitterbot/management/commands/load_tweetinfo.py
```diff
@@ -0,0 +1,70 @@
...
+        )
```
GresilleSiffle, line 20: Running the command fails when the -f option is not provided. We should mark the -f argument as required to make the command display a nice error message.

```suggestion
            required=True,
        )
```

backend/twitterbot/management/commands/load_tweetinfo.py
GresilleSiffle: :+1: Something like this?

```
help="Relative path to the input file. Each line must follow the structure {yt_video_id};{BOT_NAME}.",

Comment (GresilleSiffle) [reply backend/twitterbot/management/commands/load_tweetinfo.py:19]: :+1: Something like this?

```
help="Relative path to the input file. Each line must follow the structure {yt_video_id};{BOT_NAME}.",

GresilleSiffle (changes requested):

GresilleSiffle (approved):

Issue #832, opened 2022-04-20
Freeze Presidential Poll


lfaucon:
- [x] Set poll to an inactive state  
 * API to create/edit comparisons should return an error
 * ML should not run on this poll
- [x] Disable button for submitting comparisons on presidentielle comparison page
- [x] Add a warning box that explains the freeze on the presidentielle comparison page
- [x] Update Home page, replace "Start" button with "See results"

PR #877, opened 2022-05-04
`active` status for Poll

amatissart:
Related to #832 

* New "active" flag in model `Poll`
  * ML runs on active polls only
  * Comparison creation/update/delete is not allowed in inactive polls
  * "active" status is returned in `/polls/{name}` API response

* Changes on the Comparison page when the current poll is inactive:
  * the submit/edit button is disabled (It's still possible to access to existing comparisons).
  * a warning message is displayed at the top of the page.

#### To do in a future PR
- [ ]  Update Home page, replace "Start" button with "See results" if the poll is inactive


GresilleSiffle (approved):
:+1:

In the future, should we also disable the following route?

`POST /users/me/contributor_ratings/{poll_name}/`

Currently initializing a contributor rating shouldn't have any impact in an inactive poll.

PR #880, opened 2022-05-05
Add script to update openapi with dev-env

sigmike:
Updating openapi when using dev-env is not easy. The `yarn update-schema` command doesn't work within the container (`wget` is not installed and the API is not accessible through the default `localhost:8000`). So I made a small script to do it.

It may be better to make `update-schema` work in the container, but this solution is simple (and `curl` is already a requirement). I'm not sure where it should be documented though.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,8 @@ end representation the back end OpenAPI schema.
 **step 3** Don't forget to re-generate the service files by running
 `scripts/generate-services-from-openapi.sh`
 
+If you're using `dev-env` you can update the schema and the service files by running the script `dev-env/update-openapi.sh`.
+
 ### Front end
 
 To correct the lint of frontend files, it is possible to activate plugin
--- a/dev-env/update-openapi.sh
+++ b/dev-env/update-openapi.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -Eeuo pipefail
+
+DIR="$(realpath -e "$(dirname "$0")")"
+cd "$DIR"
+
+set -x
+
+curl -o ../frontend/scripts/openapi.yaml http://localhost:8000/schema/
+docker exec tournesol-dev-front scripts/generate-services-from-openapi.sh


dev-env/update-openapi
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
amatissart, line 3: In other scripts, such as "run-docker-compose.sh", we are using
```bash
SCRIPT_DIR="$(realpath -e "$(dirname "$0")")"
```
which seems more readable. Is it equivalent?

dev-env/update-openapi
```diff
@@ -0,0 +1,8 @@
...
+
+set -eux
```
amatissart, line 5: out of curiosity, any reason not to set the options at the start of the script?

amatissart, line 5: out of curiosity, any reason not to set the options at the start of the script?

amatissart (commented):
This scripts seems useful indeed. I agree that ideally `update-schema` would work directly in the container, but at first glance even with curl installed, it would also require additional configuration to access the API via the `http://api:8000`.So this approach looks good to me for now.

@sigmike Could you also mention this script in `dev-env/README` and/or `CONTRIBUTING.md`, to make this script more discoverable (and ensure it will be maintained).

dev-env/update-openapi
sigmike: There are a lot of comments on https://stackoverflow.com/questions/59895/how-can-i-get-the-source-directory-of-a-bash-script-from-within-the-script-itsel. I took the most upvoted solution while still being a one-liner. It seems to handle more cases, but we probably don't need to handle all these situations so the simpler one is probably enough.

Apparently using `${BASH_SOURCE[0]}` instead of `$0` makes it work if the script is evaluated with `source my_script`, and `realpath` is not always available (requires coreutils on macOS).

I didn't think to look at the other scripts. I'll copy them for consistency.

dev-env/update-openapi
sigmike: Because I didn't want the whole `SCRIPT_DIR` evaluation to be printed (with `-x`). I'll instead do the same as in other scripts.

sigmike:
@amatissart I made the changes.

amatissart (approved):

PR #886, opened 2022-05-08
inconsistencies part 2/2 : cyclic inconsistencies (#671)

glerzing:
Please double-check that there is no mistake in the function _count_cycles_and_comparison_trios.

I actually prefer to avoid also implementing an API endpoint that returns a single cycle, but of possibly longer length, because:
- it requires time and effort, and complicates things
- It would be complicated to display on the frontend
- It doesn't provide overview on how many cycles are left (we can't computationally count the cycles of any length, this would be NP-complex)

### DIFF
--- a/backend/tournesol/serializers/inconsistencies.py
+++ b/backend/tournesol/serializers/inconsistencies.py
@@ -2,6 +2,49 @@ from rest_framework import serializers
 from rest_framework.serializers import Serializer
 
 
+class CycleSerializer(Serializer):
+    """
+    Serializer for one cycle of length 3
+    """
+    criterion = serializers.CharField()
+    entity_1_uid = serializers.CharField()
+    entity_2_uid = serializers.CharField()
+    entity_3_uid = serializers.CharField()
+
+
+class CyclesStatsSerializer(Serializer):
+    """
+    Stats for each criterion
+
+    Includes the number of cycles, and the number of "comparison trios". This refers
+    to any 3 entities that all have a comparison to each other.
+
+    A useful metric could be the ratio of cycles (= cycles_count / comparison_trios_count).
+    We could consider that the better this ratio is, the more consistent the comparisons are.
+    """
+    cycles_count = serializers.IntegerField()
+    comparison_trios_count = serializers.IntegerField()
+
+
+class Length3CyclesSerializer(Serializer):
+    """
+    Return statistics about length-3 cycles in
+    the contributor's comparisons.
+    """
+    count = serializers.IntegerField()
+    results = CycleSerializer(many=True)
+    stats = serializers.DictField(child=CyclesStatsSerializer(), label="criterion")
+
+
+class Length3CyclesFilterSerializer(serializers.Serializer):
+
+    date_gte = serializers.DateTimeField(
+        default=None,
+        help_text="Restrict the search to entities created or edited after this date.\n"
+                  "Accepted formats: ISO 8601 datetime (e.g `2021-12-01T12:45:00`).",
+    )
+
+
 class ScoreInconsistencySerializer(Serializer):
     """
     Serializer for one element
--- a/backend/tournesol/serializers/inconsistencies.py
+++ b/backend/tournesol/serializers/inconsistencies.py
@@ -6,7 +6,7 @@ class CycleSerializer(Serializer):
     """
     Serializer for one cycle of length 3
     """
-    criterion = serializers.CharField()
+    criteria = serializers.CharField()
     entity_1_uid = serializers.CharField()
     entity_2_uid = serializers.CharField()
     entity_3_uid = serializers.CharField()
@@ -14,7 +14,7 @@ class CycleSerializer(Serializer):
 
 class CyclesStatsSerializer(Serializer):
     """
-    Stats for each criterion
+    Stats for each criteria
 
     Includes the number of cycles, and the number of "comparison trios". This refers
     to any 3 entities that all have a comparison to each other.
@@ -33,14 +33,14 @@ class Length3CyclesSerializer(Serializer):
     """
     count = serializers.IntegerField()
     results = CycleSerializer(many=True)
-    stats = serializers.DictField(child=CyclesStatsSerializer(), label="criterion")
+    stats = serializers.DictField(child=CyclesStatsSerializer(), label="criteria")
 
 
 class Length3CyclesFilterSerializer(serializers.Serializer):
 
     date_gte = serializers.DateTimeField(
         default=None,
-        help_text="Restrict the search to entities created or edited after this date.\n"
+        help_text="Restrict the search to comparisons done or edited after this date.\n"
                   "Accepted formats: ISO 8601 datetime (e.g `2021-12-01T12:45:00`).",
     )
 
@@ -50,7 +50,7 @@ class ScoreInconsistencySerializer(Serializer):
     Serializer for one element
     """
     inconsistency = serializers.FloatField()
-    criterion = serializers.CharField()
+    criteria = serializers.CharField()
     entity_1_uid = serializers.CharField()
     entity_2_uid = serializers.CharField()
     entity_1_rating = serializers.FloatField()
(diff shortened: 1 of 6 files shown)


lfaucon:
> I actually prefer to avoid also implementing an API endpoint that returns a single cycle, but of possibly longer length, because:

> it requires time and effort, and complicates things
> It would be complicated to display on the frontend
> It doesn't provide overview on how many cycles are left (we can't computationally count the cycles of any length, this would be NP-complex)

Agreed & Fine by me. The choice to not do something that would take time can always be changed in the future in any case. 

Thanks for this initial version that brings 99% of the value of the feature for cycle detection :)

backend/tournesol/views/inconsistencies.py
```diff
@@ -8,12 +8,188 @@
...
+        Count the cycles of length 3 and the comparison trios.
+        Add the cycles found to the cycles list.
+
+        The cycles returned are in the order 1, 2, 3 with 1 > 2 > 3 > 1.
+        """
+        cycles_count = 0
+        comparison_trios_count = 0
+        for entity_1 in self.all_entities:
+            # Search the cycles
+            for entity_2 in self.outgoing_connections[entity_1].copy():
```
lfaucon, line 166: Why do you need the `copy` operation here? Isn't it very slow to do this copy for every entity?

backend/tournesol/views/inconsistencies.py
```diff
@@ -8,12 +8,188 @@
...
+
+        for comparison in criterion_comparisons:
+            entity_1 = comparison["comparison__entity_1__uid"]
+            entity_2 = comparison["comparison__entity_2__uid"]
+
+            if entity_1 not in self.all_entities:
+                self.all_entities.add(entity_1)
+                self.incoming_connections[entity_1] = set()
+                self.outgoing_connections[entity_1] = set()
+                self.all_connections[entity_1] = set()
```
lfaucon, line 132: Using `defaultdict` would simplify the code as you would not need to initialize all dictionaries.

backend/tournesol/views/inconsistencies.py
```diff
@@ -8,12 +8,188 @@
...
+
+        The cycles returned are in the order 1, 2, 3 with 1 > 2 > 3 > 1.
+        """
+        cycles_count = 0
+        comparison_trios_count = 0
+        for entity_1 in self.all_entities:
+            # Search the cycles
+            for entity_2 in self.outgoing_connections[entity_1].copy():
+                for entity_3 in self.incoming_connections[entity_1] & \
+                                self.outgoing_connections[entity_2]:
```
lfaucon, line 168: The code would feel more readable to me with something like this:
```
                for entity_3 in self.outgoing_connections[entity_2]:
                    if entity_1 in self.outgoing_connections[entity_3]:
```
I don't think it reduces performance, does it?
Maybe we don't need to keep track of incoming edges with that change.

backend/tournesol/views/inconsistencies.py
```diff
@@ -8,12 +8,188 @@
...
+                    self.cycles.append({
+                        "criterion": self.criterion,
+                        "entity_1_uid": entity_1,
+                        "entity_2_uid": entity_2,
+                        "entity_3_uid": entity_3,
+                    })
+
+                # Remove the edge, so that these cycles won't be counted again
+                self.outgoing_connections[entity_1].remove(entity_2)
+                self.incoming_connections[entity_2].remove(entity_1)
```
lfaucon, line 179: Mmmm, I see. Having these operations here makes it much harder for me to keep track of what's hapening during the multiple loops.

I would find something that does not modify the list of connections but keeps track of a `visited` object simpler to reason about.

What do you think?

lfaucon (commented):
This looks quite good to me. I made a few comments but they are minor. I would prefer a second pair of eyes on this feature before merging it. Maybe @farnots or @amatissart would be good candidates for reviewing.

backend/tournesol/views/inconsistencies.py
glerzing: It was a problem to loop over a set, and modify it inside the loop. I think it was this error:
RuntimeError: Set changed size during iteration

backend/tournesol/views/inconsistencies.py
glerzing: Actually, the & operator between sets is very efficient, especially when the sets have few elements in common. I made some experiments on python.

glerzing: Honestly, the sets in incoming_connections and outgoing_connections must be pretty small in practice, so I'm not sure that the gain is significant, but I prefer to keep it like this.

backend/tournesol/views/inconsistencies.py
glerzing: agreed

backend/tournesol/views/inconsistencies.py
glerzing: outgoing_connections[entity_1] should usually contain less than a dozen of entities, so it will be pretty small copies.

backend/tournesol/views/inconsistencies.py
```diff
@@ -8,12 +8,188 @@
...
+                # Entity_2 is preferred
+                self.incoming_connections[entity_2].add(entity_1)
+                self.outgoing_connections[entity_1].add(entity_2)
+
+            elif comparison["score"] < 0:
+                # Entity_1 is preferred
+                self.incoming_connections[entity_1].add(entity_2)
+                self.outgoing_connections[entity_2].add(entity_1)
+            else:
+                continue
```
glerzing, line 150: I shouldn't have put an "else: continue" there. all_connections is used for comparison trios, for which the score doesn't matter.

backend/tournesol/views/inconsistencies.py
glerzing: There were 2 ways I think to do this:
1 - using simple for loops and if (the suggestion of the previous comment), which is better for readability 
2 - using the "&" operator on sets, which is better for performances

I don't know if I can keep using the 2nd method without progressively "removing the edges" where all the cycles were counted, even if it's not great for readability. So I believe it is a choice between these 2 techniques.

amatissart: I agree that using immutable connections here would help to reason about the algorithm, especially for future readers and maintainers.

If "removing the edges" is only used to avoid counting the same cycle multiple times, then an alternative solution would be to consider only the cycles where `entity_1` has the smallest uid among the 3 (i.e adding conditions `entity_1 < entity_2` and `entity_1 < entity_3` when iterating on the connections).

(This also applies to "comparison trios").

backend/tournesol/serializers/inconsistencies.py
```diff
@@ -2,6 +2,49 @@
 from rest_framework.serializers import Serializer
 
 
+class CycleSerializer(Serializer):
+    """
+    Serializer for one cycle of length 3
+    """
+    criterion = serializers.CharField()
```
amatissart, line 9: For consistency with existing API responses, I would prefer to use `criteria` here.   
(We are not very consistent across the codebase :grimacing:, but in all serializers at least `criteria` seems to be used as an invariable noun).

backend/tournesol/views/inconsistencies.py
lfaucon: Ok

backend/tournesol/views/inconsistencies.py
lfaucon: Agreed! Thanks

backend/tournesol/serializers/inconsistencies.py
glerzing: I don't think this is the right spelling in theory, but ok, it is simpler to write it as an invariable noun everywhere

backend/tournesol/views/inconsistencies.py
glerzing: For comparisons trios, I applied a different trick. It turns out that, without removing the edges, each cycle is counted 3 times, and each comparisons trio is counted 6 times. For the comparisons trios I only need a count, so I can just divide it by 6, which is more efficient than incrementing the count only when uid1 < uid2 < uid3.

amatissart (approved):

GresilleSiffle:
I just updated the OpenAPI schema in `main`. I'm going to update this branch to fix the conflict.

GresilleSiffle:
Should we merge @amatissart @glerzing ?

glerzing:
> Should we merge @amatissart @glerzing ?
ok

PR #890, opened 2022-05-08
[front] feat: add the candidate analysis page

lfaucon:
**related to** #830

---

Done:
- wiring the router for the entity analysis page

Todo:
- [x] Review, improve and merge PR #888 about `/polls/:poll_name/entities/uid` endpoint (⚠️ this can be optional if instead we build the page for one candidate out of pulling the scores for all 12 candidates from the recommendation endpoint)
- [ ] Implement minimalistic layout for the candidate analysis that includes:
    - [ ] Large picture of the candidate **// will rework the global layout at the end, see my comments below**
    - [x] The candidates' card without the avatar
    - [ ] The criteria score bar chart **// will do in a second PR, see my comments below**
    - [ ] (nice to have) #881 The ratings distribution per criteria
- [ ] (nice to have) Option to change the scores based on the multiple score modes
- [ ] (nice to have) Candidates rank (1st to 12th) per rating criteria

PR #881, opened 2022-05-05
Criteria score distribution chart

sigmike:
Related issue: https://github.com/tournesol-app/tournesol/issues/866

image

PR #895, opened 2022-05-10
Add score distributions chart to candidate analysis page

sigmike:
Related to https://github.com/tournesol-app/tournesol/pull/890. Imported from https://github.com/tournesol-app/tournesol/pull/881.

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -2,6 +2,12 @@
   "components": {
     "filtersButton": "filters"
   },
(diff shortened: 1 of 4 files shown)


GresilleSiffle:
I just discover that it's currently not possible to see the criteria scores distribution for the candidates, as the back end returns only scores related to public ratings, and all ratings of the `presidentielle2022` poll are private.

Maybe we can simply merge the distribution chart in the analysis page of videos for now.

PR #907, opened 2022-05-12
[front] refactor: sync' the style of candidate & video analysis page

GresilleSiffle:
This small PR aims to harmonize the look and feel of the analysis page of all polls.

**(1)**

The pages have now a dedicated section for the main actions they provide. The candidate page slightly invite the users to check the global ranking (using an outlined button), while the video page invite more frankly the users to compare the video (using a contained button).

**idea**: what about adding a share button next the compare button in the top right corner of the video analysis page? This will allow users to quickly copy / paste link (maybe share the page on social network), and may create a virtuous circle where people share & compare videos more easily.

**(2)**
 
The candidate and video analysis pages now use the same `<Container>`, the same width, and the same global padding. Both page are now more "relaxed" as they have an horizontal size of 1000 pixels, instead of the previous 900.

On the video analysis page, the size 1000 pixels allows the users to see the video in a wide player, while keeping visible all action buttons (compare, and add to the rate later list), and the titles of the charts.

**note**: I also removed the hard-coded maximum width of 1000 pixels in the entity card style, as it prevents it to fit the parent container, making it harder to include it in large layout. I personally think it's a good rule of thumb make this kind of component always take 100 % of the available space, as it make them very re-usable and fully delegate the layout design to the parent page / container.

As a consequence, if this PR is merged, the entity cards will take more horizontal space on the recommendations page. If this is not desirable, I think we should add a bit of CSS in the recommendations page container, instead of tweaking the entity card which will impact several other pages in the application.

### preview
screencapture-localhost-3000-entities-yt-VKsekCHBuHI-2022-05-12-14_30_50

screencapture-localhost-3000-presidentielle2022-entities-wd-Q5829-2022-05-12-14_32_25

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -345,6 +345,9 @@
       "criteriaScores": {
         "title": "Scores per criterion"
       }
+    },
+    "generic": {
+      "compare": "Compare"
     }
   },
   "entityNotFound": {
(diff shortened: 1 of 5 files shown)


frontend/public/locales/en/translation.json
```diff
@@ -371,10 +374,7 @@
     "getTheExtensionButton": "Get the extension",
     "extensionNotAvailableOnYourBrowser": "The extension is not available on your web browser. You may use it on <2>Firefox</2>, <4>Google Chrome</4> or <6>Chromium</6>.",
     "collaborativeContentRecommendations": "Collaborative Content Recommendations",
-    "tournesolPlatformDescription": "Tournesol is an <1>open source</1> platform which proposes a tool for collaborative decision. The main aim of this tool is to <3>collaboratively</3> identify top videos of public utility by eliciting contributors' judgements on content quality. Currently we are also exploring the possibility of using Tournesol's algorithms to evaluates candidates in an election.<br><br>Thanks to the data collected on Tournesol, we hope to contribute to making today's and tomorrow's large-scale algorithms <6>robustly beneficial</6> for all of humanity.",
-    "videos": {
-      "start": "Start"
```
adrausch, line 376: Just to explain the reason of these 3 lines: it's not used because it was supposed to be the home button for the tutorial. But as long as it is not used and we have in mind that uncommenting the tuto button of the home page will not be enough to make it functional, I have no objection to delete these unused lines

adrausch (approved):

frontend/public/locales/en/translation.json
GresilleSiffle: Hello @adrausch 

That's true. In fact those lines have been automatically deleted by `yarn i18n:parse` because the parser found no non-commented occurrence of the translation key `videos.start` in the code. They will be added automatically again when we will enable the video tutorial : )

Issue #669, opened 2022-03-13
[front] feat: add a duration filter in the videos recommendations


glerzing:
A filter on the maximum (and potentially minimum) duration of videos could be added to the recommendations filters. This would be useful for users that don't want to watch long videos.

(if this would overload too much the filters interface, you can reject this issue )

lfaucon:
- [ ] Refactor the recommendation endpoint to enable all current filtering that exist on the route `video`
- [ ] Adapt frontend to use `poll/{name}/recommendations`
- [ ] Adapt extension to use `poll/{name}/recommendations`
- [ ] Deprecate all three `video/` endpoints
- [ ] Implement duration filter on `poll/{name}/recommendations` :)

GresilleSiffle:
I think we have space for an additional filter on nearly all kind of screen size (see the captures below).

What do we think is the best for the users?

- being able to filter by max duration?
- being able to filter by min duration?
- both but exclusive?

In order of usefulness, from the most useful to the less useful for me it would be:
1. be able to filter by max duration (with a textual input where I can set the number of minutes, it's precise and take a small space)
2. be able to filter by min duration

May I propose to create a filter, with a text field, working only with positive integer, able to interact with the API parameter `metadata[duration:lte:int]`?

**mobile**

Below the language filter.

![capture3](https://user-images.githubusercontent.com/39056254/172848622-d319ed9d-9282-4730-a16e-52982ebc9f06.png)

**tablet**

Below the language filter.

![capture2](https://user-images.githubusercontent.com/39056254/172848750-8f62b057-392e-4909-af6b-3d4e5d016828.png)

**desktop**

Between language and criteria or just below language.

![capture](https://user-images.githubusercontent.com/39056254/172848798-0fa293b5-fcc6-4d95-9ad9-8d9ca6e7bdb0.png)

glerzing:
I like the sliders that give an interval, like the range slider or the minimum distance slider (https://mui.com/material-ui/react-slider/). 
But ideally, you would want a slider where the minutes 0 to 5 take more place than the minutes 55 to 60. So, maybe something like the "minimum distance" slider but with discrete values (e.g. 0, 5, 10, 30, 60, ∞) could be cool.

lfaucon:
I agree that filtering by max duration is the most important. Whether a number field (as in https://mui.com/material-ui/react-text-field/#inputs) or the choice options <5min / <20min / unlimited (with unlimited as the default) would be great. 

[...]

PR #912, opened 2022-05-13
[back] feat: rework the metadata filter API

farnots:
**related to** #669

**edited by** @GresilleSiffle 

---

This pull requests adds an internal metadata filter API allowing to:
- apply any explicitly allowed Django's field lookups to the filtered field
- apply any explicitly allowed type casting functions to the filtered value
- protect the back end from all metadata filter lookups and functions not explicitly allowed
- prevent the usage of the Django's ORM special string `__` allowing to follow foreign keys and fields' sub-properties (and maybe more in the future)

This API is entirely overload-able so that the metadata filter of different entity types can be precisely configured.

As a result, it's now possible to filter YT videos by duration using the following syntax:

```
# less than or equal to 600 seconds
metadata[duration:lte:int]=600

# exactly equal to 600 seconds (integer)
metadata[duration::int]=600

#exactly equal to the string "600"
metadata[duration]=600
```

### DIFF
--- a/backend/tournesol/entities/base.py
+++ b/backend/tournesol/entities/base.py
@@ -42,6 +42,16 @@ class EntityType(ABC):
 
     @classmethod
     @abstractmethod
+    def filter_duration_lte(cls, qs, duration):
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def filter_duration_gte(cls, qs, duration):
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
     def filter_search(cls, qs, query: str):
         raise NotImplementedError
 
--- a/backend/tournesol/entities/base.py
+++ b/backend/tournesol/entities/base.py
@@ -10,8 +10,8 @@ from tournesol import models
 
 UID_DELIMITER = ":"
 
-DEFAULT_ALLOWED_FILTER_FUNCS: {'int': int, 'str': str}
-DEFAULT_ALLOWED_FILTER_LOOKUPS: ['gt', 'gte', 'lt', 'lte']
+DEFAULT_ALLOWED_FILTER_FUNCS = {'int': int, 'str': str}
+DEFAULT_ALLOWED_FILTER_LOOKUPS = ['gt', 'gte', 'lt', 'lte']
 
 
 class EntityType(ABC):
@@ -111,10 +111,10 @@ class EntityType(ABC):
         lookup = None
         func = None
 
-        if len(split_op) > 0:
+        if len(split_op) > 1:
             lookup = split_op[1]
 
-        if len(split_op) > 1:
+        if len(split_op) > 2:
             func = split_op[2]
 
         return field, lookup, func
@@ -132,20 +132,22 @@ class EntityType(ABC):
         for operation, values in filters:
 
             field, lookup, func = cls.get_filter_operation(operation)
-            qstring = field
 
             if len(values) > 1:
                 qst = qst.filter(**{"metadata__" + field + "__in": values})
             else:
+                qstring = field
+
                 # The lookup must be explicitly allowed to be applied.
                 if lookup and lookup in cls.get_allowed_filter_lookups():
                     qstring += f'__{lookup}'
 
+                filtered_value = values[0]
                 # The function must be explicitly allowed to be applied.
                 if func:
-                    sanitized_value = cls.cast_filter_value(values[0], func)
+                    filtered_value = cls.cast_filter_value(filtered_value, func)
 
-                qst = qst.filter(**{"metadata__" + qstring: sanitized_value})
+                qst = qst.filter(**{"metadata__" + qstring: filtered_value})
 
         return qst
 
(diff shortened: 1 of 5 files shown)


GresilleSiffle:
Currently reviewing :robot:

GresilleSiffle:
Thanks @farnots for your work :+1: 

This PR adds the methods `filter_duration_lte` and `filter_duration_gte` directly in the `EntityType` abstract base class, and it looks a bit hard to maintain. Each time we will want to filter with less than / greater than on a new metadata, we will need to add more methods, making the current implementation non-reusable and specific.

How about having a generic less than / greater than filtering system, like the current one with `metadata`?

To be more precise, the current `metadata` filter is able to filter on any metadata, regardless of the metadata name. We could use a similar system, with less code, to filter with less than / greater than regardless of the metadata name, with only one method.

To achieve this, we can either rework the implementation of `EntityType.filter_metadata`, to check if the provided metadata key ends with `_lte` or `_gte` ; or we could add new filters as URL parameters, named `metadata_lte` / `metadata_gte` which will require more code I think.

Here are the options we have:
1. the current implementation ( specific, non-reusable )
2. rework the `EntityType.filter_metadata` method to enhance the capabilities of the current `metadata` URL parameter
3. add new URL parameters, `metadata_lte` and `metadata_gte`
4. maybe something else?

What are your opinion about this @farnots @amatissart and @lfaucon ?

When I first thought about this feature I initially imagined a generic filter, so I'm in favor of the option 2.

lfaucon:
Funny that it already works: https://api.tournesol.app/polls/videos/recommendations/?metadata[language__lte]=e 

`language__lte="e"` returns mostly german (`"de"`) videos

farnots:
Thanks a lot @GresilleSiffle and @lfaucon for your feedback.

I've implented your suggestion and a call `/polls/videos/recommendations/?metadata[duration__gte]=120&metadata[duration__lte]=200` directly on the metadata field will works for any other numeric values. 

The only difference is that the parameter `duration__lte` have double `_` and not only one in previous implementation. Does it looks ok for you ?

GresilleSiffle:
Hello @farnots 

After a discussion with @amatissart and @lfaucon we realized that the current approach relies too much on the trust we have in the user input. Not sanitizing the content inside the bracket will allow users to perform buggy operations, and potentially even worse.

If there is no objection I'll continue this PR, after the yesterday discussion I have clear ideas in mind.

To sum-up we decided to standardize how the API consumer can ask for an operation, like `lte` that targets a particular data type, like string or number. I'm going to test the following approaches, and see how complex they are:
- `metadata[duration]lte:int=600`
- `metadata_numeric_lte[duration]=600`

Depending on the difficulty I might also test:
- `metadata[duration::lte::int]=600`

lfaucon:
> After a discussion with @amatissart and @lfaucon we realized that the current approach relies too much on the trust we have in the user input. Not sanitizing the content inside the bracket will allow users to perform buggy operations, and potentially even worse.

Note that this was a problem with the previous implementation, not this PR changes.

> If there is no objection I'll continue this PR

Please go ahead 👍

GresilleSiffle:
I reworked the metadata filter API to define a standard way for the back end API consumers to safely make requests involving complex metadata filters. 

Now:
- it's possible to cast the submitted filter value to `str` or `int`
- the available casting functions are defined in an authorized list
- it's possible to use Django's fields lookups like `lt`, `lte`, `gt` and `gte`
- the available field lookups are defined in an authorized list

To get all videos with a duration inferior or equal to 600:

```
metadata[duration:lte:int]=600
```

To get all videos with a duration strictly equal to 600:

```
metadata[duration::int]=600
```

To get all videos with a duration strictly equal to the string "600":

```
metadata[duration]=600
# or
metadata[duration::str]=600
```

There are still few things to rework:
- before this PR it was possible to use the Django delimiter `__` to access a child property of a metadata, it's still possible
- before this PR it was possible to use any Django's field lookups, and it's still possible

We need to decide how to deal with this Django "pass through" syntax.

One way could be to completely forbid the usage of `__` in the metadata field name. It will have the following effects:
- **(a)** it won't be possible to maliciously use any Django's field lookups, only those authorized by the metadata API will work
- **(b)** it won't be possible to access a child property of a metadata (without defining a new convention)

Regarding **(a)** the following syntax should raise a DRF `ValidationException`.

```
metadata[views__lte]=600
# or
metadata[name__contains]=600
# or
metadata[date__week]=600
```

While this one will be accepted if explicitly configured:

```
metadata[views:lte]=600
# or
metadata[name:contains]=600
# or
metadata[date:week]=600
```

backend/tournesol/entities/base.py
```diff
@@ -19,27 +20,166 @@ class EntityType(ABC):
...
+    def _get_meta_filter_func(cls, asked_func: str):
+        """
+        If `asked_func` is present in the allowed metadata filter functions,
+        return the matching callable, return `None` instead.
+        """
+        allowed_funcs = cls.get_allowed_meta_filter_funcs()
+
+        if asked_func in allowed_funcs:
+            return allowed_funcs[asked_func]
+        return None
```
lfaucon, line 86:
```suggestion
        return allowed_funcs.get(asked_func)
```

lfaucon (approved):

Issue #530, opened 2022-02-02
refactor: use the new API `/polls/` instead of `/video/` to get recommendations


GresilleSiffle:
**requires** #529

also see the issue #509 

---

The new recommendations can be obtain from the endpoint `/polls/{pollname}/recommendations/`.

`{pollname}` will be equal to `videos`

PR #913, opened 2022-05-13
#530 Video recommandation path in extension

farnots:
Relate to #530

I don't know how to test the browser extension. I don't know if the
changes are enough in this area. If you have any feedback thanks.

### DIFF
--- a/browser-extension/src/addTournesolRecommendations.js
+++ b/browser-extension/src/addTournesolRecommendations.js
@@ -130,7 +130,7 @@ const getTournesolComponent = () => {
 
     const video_thumb = document.createElement('img');
     video_thumb.className = 'video_thumb';
-    video_thumb.src = `https://img.youtube.com/vi/${video.video_id}/mqdefault.jpg`;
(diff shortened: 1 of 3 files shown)


amatissart:
@farnots A few more changes will be required to make this work. 

For example, the date format and the metadata filters need to be adapted in the request parameters:
https://github.com/tournesol-app/tournesol/blob/7294b1aade989dcf059c6bc7eff7c291a93db476/browser-extension/src/background.js#L268-L278

And you can add the "unpacked" extension to your browser to check the behavior while you are making your changes.  
e.g on Chrome/Chromium: https://developer.chrome.com/docs/extensions/mv3/getstarted/#unpacked

farnots:
Thanks for you advice @amatissart. I update the date format and the structure use in for the card display (now using metada). I think it should be OK now, at least it's working well in local : 

image

GresilleSiffle:
Currently testing...

GresilleSiffle (approved):

PR #920, opened 2022-05-15
Fix Github FUNDING.yml

aidanjungo:
To fix that:

image


aidanjungo:
@lfaucon already did the same PR

PR #924, opened 2022-05-16
Video suggestions for comparisons

Foebus:
[Backend] Add a new route to get recommendations on video to compare

This is somewhat a not so small PR, but I will try to explain its structure here.

The PR is very separated from what is already in place, with as only junction points, the Recommender class that gets information from the DB and a dedicated View that calls functions in the same class. Also the data classes are built using models from the DB, but the complex algorithmic part is supposed to be completely agnostic of the organization of the data in the DB, allowing easier changes in the future.

More practically, the inputs and outputs of this part are the following :
Input => User making the request, concerned poll, optionally the first selected video and comparisons from DB
Output => list of recommended video ids (it may be a good idea to pair those with the proportion of information we gain from those / how those may be liked by the user)

The class Recommender is the only one that should be used outside of the recommendation module, as it has the role of interfacing, handling and lazily loading the graphs that are specific to the users. It should be instantiated once and then only its get_XXXX_video_recommendation functions called to get the relevant video recommendation list.
This class may also be maintained up to date by calling regularly do_offline_computation and registering any new comparison

The class Graph is the main one, doing most of the computations and algorithmic application. It is completely handled by Recommander. For now it is specifically thought to handl videos, but it should be easy to generalize it.

The class Video is a data class used to store only the few information needed for this purpose, some of the base function being overwritten to simplify the comprehension of the algorithms present in the Graph module

The recommended_user file is similar to the Video one, but even simpler

What is left to do for now:
- [x] Review what has been done
- [x] Maybe refactor/reorganise the module to match the actual project organization (I would greatly take advises on that)
- [x] Write tests for this part
- [x] Comment and improve readibility of the code
- [ ] Add randomization so that the recommended videos are not deterministic
- [x] Apply any improvement you may see :)


GresilleSiffle (commented):
Hello @Foebus 

Thanks a lot for you work :+1: :+1: 

Can you link in the description the related issue? if there is one of course.

I didn't test this new endpoint yet, but I think we can already re-organize some part of the code.

>  Maybe refactor/reorganise the module to match the actual project organization (I would greatly take advises on that)

**(1)** the `recommendation` folder

The new view `entities_to_compare` from the module `tournesol` requires the `Recommender` from the module `recommendation`. This `Recommender` requires the models `Entity`, `Poll` and more from the `tournesol` module.

This is a cyclic dependency between the Django app `tournesol` and the new module `recommendation`.

I suggest to avoid as much as possible this kind of design. It will be a weakness in the future, as for now everything is working, but mainly because of luck: there is a cyclic dependency between folders but not between files **yet**. That's why the back end is able to start, but it won't the day we will have a cyclic import. Unfortunately this design encourages cyclic imports and might encourages developers to reproduce it.

So where should we move this `recommendation` folder?

To help us we can ask ourselves the following question: is the module independent?

Currently `recommendation` is not an independent module as it is tightly coupled with the `tournesol` models and features. The `Graph` and `Recommender` classes clearly require models from `tournesol` and are not usable by other Django apps that don't rely on these models.

Because of this, it appears to me the `recommendation` module should be part of the `tournesol` Django app. Also, suggesting videos to compare is a feature that has its place in the `tournesol` app.

**(2)** the recommendation concept

In the Tournesol project when we talk about a recommendation we refer most of the time to the recommended entities resulting of the aggregation of users' judgement.

Things might be confusing if we start to use this world in new variables, classes that represent another concept: the suggested entities to compare.

My proposition is use a different word. If the API is returning a couple of entities (a comparison), maybe `SuggestedComparison` could do the trick. If the API is returning a single entity, maybe `SuggestedEntity` is better.

I the code we have already a recommendation serializer, a recommendation endpoint, etc. I strongly believe keeping `Recommender` and `recommendation` will confuse new developers joining the project (and even us when we are searching with our IDE strings like `Recommendation` in the source code).

Foebus:
Thanks a lot for the comments :)
I applied the organization ones and will work on the docstring ! (this is what I had in mind for the point "Comment and improve readability of the code")

GresilleSiffle:
> Thanks a lot for the comments :)
I applied the organization ones and will work on the docstring ! (this is what I had in mind for the point "Comment and improve readability of the code")

My pleasure, thanks a lot for asking :ok_hand: 

I'll be available on the Discord vocal chat in case of need.

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+    # well actually yes but not used in most of the graphs
+    def update_preferences(self):
+        """
+        Function computing online the estimated preferences with respect to another video
+        """
+        for v in self.nodes:
+            for u in set(self.graph[v]):
+                v.nb_comparison_with[u.uid] = 0
+            for u in self.graph[v]:
+                v.nb_comparison_with[u.uid] += 1
```
amatissart, line 307: still useful?

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+from tournesol.suggestions.suggested_user import SuggestedUser as RecommendationUser
+from tournesol.suggestions.suggested_video import SuggestedVideo
+
+
+class Suggester:
+    """
+    An interface class to ask for suggestions of videos to compare
+    """
+    # Entity to specific video structure dictionary
+    _entity_to_video: dict[Entity, SuggestedVideo]  # todo Use only uid as key
```
amatissart, line 15: Using uids as keys is an important improvement to reduce the memory usage of this structure. No reason to keep all entities metadata in RAM.

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+        # build complete graph
+        comparison_queryset: QuerySet = ComparisonCriteriaScore.objects \
+            .filter(comparison__poll__name=self.poll.name) \
+            .filter(criteria=self.criteria)
+        self._complete_graph = Graph(None, self.poll, self.criteria)
+
+        for c in comparison_queryset:
+            # Checks if each compared Entity has already been translated to a SuggestedVideo object
+            # and translates it otherwise
+            if c.comparison.entity_1 not in self._entity_to_video.keys():
```
amatissart, line 42: [minor]
`.keys()` is not required here
```suggestion
            if c.comparison.entity_1 not in self._entity_to_video:
```

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+        return list(map(lambda entity: self._entity_to_video[entity], req_entities))
+
+    def _get_user_rate_later_video_list(self, user: User) -> list[SuggestedVideo]:
+        """
+        Function to get the list of videos of the user's rate later list
+        """
+        req_entities = Entity.objects \
+            .filter(type=self.poll.name) \
+            .filter(videoratelater__user__email=user.email)
+        return list(map(lambda entity: self._entity_to_video[entity], req_entities))
```
amatissart, line 90: Still useful?

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+            local_user=recommendation_user,
+            local_poll=self.poll,
+            local_criteria=self.criteria,
+        )
+        # build user graph
+        query: QuerySet = (
+            ComparisonCriteriaScore.objects
+                                   .filter(comparison__poll__name=self.poll.name)
+                                   .filter(criteria=self.criteria)
+                                   .filter(comparison__user__email=new_user.email)
```
amatissart, line 121: Here tool, email should not be used, the user can be used directly. Django will use the primary key automatically.
```suggestion
                                   .filter(comparison__user=new_user)
```

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    An interface class to ask for suggestions of videos to compare
+    """
+    # Entity to specific video structure dictionary
+    _entity_to_video: dict[Entity, SuggestedVideo]  # todo Use only uid as key
+    _comparison_reference_video: SuggestedVideo
+    # Graph containing all videos and existing comparisons, used to get the video preferences
+    _complete_graph: Graph
+    # Dictionary linking a user to its comparison graph, used to get its information gain
+    _user_specific_graphs: dict[User, Graph]
```
amatissart, line 20: `user.id` can be used as keys instead of storing the whole User instance.

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+        user_graph.prepare_for_sorting()
+
+        # Prepare the set of videos to sort, taking the videos present in the graph
+        # and append the ones that are not yet compared by the user
+        considered_vids_list = self._prepare_video_list(user)
+
+        max_vid_pref = 0
+        # Todo : take into account the rate later list / already seen videos ?
+        for v in considered_vids_list:
+            v.user_pref = max(v.nb_comparison_with.values()) / v.comparison_nb
```
amatissart, line 163: How could this max() be different from 1?

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+                max_vid_pref = v.user_pref
+
+        for i in range(nb_video_required):
+            for v in considered_vids_list:
+                act_user_pref = v.user_pref
+                if (
+                        act_user_pref
+                        >= (nb_video_required - i) / (nb_video_required + 1) * max_vid_pref
+                        and v not in result
+                ):
```
amatissart, line 174: I don't understand what's going on here?

backend/tournesol/views/entities_to_compare.py
```diff
@@ -0,0 +1,37 @@
...
+
+        opt_first_entity = self.request.query_params.get("first_entity_uid")
+        limit = int(self.request.query_params.get("limit", 10))
+        if opt_first_entity is None:
+            entities = suggester.get_first_video_recommendation(user, limit)
+        else:
+            entities = suggester.get_second_video_recommendation(user, opt_first_entity, limit)
+
+        ser = self.get_serializer(entities, many=True)
+        return Response(ser.data)
```
amatissart, line 37: The response format is not consistent with the generated documentation. Even if we don't need to implement the pagination for now, the list of results should be wrapped in an object:
```suggestion
        return Response({"results": ser.data})
```

amatissart (commented):
Cool to see the progress here!

My understanding of the algorithm is still superficial. So most of my comments are actually questions about how all the pieces are working together. I think there is room for simplification in the data structures and how they are shared from one component to another. To discuss later on.

More importantly though, I am facing a few errors when running this branch on my local environment. (see my comments about exceptions). So it could be worth spending some time writing unit tests with a larger "set up" to improve the coverage of the algorithm.

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+
+        act_root = starting_node
+        depth = 0
+
+        while len(unvisited) > 0:
+            future_visits = set()
+            for act_vid in waiting_for_visit:
+                # act_vid = waiting_for_visit.pop()
+                visited.append(act_vid)
+                unvisited.remove(act_vid)
```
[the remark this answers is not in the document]
Foebus: This is by design, this error should not be raised and if it is it shows that there is an error somewhere else
Did it raise when you ran the tests ?

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def __init__(self, local_user: Optional[SuggestedUser], local_poll: Poll, local_criteria):
+        self.local_user_mean: float = 0
+        self.video_comparison_reference = SuggestedVideo(None)
+        self.dirty = True
+        self.edges = []
+        self.graph = {}
+        # init absent node values
+        self.ABSENT_NODE.video1_score = -1
+        self.ABSENT_NODE.estimated_information_gains = 0.75
```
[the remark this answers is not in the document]
Foebus: absolutely, fixed!

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+            self.adjacency_matrix[v_index][u_index] = 1
+        for u, v in self.edges:
+            u_index = self.nodes.index(u)
+            v_index = self.nodes.index(v)
+            self.normalized_adjacency_matrix[u_index][v_index] = 1 / sum(
+                self.adjacency_matrix[u_index]
+            )
+            self.normalized_adjacency_matrix[v_index][u_index] = 1 / sum(
+                self.adjacency_matrix[u_index]
+            )
```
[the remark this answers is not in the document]
Foebus: The idea here is to have the sum of each row(or column) being equal to 1, with each value being either 0 or the same value as the other in the same row(or column), and as the elements of the adjacency matrix are either equal to 1 or 0, it implies that 1 divided the sum over the row is the value to put in the normalized adjacency matrix
So the function on numpy is probably completely adapted, I will check

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+    distance_matrix: dict[SuggestedVideo, dict[SuggestedVideo, int]]
+    similarity_matrix: np.array
+
+    @property
+    def nodes(self) -> list[SuggestedVideo]:
+        return self._nodes
+
+    def __init__(self, local_user: Optional[SuggestedUser], local_poll: Poll, local_criteria):
+        self.local_user_mean: float = 0
+        self.video_comparison_reference = SuggestedVideo(None)
```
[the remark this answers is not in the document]
Foebus: Yes, it is, and will be even more once we de-couple the offline

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+        self.dirty = True
```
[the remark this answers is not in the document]
Foebus: The dirty means here that the information has changed since the last offline computation, and when the class is built, the computed elements are not yet computed so it makes sense to initialize it to True while the offline computation is not done (allowing for example to avoid sending a response that wouldn't make sense if a request is received while the first offline computation is in progress)
The term dirty is often used for that in low level programming, so I thougth it was clear, but it may make sense to rename it ?

backend/tournesol/suggestions/suggested_user.py
```diff
@@ -0,0 +1,42 @@
...
+    score_uncertainties: dict[SuggestedVideo, float]
+
+    def __init__(
+        self,
+        entity_to_video: dict[Entity, SuggestedVideo],
+        base_user: UserDB,
+        local_criteria: str,
+        concerned_poll: Poll,
+    ):
+        self.uid = base_user.email
```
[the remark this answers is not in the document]
Foebus: Ok, I didn't find that one, thanks :)

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+        the video scores otherwise
+        """
+        if self.dirty and self._local_user is not None:
+            self.dirty = False
+            self.build_adjacency_matrix()
+            self.build_distance_matrix()
+            self.build_similarity_matrix()
+            self.local_user_scaling = ContributorScaling.objects.filter(
+                user__email=self._local_user.uid
+            )[0]
```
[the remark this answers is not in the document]
Foebus: Thanks :)

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+            self.local_user_mean = (
+                ContributorRatingCriteriaScore.objects.filter(
+                    contributor_rating__user__email=self._local_user.uid
+                )
+                .filter(contributor_rating__poll__name=self._local_poll.name)
```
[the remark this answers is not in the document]
Foebus: Thanks :)

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+                .aggregate(mean=Avg("score"))
+            )["mean"]
+
+            self.compute_information_gain(scaling_factor_increasing_videos)
+
+        elif self._local_user is None:
+            entity_criteria_scores: QuerySet = EntityCriteriaScore.objects.filter(
+                comparison__poll__name=self._local_poll.name
+            ).filter(criteria=self._local_criteria)
```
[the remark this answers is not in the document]
Foebus: Thanks again :)

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+                        va.beta[vb] = user.delta_theta.get(vb, actual_scaling_accuracy) + \
+                            user.delta_theta.get(va, actual_scaling_accuracy) / (
+                            user.theta.get(vb, self.local_user_mean) -
+                            user.theta.get(va, self.local_user_mean) + 1
+                            )
+                        if max_beta < va.beta[vb]:
+                            max_beta = va.beta[vb]
+                for vb in sg.nodes:
+                    for va in self.nodes:
+                        va.video2_score[vb] += va.beta[vb] / max_beta
```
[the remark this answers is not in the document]
Foebus: This is really weired, it should never happen, if you agree, it would be great if we could look what happens together (the only case where I see this possible is if all the videos compared by the user have zero as uncertainty for the score)

backend/tournesol/suggestions/suggester.py
```diff
@@ -0,0 +1,236 @@
...
+    # Dictionary linking a user to its comparison graph, used to get its information gain
+    _user_specific_graphs: dict[User, Graph]
+
+    def __init__(self, concerned_poll: Poll):
+        """
+        Function used to initialize the class
+        It must not be called before the DB is ready, as it will call it while constructing the
+        complete graph
+        """
+        self._comparison_reference_video = SuggestedVideo(None)
```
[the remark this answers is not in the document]
Foebus: This is a special video, not a real one, used to represent the one to compare on sorting, so the reference one doesn't have a reference itself

backend/tournesol/suggestions/graph.py
Foebus: It was to recompute all the preferences, but no more useful, that's right, thanks :)

backend/tournesol/suggestions/suggester.py
Foebus: It is not yet used, but will, I should take it into account to recommend videos

backend/tournesol/suggestions/suggester.py
Foebus: I refactored this part, let me know if it is clearer this way

backend/tournesol/suggestions/suggester.py
Foebus: The nb_comparison_with is once again maybe not the best name for the variable, it is the comparability between the two videos, and not only the number of comparison between the two. And even if it was, the user can only have one but the graph of all users may have multiple comparisons between the same two videos

backend/tournesol/suggestions/suggester.py
Foebus: done thanks

backend/tournesol/suggestions/suggester.py
Foebus: Done :)

backend/tournesol/suggestions/suggester.py
Foebus: Done :)

backend/tournesol/suggestions/suggester.py
Foebus: It is not necessary, but isn't it better to be more explicit if it doesn't have cost on the execution ?

backend/tournesol/suggestions/graph.py
```diff
@@ -0,0 +1,310 @@
...
+
+class Graph:
+    """
+    Class representing a comparison graph
+    Each node is a video and each node is a comparison
+    """
+    # not necessarily the best data structure,
+    # I still have to find out what is the best one between the two
+    edges: list[tuple[SuggestedVideo, SuggestedVideo]]
+    _nodes: list[SuggestedVideo]  # todo clean that, use default dict functions
```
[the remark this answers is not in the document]
Foebus: Ok, I applied this :)

backend/tournesol/suggestions/suggester.py
amatissart: Ok, but what is still unclear is why we need to define this arbitrary threshold when including the videos in the results. What's the motivation for this instead of simply using the top results ranking by highest "user_pref"?

backend/tournesol/suggestions/suggested_video.py
```diff
@@ -0,0 +1,59 @@
...
+    def __repr__(self) -> str:
+        return "Video " + self.uid
+
+    def __hash__(self):
+        return self.uid.__hash__()
+
+    def score_computation(self, reference: SuggestedVideo):
+        return (self.score_uncertainty + reference.score_uncertainty / (
+                    self.score + reference.score + 1)
+                ) / self.suggestibility_normalization
```
amatissart, line 43: Here `score_uncertainty`, `reference.score_uncertainty` may be uninitialized if no `EntityCriteriaScore` exists for the video yet. 
Should be define a default score here, or ignore the video before sorting by score?

Foebus: Added a default score, sounds more logical for the meaning of this variable

amatissart:
@Foebus I added a fix in 2fd3b6b to avoid exceptions when computing suggestions for a new user. Could you confirm it looks good to you and does not cause or side other edge cases?

amatissart (approved):

Foebus:
> @Foebus I added a fix in [2fd3b6b](https://github.com/tournesol-app/tournesol/commit/2fd3b6b4f1580cb7276f100d4197bbc0dc96aed4) to avoid exceptions when computing suggestions for a new user. Could you confirm it looks good to you and does not cause or side other edge cases?

This looks all good to me :)

PR #947, opened 2022-05-23
Twitterbot bug fix and improvements

aidanjungo:


backend/twitterbot/admin.py
```diff
@@ -19,14 +19,24 @@ class TwitterBotAdmin(admin.ModelAdmin):
...
     list_filter = ["bot_name"]
 
+    @admin.display(description="URL of the tweet")
+    def get_twitter_link(self):
+        """Returns URL of the tweet"""
+        return format_html(
+            '<a href="https://twitter.com/{}/status/{}" target="_blank">Tweet</a>',
+            self.bot_name,
+            self.tweet_id,
+        )
```
amatissart, line 38: Here `self` is the admin instance. The admin page is failing because the tweet object need to be passed as the second argument.  

Either use a `@staticmethod` and replace `self` with `obj` (just like other methods below). Or use `obj` as the second argument after `self`.

aidanjungo: fixed!

amatissart (approved):

PR #951, opened 2022-05-24
Fix metadata language filter in browser extension requests

amatissart:
Following #913 the preferred languages were ignored in the extension recommendations.  

`metadata[language]` needs to be used instead of `language` and the separator `,` is no longer supported to pass multiple values.


GresilleSiffle (approved):
It works!

Thanks for taking the time to fix it, I forgot to test the `language` parameter, my fault.

PR #954, opened 2022-05-27
[back] feat: automatically remove videos from rate later list

lfaucon:
My rate later list is growing unsustainably which makes it lose its purpose. I propose to automatically remove the videos after 4 or more comparisons.

### DIFF
--- a/backend/tournesol/models/entity.py
+++ b/backend/tournesol/models/entity.py
@@ -121,6 +121,19 @@ class Entity(models.Model):
         )
         self.save(update_fields=["rating_n_ratings", "rating_n_contributors"])
 
+    def auto_remove_from_rate_later(self, user):
+        """
+        When called, the entity will be removed from the VideoRateLater
+        list if it has 4 or more comparisons
+        """
+        from .comparisons import Comparison
+
+        n_comparisons = Comparison.objects.filter(user=user).filter(
+            Q(entity_1=self) | Q(entity_2=self)
+        ).count()
+        if n_comparisons >= 4:
+            VideoRateLater.objects.filter(user=user, video=self).delete()
+
     @property
     def entity_cls(self):
         return ENTITY_TYPE_NAME_TO_CLASS[self.type]
--- a/backend/tournesol/tests/test_api_video_rate_later.py
+++ b/backend/tournesol/tests/test_api_video_rate_later.py
@@ -6,6 +6,7 @@ from rest_framework.test import APIClient
 from core.models import User
 from core.tests.factories.user import UserFactory
 from tournesol.models import Entity, VideoRateLater
+from tournesol.tests.factories.comparison import ComparisonFactory
 from tournesol.tests.factories.entity import VideoFactory, VideoRateLaterFactory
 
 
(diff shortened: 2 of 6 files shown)


GresilleSiffle:
I like the idea, I think I missed this feature : )) and I'd like to suggest to make it configurable for the user.

I tend to think automatic behaviours are great as long as the users keep the control of their environment, and are not encouraged/forced to change their habits to match the software "expectations".

As a user I personally rate videos differently depending on few criteria, and an automatic removal after 4 comparisons would be great for most cases, but not all.

I also think we will benefit in the future from a simple preference form in the settings menu, so that we can make Tournesol more accessible through customization. This issue remind me of this.

What do you think of this plan @lfaucon @aidanjungo 
1. adding a preference page with a preference form in the settings menu first (which require designing a preference model in the back end, maybe a JSON property bag pattern could be appropriate)
2. implementing the auto-removal feature

backend/tournesol/models/entity.py
```diff
@@ -121,6 +121,19 @@ def update_n_ratings(self):
         )
         self.save(update_fields=["rating_n_ratings", "rating_n_contributors"])
 
+    def auto_remove_from_rate_later(self, user):
```
GresilleSiffle, line 124: I think we can simply call the method `remove_from_rate_later`.

Because in fact, it's as automatic as the metadata refresh method: it runs only when it's called, so it's not really automatic. Even the method docstring suggests it's an activated effect ("When called, the entity...") and not a triggered effect.

The automatic behaviour is made possible by the call of this method by views, so `remove_from_rate_later` feels more appropriate to me.

backend/tournesol/views/comparison.py
```diff
@@ -117,8 +117,10 @@ def perform_create(self, serializer):
         comparison: Comparison = serializer.save()
         comparison.entity_1.update_n_ratings()
         comparison.entity_1.inner.refresh_metadata()
+        comparison.entity_1.auto_remove_from_rate_later(self.request.user)
         comparison.entity_2.update_n_ratings()
         comparison.entity_2.inner.refresh_metadata()
+        comparison.entity_2.auto_remove_from_rate_later(self.request.user)
```
amatissart, line 123: `auto_remove_from_rate_later()` is currently specific to the poll "videos" and relies on the model `VideoRateLater`. Should we add a condition on the poll name here (or inside the method) to avoid unnecessary checks for other polls?

backend/tournesol/models/entity.py
lfaucon: I was trying to find a better name. 

When calling a method called `remove_from_rate_later` I expect the entity is removed from the rate later list, not that it is removed condittionally to having been compared 4 times. Leaving the keyword `auto` make it sound like it doesn't always do the removal.

Any other naming ideas?
- `check_should_remove_rate_later`
- `remove_rate_later_if_enough_comparisons` (<- serious)

backend/tournesol/views/comparison.py
lfaucon: It currently works because `VideoRateLater.video` is a foreign key to `Entity` so any entity could be removed from it.

VideoRateLater seems very easy to convert into EntityRateLater and might be a useful feature for other polls? What do you think of instead of adding the condition here, we'd fully migrate VideoRateLater to Entities? Needs not much more than to add a new field `poll`.

lfaucon: Or, are you suggesting it is not desirable because it hinders performance unnecessarily?

backend/tournesol/models/entity.py
GresilleSiffle: You're right, `auto` suggests the entity is not always removed.

Maybe this?

- `remove_rl_when_n_comparisons(n_comparisons=4)`

I have no better idea, so I'm fine with the original `auto_remove_from_rate_later`.

backend/tournesol/views/comparison.py
GresilleSiffle: > VideoRateLater seems very easy to convert into EntityRateLater and might be a useful feature for other polls?

Yes I think the rate-later list could be relevant for any poll.

> What do you think of instead of adding the condition here, we'd fully migrate VideoRateLater to Entities?

I like this idea, it will make the code base even more clean by removing specific code that is not that specific. It will also implies to move the endpoint rate later to a poll-scoped endpoint. (can be done in several PR.)

backend/tournesol/tests/test_api_video_rate_later.py
```diff
@@ -242,3 +243,57 @@ def test_authenticated_can_delete(self):
         )
         self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
         self.assertEqual(user.videoratelaters.count(), 0)
+
+    def test_automatically_removed_after_4_comparisons(self):
```
GresilleSiffle, line 247: I think this should be part of the comparisons API's tests, because it's expected that `auto_remove_from_rate_later` will be called by the comparisons endpoints, not the rate-later ones.

GresilleSiffle:
To sum-up my previous comments:
- I'm OK with the name of the method `auto_remove_from_rate_later`
- I'm OK with the fact that we call this method for every poll **if** we create an issue to refactor the `VideoRateLater` 
- I think we should move the new test to the comparisons API's test case

Thanks for adding this very exhaustive test :+1: :+1:

lfaucon:
Because `auto_remove_from_rate_later` is not specific to videos (except for the legacy keyword `video`). I would prefer to not add the condittion that we will prefer removing later.

@GresilleSiffle I added one extra test using the comparison API, but I'd propose to keep the test that manually calls  `auto_remove_from_rate_later`. Maybe we want `auto_remove_from_rate_later` to be triggered in another way than through comparison (e.g. nightly) thus, I like that it is tested separately and I think it is in the right place next to rate later list tests.

GresilleSiffle:
> @GresilleSiffle I added one extra test using the comparison API, but I'd propose to keep the test that manually calls auto_remove_from_rate_later. Maybe we want auto_remove_from_rate_later to be triggered in another way than through comparison (e.g. nightly) thus, I like that it is tested separately and I think it is in the right place next to rate later list tests.

Perfect. I also think it's good to have tests focused on the method itself in addition to those related to the API. I would have place these tests in a specific file related to the `Entity` model, so that the behaviour of an entity (its methods) can be found in one place instead of being spread in the different API test cases, which is in my opinion harder to maintain. But I'm fine with the current code :ok_hand:

Issue #949, opened 2022-05-24
[back] feat: Tournesol Bot tweets about top contributors


lfaucon:
Idea: Display the nice sunflower-bar chart that shows the number of comparisons in the last month for the ten top contributors (probably restrict to public)

aidanjungo:
Some questions link to that:

- Should we use only public comparisons?
- Do we ask people to be feature in this graph?
- Should we have this feature somewhere on Tournesol website and once a month take a "screenshot" and tweet it? or should it be only linked the the TwitterBot model?

lfaucon:
Proposition of answers


> * Should we use only public comparisons?

Public is safer

> * Do we ask people to be feature in this graph?

No, because it's already public data

> * Should we have this feature somewhere on Tournesol website and once a month take a "screenshot" and tweet it? or should it be only linked the the TwitterBot model?

That would be nice to also see it in a page of the website.
I'd propose to have a cronjob that generates the right png in the server's static file such that it is easily accessible
I think it is easiest to start implementing this feature inside the Django twitter app for now. If we get more data-analytics close to the django app, we would later create a separate app for it.

GresilleSiffle:
Thanks to #956 the back end is now able to compute the top n contributors of any month of any year. It also prives a shortcut to get the top n contributors of the last month.

lfaucon:
Thanks @aidanjungo !

PR #956, opened 2022-05-29
[back] feat: get top contributors from the previous month

aidanjungo:
First part of  #949 to retrieve the top contributor of the previous month

If this part is accepted, the second part will be to generate the graph.


backend/twitterbot/top_contributors.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def get_previous_month_top_public_contributor():
+    """Return the top public contributor of the previous month."""
+
+    now = datetime.now()
+    last_month = datetime(now.year, now.month, 1) - timedelta(days=15)
+
+    public_data = ContributorRating.objects.filter(
+        is_public=True, poll__name=DEFAULT_POLL_NAME
```
GresilleSiffle, line 16: Instead of `DEFAULT_POLL_NAME` here, I'd use a variable received as a parameter by `get_previous_month_top_public_contributor`, mainly because this method could be useful anywhere in the back-end, for any poll. 

```python
def get_previous_month_top_public_contributor(poll_name: str)
```

backend/twitterbot/top_contributors.py
```diff
@@ -0,0 +1,40 @@
+import collections
+from datetime import datetime, timedelta
+
+from tournesol.models.comparisons import Comparison
+from tournesol.models.poll import DEFAULT_POLL_NAME
+from tournesol.models.ratings import ContributorRating
+
+
+def get_previous_month_top_public_contributor():
```
GresilleSiffle, line 9: If this method is able to return several users, we should use the plural.

```python
def get_last_month_top_public_contributors(poll_name: str):
```

backend/twitterbot/top_contributors.py
```diff
@@ -0,0 +1,40 @@
...
+    """Return the top public contributor of the previous month."""
```
GresilleSiffle, line 10: Same here, we can use the plural:

```suggestion
    """
    Return the top contributors of the previous month, based on the number of
    their public comparisons.
    """
```

backend/twitterbot/top_contributors.py
```diff
@@ -0,0 +1,40 @@
...
+
+    now = datetime.now()
+    last_month = datetime(now.year, now.month, 1) - timedelta(days=15)
```
GresilleSiffle, line 13: Should we use `timedelta(days=30)`?

backend/twitterbot/top_contributors.py
```diff
@@ -0,0 +1,40 @@
+import collections
```
GresilleSiffle, line 1: I think this file should be in `backend/tournesol/utils/top_contributors.py`.

- This file has currently no "grip" with the twitter bot
- It is very generic, and could be used anywhere in the application
- Computing the top contributors should be the responsibility of the application providing the `Comparison` and `ContributorRating` model

GresilleSiffle (dismissed):
Hello @aidanjungo 

Thanks for the PR : )) I made a first review with some suggestions, tell me what do you think.

After that I'll focus on the database request, I think we might be able to get the public comparisons with only one request.

backend/twitterbot/top_contributors.py
aidanjungo: No, it's a trick!
With a time delta of 30 days we could get e.g.:
- May instead of April by asking on the 31st of May
- January instead of February by asking on the 1st of March

backend/twitterbot/top_contributors.py
aidanjungo: Cool, I wasn't sure about this one, it is indeed better as you propose.

backend/twitterbot/top_contributors.py
aidanjungo: 👍

backend/twitterbot/top_contributors.py
aidanjungo: I agree, I changed it.

backend/twitterbot/top_contributors.py
GresilleSiffle: Oh I see.

Should we make the month a parameter of the function then? For instance we can ask for the top contributors of may, and we are sure the function will return the same result independently of the current time. I think it's better to rely a maximum on predictable functions, but as I don't want to give you more work we could do this in two PR : )

aidanjungo: Yes we could, but as we want to automatize that, we will have the same issue in the function which call this one. 
In the issue @lfaucon propose that we create automatically a PNG (I imagine the 1st of each month for the previous month) and we use this PNG to create the tweet:

> I'd propose to have a cronjob that generates the right png in the server's static file such that it is easily accessible
I think it is easiest to start implementing this feature inside the Django twitter app for now. If we get more data-analytics close to the django app, we would later create a separate app for it.

GresilleSiffle: Ok I understand now.

As long as we don't need to run the function for specific months, using the `timedelta(days=15)` will work perfectly :ok_hand:

GresilleSiffle:
Currently measuring the performance on staging.

With 8821 comparisons made during the last month, it takes approximately 330 ms to return the top contributors. 

```python
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=10)
3.3542445061029866
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=100)
33.7215522590559
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=1000)
337.31275870499667
```

According to `htop`, it requires 50 % of each of the two CPUs of the server.

I was able to display the recommendations in the front end with no perceptible delay.

GresilleSiffle:
To sum, with the original implementation we have the following results

```python
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=1)
1.6593089470406994
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=10)
16.540582462097518
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=100)
168.96002322691493
```

---

Using a raw SQL query to retrieve the public comparisons, and computing the top contributors in Python:

```python
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=1)
0.42846800200641155
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=10)
4.406373874982819
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=100)
44.22426020994317
```

---

Using a raw SQL query to retrieve the public comparisons **and** to compute the top contributors:

```python
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=1)
0.3961547949584201
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=10)
4.029838540009223
>>> timeit.timeit(partial(get_last_month_top_public_contributors, 'videos'), number=100)
38.05701820994727
```

Issue #933, opened 2022-05-18
Improve Criteria Bar Chart on Video Analysis page


adrausch:
TODO:

- [x] Include main criteria in Bar Chart
- [x] Display scores and labels on bars
- [x] Make the chart denser (smaller images, smaller font, less space)
- [x] When clicking (or hovering?) on a criteria, changes the selected criteria on the Histogram. (Maybe always show that one criteria is selected?)

sigmike:
> Display scores and labels on bars

I'm not sure how to do that. I made a few attempts but none of them seems satisfactory:
* adding the criteria to the left (adding them above or below doesn't seem to be possible, and adding the score on the bar doesn't seem to work reliably):
![2022-05-20_18-31](https://user-images.githubusercontent.com/28259/169572268-b2c48a50-8b54-4f62-a9e2-c493c476e584.png)
* using sliders:
![2022-05-20_17-56](https://user-images.githubusercontent.com/28259/169566527-543f7134-b8f4-4dbd-ae95-cb97edce531a.png)
* using custom made bars instead of recharts:
![2022-05-20_18-09](https://user-images.githubusercontent.com/28259/169568672-3df4d306-03a4-47f5-bfb5-95e81419970e.png)

The last 2 make the chart taller, so they don't help making the chart denser.

Any thoughts @lfaucon @adrausch @GresilleSiffle @amatissart @aidanjungo?

amatissart:
@sigmike My preference would be for your 3rd version with the custom bars. However, personally I would find acceptable to remove the criteria names, and keep them visible on hover only. 

To make the chart denser, after some thinking, we could consider drawing the bars on this graph as a "thermometer chart", without any visible minimal bound. Compared to the current bars centered on 0, I see 2 advantages:
 * the layout would be more predictable whatever the exact scores are. The score numbers can always be displayed on the right of each bar, and the bars themselves can be made thinner.
 * it would make a clear distinction between the score 0 and the absence of score, which is not obvious with the current chart. And as we expect most scores to be close to 0, the chart would be more aesthetically pleasing.

Something like this, roughly:
![thin_bars](https://user-images.githubusercontent.com/4726554/169637826-4e86febd-b3aa-4e27-9341-8922e0227937.png)

I am not sure about the tooltips though. Maybe it would be better to use a larger tooltip with the criterion name and some explanation about the scores values?

[...]

PR #958, opened 2022-05-30
Improve bar chart on video analysis page

sigmike:
Related issue: https://github.com/tournesol-app/tournesol/issues/933

For now it only includes the main criteria and changes the design.

I added a dependency to [@react-hook/resize-observer](https://github.com/jaredLunde/react-hook/tree/master/packages/resize-observer#readme) because it was easier, it looks light, and it could be useful elsewhere. But we can also make something similar with custom code.

The new design:
image
image
image

### DIFF
--- a/frontend/src/utils/color.ts
+++ b/frontend/src/utils/color.ts
@@ -1,0 +1,16 @@
+// Copied from https://stackoverflow.com/a/59387478
+export const lighten = (color: string, lighten: number) => {
+  const c = color.replace('#', '');
+  const rgb = [
+    parseInt(c.substr(0, 2), 16),
+    parseInt(c.substr(2, 2), 16),
+    parseInt(c.substr(4, 2), 16),
+  ];
+  let returnstatement = '#';
+  rgb.forEach((color) => {
+    returnstatement += Math.round(
+      (255 - color) * (1 - Math.pow(Math.E, -lighten)) + color
+    ).toString(16);
+  });
+  return returnstatement;
+};
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -8,6 +8,7 @@
     "@mui/icons-material": "^5.2.5",
     "@mui/material": "^5.2.6",
     "@mui/styles": "^5.2.3",
+    "@react-hook/resize-observer": "^1.2.5",
     "@reduxjs/toolkit": "^1.7.1",
     "@types/node": "^12.0.0",
     "@types/react": "^17.0.38",
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -3,12 +3,15 @@
     "filtersButton": "filters"
   },
   "personalCriteriaScores": {
-    "chartTooltip": "Personal score: {{score}}",
     "reasonWhyItCannotBeActivated": {
       "notLoggedIn": "You must be logged in to display personal scores.",
       "noPersonalScore": "You have not compared this video yet."
     },
     "activatePersonalScores": "Display personal scores"
+  },
+  "criteriaBarChart": {
+    "scoreTooltip": "Score calculated by Tournesol",
+    "personalScoreTooltip": "Your personal score"
   },
   "dialogBox": {
     "continue": "continue"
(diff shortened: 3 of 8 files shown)


frontend/src/utils/color.ts
```diff
@@ -0,0 +1,16 @@
+// Copied from https://stackoverflow.com/a/59387478
+export const lighten = (color: string, lighten: number) => {
+  const c = color.replace('#', '');
+  const rgb = [
+    parseInt(c.substr(0, 2), 16),
+    parseInt(c.substr(2, 2), 16),
+    parseInt(c.substr(4, 2), 16),
```
GresilleSiffle, line 7: I discovered that `substr` should be avoided, replaced by `substring`.

```suggestion
    parseInt(c.substring(0, 2), 16),
    parseInt(c.substring(2, 4), 16),
    parseInt(c.substring(4, 6), 16),
```

See: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/substr
See: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/substring

frontend/src/utils/color.ts
```diff
@@ -0,0 +1,16 @@
+// Copied from https://stackoverflow.com/a/59387478
+export const lighten = (color: string, lighten: number) => {
```
GresilleSiffle, line 2: We can add a little documentation string, to precise what is expected in `color` and `lighten`. 

Example, with a non-documented `lighten` parameter.

```suggestion
/**
 * Return a lighten version of an hexadecimal color.
 *
 * Adapted from: Copied from https://stackoverflow.com/a/59387478
 * 
 * @param color An hexadecimal color to lighten.
 * @param lighten A positive integer? number? representing...
 * @returns An hexadecimal lighten version of the original `color`. 
 */
 export const lighten = (color: string, lighten: number): string => {
```

frontend/src/utils/color.ts
GresilleSiffle, line 7: I discovered that `substr` should be avoided, replaced by `substring`.

```suggestion
    parseInt(c.substring(0, 2), 16),
    parseInt(c.substring(2, 4), 16),
    parseInt(c.substring(4, 6), 16),
```

See: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/substr
See: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/substring

frontend/src/utils/color.ts
```diff
@@ -0,0 +1,16 @@
...
+    parseInt(c.substr(2, 2), 16),
+    parseInt(c.substr(4, 2), 16),
+  ];
+  let returnstatement = '#';
+  rgb.forEach((color) => {
+    returnstatement += Math.round(
+      (255 - color) * (1 - Math.pow(Math.E, -lighten)) + color
+    ).toString(16);
+  });
+  return returnstatement;
```
GresilleSiffle, line 15: The `returnstatement` is not really a statement, but rather an expression. We can make it more explicit by describing what kind of thing it represents.

```suggestion
  let lightenColor = '#';
  rgb.forEach((color) => {
    lightenColor += Math.round(
      (255 - color) * (1 - Math.pow(Math.E, -lighten)) + color
    ).toString(16);
  });
  return lightenColor;
```

GresilleSiffle:
LGTM.

It works well on my local dataset.

Does any @tournesol-app/maintainers  want to have a second look at the new `CriteriaBarChart` component?

frontend/src/utils/color.ts
sigmike: Done in 4325c03ea6a76d404b7153c711f827cc3bd626d4.

frontend/src/utils/color.ts
sigmike: Done in 4325c03ea6a76d404b7153c711f827cc3bd626d4.

frontend/src/utils/color.ts
sigmike: Done in 4325c03ea6a76d404b7153c711f827cc3bd626d4.

GresilleSiffle (approved):

PR #968, opened 2022-06-06
extension: Look for new buttons container on Youtube redesign

amatissart:
The rate-later button integration was broken on Youtube new design. And I finally managed to access this new version (on Chromium private browsing only).

The rate-later button will be inserted into the container used by the new design if it's found. The previous design (still accessed by most users) should not be impacted.

And this should also fix the sporadic failures of e2e tests.


GresilleSiffle (approved):
I wasn't able to access the new YT interface, but I confirm the extension works well with the "old" YT selector.

PR #969, opened 2022-06-08
[dev-env] feat: support docker-compose-plugin in addition to docker-compose

cthiounn:
Docker has released a V2 for Docker compose, integrated in Docker CLI, instead of the previous docker-compose binary.

https://docs.docker.com/compose/

By default, Docker suggest to install the docker compose plugin, like in this tutorial : 

https://docs.docker.com/engine/install/ubuntu/

Tournesol's dev_env sh script use the docker-compose binary to set up the dev env. The script can be adapted to automatically detect if that binary is either present or either docker compose plugin is.

enhancements :

* auto detection user's config :
    * A docker-compose is present
    * docker compose plugin is installed
    * None, script should advise to install one
* tear down command

### DIFF
--- a/dev-env/README.md
+++ b/dev-env/README.md
@@ -3,7 +3,7 @@
 ## Quick start
 
 ### Requirements:
-* docker-compose (or docker compose plugin)
+* `docker-compose-plugin` (or `docker-compose`)
 * curl
 * expect
 
--- a/dev-env/README.md
+++ b/dev-env/README.md
@@ -3,7 +3,7 @@
 ## Quick start
 
 ### Requirements:
-* docker-compose
+* docker-compose (or docker compose plugin)
 * curl
 * expect
 
--- a/dev-env/run-docker-compose.sh
+++ b/dev-env/run-docker-compose.sh
@@ -32,7 +32,7 @@ function compose_up_with_compose_plugin() {
 }
 
 function compose_up(){
-  if docker compose version ; then
+  if docker compose version 2>/dev/null; then
       echo "compose_up : docker-compose-plugin found"
       compose_up_with_compose_plugin "$@"
   else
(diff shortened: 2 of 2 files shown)


dev-env/README.md
```diff
@@ -3,7 +3,7 @@
 ## Quick start
 
 ### Requirements:
-* docker-compose
```
GresilleSiffle, line 6:
```suggestion
* docker-compose-plugin (or docker-compose)
```

dev-env/run-docker-compose.sh
```diff
@@ -49,6 +97,13 @@ if [[ "${1:-""}" == 'restart' ]]; then
   exit
 fi
 
+if [[ "${1:-""}" == 'stop' ]]; then
```
GresilleSiffle, line 100: The commands `down` and `stop` have a different meaning for `compose`.

- `down` stops the services and remove the containers
- `stop` only stops the services

Here the script if answering to a `stop` positional argument with a `down`, which can be confusing.

As we have already a way to recreate the container with the argument `restart` (the `restart` triggers a `docker compose up --force-recreate`), I suggest to simply call `docker compose stop` and rename the related function accordingly.

dev-env/run-docker-compose.sh
```diff
@@ -49,6 +97,13 @@ if [[ "${1:-""}" == 'restart' ]]; then
...
+  echo "Stopping dev containers..."
+  compose_down
```
GresilleSiffle, line 102: This suggestion also implies to replace all occurrences of  `compose_down` in the script.

```suggestion
  compose_stop
```

GresilleSiffle (dismissed):

GresilleSiffle (commented):
Works well on Debian 11, with

- `docker-compose-plugin` version 2.6.0
- and also with `docker-compose` version 1.29.2

adrausch:
Tested on my MacOS and it worked for me

dev-env/README.md
```diff
@@ -3,7 +3,7 @@
...
+* docker-compose (or docker compose plugin)
```
GresilleSiffle, line 6:
```suggestion
* `docker-compose-plugin` (or `docker-compose`)
```

GresilleSiffle (approved):

PR #970, opened 2022-06-08
[back] fix: server error on `entities_to_compare` 

amatissart:
A server error could be raised when a new video is present in the database with comparisons, but without contributor score.
This should fix the error reported by @GresilleSiffle in #856.

@Foebus Is `0.0` an appropriate default value for the `score_uncertainty` in `SuggestedUserVideo` ?


GresilleSiffle (approved):
Looks good to me.

I approve so it can be merged quickly after the @Foebus review.

PR #976, opened 2022-06-09
WIP / [front] feat: add duration lte filter (WIP)

GresilleSiffle:
**related to** #669 

---

still very WIP


adrausch (approved):
I'm not really fan of the max before the min but everything work.

Just one question: why not a range slider?

PR #978, opened 2022-06-10
[back] fix: entities_to_compare fails after a new video is compared by a supertrusted

amatissart:
Although a "staff" user was created in the setup of `SuggestionAPITestCase`, it was not correctly configured as a super-trusted (no trusted email).  The usage of `self._entity_to_video` should be safe now and is covered by the unit tests.

Thanks @GresilleSiffle for reporting this bug.

### DIFF
--- a/backend/tournesol/suggestions/suggestionprovider.py
+++ b/backend/tournesol/suggestions/suggestionprovider.py
@@ -75,16 +75,25 @@ class SuggestionProvider:
                     Q(comparisons_entity_2__user__in=User.supertrusted_seed_users()))\
             .distinct() \
             .values_list("uid", flat=True)
-        return [self._entity_to_video[uid] for uid in req_entities]
+        return [
+            video
+            for uid in req_entities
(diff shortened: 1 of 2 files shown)


backend/tournesol/suggestions/suggestionprovider.py
```diff
@@ -75,16 +75,25 @@ def _get_user_comparability_augmenting_videos(self) -> list[SuggestedVideo]:
...
-            .filter(videoratelater__user=user)
-        return list(map(lambda entity: self._entity_to_video[entity.uid], req_entities))
+        rate_later_uids = Entity.objects \
+            .filter(type=self.poll.entity_type) \
+            .filter(videoratelater__user=user) \
+            .values_list("uid", flat=True)
+        return [
+            video
+            for uid in rate_later_uids
+            if (video := self._entity_to_video.get(uid)) is not None
```
lfaucon, line 95: 🤩

lfaucon (approved):

PR #982, opened 2022-06-11
Suggestion to remove Expect

glerzing:
Expect does not exist on Windows. And for Linux users it would remove the need for a dependency and make the environment installation a bit easier.

The solution comes from https://stackoverflow.com/questions/6244382/how-to-automate-createsuperuser-on-django

As always, feel free to criticize if I missed something. I suggest that you also check quickly if it works well in your environment.


glerzing:
According to Wikipedia (https://en.wikipedia.org/wiki/Usage_share_of_operating_systems), the desktop operating system share among those identifying as professional developers was:

Windows: 45.3%
macOS: 29.2%
Linux: 25.3%
BSD/Unix: 0.1%

I think we should open a ticket to make our docker environment more compatible with Windows. 
JST showed me some trick to make it work (removing the line "user: ${DB_UID}:${DB_GID}" in docker-compose.yml and executing run-docker-compose.sh with the MinGW64 bash instead of the WSL2 bash ("./run-docker-compose.sh" instead of "bash run-docker-compose.sh", for some unknown reason, it solves permissions errors)). 
But if almost half of the developers use Windows, it's important not to discourage these potential new contributors. Maybe I will try to propose a solution, since I may be the only one currently using Windows in the team.

GresilleSiffle:
I have no objection.

I also saw that `expect` is mentioned in the back end's `README.me`. We must remove it there too.

amatissart (commented):
Thanks for removing this dependency :+1:

glerzing:
I opened #993. The end-to-end check fail looks like it is random and due to temporarily bad network connection. But I don't know how to just retry, so I rebased and re-pushed.

amatissart (approved):

Issue #950, opened 2022-05-24
EPIC: Migrate videos poll to Mehestan algorithm


lfaucon:
A first problem identified so far is the scaling of scores that is very different to Lichavi's. However this is only a small problem and we can easy scale manually somehow or adapt the visualizations on the frontend.

The Lichavi's algorithm fails to correctly estimate individual preferences and it appears that all individual contributors are very close to the global ranking, which shouldn't be the case.

Before making this change, it would be best to test on staging with data from prod, and check that rankings look roughly correct.

If more tasks are needed to achieve the migration, let's add them on this ticket.

amatissart:
I ran Mehestan with data from a production backup dated today (2022-05-26).
The job run on my local VM (with a single process) in 12min46s on the branch [`mehestan-tournesol-score`](https://github.com/tournesol-app/tournesol/compare/3999c55)
Even though the CPUs may not be as performant, we can expect the job to run significantly faster on the production VM thanks to the criteria parallelization among multiple processes.

<details>
<summary> Click to expand ML logs</summary>

```
May 26 20:19:12 tournesol systemd[1]: Starting Train ML...
May 26 20:19:14 tournesol bash[21329]: NumExpr defaulting to 2 threads.
May 26 20:19:14 tournesol bash[21329]: Mehestan for poll 'videos': Start
May 26 20:19:14 tournesol bash[21337]: Mehestan for poll 'videos': computing scores for crit 'largely_recommended'
May 26 20:19:22 tournesol bash[21337]: Individual scores computed for crit 'largely_recommended'
May 26 20:19:43 tournesol bash[21337]: Computing scaling for 2940 non_supertrusted, based on 72 supertrusted
May 26 20:21:01 tournesol bash[21337]: Mehestan for poll 'videos': scores computed for crit 'largely_recommended' and mode 'default'
May 26 20:21:05 tournesol bash[21337]: Mehestan for poll 'videos': scores computed for crit 'largely_recommended' and mode 'all_equal'
May 26 20:21:07 tournesol bash[21337]: Mehestan for poll 'videos': scores computed for crit 'largely_recommended' and mode 'trusted_only'
May 26 20:21:08 tournesol bash[21337]: Mehestan for poll 'videos': done with crit 'largely_recommended'
May 26 20:21:08 tournesol bash[21337]: Mehestan for poll 'videos': computing scores for crit 'reliability'
[...]

PR #983, opened 2022-06-11
[back] refactor: move constants from core/utils to tournesol/utils

glerzing:
It is a bit frivolous, but Tournesol constants should not logically be in core/utils.


glerzing:
RATING_MAX isn't used anywhere. Should it be removed, or used somewhere?
Is the description even correct? (are ratings in [-100, 100] ?) In ml/managements/commands/ml_train.py, it is said to be "in [0,100] or [-1,1]".

amatissart:
No objection for moving this constant file, more related to "tournesol" indeed.  
I suppose we could still argue whether we could get find a better naming/structure for this "utils" stuff (ping @GresilleSiffle :wink: ).

> RATING_MAX isn't used anywhere. Should it be removed, or used somewhere?

Yes, it seems to be unused for a long time. And it's not relevant neither for the current algorithms, or the future score scale that is being implemented for Mehestan (#950).

Another obsolete constant is `featureIsEnabledByDeFault`: an "optional" flag is now defined for each criterion in the database.

GresilleSiffle:
Yes I think we can improve the structure of the `utils` folder, we can do this in another PR so that we can merge these improvements before.

I'm going to create a task for this.

Concerning the `constants.py` file. I think we can also improve this file, maybe directly in this PR.

For instance, the variables `COMPARISON_MAX` and `RATING_MAX` look more like configuration we might want to change rather than "utils". On the other hand `YOUTUBE_VIDEO_ID_REGEX_SYMBOL` is something we do not want to change, as it must match the behaviour of YouTube.

So, maybe we can leave it as is for now, because this PR is already a satisfying improvement, then, we could decide how to re-organize this `constants.py` file. The configuration could go in the Django's settings file, under the `TOURNESOL_APP` key. The `YOUTUBE_VIDEO_ID_REGEX_SYMBOL` and friends could go in `utils.youtube.py` (or even in a `utils.youtube.__init__.py` so that we can move the existing `api_youtube.py` file inside this new `utils.youtube` folder).

Concerning the `utils`, we could ask ourselves: are these functions very related to the `tournesol` app, or are they relevant even for other Django apps?

If they are only related to `tournesol`, we could rename `utils` by `lib` which will force us to organize all of its sub-folders into independent python module.

If these functions are relevant everywhere in the Django project, and if they might create cyclic dependencies while living in the `tournesol` app, we could create a folder `library` or `lib` at the same level of `core` and `tournesol`. This way the library can become complex at will, and will be able to contain its own `tests` folder. I just don't know if Django is able to run the tests located in a non app folder.

GresilleSiffle (approved):

PR #984, opened 2022-06-12
[back] fix: Unexpected side-effect in unconnected_entities view

amatissart:
Some of the data structures (including the set of "visited nodes") would persist across requests. This is due to how in Python, attributes initialized in the class definition are owned by the Class object itself, and not by the instances.

These attributes initialization were expected to be reset for each request (i.e view instances)
```python
    _already_visited_node_ = set()
    _all_connections_ = defaultdict(set)
```

Here it seems less error-prone to instead move the algorithm inside `get_queryset()` and make the scope of these structures more explicit.

### DIFF
--- a/backend/tournesol/views/unconnected_entities.py
+++ b/backend/tournesol/views/unconnected_entities.py
@@ -30,35 +30,36 @@ class UnconnectedEntitiesView(
     def get_queryset(self):
         # Get related entities from source entity
         source_node = get_object_or_404(Entity, uid=self.kwargs.get("uid"))
-        comparison_pairs = list(
+        comparisons = list(
             Comparison.objects.filter(
                 poll=self.poll_from_url, user=self.request.user
             ).values_list("entity_1_id", "entity_2_id")
         )
 
         all_connections = defaultdict(set)
-        already_visited_nodes = set()
 
-        for (entity_1_id, entity_2_id) in comparison_pairs:
+        for (entity_1_id, entity_2_id) in comparisons:
             all_connections[entity_1_id].add(entity_2_id)
             all_connections[entity_2_id].add(entity_1_id)
 
         def get_related_entities(entity_id):
-            already_visited_nodes.add(entity_id)
-            related_entities = set()
-            if entity_id not in all_connections:
-                return set()
+            already_visited_nodes = set()
+            related_entities = {entity_id}
+            to_visit = {entity_id}
 
-            for other_node in all_connections[entity_id]:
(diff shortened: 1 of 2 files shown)


backend/tournesol/views/unconnected_entities.py
```diff
@@ -27,42 +27,39 @@ class UnconnectedEntitiesView(
...
-
-        self._already_visited_node_.add(entity_id)
-
-        related_entities = set()
-        if self._all_connections_.get(entity_id) is None:
-            return set()
+    def get_queryset(self):
+        # Get related entities from source entity
+        source_node = get_object_or_404(Entity, uid=self.kwargs.get("uid"))
+        comparison_pairs = list(
```
lfaucon, line 33: Why not just `comparisons` ?

backend/tournesol/views/unconnected_entities.py
```diff
@@ -27,42 +27,39 @@ class UnconnectedEntitiesView(
...
+            if entity_id not in all_connections:
+                return set()
 
-        source_node = get_object_or_404(Entity, uid=self.kwargs.get("uid"))
-        comparisons = list(Comparison.objects.filter(
-            poll=self.poll_from_url, user=self.request.user
-        ))
+            for other_node in all_connections[entity_id]:
+                if other_node not in already_visited_nodes:
+                    related_entities.update(get_related_entities(other_node))
```
lfaucon, line 54: `entity_id` should be part of related entities. Right now, it is part of it only because comparisons are bi-directional. That works, but I'd prefer the below implementation

```Python
        def get_related_entities(entity_id):
            already_visited_nodes.add(entity_id)
            related_entities = {entity_id}

            for other_node in all_connections.get(entity_id, []):
                if other_node not in already_visited_nodes:
                    related_entities.update(get_related_entities(other_node))

            return related_entities
```

lfaucon (approved):

backend/tournesol/views/unconnected_entities.py
amatissart: I wanted to describe that this list contains pairs of entities id, but it may bring more confusion. Reverted.

backend/tournesol/views/unconnected_entities.py
amatissart: I agree. After thinking about it a bit more, I also replaced the recursion with a `while` loop. The recursion limit is quite low in Python (1000 by default in most cases).

GresilleSiffle (approved):

Issue #988, opened 2022-06-14
[back] optimize request to public dataset


lfaucon:
With the production database, the public dataset takes about 30 seconds to be generated. We probably should be able to do better than this.

Secondly the frontend could show feedback that the download is already in progress.

GresilleSiffle:
- investigate the SQL perf, possibly switch to a raw SQL
- avoid working with model instances when only values are relevant

GresilleSiffle:
The draft of the final query:

```sql
SELECT
    tournesol_comparison.id,

    core_user.username,
    entity_1.uid AS uid_a,
    entity_2.uid AS uid_b,
    criteriascore_1.criteria,
    criteriascore_1.score

FROM tournesol_comparison

-- this JOIN allows to filter on the desired poll
JOIN tournesol_poll
  ON tournesol_poll.id = tournesol_comparison.poll_id

JOIN core_user
  ON core_user.id = tournesol_comparison.user_id

JOIN tournesol_entity AS entity_1
  ON entity_1.id = tournesol_comparison.entity_1_id

JOIN tournesol_entity AS entity_2
  ON entity_2.id = tournesol_comparison.entity_2_id

-- this JOIN allows to filter by public ratings for the entity_1
-- the poll has been already filtered, no need to filter it again
JOIN tournesol_contributorrating AS rating_1
  ON rating_1.entity_id = tournesol_comparison.entity_1_id
 AND rating_1.user_id = tournesol_comparison.user_id

JOIN tournesol_contributorratingcriteriascore AS criteriascore_1
  ON criteriascore_1.contributor_rating_id = rating_1.id

-- this JOIN allows to filter by public ratings for the entity_2
-- the poll has been already filtered, no need to filter it again
JOIN tournesol_contributorrating AS rating_2
  ON rating_2.entity_id = tournesol_comparison.entity_2_id
 AND rating_2.user_id = tournesol_comparison.user_id

WHERE tournesol_poll.name = 'videos'
  -- keep only public ratings
  AND rating_1.is_public = true
  AND rating_2.is_public = true
```

PR #996, opened 2022-06-15
[back] perf: speedup `ExportPublicComparisonsView` with raw SQL

GresilleSiffle:
**related to** #988 

---

The PR reworks the way the `ExportPublicComparisonsView` used to retrieve the public dataset.

Instead of doing several non-optimized SQL queries due to the Django's ORM limitations, I used a raw SQL query to directly retrieve the public comparisons.

I created a folder `lib`, to avoid storing the raw SQL query in the view, and to invite us to factorize similar functions when relevant.

### DIFF
--- a/backend/tournesol/views/exports.py
+++ b/backend/tournesol/views/exports.py
@@ -69,7 +69,7 @@ def write_public_comparisons_file(poll_name: str, write_target) -> None:
             "weight": comparison.weight,
             "score": comparison.score,
         }
-        for comparison in get_dataset(poll_name)
+        for comparison in get_dataset(poll_name).iterator()
     )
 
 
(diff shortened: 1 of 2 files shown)


backend/tournesol/views/exports.py
```diff
@@ -60,40 +60,16 @@ def write_public_comparisons_file(request, write_target):
...
+            "video_b": comparison.uid_b.split(UID_DELIMITER)[1],
+            "criteria": comparison.criteria,
+            "weight": comparison.weight,
+            "score": comparison.score,
         }
-        for (public_username, comparison) in zip(
-            public_usernames, serialized_comparisons
-        )
-        for criteria_score in comparison["criteria_scores"]
+        for comparison in get_dataset(poll_name)
```
amatissart, line 72: An iterator can be used here to reduce the memory usage while writing the csv content:
```suggestion
        for comparison in get_dataset(poll_name).iterator()
```

amatissart (approved):
Nice :+1:

Issue #998, opened 2022-06-16
[front] feat: add a top-right profile menu


GresilleSiffle:
subtask of https://github.com/tournesol-app/tournesol/issues/980 (top-right personal menu)

---

A click on the top-right profile button must display a menu, similar to the top-right menu used by GitHub.

Two items can be added already:
- `Profile` a link to the settings > profile page
- `Account` a link to the settings > account page

PR #999, opened 2022-06-16
[front] feat: add a top-right menu

GresilleSiffle:
**related to** #998 

---

This PR adds a personal menu, displayed when the logged user click on its profile button.

I deleted the logout button to included it directly in the personal menu:
- it saves space in the top bar, especially on mobile, could be useful for other things (a comparison cart ?)
- the logout action is not part of the main use cases of the application

**to-do**
- [x] move the personal menu in its own component
  - [x] translate this new menu 
  - [x] make the items selected depending on the URL
- [x] use the same configuration to display the `PersonalMenu` and the `SettingsMenu`
- [x] rework the e2e tests

### preview

capture

capture2

### DIFF
--- a/frontend/src/features/frame/components/topbar/AccountInfo.tsx
+++ b/frontend/src/features/frame/components/topbar/AccountInfo.tsx
@@ -3,7 +3,7 @@ import { Link } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 
 import { Button, Grid } from '@mui/material';
-import { AccountCircle } from '@mui/icons-material';
+import { AccountCircle, ArrowDropDown, ArrowDropUp } from '@mui/icons-material';
 
 import { useLoginState, useNotifications } from 'src/hooks';
 import { revokeAccessToken } from '../../../login/loginAPI';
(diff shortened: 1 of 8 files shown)


frontend/src/features/frame/components/topbar/AccountInfo.tsx
```diff
@@ -22,6 +23,16 @@ const LoggedInActions = () => {
 
   const { logout, loginState } = useLoginState();
 
+  const [menuAnchor, setMenuAnchor] = React.useState<null | HTMLElement>(null);
```
lfaucon, line 26: I'm not familiar with the fact that having an anchor means that the menu is open and visible. To me the anchor is where the menu will be, not whether it is closed or open. Would it be worth it to add one line of comment above this state to explain its purpose?

As I read the code below, I would even recommned the refactor:
- Anchor is not a state that changes, but is passed as a props (in this case the top-right button in the header)
- `PersonalMenu` takes open as a props
- The state here is replaced by `isMenuOpen`

Unless you have reasons not to?

lfaucon (approved):
Looks great. Thank you

frontend/src/features/frame/components/topbar/AccountInfo.tsx
GresilleSiffle: Thanks for the review.

The anchor is indeed not a state that changes during the life cycle of the component. I used it because I followed the MUI documentation, and I agree we can simplify the code by relaying only on `open`.

I'll refactor it before merging.

GresilleSiffle: > Unless you have reasons not to?

I made the refactor. I find the new code slightly more complex now, as it have two independent notions, anchor and open, nevertheless more readable with the presence of the open notion in the parent component.

Can a menu be open without having an anchor? The original implementation didn't allow this as the open state was deduced directly by reading the anchor prop.

I feel equally comfortable with the new and the original implementation. I like the fact the original one has less code, and I like the readability of the new one.

Unless objection @lfaucon , I'll merge it as is (with the new implementation).

PR #1002, opened 2022-06-20
[ext] fix: the LICENSE file was in a sub-folder

GresilleSiffle:
This PR moves the LICENSE file from the `browser-extension/src` folder to the root of the extension.

This make the folder structure coherent with the other Tournesol's software, and will fix the `README.md` broken extension link.

The `build.sh` script includes the license file as before.

The `package.json` now clearly informs the users about the licence and the package name.


browser-extension/build.sh
```diff
@@ -1,5 +1,18 @@
...
 
-cd src
-zip -r -FS ../tournesol_extension.zip *
+TARGET='tournesol_extension.zip'
+
+pushd src
+zip -r -FS ../${TARGET} *
+popd
+
+zip ${TARGET} LICENSE
```
amatissart, line 17: The script relies on being run from the "browser-extension" directory. 
Not really a problem, it was already the case before. But using the script path as the reference in these commands could be a slight improvement (similar to infra scripts).

amatissart (approved):

PR #1005, opened 2022-06-20
Implement poll-level scaling for Mehestan

amatissart:
Related to the scaling definition discussed in #950 

These new scaling values will be stored in the `Poll` model. Global scores (i.e `EntityCriteriaScores`) are computed after the appropriate scaling is applied on individual scores. So these entity scores will always be saved as scaled scores in the database.

Following the discussion with @cthiounn last week, it seems more relevant to keep individual scores non-scaled in the database, in order to make future online individual updates more straightforward. As a consequence this branch needs to implement quite a convoluted ORM query to dynamically compute individual scaled scores returned by contributor ratings APIs.


backend/tournesol/tests/test_api_ratings.py
```diff
@@ -300,9 +300,15 @@ def test_authenticated_can_update_public_status(self):
         rating = ContributorRating.objects.get(
             poll=self.poll_videos, user=self.user1, entity=self.video1
         )
-
         self.assertEqual(rating.is_public, False)
 
+        # Create contributor score, that will be returned in the PATCH response
+        ContributorRatingCriteriaScoreFactory(
```
cthiounn, line 306: the test "test_authenticated_can_update_public_status" is also working without these lines. An assert statement about this object is missing  ?

GresilleSiffle:
Worked well on my environment :ok_hand:

backend/tournesol/tests/test_api_ratings.py
cthiounn: Indeed, with that fix, like in "test_authenticated_can_fetch_existing_entity", the data is tested. I'm quite puzzled since the main purpose as far I understand from test_authenticated_can_update_public_status is to test if one user can update from private to public a entity, which was already the case before. The added lines got me confused in the review process :

@amatissart What was the intent behind these lines ?

Is it Mehestan related or a minor quick fix ? Either way, I am good with any decision you choose to make (if you want to make a commit or to keep as is before the merge)


old version : 
{'entity': {'uid': 'yt:QZZTAXCWY0C', 'type': 'video', 'metadata': {'name': 'Entity title 0', 'source': 'youtube', 'duration': 60, 'language': 'aa', 'uploader': 'Uploader 0', 'video_id': 'QZZTAXCWY0C', 'description': 'Entity description 0', 'publication_date': '2022-07-05'}}, ***'is_public': True***, **'criteria_scores': []**, 'n_comparisons': 1}

newer version : 

{'entity': {'uid': 'yt:24KFDVYJBNB', 'type': 'video', 'metadata': {'name': 'Entity title 0', 'source': 'youtube', 'duration': 60, 'language': 'aa', 'uploader': 'Uploader 0', 'video_id': '24KFDVYJBNB', 'description': 'Entity description 0', 'publication_date': '2022-07-05'}}, ***'is_public': True***, **'criteria_scores': [{'criteria': 'test-criteria', 'score': 3.0, 'uncertainty': 0.0}]**, 'n_comparisons': 1}

GresilleSiffle (approved):

backend/tournesol/tests/test_api_ratings.py
amatissart: In one of the iterations on this branch, the rating serialization would fail and cause a server error specifically when a rating is updated while containing at least 1 criteria score. That's why I added this entry in the database.

It's no longer related to Mehestan with the final implementation. But it's still useful to cover that particular situation. I will add the check suggested by @GresilleSiffle to make this more explicit.

cthiounn (approved):

Issue #938, opened 2022-05-18
Add side component for contextual recommandations on Video Analysis page


adrausch:
- [ ] List of recommended videos with tab system :
    - The current video YouTube channel (used here: https://tournesol.app/recommendations?language=&uploader=Kurzgesagt%20%E2%80%93%20In%20a%20Nutshell)
    - Compared with (used here: https://tournesol.app/comparisons/?uid=yt:OH0P1AraqAc)
    - (future) The current video topic (Note this is blocked by the search feature)
    - (future) Active learning recommandation (Note this currently disabled due to performance)
- [ ] Clicking the thumbmail links to watching the selected video
- [ ] (optionaly) immediately comparing the current video with the selected video

PR #1006, opened 2022-06-21
Contextual recommendations on video analysis page

sigmike:
Related issue: #938 

image

It adds a side panel to the video analysis page with 2 tabs : videos of the same channel and videos it has already been compared to.

Some notes:
* I added `unsafe: true` to the same channel request to get results in dev-env, but I don't know if it should be used in production.
* I added some style props to `EntityTabsBox` to have a correct width and height. You may prefer another way to do it.
* I also added a `withLink` prop to `EntityTabsBox` to have real links (instead of a `onClick` that would change the URL). The link is only on the thumbnail and title like in video cards.
* The layout is not exactly the one in #928. I made the side panel continue on the whole page instead of only up to the collapsible description, because doing it like in #928 is more difficult (we probably need a dynamic max height) so I wanted to check first whether that's really what we want.

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -219,6 +219,9 @@
     "compareVideosButton": "Compare videos"
   },
   "options": "Options",
+  "contextualRecommendations": {
+    "channel": "Channel"
+  },
   "settings": {
     "typeKeywordDeleteAccount": "Please type the words <1>{{DELETE_ACCOUNT_KEYWORD}}</1> in the text box below to enable deleting your account.",
     "deleteYourAccount": "Delete your account",
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -220,7 +220,8 @@
   },
   "options": "Options",
   "contextualRecommendations": {
-    "channel": "Channel"
+    "channel": "Channel",
+    "comparedWith": "Compared with"
   },
   "settings": {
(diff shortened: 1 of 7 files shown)


GresilleSiffle:
Looks good for me :+1: 

> I added `unsafe: true` to the same channel request to get results in dev-env, but I don't know if it should be used in production.

Usually we don't want to recommend unsafe videos, but for me it really depends of the purpose of the analysis page. If we want to invite the users to see and compare more videos, display unsafe videos can make sense to gather more ratings about them.

If the purpose of the analysis page is to recommend content, like the recommendations page, maybe we want to avoid recommending unsafe videos.

What do you think @aidanjungo @lfaucon @amatissart ?

> I added some style props to `EntityTabsBox` to have a correct width and height. You may prefer another way to do it.

Ok for me.

> I also added a `withLink` prop to `EntityTabsBox` to have real links (instead of a `onClick` that would change the URL). The link is only on the thumbnail and title like in video cards.

Ok for me.

> The layout is not exactly the one in https://github.com/tournesol-app/tournesol/issues/928. I made the side panel continue on the whole page instead of only up to the collapsible description, because doing it like in https://github.com/tournesol-app/tournesol/issues/928 is more difficult (we probably need a dynamic max height) so I wanted to check first whether that's really what we want.

I see, what did we decide the last time? Should I rework the layout so that the video + the lists both fit on top of the metadata?

GresilleSiffle (approved):

PR #1010, opened 2022-06-22
WIP : [Back][ML] Online heuristics for Mehestan

cthiounn:
**Goal** :  This PR is about adding an light online heuristics on Tournesol scores for Mehestan

**Remaining** :

- [x]  continue testing and auditing the implemented code

- [x] fix test test_update_individual_scores_after_new_comparison_with_mehestan_run_after_update

- [x] add a specific test about online heuristics

- investigating parallel run and compatibility with back and test > to another version

- [x] implement post-scaling when that feature is implemented

- [ ] test with a large number of data 

**Done** :

- [x] first overall code (draft to be refined and refactored)

- [x] ML command (ml_train with manage.py) for debugging/testing purpose

**Remarks** : 

* It might have some minor formatting diff since i'm using black's default config in vscode


cthiounn:
fix test test_update_individual_scores_after_new_comparison_with_mehestan_run_after_update

Since i've put the online heuristic in update_user_scores and is called each time there is a new comparison in the back api, the test nature was changed and failed. I took the initiative to tweek it.

However, I can go back to the previous version and add another test and another update_user_scores function with online heuristic.

I'm also thinking about a general attribute in poll to enable or not online heuristic, as an option case by case

backend/ml/mehestan/online_heuristics.py
```diff
@@ -0,0 +1,519 @@
...
+    if parallel_computing:
+        os.register_at_fork(before=db.connections.close_all)
+
+        cpu_count = os.cpu_count() or 1
+        with Pool(processes=max(1, cpu_count - 1)) as pool:
+            for _ in pool.imap_unordered(
+                partial_online_heuristics,
+                remaining_criteria,
+            ):
+                pass
```
amatissart, line 493: When would we use parellel_computing in this context? Is there a significant gain to expect when running the online heuristics for a single user?
If that's the case, we need to be cautious about the overhead of initializing the threads. At least, the `Pool` could be created once. And as the API is already deployed with multiple workers, the number of processes should be fine-tuned to avoid DoS on the server.

backend/ml/mehestan/online_heuristics.py
```diff
@@ -0,0 +1,519 @@
...
+    ].values.squeeze()[()]
+    delta_star_a = new_raw_uncertainties.loc[
+        new_raw_uncertainties.index == entity_id_a, "raw_uncertainty"
+    ].values.squeeze()[()]
+    theta_star_b = new_raw_scores.loc[
+        new_raw_scores.index == entity_id_b, "raw_score"
+    ].values.squeeze()[()]
+    delta_star_b = new_raw_uncertainties.loc[
+        new_raw_uncertainties.index == entity_id_b, "raw_uncertainty"
+    ].values.squeeze()[()]
```
amatissart, line 207: I don't understand what's going on here. I suppose `.squeeze()` is used to extract the single value. But why `[()]` ?

backend/tournesol/views/comparison.py
```diff
@@ -132,9 +132,17 @@ def perform_create(self, serializer):
...
+            settings.UPDATE_MEHESTAN_SCORES_ON_COMPARISON
+            and poll.algorithm == ALGORITHM_MEHESTAN
+        ):
+            update_user_scores(
+                poll,
+                user=self.request.user,
+                uid_a=self.request.data["entity_a"]["uid"],
+                uid_b=self.request.data["entity_b"]["uid"],
+                delete_comparison_case=False,
+            )
```
amatissart, line 145: What's happening if the main mehestan job is running while a new comparison will be submitted? Or 2 concurrent comparisons are submitted about the same entities ?
Is there not a risk that these updates may conflict ? Do we need some lock mechanism ?

amatissart (commented):
I added a few questions. But overall the approach looks good to me.  
Of course we will need to make some benchmark to measure what is the expected latency on real-world data.

backend/ml/mehestan/online_heuristics.py
```diff
@@ -0,0 +1,519 @@
...
+
+    r_set = r.loc[list(set_of_entity_to_update)]
+    r_tilde = r_set / (1.0 + R_MAX)
+    r_tilde2 = r_tilde**2
+    # r.loc[a:b] is negative when a is prefered to b.
+    l = -1.0 * r_tilde / np.sqrt(1.0 - r_tilde2)  # noqa: E741
+    k = (1.0 - r_tilde2) ** 3
+
+    L = k.mul(l).sum(axis=1)
+    Kaa_np = np.array(k.sum(axis=1) + ALPHA)
```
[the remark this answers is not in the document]
cthiounn: Indeed, I'll patch a refactor with compute_individuel_score() in individual.py

backend/ml/mehestan/online_heuristics.py
cthiounn: I did not success to make the parallel computing working , maybe I should delete that try for the first version to focus the PR on the algorithm rather some technical details and let you investigate if needed in a enhanced version

backend/tournesol/views/comparison.py
cthiounn: Good question about entity conflict/lock. 


For the main mehestan job, we should disable the heuristic in that windows frame, I guess it is a safe option, depending on the duration of the job


For two concurrent comparisons, I think it can happen when computing global scores of A and B if two users are to collide at the same time (one can read before the save of the other). For the individual scores, since it is user-based, it should not be an issue.

Since the online heuristic purpose is ultimately to update the global score approximatively, I don't how strict we should be about that error. Hopefully another user compare either A and B and global score should be "corrected" by that error.

A test load should be able to detect if that conflict will happen or the heuristic is fast enough to hold ACID-database properties

backend/ml/mehestan/online_heuristics.py
cthiounn: with [()] : it returns  <class 'numpy.float64'>
without : it returns  <class 'numpy.ndarray'>

if you have an much more elegant solution, I will take it for sure than an obscure [()]   :+1:

backend/tournesol/views/comparison.py
cthiounn: the heuristic take 8 seconds to complete with public database real data, with 7 seconds for save_tournesol_scores(poll). I'm patching right now to only save tournesol_score of entity A and B.

reapplying the individual scaling take nearly the second left. I'm looking into it to make it faster

amatissart: This total number of entities is used for a single purpose: make the difference between "large" polls (eg "videos") and "small" polls (with very few entities) because they use a slightly different definition for supertrusted users.

As the exact number doesn't really matter, another approach (if necessary) would be to keep this value in cache, to avoid recomputing on each run.

backend/ml/mehestan/online_heuristics.py
amatissart: ok, so it could simply be written as 
```py
delta_star_b = new_raw_uncertainties.loc[entity_id_b, "raw_uncertainty"]
```
or I'm missing something?

cthiounn: new_raw_uncertainties.loc[entity_id_b, "raw_uncertainty"]  give a series with one or multiple lines (with this syntax, the filter can give multiple rows):

1 entity_a
16    0.717044
Name: raw_uncertainty, dtype: float64 <class 'pandas.core.series.Series'>

 with values : [0.71704443] <class 'numpy.ndarray'>
with values.unsqueeze :  0.71704443 <class 'numpy.ndarray'>
and with values.unsqueeze[()]:  0.71704443 <class 'numpy.float64'>

lfaucon:
Plan for this:
- [ ] Rebase the branch to solve conflicts (@GresilleSiffle and @amatissart )
- [ ] review & approval by @GresilleSiffle and @amatissart 
- [ ] Merge and deploy but does not activate in prod for performance reasons
- [ ] Activate after work on asynchronous jobs

PR #1015, opened 2022-06-27
[back] fix: vouching made more human-readable

lenhoanglnh:
I rewrote the code to make it more human-readable. I verified that it successfully pass the tests.
There may be a need to go further to avoid ambiguities, as the word "trust" is used with too many slightly different meanings.
    
In particular, I would suggest:
1. The user's field "trust_score" could be replaced with "voting_right".
2. The vouch's field "trust_value" could be replaced by "vouch_value".
3. The vouch's field "by" could be replaced by "from".
    
As a result, while "trust" remains not fully well defined (currently, in the code, it only matters if it is > 0), the word "relative_trust" implies a sum of all users' relative trusts equals 1. Conversely, a user's "voting_right" an take any positive number. Moreover pretrusted users (by email domains) are guaranteed to have a voting right of at least 1.

What do you think?

### DIFF
--- a/backend/vouch/tests/test_trust_algo.py
+++ b/backend/vouch/tests/test_trust_algo.py
@@ -9,9 +9,10 @@ from core.models import user
 from core.models.user import EmailDomain
 from core.tests.factories.user import UserFactory
 from vouch.models import Voucher
-from vouch.trust_algo import normalize_vouch_matrix, compute_relative_posttrust, \
+from vouch.trust_algo import normalize_vouch_matrix, compute_relative_posttrusts, \
     compute_voting_rights, trust_algo, PRETRUST_BIAS, MIN_PRETRUST_VOTING_RIGHT
 
+EPSILON = 0.0000001
 
 class TrustAlgoTestCse(TestCase):
     _user_0 = "username_0"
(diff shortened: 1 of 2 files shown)


backend/vouch/tests/test_trust_algo.py
```diff
@@ -68,55 +70,77 @@ def setUp(self) -> None:
...
+        normalized_vouch_matrix = normalize_vouch_matrix(vouch_matrix, pretrust)
+        relative_posttrust = compute_relative_posttrust(normalized_vouch_matrix, relative_pretrust)
+        self.assertAlmostEqual(relative_posttrust[untrusted_user], 0)
 
-            if trust_stat[i] > 0:
-                self.assertTrue(rescaled[i] >= 0.999999999)
+    # The voting rights of pretrusted users must be at least MIN_PRETRUST_VOTING_RIGHT
+    # Moreover, unless the voting right is clipped to 1, it must be proportional to posttrust
+    def test_compute_voting_rights(self):
+        pretrust = np.random.randint(2, size=self._nb_users)
```
aidanjungo, line 115: I think this line will produce a number between 2 and `_nb_users` and not a list. Is it what want? 
If yes, what is the point of using a for loop over one number on line 117?

backend/vouch/tests/test_trust_algo.py
```diff
@@ -68,55 +70,77 @@ def setUp(self) -> None:
...
+        posttrust = np.random.randint(20, 100, self._nb_users)
+        for u in range(len(pretrust)):
+            if pretrust[u] == 0 and np.random.rand() < 0.5:
+                posttrust[u] = 0
```
aidanjungo, line 119: Depending on your answer to the previous comment, I think these lines could be simplified

aidanjungo (commented):
Not a complete review, just a few comments :)

backend/vouch/trust_algo.py
```diff
@@ -1,110 +1,128 @@
...
+def compute_voting_rights(relative_posttrusts, pretrusts):
     """
     Go from ratio of trust to actual voting weight that should be assigned to
     users given the trust the network puts in them.
     """
-    mint = np.amin(
-        [trust_vector[i] for i in range(len(trust_vector)) if trust_status[i] > 0]
+    min_relative_trusts_of_pretrusteds = np.amin(
+        [relative_posttrusts[u] for u in range(len(relative_posttrusts)) if pretrusts[u] > 0]
     )
```
amatissart, line 84: Similarly, this can be simplified with [`np.amin(where=...)`](https://numpy.org/doc/stable/reference/generated/numpy.amin.html#numpy.amin). 

```suggestion
    min_relative_trusts_of_pretrusteds = np.amin(
        relative_posttrusts,
        where=pretrusts > 0,
        initial=MIN_PRETRUST_VOTING_RIGHT,
    )
```

And the "initial" value prevents `np.amin()` from failing if no "pretrusted" user exists.

backend/vouch/trust_algo.py
```diff
@@ -1,110 +1,128 @@
...
-    # List of users, list of their trust status (regarding their email) in the
-    # same order, and number of users.
+    # Import users and pretrust status
     users = list(
         user.User.objects.all().annotate(
             _is_trusted=Q(pk__in=user.User.trusted_users())
         )
     )
-    trust_status = [int(u._is_trusted) for u in users]
+    pretrusts = [int(u._is_trusted) for u in users]
```
amatissart, line 108: To apply the previous suggestions, you may need to initialize these values as a numpy array
```suggestion
    pretrusts = np.array([int(u._is_trusted) for u in users])
```

amatissart (commented):
Looks good to me.

Minor suggestions for optimizations.

amatissart (approved):

BerangereColbois (approved):
Thank you a lot @lenhoanglnh, it is indeed clearer now!
Regarding the values of the hyperparameters, is it fine if we let them like this for now, and maybe change them in the future depending on how the vouching mechanism is used in practice? For instance, the pre-trust bias is pretty low for now, but I assumed that the vouching option would mostly be used by trusted users, does it seem ok for you like this?

PR #1017, opened 2022-06-27
[front] fix: flex layout on analysis page not good on mobile

GresilleSiffle:
This PR removes the horizontal scroll of the analysis page when displayed on mobile.

The usage of `<Box>` instead of `<Grid>` did the trick.

The width definition of the two flex items are a bit verbose. I wanted to make the first item (the video) take approx 65 % and I expected the second item (the contextual recommendations) to take the remaining space thanks to `flexGrow={1}` but I didn't manage to make it work.

Without the `flexGrow` the layout is still nice but there is and extra blank space at the end of the flex container that could be used by the contextual recommendations (for now, I moved this space between the items with `justifyContent="space-between"`, but it doesn't solve the problem).

capture

### DIFF
--- a/frontend/src/pages/videos/VideoAnalysisPage.tsx
+++ b/frontend/src/pages/videos/VideoAnalysisPage.tsx
@@ -47,9 +47,7 @@ export const VideoAnalysis = ({
       gap="16px"
       justifyContent="space-between"
     >
-      <Box
-        width={{ sm: '100%', xs: '100%', md: '65%', lg: '64.9%', xl: '64.8%' }}
-      >
+      <Box flex={2} minWidth={{ xs: '100%', md: null }}>
         {/* Top level section, containing links and maybe more in the future. */}
         <Box mb={2} display="flex" justifyContent="flex-end">
           <Button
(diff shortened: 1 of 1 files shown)


frontend/src/pages/videos/VideoAnalysisPage.tsx
```diff
@@ -39,112 +39,117 @@ export const VideoAnalysis = ({
...
+                    <CriteriaScoresDistribution uid={uid} />
+                  </Box>
+                </Paper>
+              </Grid>
+            </SelectedCriterionProvider>
+          )}
         </Grid>
-      </Grid>
+      </Box>
+      <Box width={{ sm: '100%', xs: '100%', md: '33%', lg: '33%' }}>
```
GresilleSiffle, line 147: If any of us knows how to replace this verbose width by the proper usage of `flex-grow`, it would be great.

amatissart: A potential solution seems to be 
First Box (line 50): `<Box flex={2} minWidth={{ xs: '100%', md: null }}>`
Second Box (line 147): `<Box flex={1}>`

GresilleSiffle: Thank u, it works!

adrausch (approved):

Issue #997, opened 2022-06-16
[back] refactor: the suggestions API has low performance


GresilleSiffle:
### context

We realized while using the new video selector in production, that the suggestions tab calling the `entities_to_compare` API takes a long time to load.

We have disabled this tab since.

### proposition

**(a)** investigate what are the costly operations done by the `entities_to_compare` API and optimize them when relevant

**(b)** the `entities_to_compare` API shouldn't be attack vector, if it turns out to be a naturally slow endpoint after the optimizations, requiring a lot of CPU or memory, we could add a specific throttle rate (may be called `RessourceHeavyThrottle` to be shared by other endpoints).

**(c)** let's test more the suggestions tab, to see if the recommended videos seem relevant (and engaging ?)

**further improvements:** we could also transforms some parameters of the suggestion algorithm into Django settings, so that we can control more easily its behaviour.

PR #1018, opened 2022-06-28
[back] perf: small fixes in the suggestion algorithm

Foebus:
Added a missing minus on the exponent of the similarity matrix computation
Simplified computation of score computation

### DIFF
--- a/backend/tournesol/suggestions/graph.py
+++ b/backend/tournesol/suggestions/graph.py
@@ -303,10 +303,10 @@ class Graph(CompleteGraph):
                             and vb in scaling_factor_increasing_videos
                     ):
                         va.video1_score = 1
-                        va._graph_sparsity_score = 1
+                        va._graph_sparsity_score[vb] = 1
                     else:
                         va.video1_score = 0
-                        va._graph_sparsity_score = 0
+                        va._graph_sparsity_score[vb] = 0
         # Once the scaling factor is high enough, check what video should gain
         # information being compared by the user
         else:
@@ -326,11 +326,11 @@ class Graph(CompleteGraph):
                         if eigenvalues[1] > self.LAMBDA_THRESHOLD:
                             u_index = sg.uid_to_index[vb.uid]
                             v_index = sg.uid_to_index[va.uid]
-                            va._graph_sparsity_score = sg.similarity_matrix[u_index, v_index]
+                            va._graph_sparsity_score[vb] = sg.similarity_matrix[u_index, v_index]
                         elif va.uid not in sg.uid_to_index:
-                            va._graph_sparsity_score = 1
+                            va._graph_sparsity_score[vb] = 1
                         else:
-                            va._graph_sparsity_score = 0
+                            va._graph_sparsity_score[vb] = 0
                         # Compute estimated information gain relative to the respective
                         # uncertainties in both scores
                         va.beta[vb] = va.uncertainty_diminution(vb)
--- a/backend/tournesol/suggestions/graph.py
+++ b/backend/tournesol/suggestions/graph.py
@@ -4,6 +4,7 @@ from typing import Optional
 
 import numpy as np
 from django.db.models import Avg, F, QuerySet
+from scipy.sparse.csgraph import shortest_path
 
 from tournesol.models import (
     ContributorRatingCriteriaScore,
@@ -104,7 +105,6 @@ class Graph(CompleteGraph):
     _nodes: list[SuggestedUserVideo]  # todo clean that, use default dict functions
     uid_to_index: dict[str, int]
     graph: dict[SuggestedVideo, list[SuggestedVideo]]
-    sigma: float
     dirty: bool
     _local_user: SuggestedUser
     local_user_scaling: ContributorScaling
@@ -112,7 +112,6 @@ class Graph(CompleteGraph):
     adjacency_matrix: np.array
     normalized_adjacency_matrix: np.array
 
-    distance_matrix: dict[SuggestedVideo, dict[SuggestedVideo, int]]
     similarity_matrix: np.array
 
     def __init__(self, local_user: SuggestedUser, local_poll: Poll, local_criteria):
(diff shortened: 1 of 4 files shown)


GresilleSiffle:
Thanks @Foebus 

Was the missing minus a bug? If yes, does it mean we have a missing test?

Foebus:
> Thanks @Foebus
> 
> Was the missing minus a bug? If yes, does it mean we have a missing test?

Good question, actually it changes the computed results...It may be possible to detect it, I will look to add a test that should help (the idea is for this value to always be <= 1 but I can't really test a lot more than that)

amatissart:
@Foebus I just opened #1025  based on this branch, to suggest an optimization that could bring a significant improvement regarding #997.

> Good question, actually it changes the computed results...It may be possible to detect it, I will look to add a test that should help (the idea is for this value to always be <= 1 but I can't really test a lot more than that)

Ideally, it would be nice to add a test case that illustrates how the values of this similarity matrix affect the actual suggested entities. It's unclear for me how they interact. It seems the similarity value is only used here:
https://github.com/tournesol-app/tournesol/blob/3c9e11b6804b682c8a3b830446fa926bdbe18909/backend/tournesol/suggestions/graph.py#L329

But the `.graph_sparsity` property doesn't seem to be used in the score computation. Am I missing something?

Foebus:
> Thanks @Foebus
> 
> Was the missing minus a bug? If yes, does it mean we have a missing test?

Good question, actually it changes the computed results...It may be possible to detect it, I will look to add a test that should help (the idea is for this value to always be <= 1 but I can't really test a lot more than that)

> @Foebus I just opened #1025 based on this branch, to suggest an optimization that could bring a significant improvement regarding #997.
> 
> > Good question, actually it changes the computed results...It may be possible to detect it, I will look to add a test that should help (the idea is for this value to always be <= 1 but I can't really test a lot more than that)
> 
> Ideally, it would be nice to add a test case that illustrates how the values of this similarity matrix affect the actual suggested entities. It's unclear for me how they interact. It seems the similarity value is only used here:
> 
> https://github.com/tournesol-app/tournesol/blob/3c9e11b6804b682c8a3b830446fa926bdbe18909/backend/tournesol/suggestions/graph.py#L329
> 
> But the `.graph_sparsity` property doesn't seem to be used in the score computation. Am I missing something?

Absolutely, thanks for noticing, I added it and will add some test cases

backend/tournesol/tests/test_suggestions.py
```diff
@@ -271,6 +288,24 @@ def setUp(self):
...
+                    datetime_lastedit=now + datetime.timedelta(minutes=2),
+                ))
+
+        self.comparisons.append(ComparisonFactory(
+            user=self.sparsity_comparison_user,
+            entity_1=self.videos[0],
+            entity_2=self.videos[4],
+            duration_ms=501,
+            datetime_lastedit=now + datetime.timedelta(minutes=2),
+        ))
```
amatissart, line 307: What's the intention behind this setup? It's not clear to me at first glance, a comment would be welcome.

(`duration` and `datetime_lastedit` could probably be omitted as they are optional in the factory).

backend/tournesol/tests/test_suggestions.py
```diff
@@ -359,7 +394,7 @@ def test_most_informative_vid_given_first(self):
         )
         last_vid_score = 1000
         for v in user2_videos:
-            v_score = v.score_computation(user_videos[0]) + v.graph_sparsity
+            v_score = v.score_computation(user_videos[0]) + v.graph_sparsity(user_videos[0])
```
amatissart, line 397: Do we really want to call `.graph_sparsity()` here, now that it's included into `.score_computation()`

backend/tournesol/tests/test_suggestions.py
```diff
@@ -399,6 +434,30 @@ def test_suggestions_for_new_user(self):
...
+
+        for n in user_graph.nodes:
+            if n.uid == self.videos[0].uid:
+                alone_node = n
+                break
+
+        for n in user_graph.nodes:
+            max_value = n.graph_sparsity(alone_node)
+            for m in user_graph.nodes:
+                assert n.graph_sparsity(m) <= max_value
```
amatissart, line 450: What's the intention here again? It seems that in this particular setup, all graph sparsity values are 1.
So this could be simplified as
```suggestion
        for n in user_graph.nodes:
            for m in user_graph.nodes:
                assert n.graph_sparsity(m) == 1
```

But is it really the expected result?

backend/tournesol/suggestions/graph.py
```diff
@@ -327,18 +326,14 @@ def compute_information_gain(self, scaling_factor_increasing_videos: list[Sugges
...
                             u_index = sg.uid_to_index[vb.uid]
                             v_index = sg.uid_to_index[va.uid]
-                            va._graph_sparsity_score = sg.similarity_matrix[u_index, v_index]
+                            va._graph_sparsity_score[vb] = sg.similarity_matrix[u_index, v_index]
                         elif va.uid not in sg.uid_to_index:
-                            va._graph_sparsity_score = 1
+                            va._graph_sparsity_score[vb] = 1
                         else:
-                            va._graph_sparsity_score = 0
+                            va._graph_sparsity_score[vb] = 0
```
amatissart, line 333: The first condition `if eigenvalues[1] > self.LAMBDA_THRESHOLD` is never reached in the test suite.

`np.linalg.eigvalsh` returns the eigenvalues in _ascending_ order. Is it deliberate to use the second smallest value?

amatissart (commented):
@Foebus I have a few more questions below about how this suggestion matrix is tested.

(I also rebased this branch on "main" to remove the merge commit and make the diff easier to read on GitHub. No change in the implementation.)

backend/tournesol/tests/test_suggestions.py
Foebus: Absolutely !

backend/tournesol/tests/test_suggestions.py
Foebus: Actually no, the values should be all < 1 and especially low for all nodes except when these are compared with the one containing video 0 (this is due to the comparison graph, all nodes are densely connected but this one)
If all are equal to 1 I may have inverted a comparison

backend/tournesol/suggestions/graph.py
Foebus: Thanks for noticing, I changed it to get the correct eigenvalue

backend/tournesol/tests/test_suggestions.py
Foebus: I checked, there is a problem with the normalization applied, in fact there should be a symmetric normalized adjacency matrix, I look to fix it

Foebus: Ok, it looks way better now :)

Issue #915, opened 2022-05-14
[front] Add a share button on the video analysis page


aidanjungo:
from @GresilleSiffle idea in #907:
What about adding a share button next the compare button in the top right corner of the video analysis page? This will allow users to quickly copy / paste link (maybe share the page on social network), and may create a virtuous circle where people share & compare videos more easily. 🌻

Sbyspaceopera:
If you are ok with adding a dependency I think this library ccould be a great time saver when this feature will leaves the backlogs :

https://www.npmjs.com/package/react-share

Types are includes and it seems easy to use.

PR #1019, opened 2022-06-28
[front] feat: add a copy URI button on the analysis page

GresilleSiffle:
**related to** #915

**estimated BL** 10 per year

---

This PR adds a button on the video analysis page, that allows to directly copy the current URI to the browser clipboard.

This is the first step of the issue #915. This button could be replaced in the future by a proper share button, as described in the issue.

https://user-images.githubusercontent.com/39056254/176186724-022f58a7-a3cb-4ca9-99d7-31ea33abb3d1.mp4

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -1,4 +1,22 @@
 {
+  "entityAnalysisPage": {
+    "generic": {
+      "copy": "Copy URI",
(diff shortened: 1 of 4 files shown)


adrausch (approved):
I cannot watch the video you shared because my 3 browsers interpreted it as corrupted. But everything worked for me when I tested it.
So I approve :)

Issue #844, opened 2022-04-22
[front] feat: adapt the `presidentielle` tutorial to the videos poll


lfaucon:
We have found that it was very useful for users to understand the core functionalities of Tournesol to start with this tutorial so we want to adapt it to the videos too.

What roughly needs to be done:
- [x] Decide on how videos are proposed for comparisons. An idea would be to select good (score > 0) and short (duration < 4min) videos in the language of the user.
  - see the PR https://github.com/tournesol-app/tournesol/pull/1021
  - videos are **safe recommendations**
  - from the **top 100 of all-time** recommendation...
  - ...having a **max duration 5 minutes** 
- [x] Decide on the length of the tutorial? Maybe 7 comparisons is too much?
  - we decided a **length of 4**
- [x] The videos must use the user's language
  - see #1035 
- [x] Adapt the text of the tutorial when used in the Video Poll
- [x] Add a link to start the tutorial on the Home Page of the Video Poll
  - see: https://github.com/tournesol-app/tournesol/issues/1046

adrausch:
> 

Selecting short duration is a good idea but I asked @GresilleSiffle and he explained me that we need to change the backend to filter the "duration" parameter in the entities metadata and that we would not have tool to filter with "inferior or equal" on this criterium.
So should we (me and @GresilleSiffle) implement this in the backend or do you think about an alternative that we wouldn't have thought of?

lfaucon:
We have already work in progress concerning adding the ability to filter with "greater than" / "less than" operations on the duration: https://github.com/tournesol-app/tournesol/issues/669 @farnots probably has work in progress on this. Can you try reaching out on Discord?

In order to not be blocked, it is great to do a version of this tutorial without any duration filter and add it later.

GresilleSiffle:
I think we reached this first milestone about the tutorial and more generally this first milestone about Tounesol accessibility.

Few improvements I see for the future:
- rework the Home page : reducing the amount of text, talk about the platform, not the underlying software ; add images, etc.
- sometimes videos suggested by the tutorial doesn't exist anymore on YouTube
- we could add a help button on the comparison page, display a comparison FAQ + some explanation from the tutorial inside a dialog box

lfaucon:
Created the issue #1070 concerning your second point

PR #1021, opened 2022-06-29
[front] feat: use only 5 min. max videos for the tutorial

GresilleSiffle:
**related to** #844 

---

This PR changes the type of entity recommended by the `videos` tutorial.

Suggested entities:
- are now **safe recommendations**
- are in the **top 40 of all-time** recommendations...
- ...having a maximum duration of **5 minutes**

After having submitted their last comparison in the `videos` tutorial, the users were abruptly redirected to a blank comparison page, without any message or feedback of any kind. Even for me, as a confirmed user, it was unexpected and confusing. I didn't know if my last comparison was successfully submitted or not.

In this PR I decided to redirect the users to their own comparisons page, to allow them to see their progress, and to confirm their last comparison was correctly saved.

To create a virtuous circle, we could add the same feature we added in the `presidentielle2022` feedback page: add a button compare on the top right corner of the My comparisons page, allowing the users to continue comparing after the tutorial. I think it will help a lot to have a button like this, so that the users don't have to adventure themselves in the sidebar to explore the application.

```
tutorial -- (auto redirect) --> feedback with My comparisons page -- (a button invite to) --> comparison page
```

to-do
- [x] check if the `presidentielle2022` tuto still works
- [x] check if there is enough diversity in production with those criteria ( safe & less than 5 minutes )
  - at the time of writing, there are 116 safe reco. with a duration inferior to 5 minutes
  - it seems enough for a tutorial of 4 comparisons

### technical points

The tutorial are now compatible with both the entities API and the recommendations API. This allows to have a tutorial for polls similar to `presidentielle2022` using the entities API (small amount of alternatives, all alternatives are suggested regardless of their score), and for polls similar to `videos` (a lot of alternatives, not all of them are recommendable).

### next steps

**to-do**

Add a link in the home page to the tutorial, if the user has not finished it yet (same behaviour as in the `presidentielle2022` poll).

Use the user's language to determine the language of the suggested videos.

See #844

### DIFF
--- a/frontend/src/features/comparisonSeries/ComparisonSeries.tsx
+++ b/frontend/src/features/comparisonSeries/ComparisonSeries.tsx
@@ -5,7 +5,7 @@ import DialogBox from 'src/components/DialogBox';
 import LoaderWrapper from 'src/components/LoaderWrapper';
 import Comparison, { UID_PARAMS } from 'src/features/comparisons/Comparison';
 import { useCurrentPoll } from 'src/hooks/useCurrentPoll';
-import { Entity } from 'src/services/openapi';
+import { Entity, Recommendation } from 'src/services/openapi';
 import { alreadyComparedWith, selectRandomEntity } from 'src/utils/entity';
 import { getUserComparisons } from 'src/utils/api/comparisons';
 import { OrderedDialogs } from 'src/utils/types';
@@ -19,7 +19,7 @@ const MIN_LENGTH = 2;
 interface Props {
   dialogs?: OrderedDialogs;
   generateInitial?: boolean;
-  getAlternatives?: () => Promise<Array<Entity>>;
+  getAlternatives?: () => Promise<Array<Entity | Recommendation>>;
   length: number;
   // redirect to this URL when the series is over
   redirectTo?: string;
@@ -63,7 +63,9 @@ const ComparisonSeries = ({
   // tell the `Comparison` to refresh the left entity, or the right one
   const [refreshLeft, setRefreshLeft] = useState(false);
   // a limited list of entities that can be used to suggest new comparisons
-  const [alternatives, setAlternatives] = useState<Array<Entity>>([]);
+  const [alternatives, setAlternatives] = useState<
+    Array<Entity | Recommendation>
+  >([]);
   // an array of already made comparisons, allowing to not suggest two times the
   // same comparison to a user, formatted like this ['uidA/uidB', 'uidA/uidC']
   const [comparisonsMade, setComparisonsMade] = useState<Array<string>>([]);
@@ -87,7 +89,9 @@ const ComparisonSeries = ({
    * series.
    */
   useEffect(() => {
-    async function getAlternativesAsync(getAlts: () => Promise<Array<Entity>>) {
+    async function getAlternativesAsync(
+      getAlts: () => Promise<Array<Entity | Recommendation>>
+    ) {
       const alts = await getAlts();
       if (alts.length > 0) {
         setAlternatives(alts);
@@ -197,7 +201,7 @@ const ComparisonSeries = ({
    * @param uidB The current value of the `uidB` URL parameter
    */
   const genInitialComparisonParams = (
-    from: Array<Entity>,
+    from: Array<Entity | Recommendation>,
     comparisons: Array<string>,
     uidA: string,
     uidB: string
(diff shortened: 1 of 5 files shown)


frontend/src/features/comparisonSeries/ComparisonSeries.tsx
```diff
@@ -87,7 +89,9 @@ const ComparisonSeries = ({
    * series.
    */
   useEffect(() => {
-    async function getAlternativesAsync(getAlts: () => Promise<Array<Entity>>) {
+    async function getAlternativesAsync(
+      getAlts: () => Promise<Array<Entity | Recommendation>>
```
lfaucon, line 93: You wrote many times `Entity | Recommendation`, do you think it is worth it to define this once? It would get more complex if we get a third wrapper around entity objects.

Another option maybe to keep using the more restricted type `Entity` here and convert Recommendation to Entity ?

frontend/src/utils/polls/videos.ts
```diff
@@ -1,16 +1,37 @@
...
+ * Return a list of `Recommendation` for the tutorial.
+ *
+ * The recommendations returned:
+ *   - are safe recommendations
+ *   - are in the TOP 40 of all-time recommendations
+ *   - have a maximum duration of 5 minutes
+ *
+ * TODO:
+ *   - make the recommendations' language match the user language
+ *   - if no reco. are available in this language, use `en`?
```
lfaucon, line 17: Good catch. We probably could do this also on the recommendation page. Showing a list of English video on an english website seems better than an empty list.

frontend/src/utils/polls/videos.ts
```diff
@@ -1,16 +1,37 @@
...
   }
-  VIDEOS = EntitiesService.entitiesList({
-    type: TypeEnum.VIDEO,
-    limit: 8,
+
+  const minutesMax = 5;
+  const top = 40;
+
+  const metadata = {
+    'duration:lte:int': 60 * minutesMax,
```
lfaucon, line 28: 🤩

lfaucon (commented):
Waiting for the TODOs to be done before approving 👍

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
GresilleSiffle: Good question.

I didn't convert the `Recommendation` to `Entity` on purpose. First because the tutorial must work with both entities (it's the présidentielle use case), and recommendations (videos use case). It's what I expect from it, to be able to interact with both API. Keeping these two types make the behaviour of the code more explicit.

Second because I think we may be able to simplify the tutorial by working only with UID, and not directly with entities and recommendations. I think we can avoid more complexity by not using casting functions.

About the possible wrapper around those two types. Usually in the front end, dealing with a variable that can have many types introduce more complexity. We are not sure all desired attributes are available, we must test their existence, etc. etc. I think I'd like to simplify the tutorial by using only UIDs, before thinking about adding a wrapper.

What do you think?

amatissart: Note that `RelatedEntityObject` defined in "src/utils/types.ts" could probably be used here instead of `Entity | Recommendation`. 

(Arguably, we might want to rename `RelatedEntityObject` to a more expressive name, such as `GenericEntity` or `EntityObject`)

GresilleSiffle:
> Should we do "Use the user's language to determine the language of the suggested videos" in this PR, or do you really prefer keeping it as a next step?

Thanks a lot for the comments @lfaucon 

As you wish. I think the current state of the tutorial in production is problematic : possible suggestion of non recommended videos. I tend to think we should fix this as soon as possible and then work on the language.

frontend/src/utils/polls/videos.ts
```diff
@@ -1,16 +1,37 @@
...
+export function getTutorialVideos(): Promise<Recommendation[]> {
   if (VIDEOS != null) {
     return VIDEOS;
   }
-  VIDEOS = EntitiesService.entitiesList({
-    type: TypeEnum.VIDEO,
-    limit: 8,
+
+  const minutesMax = 5;
+  const top = 40;
```
amatissart, line 25: Isn't 40 a bit restrictive? I fear that excessively promoting a small number of videos which would receive a lot of comparisons would possibly reinforce their ranking among the top recommendations. 

In the longer term a filter on the tournesol score may be more relevant, but in the meantime why not setting a higher limit? 100 ?

GresilleSiffle: You're right, 100 will be better to avoid promoting the same content.

lfaucon (approved):

frontend/src/utils/polls/videos.ts
GresilleSiffle: As I don't have my computer with me, feel free to update this value.

PR #1028, opened 2022-07-05
WIP / [front] refactor: range slider in `DurationFilter`

adrausch:
**blocked by PR** https://github.com/tournesol-app/tournesol/pull/1029
 
---

Add non-linear range slider to the SearchFilter
Screenshot 2022-07-05 at 09 11 33

Pagination seems to work when we remove these lines in useListFilter.ts :
```
if (key !== 'offset') {
    searchParams.delete('offset');
}
```

WIP problems:
- the pagination issue is not solved properly

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -208,15 +208,7 @@
     "allTime": "All time",
     "uploadDate": "Uploaded",
     "duration": {
-      "title": "Duration (minutes)",
-      "max": {
-        "label": "max",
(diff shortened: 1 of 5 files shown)


GresilleSiffle:
> the pagination issue is not solved properly

This will be fixed by https://github.com/tournesol-app/tournesol/pull/1029

GresilleSiffle:
I really like this design.

I think we can even get rid of the two input fields if:
- we opt for a slider
- the slider steps are easy to play with
- we make the last step >2 hours work

PR #1029, opened 2022-07-05
[front] fix: prevent `DurationFilter` infinite rendering

GresilleSiffle:
This PR fixes the infinite rendering loop caused by the `<DurationFilter>`.

This should fix:
- the `<DurationFilter>` filter on Firefox
- the reset pagination bug, when clicking on next with a duration filter active

To achieve this I removed the `onChangeCallback` from the `<DurationFilter>` effects' dependencies.

(I also moved the code responsible of the state `min` before the state `max` just to please my brain.)

### DIFF
--- a/frontend/src/hooks/useListFilter.ts
+++ b/frontend/src/hooks/useListFilter.ts
@@ -1,4 +1,3 @@
-import { useCallback } from 'react';
 import { useLocation, useHistory } from 'react-router';
 
 export const useListFilter = ({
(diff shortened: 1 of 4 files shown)


frontend/src/hooks/useListFilter.ts
```diff
@@ -10,18 +11,22 @@ export const useListFilter = ({
...
+      }
+      // Reset pagination if filters change
+      if (key !== 'offset') {
+        searchParams.delete('offset');
+      }
+      history.push({ search: searchParams.toString() });
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [setEmptyValues]
+  );
```
lfaucon, line 29: This creates very weird issues.

useCallback is about memoization, but the way you use it here without passing `searchParams` causes the method `setFilter` to not be updated when searchParams changes. Weirdly as long as only this `setFilter` method changes the `searchParams` then the newly updated `searchParams`and the `searchParams` memoized stay in sync and all works well.

It starts breaking when search params are updated somewhere else, for example by navigating back or by opening a link that does not specify the language filter.

Example to reproduce:

1. Home at http://localhost:3000/ 
2. Select Recommendations
3. Open filters and select "All Time"
4. Navigate back
5. Observe that the filter is back to "A month ago" as expected
6. Select max duration 20
7. /!\ Observe that the data filter is set to "All Time" but should have stayed "A month ago"

Other issue:

1. Home at http://localhost:3000/ 
2. Select Recommendations
4. Select max duration 20
5. Navigate back
6. Observe that the filtering and url correctly do not have any max duration, but the TextInput still shows max duration at 20

lfaucon (dismissed):
Heyhey, it's not easy :/ 

We probably should try including some back navigation in frontend tests?

frontend/src/hooks/useListFilter.ts
GresilleSiffle: I see, it's more complex than I expected.

I'll give it a second look after the research PR.

GresilleSiffle: I still need to do more tests to be sure I didn't break anything else, but the two issues mentioned above should be fixed by: https://github.com/tournesol-app/tournesol/pull/1029/commits/4581aa76da0025847adcbdbba8a5202bb218df68

lfaucon (approved):
LGTM, but let's pay attention as there may be still some unexpected behaviours.

Maybe that calls for using a smarter state management?

PR #1031, opened 2022-07-06
[back] chore: update 3 dependencies

GresilleSiffle:
django-cors-headers: from 3.11.0 to 3.13.0
see: https://github.com/adamchainz/django-cors-headers/blob/main/HISTORY.rst#3130-2022-06-05

django-countries: from 7.2.1 to 7.3.2
https://github.com/SmileyChris/django-countries/blob/master/CHANGES.rst#732-4-march-2022

django-filter: from 2.4.0 to 22.1
https://github.com/carltongibson/django-filter/blob/main/CHANGES.rst#version-221-2022-6-17

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -1,8 +1,8 @@
 Django==4.0.6
 django-computed-property==0.3.0
-django-cors-headers==3.11.0
-django-countries==7.2.1
-django-filter==2.4.0
+django-cors-headers==3.13.0
+django-countries==7.3.2
+django-filter==22.1
 django-oauth-toolkit==1.7.0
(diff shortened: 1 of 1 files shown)


amatissart (approved):
Thanks :+1: 

At some point we had considered using pdm or pipenv to manage the dependencies and make this kind of maintenance easier. Not sure it's worth it for now. But to keep in mind if we feel outdated dependencies become unmanageable.

GresilleSiffle:
Sure, thanks @amatissart .

I'm thinking about PDM sometimes, after the summer I'd like to test a setup for Tournesol.

Issue #1013, opened 2022-06-22
[front] feat: Pasting Tournesol URL should work on Tournesol


lfaucon:
In Rate Later list or Comparison interface we can paste youtube urls, we want to make it possible to paste Tournesol analysis page urls too

PR #1033, opened 2022-07-06
[front] feat: analysis URLs are now accepted by the entity selector

Sbyspaceopera:
**related to** #1013

---

I kept the original way, just improved the RegExp in `./frontend/src/utils/video.ts`

You can visualize the new RegExp workflow [here](https://regexper.com/#%2F%28%3F%3Ahttps%3F%3A%5C%2F%5C%2F%29%3F%28%3F%3Awww%5C.%29%3F%28%3F%3Ayoutube%5C.com%5C%2Fwatch%5C%3Fv%3D%7Cyoutu.be%5C%2F%7Ctournesol%5C.app%5C%2Fentities%5C%2Fyt%3A%29%28%5BA-Za-z0-9-_%5D%2B%29%2F)  on [RegExper.com](url)

### DIFF
--- a/frontend/src/utils/video.ts
+++ b/frontend/src/utils/video.ts
@@ -3,8 +3,15 @@ import { YOUTUBE_POLL_NAME } from './constants';
 import { RelatedEntityObject, VideoObject } from './types';
 
 export function extractVideoId(idOrUrl: string) {
+  const host = process.env.PUBLIC_URL || location.host;
+  const escapedCurrentHost = host.replace(/[.\\]/g, '\\$&');
+
   const matchUrl = idOrUrl.match(
-    /(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu.be\/|tournesol\.app\/entities\/yt:)([A-Za-z0-9-_]+)/
+    new RegExp(
+      '(?:https?:\\/\\/)?(?:www\\.)?(?:youtube\\.com\\/watch\\?v=|youtu\\.be\\/|' +
+        escapedCurrentHost +
+        '\\/entities\\/yt:)([A-Za-z0-9-_]+)'
+    )
   );
   const id = matchUrl ? matchUrl[1] : idOrUrl.trim();
   if (isVideoIdValid(id)) {
--- a/frontend/src/utils/video.ts
+++ b/frontend/src/utils/video.ts
@@ -4,7 +4,7 @@ import { RelatedEntityObject, VideoObject } from './types';
 
 export function extractVideoId(idOrUrl: string) {
   const matchUrl = idOrUrl.match(
-    /(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu.be\/)([A-Za-z0-9-_]+)/
+    /(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu.be\/|tournesol\.app\/entities\/yt:)([A-Za-z0-9-_]+)/
   );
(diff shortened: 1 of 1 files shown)


frontend/src/utils/video.ts
```diff
@@ -4,7 +4,7 @@ import { RelatedEntityObject, VideoObject } from './types';
 
 export function extractVideoId(idOrUrl: string) {
   const matchUrl = idOrUrl.match(
-    /(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu.be\/)([A-Za-z0-9-_]+)/
+    /(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu.be\/|tournesol\.app\/entities\/yt:)([A-Za-z0-9-_]+)/
```
GresilleSiffle, line 7: I found an old bug in the original regex.

Look the middle path:
 
capture

It shouldn't accept any character, only the dot:

capture2

Accepting this suggestion will fix the problem.

```suggestion
    /(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu\.be\/|tournesol\.app\/entities\/yt:)([A-Za-z0-9-_]+)/
```

GresilleSiffle (changes requested):
This looks good, thanks for the work :+1: 

I found few limitations:
- it won't work on staging, because the staging hostname is `staging.tournesol.app`
- it won't work in local, because the string `localhost:3000` != `tournesol.app`

We have several options:
1. use the `location.host` variable instead of the string `tournesol.app`
2. hardcode `staging` and `localhost:{port}` in the regex
3. something else?

The side effects of the option 1:
- copy/pasting URLs between local and staging and production will work.
- we can change of domain name, the copy/paste will still work

Sbyspaceopera:
@GresilleSiffle I think that the first solution is quite clean. I'll work on it this night or tomorow.

Note that `location.host` can't be used as it is in a regex litteral : 

- This can be done with the `RegExp` Object constructor. 
- Some characters in the value returned by `location.host` needs to be escaped automatically to not conflict with the regex, just as we seen it with `youtu.be`.

GresilleSiffle:
It looks good, we can even use this:

```javascript
  let host = process.env.PUBLIC_URL;
  if (!host) {
    host = location.host;
  }

  const escapedCurrentHost = host.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');

  const matchUrl = idOrUrl.match(
    new RegExp(
      '(?:https?:\\/\\/)?(?:www\\.)?(?:youtube\\.com\\/watch\\?v=|youtu\\.be\\/|' +
        escapedCurrentHost +
        '\\/entities\\/yt:)([A-Za-z0-9-_]+)'
    )
  );
```

Here we use the public URL configured by the environment variables. This variable may not be defined in development environments (like the `dev-env`) so there is a fallback on the browser API `location`.

Sbyspaceopera:
I just commit the @GresilleSiffle version.

I also reduce this line :

`const escapedCurrentHost = host.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');`

To this :

`const escapedCurrentHost = host.replace(/[.\\]/g, '\\$&');`

As "jst" mentionned on the Discord, there was a lot of unnecessary escaping in this regex regarding our urls.

Is it ok like this for you?

frontend/src/utils/video.ts
GresilleSiffle: Fixed by 7b11bff05fb208e35503b6f9c8b8b90425ec2926.

GresilleSiffle (approved):

Issue #1022, opened 2022-06-29
Rename vouching user fields 


lenhoanglnh:
I rewrote the code to make it more human-readable. I verified that it successfully pass the tests.
There may be a need to go further to avoid ambiguities, as the word "trust" is used with too many slightly different meanings.

In particular, I would suggest:

    The user's field "trust_score" could be replaced with "voting_right".
    The vouch's field "trust_value" could be replaced by "vouch_value".
    The vouch's field "by" could be replaced by "from".

As a result, while "trust" remains not fully well defined (currently, in the code, it only matters if it is > 0), the word "relative_trust" implies a sum of all users' relative trusts equals 1. Conversely, a user's "voting_right" an take any positive number. Moreover pretrusted users (by email domains) are guaranteed to have a voting right of at least 1.

PR #1034, opened 2022-07-06
[back] refactor: rename few vouch fields

GresilleSiffle:
**related to** #1022 

---

I didn't rename `by` into `from`, because `from` is a reserved key-word by Python.

I think we could use only `value` instead of `vouch_value`:
- writing `voucher.value` feels more natural than `voucher.vouch_value`
- the object `voucher` is acting as a namespace, so we know we are accessing the value of the voucher

What do you think @lenhoanglnh @amatissart ?


backend/core/models/user.py
```diff
@@ -187,10 +187,10 @@ class User(AbstractUser):
         validators=[validate_avatar],
         null=True,
     )
-    trust_score = models.FloatField(
+    voting_right = models.FloatField(
```
lfaucon, line 190: I'm not very convinced by this name change. Isn't this field exactly how much we trust that the user is a real user?

I don't know well enough how we go from (certified emails + vouching) -> trust scores -> voting weight -> impact on the aggregation of votes. But, I find that it is meaningful to keep track of "trust scores" through the process.

If this renaming was agreed by @lenhoanglnh and @amatissart , then go ahead :)

lfaucon (commented):
Letting other reviewers approve

backend/vouch/models.py
```diff
@@ -33,9 +33,9 @@ class Voucher(models.Model):
     is_public = models.BooleanField(
         default=False, help_text="Should the voucher be public?"
     )
-    trust_value = models.FloatField(
+    vouch_value = models.FloatField(
```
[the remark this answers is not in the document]
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1034/commits/22224c632efcaf50b17c05ba32171d2e9fd395a0.

backend/core/models/user.py
amatissart: As of now, I would also find preferable to keep "trust_score" as the field name. It's not clear yet how this score will be used in the ML part, and whether it will be used directly as a voting right or if other processing will be applied.   
So "trust_score" seems to be more explicit of what is currently computed.

GresilleSiffle: The name `trust_score` seems more accurate for me to.

Voting right may be the result of several parameters : a confidence that the user is not a robot, and a certain expertise in a particular domain. Trust score seems more related to how much we trust this user is a human.

How about `human_trust` or `human_confidence`?

amatissart (approved):

PR #1035, opened 2022-07-07
[front] feat: tutorial now uses the browser language

GresilleSiffle:
**related to** #844

---

This PR use the `recommendationsLanguagesFromNavigator()` function to filter the entities selected by the `videos` tutorial.

The `getTutorialVideos()` function expects to have at least one language returned by `recommendationsLanguagesFromNavigator()`.


frontend/src/utils/polls/videos.ts
```diff
@@ -11,22 +12,22 @@ let VIDEOS: Promise<Recommendation[]> | null = null;
...
+  const metadata: Record<string, string | string[]> = {};
 
   const minutesMax = 5;
   const top = 100;
 
-  const metadata = {
-    'duration:lte:int': 60 * minutesMax,
-  };
+  metadata['duration:lte:int'] = (60 * minutesMax).toString();
+  metadata['language'] = recommendationsLanguagesFromNavigator().split(',');
```
lfaucon, line 30: Wouldn't it be better to do something like in the recommendation page?

```
let loadedLanguages = loadRecommendationsLanguages();

if (loadedLanguages === null) {
  loadedLanguages = recommendationsLanguagesFromNavigator();
  saveRecommendationsLanguages(loadedLanguages);
}
```

For the scenario where a user first plays with recommendations, then goes to the tutorial?

I think we will want to give access to the tutorial even to users that have been on Tournesol for some time already. For example if we make an advance tutorial for new complex feature.

lfaucon (approved):
Approve conditionally of my comment being addressed.

frontend/src/utils/polls/videos.ts
GresilleSiffle: Good question.

I first thought about having the same behaviour on both the recommendation page and the tutorial, but finally I decided to propose a more simple approach for the tutorial.

The current implementation avoids the following problem:
- Italian-speaking users go to the recommendation page
- they play with the language filter, to see if Tournesol recommends video in Korean
- they finally start the tutorial
- suggested videos are in Korean, but they speak only Italian

This is not a problem right now, as the tutorial is not easily accessible, the link is only displayed once after the e-mail activation, so back and forth between tutorial and the recommendation page is very very uncommon ; but this will be a problem (maybe a small problem?) when the tutorial link will be displayed on the home page.

> I think we will want to give access to the tutorial even to users that have been on Tournesol for some time already. For example if we make an advance tutorial for new complex feature.

This is a good idea, but do we have already clear ideas about this new tutorial?

The comparison series allow use to create different kind of series, including tutorials, so it's even possible to have a very basic friendly tutorial for new users, using navigators' languages, and a more complete one for "old" users using the recommendation languages.

The only reason that makes me want to keep the implementation as-is, is to potentially avoid confusing new users. But I understand we may want to rework this in the future (also, it could be a good idea to have a language selector directly in the tutorial).

What is your opinion in this @lfaucon ? Would you like to sync the recommendations languages with the tutorial?

adrausch (approved):
I think developing 2 tutorials is a bit weird. I suppose that the FAQ will answer the advanced questions and the advanced tutorial would not be necessary. 
On the other hand, we definitely need a way to communicate with the non-Discord users to let people know that complex features have been added to the app. 
Suggestion: for each tutorial update, we "force" the tutorial for every users who have not their last comparison in the meanwhile (in order to be sure that they saw the updated tutorial)?

As "old" user I'd like to access the tutorial if I want to do a demo for example. Maybe in the case of old user, the objective is not to test the platform but just read the tutorial, so we could maybe reuse comparison already done (in order to let them see the tutorial without having to watch new videos or researching watched videos)?


For this PR: synchronizing recommandations languages with the tutorial seems ok for me. I personally have my navigator in english and I'd be ok to have suggestion in french. And I think people with non-native navigator language outnumber those who play with the recommendation filter languages... But I don't know: both solutions seem consistent and defensible, so I let you decide ^^'

GresilleSiffle:
> On the other hand, we definitely need a way to communicate with the non-Discord users to let people know that complex features have been added to the app.

That's a good idea. We could send an release note by email to those who subscribed to the Tournesol updates newsletter (which doesn't exist yet).

> Suggestion: for each tutorial update, we "force" the tutorial for every users who have not their last comparison in the meanwhile (in order to be sure that they saw the updated tutorial)?

I think it will depend on what changed in the tutorial. For instance:
- rewording some dialog : notifying every user might not be required
- adding a new feature on the comparison page : it should not be necessary for users to go through the whole tutorial again simply to see the new feature ; Discord announcements / Tournesol updates newsletter should be enough to present the additions

I agree that we may want to update the tutorial to make it exhaustive and bigger. Because one big tutorial may feel exhausting, if the tutorial grows bigger I'd suggest to split it in several independent parts. 
- part 1: the basis
- part 2: the entity selector
- part 3: etc. etc.

Also, I don't think it's a good idea to "force" the users to go through a comparison series, but I understand the idea of making the tutorial available again, notifying the users, and inviting them to check what's new. 

> As "old" user I'd like to access the tutorial if I want to do a demo for example. Maybe in the case of old user, the objective is not to test the platform but just read the tutorial, so we could maybe reuse comparison already done (in order to let them see the tutorial without having to watch new videos or researching watched videos)?

Yes being able to play again the tutorial is something we want, as far as I remember.

On the other hand reading again the dialog appears to me as a different use case. It may be better to make all text available somewhere (displayed a dialog appearing after a click on a help button), instead of being force to play again the tutorial to read it.

> For this PR: synchronizing recommandations languages with the tutorial seems ok for me. I personally have my navigator in english and I'd be ok to have suggestion in french. And I think people with non-native navigator language outnumber those who play with the recommendation filter languages... But I don't know: both solutions seem consistent and defensible, so I let you decide ^^'

Now, I don't know neither : p As it's not easily possible for a user to see the recommendations, change the languages, and go back to the tutorial, both approach have the same limitations for me: they don't allow the user to explicitly change the tutorial language. Using the language filter in the recommendation is a nice side effect, but is very opaque. Will users understand that playing with the language filter will affect the tutorial?

I think having a language selector in the tutorial is a good option:
- users don't have to go to the recommendation page, and guess that the language filter has an impact on the tuto.
- users don't have to leave the tutorial to select a language
- maybe the back end can provide the list of available languages

frontend/src/utils/polls/videos.ts
lfaucon: > This is a good idea, but do we have already clear ideas about this new tutorial?

No

> suggested videos are in Korean, but they speak only Italian

Good point, I did not anticipate this potential issue

> What is your opinion in this @lfaucon ? Would you like to sync the recommendations languages with the tutorial?

I'm convinced about your argument for not syncing both

I realize that even relying on browser language may break, because for many language we don't have recommended videos.

What about hard coding for the moment: "if languages don't contain fr or en, then add en (or selected interface language)" ? We can justify this as the website is only available in fr/en.

GresilleSiffle: > What about hard coding for the moment: "if languages don't contain fr or en, then add en (or selected interface language)" ? We can justify this as the website is only available in fr/en.

It sounds reasonable to me.

GresilleSiffle: Addressed by 2553f953ca24c61dc1ac8e7e63661929519e20ad.

Issue #902, opened 2022-05-10
Search index


amatissart:
### Goals
* Search robust to typos, multiple words (in any order)
* Search with synonyms (optional)

### Questions
* Where to store the index?
  *  The easiest solution would be to use Postgres (single source of truth)
  * "whoosh" writes index files on disk. When are they generated?

* How to update the index, when an entity is added / updated / removed?
  * Do we need to rebuild the whole index periodically? 

* Preprocessing steps
  * tokenization / stemming / etc.
  * synonyms, with Wordnet ? at query-time? at index-time?  

* Evaluate "whoosh" to build the index.
  * Any feature that is not already implemented in PG? (https://www.postgresql.org/docs/13/textsearch.html)

* How long to build the index?

glerzing:
> Any feature that is not already implemented in PG?

In terms of functionality, I think PostgreSQL has indeed all the needed features (indexing, weighting, synonyms...). It looks a bit rigid though, so if we want to do some advanced functionality or customization that is not directly supported, it could be a problem. 

> Where to store the index?

Django-Haystack can be used with Whoosh to store the index either in files or in PostgreSQL. But in my opinion (without much practical experience with databases), storing it in a file or temporary database element would be easier to maintain, and if the index is rebuild at startup, this would regularly clean any inconsistency between the index and the entities table. And it would make the database lighter. But I don't know yet how long it is to rebuild the whole index.

> Do we need to rebuild the whole index periodically?

No, just I guess when an entity is modified. Partial indexing is supported.

It is complicated to make a decision, but here are in my opinion the pros and cons of using PostgreSQL instead of (Django-Haystack + Whoosh) :
pros:
- less or even no dependencies if we use the PostgreSQL thesaurus
- More maintained than Django-Haystack and Whoosh
- In terms of performances, it looks like Elastic Search > PostgreSQL > Whoosh

cons:
- scalability (a benefit of Django-Haystack is that it is a common interface with multiple search engines, which facilitates migration to higher-traffic solutions, like Elastic Search)
- The Django interface in django.contrib.postgres.search looks limited, so I think it requires writing SQL. The building of the index does not look simple too.
- possibility that some advanced feature is missing or, more likely, that it does not allow enough control and customization over the index.

glerzing:
Performances avec et sans index GIN:
[explain with GIN.txt](https://github.com/tournesol-app/tournesol/files/8952088/explain.with.GIN.txt)
[...]

PR #1039, opened 2022-07-09
full-text search (#902)

glerzing:
design choices, notes, considerations... :

1 - Choice of Postgres full-text search (instead of the more basic Whoosh and the more advanced ElasticSearch)

2 - Creation of specific Postgres configurations with the unaccent filter dictionary added (the only difference is that accents are ignored). 

3 - Relevance score that does not depend on the length of the text (a bit unfair for entities with short descriptions, but it probably gives better results in my opinion, c.f. "normalization option" in https://www.postgresql.org/docs/current/textsearch-controls.html).

4 - No autocorrection using trigrams (looks more suited to searches without indexes - some extensions exist but not sure the gain is worth the additional complexity)

5 - No synonyms search. Postgres supports it, but the difficulty is to find synonyms dictionaries. It looks usually used in the litterature for purpose-specific needs, not with every synonyms of every word.

6 - The search is fast (the queryset is usually evaluated in around 30 ms), but building every search vector can be long (15 seconds on my fast computer). This is because every entity is separately updated with the new search_vector value. I tried to use the queryset update function, which would be much faster, but eventually gave up because the config parameter of SearchVector does not seem to accept `F("metadata__language")`, it looks like it needs an explicit value.

7 - Not including the subtitles into the full-text search. Mainly because it is already long to build the search vector.

8 - The entity "save" method is overridden to automatically update the SearchVectorField. This costs some extra time, but it would have been complicated to do with Postgres triggers (because of variable configs, weights, metadata formats...). 

9 - Score function that mixes the total_score with the relevance score. I chose a function that promotes more the search relevance than the total_score. It would be simpler to have a function like `relevance * (weights_sum + total_score)` , but perhaps it would downgrade relevant but badly-rated entities too much.

10 - A GIN index was added to SearchVectorField. The section "Index size/usage statistics" of https://wiki.postgresql.org/wiki/Index_Maintenance shows how to display the index sizes (7 Mbytes in dev-env) and usage statistics. It sometimes happened that Postgres randomly chose not to use it, making the queryset processing 3 times longer.
Here are the queryset exaplin logs (at some point when Postgres was not using the GIN index for querys of only 1 word), but don't waste too much time trying to make sense out of it :
[used_gin.txt](https://github.com/tournesol-app/tournesol/files/9075798/used_gin.txt)
[unused_gin.txt](https://github.com/tournesol-app/tournesol/files/9075799/unused_gin.txt)

11 - Filter via a normal search ("qs.filter(search_vector=search_query)") and then make a SearchRank. Instead of firstly making a SearchRank to annotate the relevance and then filtering on some relevance threshold. It gives faster results.

12 - If no language is given in the query (or none of the given languages are supported, Postgres supports fewer languages than our language filters), use a classical search without search vector. This can be problematic if users usually have no recommendations language filter. Possible solutions include searching for every possible languages even if it is slower, or picking a few languages by default, or creating a 2nd SearchVectorField with a language-independent config... This may be worth a short discussion on the vocal chat.

I'll be available for a Discord vocal chat most of the days this week if you have questions


GresilleSiffle (commented):
I made a first quick review, mainly focused on how the code looks like and minor suggestions.

Tell me what do you think @glerzing : )

I'm going to test the feature now.

backend/tournesol/tests/test_text_search.py
```diff
@@ -0,0 +1,712 @@
...
+            self._make_url(query, "fr"),
+            format="json"
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["count"], 1)
+
+    def test_languages_filter(self):
+        """
+        There should be a filter on the entities' languages
```
GresilleSiffle, line 326: Given the fact it's not explicit here that `self.url` contains the `en` language filter, it's not easy to understand why the `fr` entities are not returned.

I suggest to make it explicit that `fr` entities don't match the query languages.

```suggestion
        Only entities matching the query languages must be returned.
```

backend/tournesol/tests/test_text_search.py
```diff
@@ -0,0 +1,712 @@
...
+        Postgres full-text search does not support as many languages
+        as the Tournesol query languages filter.
+
+        If all the languages of the query are not supported by Postgres,
+        a classic case-insensitive, language-independent search is done.
+
+        Just check that the search still works.
+        """
+        word = "kook"
+        self._create_rated_entity(field1=word, language="et")
```
GresilleSiffle, line 364: Should we use `self.language_unknown_by_postgres` here? and in every place where "et" is used?

backend/tournesol/tests/test_text_search.py
```diff
@@ -0,0 +1,712 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["count"], len(scores))
+        self.assertEqual(len(response.data["results"]), response.data["count"])
+
+        sorted_scores = sorted(scores, reverse=True)
+        for index, score in enumerate(sorted_scores):
+            entity = entities[scores.index(score)]
+            self.assertEqual(response.data["results"][index]["uid"], entity.uid)
+
+    def test_public_recommendations_sorting_depends_on_total_score(self):
```
GresilleSiffle, line 399: I'd add the term "personal" here.

`def test_public_personal_reco_sorting_depends_on_total_score`

backend/tournesol/tests/test_text_search.py
```diff
@@ -0,0 +1,712 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["count"], len(scores))
+        self.assertEqual(len(response.data["results"]), response.data["count"])
+
+        sorted_scores = sorted(scores, reverse=True)
+        for index, score in enumerate(sorted_scores):
+            entity = entities[scores.index(score)]
+            self.assertEqual(response.data["results"][index]["uid"], entity.uid)
+
+    def test_private_recommendations_sorting_depends_on_total_score(self):
```
GresilleSiffle, line 430: I'd add the term "personal" here too.

`def test_private_personal_reco_sorting_depends_on_total_score`

backend/tournesol/tests/test_text_search.py
```diff
@@ -0,0 +1,712 @@
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["count"], len(scores))
+        self.assertEqual(len(response.data["results"]), response.data["count"])
+
+        sorted_scores = sorted(scores, reverse=True)
+        for index, score in enumerate(sorted_scores):
+            entity = entities[scores.index(score)]
+            self.assertEqual(response.data["results"][index]["uid"], entity.uid)
+
+    def test_fields_weights(self):
```
GresilleSiffle, line 462: As this test is working with videos, I'll use `video` in the title.

`def test_video_fields_weights`

backend/tournesol/tests/test_text_search.py
```diff
@@ -0,0 +1,712 @@
...
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["count"], 2)
+        self.assertEqual(response.data["results"][0]["uid"], relevant_entity.uid)
+        self.assertEqual(response.data["results"][1]["uid"], less_relevant_entity.uid)
+
+    def test_multiple_words(self):
+        """
+        Test with multiple words in the query.
+        Leave the freedom to accept or reject an entity for
```
GresilleSiffle, line 632: Leave the freedom to who? : P

If we are delegating a decision to an underlying system, we can make it more explicit by writing the name of the system.

Example: The PostgreSQL full-text search can discard results that seem less relevant to it, when only one word matches for instance. 

If it was an unexpected behaviour that we cannot configure, we can mention it too.

Example: Due to a limitation of the PostgreSQL full-text search... etc.

GresilleSiffle (commented):
Thanks a lot for the exhaustive test suite. This is really appreciated :+1: 

I made a lot of small suggestions, to make the code more explicit and hopefully more accessible. It was already accessible, but I found that some parts can benefit from a little bit of verbosity.

Mainly `self.url` which looks like an innocent variable, containing a simple URL, but in fact acting like a specific search URL related to the `en` language. I made a suggestion to make it more intriguing, and arouse curiosity of other developers.

amatissart:
>**13 (to deliberate)** - If no language is given in the query (or none of the given languages are supported, Postgres supports fewer languages than our language filters), use a classical search without search vector. This can be problematic if users usually have no recommendations language filter. Possible solutions include searching for every possible languages even if it is slower ( ~ 120 ms instead of 30 ms), or picking a few languages by default, or creating a 2nd SearchVectorField with a language-independent config... This may be worth a short discussion on the vocal chat.

This question has been discussed during the weekly meeting. Here is the suggested solution when no language filter is defined:
 * build a `SearchQuery` with at most 2 languages: English + the UI language (can be accessed via Django's `request.LANGUAGE_CODE`, based on the `Accept-Language` header)
 * in this case no explicit language filter is applied in the database query: the results may be in any language.
 
 @glerzing Does it seem a reasonable solution to you? Or should we expect irrelevant results when mixing different language configurations in the `SearchQuery` and the `SearchVector`  being queried?
 
 ---
 
>**14 (to deliberate)** - Shouldn't the frontend someway automatically remove the date filter "A month ago" to return more results? This is especially important because people (including the past me) may not realize that they get very few results because of this filter.

This is a separate issue, which may influence the behavior of the filters in other contexts. A new issue #1042 has been created (thanks @GresilleSiffle)

backend/tournesol/views/polls.py
```diff
@@ -56,7 +57,7 @@
                 examples=[
                     OpenApiExample(
                         name="Videos - some examples",
-                        value={"language": "fr,pt", "uploader": "kurzgesagtES"},
+                        value={"language": ["en", "pt"], "uploader": "Kurzgesagt – In a Nutshell"},
```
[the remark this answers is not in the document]
glerzing: I hesitated to make an empty example, used by default. So that you don't have to clear it when you don't want metadata filters. Same for the weights.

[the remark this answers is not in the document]
glerzing: Tell me if it's worth doing

backend/tournesol/tests/test_text_search.py
```diff
@@ -0,0 +1,712 @@
...
+    The tests are based on many design choices, that are likely to
+    change over time.
+    """
+
+    def setUp(self):
+        self.poll = Poll.default_poll()
+
+        self.query = "fruit"
+
+        self.field = "10 best fruit cake recipes"
```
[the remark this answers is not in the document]
glerzing: Since there are numerous occurrences, I'll use self.text. 
I'll maybe keep the term "field" for _create_rated_entity, making it more explicit that it is metadata fields. Or do you think it's better to also use "text1", "text2", .... ?

backend/tournesol/tests/test_text_search.py
glerzing: I'll use the term "contributor" instead of "personal", because that's what is used elsewhere

glerzing: Even if "personal" is arguably more adapted

backend/tournesol/tests/test_text_search.py
glerzing: Now that I make a pre-filtering ("qs.filter(search_vector=search_query)" in filter_search), all the query words (except stop words) have to appear in the entity in order to not get filtered. 
It's not obvious if it is the desired behavior. But I'll thus consider that only 1 result should be returned.

backend/tournesol/entities/base.py
```diff
@@ -220,12 +217,70 @@ def refresh_metadata(self, *, force=False, save=True):
...
+            if isinstance(languages, str):
+                languages = [languages]
+
+            for language in languages:
+                pg_language = language_to_postgres_config(language)
+                if pg_language:
+                    if search_query:
+                        search_query |= SearchQuery(text_to_search, config=pg_language)
+                    else:
+                        search_query = SearchQuery(text_to_search, config=pg_language)
```
amatissart, line 259: After digging a bit more into how `SearchQuery` works in Django, I realized that it's possible to use a dynamic expression as the config value:
https://docs.djangoproject.com/en/4.0/ref/contrib/postgres/search/#changing-the-search-configuration

It could bring some simplification: in our use case: as the SearchVector is indexed with the configuration related to the video language, we could simply reuse the same configuration (per entity) in the SearchQuery.  
With something like:
```python
search_query = SearchQuery(text_to_search, config=F('metadata__language'))
```
(The exact SQL expression would be slightly more complicated than this, to handle the "customized_" configuration name, and possibly define a fallback for unsupported languages.)

The advantage is this `SearchQuery` is now independent from the language filter: if no language filter is defined we can search among all videos: the preprocessing will be applied automatically on the correct language for each entity.

@glerzing Do you think that could work?

glerzing: A possibility would be to use the same format for customized configs and language metadata (e.g. "en" instead of "customized_english").

It does not look easy however to handle languages that exist in the filters, but that are not supported by Postgres (and the metadata for other polls may not have a language field).

amatissart: To make it simpler, why not saving the configuration name used to create the SearchVector in a new column "search_config_name"? Then, we would be free to define our own config/fallbacks, and we can simply use `F('search_config_name')` as the search config.

glerzing: And maybe we could create a generic, language-independent configuration (like "customized_simple" in reference to Postgres's "simple" config) for languages not supported by Postgres instead of having fallbacks.
I suppose that saving the metadata should also automatically update this language config field.

backend/tournesol/views/polls.py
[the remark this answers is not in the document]
GresilleSiffle: It's true that I'm often deleting the default example before using the API documentation... I think it's worth it.

backend/tournesol/tests/test_text_search.py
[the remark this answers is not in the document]
GresilleSiffle: I think it's good to use `self.text` as an attribute of the test case, and I agree with you to keep the abstract `field` arguments in the `_create_rated_entity` method as it is used to fill different fields depending on the entity created.

GresilleSiffle:
Thank you @glerzing for having addressed all my previous comments :ok_hand:

backend/tournesol/entities/base.py
glerzing: Using `F('search_config_name')` isn't very efficient because for some unknown reason, it doesn't exploit the GIN search index. Simply concatenating the `SearchQuery` is simpler and surprisingly gives faster results because it exploits the GIN index, though the SQL doesn't look satisfying. 
Sometimes also the search is 3 times longer, and after a restart of dev-env, the duration looks normal again.

I pushed the branch "search_algo_debug", @amatissart , so that you can give a check if you want.

Here are logs for a 1-word search:
[using search_config_name.txt](https://github.com/tournesol-app/tournesol/files/9126903/using.search_config_name.txt)
[collation of all configs.txt](https://github.com/tournesol-app/tournesol/files/9126904/collation.of.all.configs.txt)

Here are logs for a 2-words search:
[using search_config_name - 2 words.txt](https://github.com/tournesol-app/tournesol/files/9126905/using.search_config_name.-.2.words.txt)
[collation of all configs - 2 words.txt](https://github.com/tournesol-app/tournesol/files/9126906/collation.of.all.configs.-.2.words.txt)

glerzing: Actually, it makes sense that Postgres doesn't use the index with F('search_config_name'), @amatissart. Because in this case, to know all the configs that are being used for SearchQuery, it needs to read search_config_name in every row. And, the index is about not having to read every row. So, I guess there is no point anymore about using the index in this case.

glerzing: Still at least 3 possible solutions:
- having a 2nd SearchVectorField always using the "generic" configuration => **more code**, the time overhead of the 2nd SearchVectorField update however should be small, and not sure if the overhead of the 2nd GIN index would be significant too.
- using english + UI language => **ignoring videos** in some other languages during the search
- concatenating the SearchQuery of all the languages => **slower search** processing

glerzing: I pushed a branch that implements solution 1: "search_algo_generic_vector". The search works as expected.

Building all the search vector in this case lasts 25 seconds instead of 16 seconds. But this is only done during the migrations, so if the dump.sql is updated, this shouldn't be necessary anymore. It was possible to almost eliminate this overhead using queryset update, but it was complicating too much the code and maintainability.

amatissart: I tried some other approaches to understand exactly what are the constraints to use the index defined on "search_vector".

> It makes sense that Postgres doesn't use the index with F('search_config_name'), @amatissart. Because in this case, to know all the configs that are being used for SearchQuery, it probably needs to read search_config_name in every row. And, the index is about not having to read every row. So, I guess there is no point about using the index in this case.

It's actually possible to use the index in this case without reading every row. But that requires the set of possible language configurations to be known in advance, to call the index at most once for each language. Unfortunately Postgres can't figure it out by itself, even by modifying the type used in column "search_config_name" or defining an additional index. The only solution seems to use an explicit join of "pg_ts_config" which contains the list of available language configurations. Then, a "nested loop" can be used instead of an expensive seqscan:

```
EXPLAIN ANALYZE SELECT e.uid
                    FROM tournesol_entity e
                    INNER JOIN pg_ts_config c ON c.oid = e.search_config_name::regconfig 
                    WHERE e."search_vector" @@ (plainto_tsquery(oid, 'medecine'));
                    
                                                                     QUERY PLAN                                       
                              
----------------------------------------------------------------------------------------------------------------------
------------------------------
 Nested Loop  (cost=11.90..2849.02 rows=6 width=14) (actual time=2.468..4.022 rows=50 loops=1)
   ->  Index Only Scan using pg_ts_config_oid_index on pg_ts_config c  (cost=0.14..8.83 rows=46 width=4) (actual time=
0.014..0.028 rows=46 loops=1)
         Heap Fetches: 0
   ->  Bitmap Heap Scan on tournesol_entity e  (cost=11.76..61.73 rows=1 width=186) (actual time=0.083..0.085 rows=1 l
oops=46)
         Recheck Cond: (search_vector @@ plainto_tsquery((c.oid)::regconfig, 'medecine'::text))
         Filter: (c.oid = ((search_config_name)::regconfig)::oid)
         Rows Removed by Filter: 30
         Heap Blocks: exact=1392
         ->  Bitmap Index Scan on search_index  (cost=0.00..11.76 rows=27 width=0) (actual time=0.011..0.011 rows=32 l
oops=46)
               Index Cond: (search_vector @@ plainto_tsquery((c.oid)::regconfig, 'medecine'::text))
 Planning Time: 0.243 ms
 Execution Time: 4.071 ms
```

I tried to integrate this solution with Django's ORM. I had to extend objects manager for Entity with a raw SQL query. 
See the commit 70fab6a6 on branch `amatissart/search_algo_any_language`

Overall I think that could be a good compromise: the performance should scale correctly for any language combination in filters or data, and it remains simple to extend the query with additional configurations (stemming, etc.) or other filtering options.

glerzing: The solution looks clean, but on my environment on commit https://github.com/tournesol-app/tournesol/commit/70fab6a69ef560f1c8cecf3e04f878c8794f826a it still doesn't use the GIN index (c.f. Index size/usage statistics, https://wiki.postgresql.org/wiki/Index_Maintenance), and it is therefore pretty slow.
I'll investigate tomorrow.

[logs.txt](https://github.com/tournesol-app/tournesol/files/9176800/logs.txt)

image

glerzing: Besides, it's unrelated, but I wonder if gin_pending_list_limit  is not largely oversized (its default value is 4 Mbytes) => https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-GIN-PENDING-LIST-LIMIT
This means I suppose that it can take up to 4 Mbytes of new data before the GIN index is updated.

amatissart: > The solution looks clean, but on my environment on commit [70fab6a](https://github.com/tournesol-app/tournesol/commit/70fab6a69ef560f1c8cecf3e04f878c8794f826a) it still doesn't use the GIN index (c.f. Index size/usage statistics, https://wiki.postgresql.org/wiki/Index_Maintenance), and it is therefore pretty slow. I'll investigate tomorrow.
> 
> [logs.txt](https://github.com/tournesol-app/tournesol/files/9176800/logs.txt)
> 
> image

Yes I also observed that the index was not used consistently, probably because the planner considers that the "entity" table on my local environment is too small for the index to be efficient. 

For development purposes, using `set enable_seqscan=false` in a psql session can be useful, to discourage the planner from using "seqscan", and verify that the index can actually be used by the query.

In a second phase, if that's really an issue in production, we could try to fine tune the cost values used by the query planner: https://www.postgresql.org/docs/13/runtime-config-query.html#RUNTIME-CONFIG-QUERY-CONSTANTS
For example, `SET random_page_cost=0.3` seems to help on my local dev-env.

glerzing: With `set enable_seqscan=false`, the SQL request 
```
EXPLAIN ANALYZE SELECT e.uid
                    FROM tournesol_entity e
                    INNER JOIN pg_ts_config c ON c.oid = e.search_config_name::regconfig 
                    WHERE e."search_vector" @@ (plainto_tsquery(oid, 'medecine'));
```
uses the index as expected (once for every configuration) and is pretty fast. However, the request to the `/polls/videos/recommendations` API still doesn't use the index in my environment.

glerzing: I let you check if the new implementation is ok, @amatissart .

I added `AND c.cfgname = ANY(%s)` to filter on the customized configs. This doesn't seem to have significant effect though.

For the `ALTER DATABASE tournesol SET random_page_cost=1.1;`, I put it in the migrations. There is also effective_cache_size  that influences the index usage, but it's already set to a pretty high value by default (4 Gbytes). So I don't see other planner cost constant to modify.

backend/tournesol/entities/base.py
```diff
@@ -220,12 +217,54 @@ def refresh_metadata(self, *, force=False, save=True):
...
 
         self.update_metadata_field()
         # Ensure that the metadata format is valid after refresh
         self.instance.metadata = self.cleaned_metadata
         self.instance.metadata_timestamp = timezone.now()
         if save:
-            self.instance.save(update_fields=["metadata", "metadata_timestamp"])
+            self.instance.save(
+                update_fields=["last_metadata_request_at", "metadata", "metadata_timestamp"]
+            )
```
amatissart, line 228: @glerzing What's the motivation for this change?

I think the initial intention was to save `last_metadata_request_at` separately, as a precaution in the case when `self.update_metadata_field()` would fail unexpectedly. For videos metadata especially, it would prevent from requesting Youtube API repeatedly (and potentially abusively) because of a temporary issue (API unavailability, invalid credentials, etc).

glerzing: This was just to avoid doing unnecessary saves.
Did I miss some type of unhandled exception?
Anyway, you can revert this if you deem it safer.

backend/tournesol/views/polls.py
```diff
@@ -136,6 +146,40 @@ def filter_unsafe(self, queryset, filters):
...
+            If set to 0, total_score is ignored.
+            Set it to a higher value to take more into account the total score.
+        """
+        if filters["search"]:
+            weights_sum = self._weights_sum
+            queryset = queryset.annotate(
+                search_score=F("relevance") * (
+                     weights_sum * F("relevance") +
+                     self.search_score_coef * (weights_sum + F("total_score"))
+                )
```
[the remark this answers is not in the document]
glerzing: > For entities with negative total_score, the search_score may be negative too. As result the most relevant results will be put even lower in the ranking (due to the multiplication by "relevance").

I used `self.search_score_coef * (weights_sum + F("total_score"))` instead of `self.search_score_coef * F("total_score")`. So that, assuming that `weights_sum > 0`  and `F("total_score") <= weights_sum` (if weights_sum is not the maximum value of total_score for the given weights, replace it by this maximum value), the result should be always positive.

> the ranking is still sensitive to the scale of the weights. I would expect a similar ranking when all weights equal 10, and when all weights equal 100. With this implementation larger weights may reduce the importance of "relevance".

I may have misunderstood this. My rationale is that when the weights are multiplied by 10, weights_sum will be multiplied by 10, and F("total_score") will also be multiplied by 10. So, search_score would be exactly 10 times higher.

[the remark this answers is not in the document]
glerzing: Here is an example of what it looks like (with weights_sum equal to 100 and each curve is for a total_score in range(-100, 101, 20)):

image

backend/tournesol/entities/base.py
```diff
@@ -220,12 +217,54 @@ def refresh_metadata(self, *, force=False, save=True):
...
+
+        It is currently implemented based on the PostgreSQL full-text search.
+        If there is someday a need for high performances, we could
+        switch to ElasticSearch. But this would require managing a
+        specific server (and thus a new container).
+        """
+        # Search over every language config
+        search_query = SearchQuery(text_to_search, config=F("search_config_name"))
+        qs = qs.filter_with_text_query(text_to_search)
+        qs = qs.annotate(relevance=SearchRank(F("search_vector"), search_query))
```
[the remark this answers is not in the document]
glerzing: I just personally believe that the "density" of occurrences shouldn't have so much effect on the resulting score.

One difference between the 2 is that with cover_density, the result is not in [0, 1]. This is not necessarily a bad thing, but the current function to calculate search_score in sort_results is not well fit for unbounded values of relevance.

backend/tournesol/views/polls.py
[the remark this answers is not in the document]
amatissart: My bad, I got confused with my second point. The scaling seems to be ok.  

The mistake is due to the assumption that we always have `F("total_score") <= weights_sum`. This is not the case: with the current implementation of Mehestan, each criteria scores belongs to [-100, 100], and when no explicit weight is defined, a single criterion is considered (`weights_sum` = 1). As for Licchavi scores are not bounded, no maximum score can be defined.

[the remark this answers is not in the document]
amatissart: So, assuming we can ignore Licchavi, which is to be deprecated, we can normalize the total_score using `MEHESTAN_MAX_SCALED_SCORE`.

The implementation would be
```py
        if filters["search"] and self.poll_from_url.algorithm == ALGORITHM_MEHESTAN:
            max_absolute_score = MEHESTAN_MAX_SCALED_SCORE * self._weights_sum
            normalized_total_score = (
                (F("total_score") + max_absolute_score) / 2 * max_absolute_score
            )
            queryset = queryset.annotate(
                search_score=F("relevance")
                * (F("relevance") + self.search_score_coef * normalized_total_score)
            )
            return queryset.order_by("-search_score", "-pk")
        else:
            return queryset.order_by("-total_score", "-pk")
```

@glerzing It should be equivalent with what you described, except Mehestan scaling is now correctly taken into account. What do you think?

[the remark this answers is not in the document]
glerzing: It looks good, just don't forget the parentheses: `(F("total_score") + max_absolute_score) / 2 * max_absolute_score` => `(F("total_score") + max_absolute_score) / (2 * max_absolute_score)`

And please give a check that it returns the same search_score values in practice.

By the way, I like your other improvement commits 👍.

amatissart (approved):

backend/tournesol/entities/base.py
amatissart: I moved the `random_page_cost` value to the connection options in Django settings. This way, I think it's more visible and easier to modify/extend in the future.

backend/tournesol/entities/base.py
amatissart: I reverted the change, and added a comment to explain why it's safer.

backend/tournesol/views/polls.py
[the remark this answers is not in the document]
amatissart: > It looks good, just don't forget the parentheses

@glerzing Thanks for noticing that ^^. All questions have been addressed I think. Could you have a last look before we merge?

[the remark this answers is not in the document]
glerzing: When using Django Debug Toolbar, it returns an error on the EXPLAIN ANALYZE:
`2022-08-13 00:21:13.458 UTC [150] ERROR:  syntax error at or near "ARRAY" at character 1278`

Which is about this line:
`ON c.oid = e.search_config_name::regconfig AND c.cfgname IN ARRAY['customized_english']`

Something that solves this issue is to instead make:
`ON c.oid = e.search_config_name::regconfig AND c.cfgname = ANY(%s)`
(and to replace `tuple(search_configs)` by `search_configs`)

Do you approve this change?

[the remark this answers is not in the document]
amatissart: Ok it seems to be caused by a bug in "django-debug-toolbar", which fails to preserve tuple params: https://github.com/jazzband/django-debug-toolbar/issues/1482

I updated to query to use `= ANY(%s)`, just in case it can help with future debugging.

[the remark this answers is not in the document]
glerzing: We can always set a higher value for search_score_coef if we consider that the search doesn't take enough into account the total score (I was using this script to visualize it: [2d_model.zip](https://github.com/tournesol-app/tournesol/files/9331759/2d_model.zip) ). 

If you think the current value is ok, I have no more comments and we can push.

PR #1040, opened 2022-07-09
[back] feat: unconnected entities are now ordered by n_comparisons

GresilleSiffle:
The goal is to invite the users to compare the entities with the least comparisons first, and to make the list of unconnected entities dynamic.

Before, the front end tab "to connect" always displayed the same videos, when called with a never compared video. This was not satisfying, as it could suggest videos that have been already compared several times by the user. Now the users will see a more diverse list of videos, as its order will change after each new comparison with an unconnected entity.

I took the time to add more comments in the tests and to rename some methods, so it's now explicit what the test cases are testing: a single well connected graph, two isolated graphs, a single graph with too distant entities, etc.

**to-do**
- [x] write a test

### DIFF
--- a/backend/tournesol/tests/test_api_unconnected_entities.py
+++ b/backend/tournesol/tests/test_api_unconnected_entities.py
@@ -148,12 +148,15 @@ class SingleGraphOneIsolatedEntityTestCase(TestCase):
         self.assertEqual(response.data["count"], 3)
 
 
-class SingleGraphAllConnectedTestCase(TestCase):
+class SingleGraphNonReducibleDistanceTestCase(TestCase):
     """
     A test case of the unconnected entities API.
 
     Here, all entities are not necessarily connected together to by the tested
     user, yet forming a single graph.
+
+    No entity is distant enough from any other entity to be considered
+    connectable.
     """
 
     def setUp(self):
@@ -188,8 +191,8 @@ class SingleGraphAllConnectedTestCase(TestCase):
     def test_all_linked_must_return_empty(self):
         """
         An authenticated user must get an empty list, when all of its compared
-        entities are connected, even if they are not individually connected to
-        each other.
+        entities are connected, and close enough to each other, even if they
+        are not individually connected to each other.
         """
         self.client.force_authenticate(self.user_1)
 
@@ -216,9 +219,11 @@ class TwoIsolatedGraphsTestCase(TestCase):
         self.poll_videos = Poll.default_poll()
         self.user_base_url = f"/users/me/unconnected_entities/{self.poll_videos.name}"
 
+        # First graph.
         video_1 = VideoFactory()
         video_2 = VideoFactory()
         video_3 = VideoFactory()
+        # Second graph.
         video_4 = VideoFactory()
         video_5 = VideoFactory()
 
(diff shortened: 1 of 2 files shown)


backend/tournesol/tests/test_api_unconnected_entities.py
```diff
@@ -50,15 +52,18 @@ def setUp(self):
...
+        """
         response = self.client.get(
             f"{self.user_base_url}/{self.video_source.uid}/",
             format="json",
         )
 
         self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
 
-    def test_without_uid_show_not_found(self):
+    def test_auth_get_404_with_no_uid(self):
```
lfaucon, line 66: Minor: should it return Bad Request instead of Not Found in this case?

backend/tournesol/tests/test_api_unconnected_entities.py
```diff
@@ -140,10 +148,15 @@ def test_no_link_should_return_all(self):
...
 
-class AdvancedAllConnectedTest(TestCase):
+class SingleGraphNonReducibleDistanceTestCase(TestCase):
     """
-    TestCase for the unconnected entities API. 
-    In this Advanced test, all video are not necessary connecter to the user
+    A test case of the unconnected entities API.
+
+    Here, all entities are not necessarily connected together to by the tested
+    user, yet forming a single graph.
```
lfaucon, line 156: What you mean by "single graph" is "connected graph"

backend/tournesol/tests/test_api_unconnected_entities.py
```diff
@@ -187,9 +205,12 @@ def test_all_linked_should_return_empty(self):
         self.assertEqual(response.data["count"], 0)
 
 
-class SimpleNotAllConnectedTest(TestCase):
+class TwoIsolatedGraphsTestCase(TestCase):
```
lfaucon, line 208: The righter term would be TwoIsolatedGraphComponentsTestCase

lfaucon: There is always a single graph even if it has multiple unconnected components

backend/tournesol/tests/test_api_unconnected_entities.py
```diff
@@ -198,9 +219,11 @@ def setUp(self):
         self.poll_videos = Poll.default_poll()
         self.user_base_url = f"/users/me/unconnected_entities/{self.poll_videos.name}"
 
+        # First graph.
         video_1 = VideoFactory()
         video_2 = VideoFactory()
         video_3 = VideoFactory()
+        # Second graph.
```
lfaucon, line 226: Second component

lfaucon (approved):
Only commented nitpicky Details about language when talking about graphs. I really like the more explicit test case names!

backend/tournesol/tests/test_api_unconnected_entities.py
GresilleSiffle: I was surprised too, I need to check how the rest of the API behaves in this kind of case.

backend/tournesol/tests/test_api_unconnected_entities.py
GresilleSiffle: I see, I'll update the comment.

backend/tournesol/tests/test_api_unconnected_entities.py
GresilleSiffle: Thanks, I'll fix this.

Issue #993, opened 2022-06-14
Make dev-env compatible with Windows


glerzing:
According to Wikipedia (https://en.wikipedia.org/wiki/Usage_share_of_operating_systems), the desktop operating system share among those identifying as professional developers was:

Windows: 45.3%
macOS: 29.2%
Linux: 25.3%
BSD/Unix: 0.1%

On Windows, the docker containers usually fails with permission errors.

JST showed me some trick to make it work (removing the line user: `${DB_UID}:${DB_GID}` in docker-compose.yml and executing run-docker-compose.sh with the MinGW64 bash instead of the WSL2 bash ("./run-docker-compose.sh" instead of "bash run-docker-compose.sh", for some unknown reason, it solves permissions errors)).

But if almost half of the developers use Windows, it's important not to discourage these potential new contributors.

glerzing:
Other issues I encounter:
- Having to add a delay (e.g. `sleep 2`) before `docker exec --env PGPASSWORD=password tournesol-dev-db bash -c "psql -1 -q -d tournesol -U tournesol < /var/lib/postgresql/data/dump.sql"` for the database data to load correctly.
- `tar xvf "$CURRENT_DIR"/dump-for-dev-env.sql.tgz` not working until I replace "$CURRENT_DIR" by an absolute path

PR #1048, opened 2022-07-14
Update of dev-env, notably for Windows compatibility (#993)

glerzing:
### Modifications
1 - Not using realpath, which can give problematic results in Windows (e.g. paths that start with "C:/" instead of "/c/" in MinGW)
2 - Exporting DB_UID and DB_GID, to simplify the code => check that it doesn't have side effects on child scripts
3 - Changing the position of the "", just because it looks cleaner

It's very possible that it doesn't work on every environment, or that there still are bugs, or that it's not a clean solution.
I'll first call JST as a reviewer for his opinion, then ask someone with other OS (Mac, Windows), to see if it's an improvements and if there is no regression.
**Don't hesitate to report any remaining issue with dev-env.**

### The case of Windows
In Windows, the postgres uid and gid are fixed to 999, and can't be easily changed. So, if `$(id -u)` and `$(id -g)` were returning values different from 0 (root), or 999 (postgres), this was leading to permission issues during the container's creation. Since DB_UID  and DB_GID are now set to 0 for Windows, this should not be a problem anymore (tell me if this patch should also be applied for Mac).

So using `./run-docker-compose.sh` or `./dev-env/run-docker-compose.sh` in a Powershell or cmd should now work. At least if the shortcut to the Git bash (also called MinGW bash) `./` exists. 

However, `bash run-docker-compose.sh` or `wsl run-docker-compose.sh` should still not work. Because they will be based on WSL2, and I encountered at least 2 other issues with Docker on WSL2. It's better to stick with the Git bash than to make dirty patches to make it work on WSL2. I chose not to make a section in the README.md to explain this, it may lead to more confusion than understanding.

### DIFF
--- a/dev-env/README.md
+++ b/dev-env/README.md
@@ -4,6 +4,22 @@
 * `docker-compose-plugin` (or `docker-compose`)
 * curl
 
+
+## On Windows
+
+It is recommended on Windows to install:
+* WSL2, with a Linux distribution (from a Powershell as administrator: `wsl --install`)
+* Docker Desktop (which includes `docker-compose`)
+
+For the frontend dynamic updates to work properly, the git repository should be **cloned inside the WSL2 file system**
+
+If `git diff` displays unexpected modifications (`old mode 100755 new mode 100644`), you can execute 
+`git config core.filemode false` to indicate Git to ignore the executable bit in files permissions.
+
+Beware that `./run-docker-compose.sh` in Windows may use a shortcut to the MinGW64 bash (also called Git bash).
+If so, type `bash run-docker-compose.sh` to use the WSL2 bash instead.
+
+
 ## Start the containers
 
 ```bash
@@ -30,6 +46,7 @@ Then, the application is accessible on http://localhost:3000.
 
 The created database includes metadata about ~ 5000 videos, as well as 5 sample accounts with usernames
 `user1`, `user2`, `user3`, `user4` and `user5`, and password `tournesol`.
+
 
 ## Rebuild the containers while preserving the database
 
--- a/dev-env/README.md
+++ b/dev-env/README.md
@@ -1,12 +1,10 @@
-# Create a development environment with `docker-compose`
+# Starting the development environment
 
-## Quick start
-
-### Requirements:
+## Requirements:
 * `docker-compose-plugin` (or `docker-compose`)
 * curl
 
-### Start the containers
+## Start the containers
 
 ```bash
 ./run-docker-compose.sh
@@ -33,7 +31,7 @@ Then, the application is accessible on http://localhost:3000.
 The created database includes metadata about ~ 5000 videos, as well as 5 sample accounts with usernames
 `user1`, `user2`, `user3`, `user4` and `user5`, and password `tournesol`.
 
-#### Rebuild the containers while preserving the database
+## Rebuild the containers while preserving the database
 
 By default, the database content is initialized with test data.
 To recreate the containers (e.g to update the backend dependencies) while preserving the data, use:
--- a/dev-env/README.md
+++ b/dev-env/README.md
@@ -8,16 +8,18 @@
 ## On Windows
 
 It is recommended on Windows to install:
-* WSL2, with a Linux distribution (from a Powershell as administrator: `wsl --install`)
-* Docker Desktop (which includes `docker-compose`)
+* **WSL2**, with a Linux distribution (just execute `wsl --install` in a Powershell as administrator. More info here: https://docs.microsoft.com/en-us/windows/wsl/install)
+* **Docker Desktop** (which includes `docker-compose`. The default configuration should be fine. More info here: https://docs.docker.com/desktop/windows/wsl/)
 
-For the frontend dynamic updates to work properly, the git repository should be **cloned inside the WSL2 file system**
 
-If `git diff` displays unexpected modifications (`old mode 100755 new mode 100644`), you can execute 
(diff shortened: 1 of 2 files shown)


jstainer-kleis:
lgtm

jstainer-kleis (approved):

glerzing:
Ok, then I invite more people to give a check.

Sbyspaceopera (commented):
Work as expected on my Windows machine using GitBash.

glerzing:
> Work as expected on my Windows 10 device using GitBash.

Thanks for the confirmation. 
Just in case I wasn't clear, you can launch it from Git Bash, but if it's easier you can also use "./run-docker-compose.sh" from a cmd or Powershell and the "./" usually serves as a shortcut to the Git Bash (depending on the Git installation options).

GresilleSiffle:
I confirm it still works on Debian eh eh : )

I have question regarding the UID, if we use 0 instead of 999 in Windows, does it mean that using 999 did work?

glerzing:
At some point I though 999 was working, but actually no.

The solution works well for the backend. But there are still problems with the frontend (inotify does not work, so the frontend is not dynamically updated). The solution often suggested is to have the repository in WSL2, and to edit it remotely using VS Code in Windows (Docker Desktop would still be installed in Windows and not in Linux). 
I prefer spending one or 2 more weeks investigating before pushing to main.

GresilleSiffle:
We discussed we the rest of the team yesterday, we are just waiting for a macOS user to approve.

glerzing:
I tried to clone the repository inside the WSL2 file system, and it works much better. There can still occasionally be small things to debug, like access rights (=> `sudo chmod -R 777 tournesol`), but I chose not to say that in the README.

I hesitated to add a link to the Docker documentation (https://docs.docker.com/desktop/windows/wsl/), but it does not look entirely up-to-date and relevant for us.

I'm interested if you have remarks or improvement ideas, @Sbyspaceopera

Sbyspaceopera:
> I tried to clone the repository inside the WSL2 file system, and it works much better. There can still occasionally be small things to debug, like access rights (=> `sudo chmod -R 777 tournesol`), but I chose not to say that in the README.
> 
> I hesitated to add a link to the Docker documentation (https://docs.docker.com/desktop/windows/wsl/), but it does not look entirely up-to-date and relevant for us.
> 
> I'm interested if you have remarks or improvement ideas, @Sbyspaceopera

My systems and docker skills are quite limited.

Anyway, I tested with the WSL2 file system as you did and it just works fine. Dynamic updates are now availables for the frontend by this way.

Note that the tournesol-dev-db container returns a warning :
`chmod: changing permissions of '/var/run/postgresql': Operation not permitted`

Regarding the links. It could be nice to put the official links from the Microsoft and Docker documentations like this :
```
* [WSL2](https://docs.microsoft.com/en-us/windows/wsl/install), with a Linux distribution (from a Powershell as administrator: `wsl --install`)
* [Docker Desktop](https://docs.docker.com/desktop/windows/wsl/) (which includes `docker-compose`)
```
That will avoid new comers to look themself how to set/install those things.

Plus I think every problem that we are 100% sure it will happens when a new contributor join us should be mentionned in the README. 

What are exactly the bugs you have encountered?

Any idea why the extension's e2e test is failing about finding the RATE LATER button? It seems strange regarding the changes in this PR.

glerzing:
> chmod: changing permissions of '/var/run/postgresql': Operation not permitted 

=> from what I understood, if you don't use sudo, Docker will create the database with your user. 
It will then try to set its uid and gid with the one of user Postgres (=999, as determined by the image). If you don't use sudo, it fails to give the permissions to Postgres, prints the warning you noticed in the logs, and the database remains yours (usually, uid = 1000). 
If you used sudo (`sudo bash run-docker-compose.sh`), it will manage to give the permissions to Postgres (999), so that now if you want to use `run-docker-compose.sh` without sudo, you have to delete db-data first (`sudo rm -R db-data`), because it's not yours anymore.
I don't know if it is the same on Linux. But I think we can ignore this warning.

> Regarding the links. It could be nice to put the official links from the Microsoft and Docker documentations like this :

ok

> What are exactly the bugs you have encountered? 

=> As explained above, during debug, I sometimes had to do a `sudo rm -R db-data`.
=> other types of permission errors, on bash scripts. Maybe because I made the git clone with sudo while I shouldn't have. You shouldn't have to need sudo to execute `run-docker-compose.sh`. But I made again the while setup multiple times and didn't have this issue again.
=> at some point, an issue with docker credentials for the frontend. I removed `\\wsl.localhost\Debian\home\alain\.docker\config.json`. But if you still have this file in your WSL2 path and the frontend works, it's not a real issue.

> Any idea why the extension's e2e test is failing about finding the RATE LATER button? It seems strange regarding the changes in this PR.

The e2e test is failed this time due to a new Youtube display format. But it may have other instabilities.

glerzing:
@Sbyspaceopera , I added new commits to add the links to the official documentation, and a solution to an error with Docker credentials that I solved by deleting `~\.docker\config.json`. Another solution would be to replace "credsStore" by "_credsStore" in `~\.docker\config.json`. I don't know if it will bother people to delete a file that they don't know if it is useless. But I didn't want to enter too much into details.

Do you think this is sufficient, or do you have other suggestions for improvement?

Sbyspaceopera:
No, I have no more suggestions to improve this.
The md file seems sufficient to me but I can't help more than that about bash/docker as I said above.
As a windows user, thank you for your job @glerzing

lfaucon (approved):

PR #1052, opened 2022-07-15
[front]Fix #1050 - displayErrorForm now handle error msg datas correctly + extractVideoId works with yt:video_id

### LINKED ISSUE #1050: [front]Bug: Using yt:{video_id} in rate later page cause a JS shut down on save
### Reproduct :
You can use this text for example : `yt:F7YlkbhYq3s`
Paste it in the rate later page's input and click on "Add to my rate later list"

### Possible cause :
In `\src\features\rateLater\RateLaterAddForm.tsx`, the `handleSubmit` function use the `extractVideoId` from `\src\utils\video.ts`.
_RateLaterAddForm.tsx_ :
```javascript
const handleSubmit = async (event: React.FormEvent) => {
    event.preventDefault();
    const success = await addVideo(extractVideoId(formVideo) || '');
    if (success) {
      setFormVideo('');
    }
  };
```
The `extractVideoId` function clearly not handle the case `yt:{video_id}` and return `null` when it passed as argument cause of the `match` function behavior.
_video.ts_ :
```ts
export function extractVideoId(idOrUrl: string) {
  const matchUrl = idOrUrl.match(
    /(?:https?:\/\/)?(?:www\.)?(?:youtube\.com\/watch\?v=|youtu.be\/)([A-Za-z0-9-_]+)/
  );
  const id = matchUrl ? matchUrl[1] : idOrUrl.trim();
  if (isVideoIdValid(id)) {
    return id;
  }
  return null;
}

export function isVideoIdValid(videoId: string) {
  return !!videoId.match(/^[A-Za-z0-9-_]{11}$/);
}
```
It works in `\src\features\entity_selector\EntitySelector.tsx` because the `handleChange` function uses `trim.value()` if `extractVideoId` returns `null`. In this case, `yt:{video_id}` is exactly what the `handleChange` function needs to work correctly so it seems to work as expected.
_EntitySelector.tsx_ :
```ts
const videoIdFromValue =
      pollName === YOUTUBE_POLL_NAME ? extractVideoId(value) : null;
    const newUid = videoIdFromValue
      ? UID_YT_NAMESPACE + videoIdFromValue
      : value.trim();
    setInputValue(newUid);
    onChange({
      uid: newUid,
      rating: null,
    });
```
As this logic is not in `RateLaterAddForm.tsx`, an error occure.

### Possible solutions :
1) Extends the regex in `video.ts` to make it works explicitly with `yt:{video_id}`
2) Use the same logic from `EntitySelector.tsx` in `RateLaterAddForm.tsx`
3) Both

Solution one seems to me the best solution to avoid unnecessary duplication and avoird further errors like this one. In the other hand, making the rate later input works juste like the comparisons input can be a good aesthetic choice.

What do you think?

Sbyspaceopera:
More details on #1050 

There was two parts in this problem :

1) The rate later page's input was not calibrate for handling `yt:{video_id}`, I include this in the `extractVideoId` function to avoid further errors like this one.

2) `displayErrorsForm` function in `hooks\useNotifications.ts` used the wrong kind of value to throw in `showErrorAlert` (object instead of string).

I think it should be good like this.


Sbyspaceopera:
So, it works but the first test failure (regardless to the extension bug) let me think that this fix is probably not best way to handle this problem.

I've a lack of experience about the code base for being confident but it seems to me that the `displayErrorsFrom` should be more exclusive about its args, espacially for `reason.body`.

**_displayErrorsFrom_ in this PR** :
```javascript
const displayErrorsFrom = useCallback(
    (
      reason: ApiError | Error,
      message?: string,
      overrideStatuses?: Array<{
        status: number;
        msg: string;
        variant: VariantType;
      }>
    ) => {
        // override message from the API
        if (overrideStatuses) {
          const status = overrideStatuses.find(
            (elem) => elem.status === reason.status
          );
          if (status) {
            enqueueSnackbar(status.msg, { variant: status.variant });
            return;
          }
        }

        // HTTP 429 Too Many Requests are considered information
        if (reason.status === 429) {
          showInfoAlert(reason.body['detail']);
        } else {
          const body = reason.body;
          const newErrorMessages =
            typeof body === 'string'
              ? [body]
              : Object.values(reason['body']).flat();
          newErrorMessages.map((msg) => {
            if (typeof msg !== 'string') {
              msg = Object.values(msg as Record<string, unknown>)[0];
            }
            showErrorAlert(msg as string);
          });
        }
      } else {
        contactAdministrator('error', message);
      }
    },
    [enqueueSnackbar, contactAdministrator, showErrorAlert, showInfoAlert]
  );
```

Any advice?

amatissart:
@Sbyspaceopera `displayErrorsFrom` is actually a best-effort function to format any `ApiError` object. But as you can see, the exact format of the body cannot be known in advance: it's built dynamically by Django Rest Framework based on the request.
For example an incorrect "video_id" when adding a video in a rate-later list can cause an error such as 
```json
{
  "video": {
    "video_id": [
      "Cette valeur ne satisfait pas le motif imposé."
    ]
  }
}
```
But if the entire "video" object is missing the error structure will be different:
```json
{
  "video": [
    "Ce champ est obligatoire."
  ]
}
```

So ideally, a recursive implementation may be needed to cover more cases. But a reasonable step could also be to catch the case when the error type is not handled correctly and fallback to the generic `message` as expected.

GresilleSiffle:
Just for information : the rate-later API is going to be refactored, and this will be the good moment to harmonize its errors JSON payload with the rest of the API. 

If `displayErrorsFrom` is not adapted because of the wild format of the current rate-later API, we might want to wait the rate-later refactor before going further.

amatissart:
> Just for information : the rate-later API is going to be refactored, and this will be the good moment to harmonize its errors JSON payload with the rest of the API.
> 
> If `displayErrorsFrom` is not adapted because of the wild format of the current rate-later API, we might want to wait the rate-later refactor before going further. 

I don't think the problem is specific to the current rate-later API though. This is dependent on how DRF builds its response in case of validation errors, and it might happen with any route where nested serializers are used to parse and validate the request.  

In any case it seems relevant to extend the implementation of `displayErrorFrom` to cover these cases, and/or to define a generic fallback message for situations where the error response with status 400 can't be transformed easily into a simple human-readable message.

GresilleSiffle:
I see, that's tricky.

GresilleSiffle:
I'm thinking about two alternatives

**(a)**

How about adding a parameter to `displayErrorsFrom` to make it look for errors in each key presents in the body (instead of letting the function guess the JSON structure)?

This way we have a complete control on the process, but the process becomes a manual process. This may be helpful only if relying on the presence of a key (video) + key (video_id) + array is not accurate.

**(b)**

**note** we can override any status code in `displayErrorsFrom` thanks to the `overrideStatuses` parameter. A quick fix could be to display the same error message for each 400 errors returned by the API.
- pro: it's quick
- con: invalid video ID scheme and video ID not found in YouTube will display the same message (could be `Video not found.`)

Sbyspaceopera:
So this is the new `displayErrorsFrom` function.

I chose to implement a recursive function called `displayMessagesFromBody` to handle the @GresilleSiffle (b) option. Like this any kind of structures will be supported in `reason.body`.

Is it looks good for you?

```typescript
const displayErrorsFrom = useCallback(
    (
      reason: ApiError | Error,
      message?: string,
      overrideStatuses?: Array<{
        status: number;
        msg: string;
        variant: VariantType;
      }>
    ) => {
        // override message from the API
        if (overrideStatuses) {
          const status = overrideStatuses.find(
            (elem) => elem.status === reason.status
          );
          if (status) {
            enqueueSnackbar(status.msg, { variant: status.variant });
            return;
          }
        }

        // HTTP 429 Too Many Requests are considered information
        if (reason.status === 429) {
          showInfoAlert(reason.body['detail']);
        } else {
          const body = reason.body;
            
          function displayMessagesFromBody(errorsSource : any){
            if(typeof errorsSource === 'string'){
              showErrorAlert(errorsSource)
            }

            if(typeof errorsSource === 'object' && !Array.isArray(errorsSource)){
              Object.values(errorsSource).map( value => displayMessagesFromBody(value))
            }

            if(Array.isArray(errorsSource)){
              errorsSource.map( value => displayMessagesFromBody(value))
            }
          }
  
          displayMessagesFromBody(body)
        }
      } else {
        contactAdministrator('error', message);
      }
    },
    [enqueueSnackbar, contactAdministrator, showErrorAlert, showInfoAlert]
  );
```

Sbyspaceopera:
And what about the `any` type of `reason.bondy` / `reasonBody `?

`reason.body` type is from the `/src/services/openapi/core/ApiError.ts` code generated by the `openapi-typescript-codegen` package. It's define as `any` in the ApiError definition.

Since an `any` type is not a great practice in TS, it could be interesting to investigate in a solution to replace it. Despite this, not sure that an other type will make the job regarding the generical goal of the `displayErrorsFrom` function defined in its own comments :

>    This function provides a generic way to display all error messages returned
>    by an API call. It can be used in every form to harmonize the messages
>    processing.

GresilleSiffle:
Hello @Sbyspaceopera , I'm currently reviewing this pull request.

I added a commit, to display the errors returned by the back end, instead of displaying "This field cannot be empty" when typing an invalid video ID / UID.

GresilleSiffle:
> Since an `any` type is not a great practice in TS, it could be interesting to investigate in a solution to replace it.

Removing the explicit mention of `any` seems to appease ESLint. I think it's safe to remove it, as the function checks the type of `reasonBody` before doing anything.

GresilleSiffle (approved):

Issue #964, opened 2022-06-02
[back] refactor: make the `VideoRateLater` generic


GresilleSiffle:
**subtask of** #606

---

### technical specifications

**(a)** The current model `VideoRateLater` must be renamed into `RateLater`, and be defined into its own file in `backend/tournesol/models/rate_later.py` (instead of the current `entity.py`).

All references to the previous `VideoRateLater` must be updated accordingly.

**(b)** The new `RateLater` must contain the following fields:
- `poll` - a foreign key to the related Poll, one-to-one
- `user` - the owner of the current rate-later record
- `entity` - replace the previous `video` field
- `created_at` replace the previous `datetime_add`

The `poll` field is optional for now, but not null, and must be automatically set to the default `videos` poll (can be done by overloading the model's `save` method, and adding the property `blank=True`). Later it will be the responsibility of the front end to send the desired  poll's name to the proper serializer.

The `Meta` class of the model must be updated:

```python
    class Meta:
        unique_together = ["poll", "user", "entity"]
        ordering = ["user", "-created_at"]
```

And the `__str__` method too to include the poll id. Could be:

```python
    def __str__(self):
        return f"poll:{self.poll}/user:{self.user}/entity:{self.entity}@{self.created_at}"
```

Note that calling `__str__` must not trigger extra SQL queries. Better to keep the `__str__` very lazy, we can use another method the day we will want to display a human readable representation of a `RateLater` object.

All references to the previous fields must be updated accordingly.

PR #1055, opened 2022-07-18
[back] refactor: make the `RateLater` model generic

GresilleSiffle:
**related to** #964

**refactor of the PR** #1000 

---

This pull request makes the `RateLater` model generic, compatible with polls and entities.

While the model is generic, the related API is still legacy, tightly bound to the default poll, and representing JSON data as videos, to allow the front end and the browser extension to continue working.

This PR is the first step of the following plan:
1. make `RateLater` generic
2. add new endpoints `/users/me/ratelater/{poll_name}/` and new serializers, using entity instead of video
  - create a new test case for these endpoints
  - keep the legacy endpoints for the front end and the browser extension
  - remove the default value for the field `RateLater.poll`
3. make the front end use the new endpoints, the rate later list should now display the number of comparisons
4. make the extension use the new endpoints
5. finally delete the legacy endpoints and their tests


backend/tournesol/tests/factories/entity.py
```diff
@@ -51,5 +51,5 @@ class RateLaterFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = RateLater
 
-    video = factory.SubFactory(VideoFactory)
+    entity = factory.SubFactory(VideoFactory)
```
lfaucon, line 54: Do we need to add the poll subfactory ?

lfaucon (approved):
LGTM 👍

PR #1069, opened 2022-07-22
[front] fix: misleading "Active users" metric name

GresilleSiffle:
The front end used to display the metric active users.

Active users could be:
- users that have done comparisons on a regular basis
- users that still do comparisons on a regular basis
- users with a least few comparisons
- etc.

It was a bit misleading and I remember Lê asking for the meaning of this metric during a live on Discord or YouTube.

I think **activated accounts** is more appropriate.

### DIFF
--- a/backend/tournesol/views/stats.py
+++ b/backend/tournesol/views/stats.py
@@ -10,9 +10,8 @@ from rest_framework.response import Response
 
 from core.models import User
 from core.utils.time import time_ago
+from tournesol.models import Comparison, Entity, Poll
 from tournesol.serializers.stats import StatisticsSerializer
-
-from ..models import Comparison, Entity, Poll
 
 
 class ActiveUsersStatistics:
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -284,7 +284,7 @@
   "stats": {
     "ratedVideos": "Rated videos",
     "ratedCandidates": "Rated candidates",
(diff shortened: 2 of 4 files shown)


amatissart (approved):

lfaucon (approved):
A fun (maybe sad if too low) metric for Active Users would be users that did at least one comparison in the last 30 days (and the +X sould be the number of users that did one comparison in the last 30 days minus the number of users would did one comparison between 60 days ago and 30 days ago)

GresilleSiffle:
> A fun (maybe sad if too low) metric for Active Users would be users that did at least one comparison in the last 30 days (and the +X sould be the number of users that did one comparison in the last 30 days minus the number of users would did one comparison between 60 days ago and 30 days ago)

Yes more stats would be fun and insightful :+1: The metric you described would be very useful to see the currently active Tournesol community. I think it's also a good idea to not delete the Created accounts metric, even if we created a new one, because it shows how much Tournesol is attractive.

Note: the difference between Created accounts and the new metric Created accounts with at least 4 comparisons (the tutorial) can show how efficient the tutorial is.

Let's discuss about it next time, I'll be happy if we add an additional line of two metrics on the front end.

A possible layout could be the following (we can play the font-size / color of the second line to make it more accessible):

capture

glerzing:
"Comptes créés" is indeed more explicit than "Comptes activés".

I don't like however the idea of putting more metrics. Notably, "tutorials finished" isn't very meaningful.

PR #1074, opened 2022-07-26
[twitterbot] create systemd service and define schedule for automated tweets

amatissart:
The new "oneshot service" will run daily at 18:03 (Swiss time).  

On Monday, Tuesday, Thursday, Friday a tweet will be sent in French.
On Wednesday, a tweet will be sent in English. 

This new schedule is defined for production only. The timer is inactive on other environments (where twitter credentials are unavailable anyway).


aidanjungo (approved):

lfaucon (commented):
Looked at it after the fact. It looks great :) and it is working great too already

Issue #1044, opened 2022-07-13
Don't rely on the Youtube API in backend tests


glerzing:
The tests test_authenticated_can_create_with_new_video_id and test_authenticated_can_create_with_non_existing_entity are failing when the YOUTUBE_API_KEY is set. They should not be really using the Youtube API. 

A mock could be done on the API. 
Or, as a test-specific config, we could chose to set YOUTUBE_API_KEY to an invalid value.

GresilleSiffle:
We could indeed either mock the setting, or mock the `tournesol.utils.api_youtube.get_video_metadata` to force it to always return `{}`.

amatissart:
The simpler fix would probably be to decorate the affected test cases with `@override_settings(YOUTUBE_API_KEY=None)` to make sure no actual API key is used.

Mocking the API calls could also be a good idea if we'd like to check the API response is used correctly, but I think this is already covered by other test cases.

glerzing:
Overriding YOUTUBE_API_KEY does not look sufficient, because it does not cause the "youtube" variable in api_youtube.py to be recreated.

glerzing:
So I think I'll use `@patch('tournesol.utils.api_youtube.youtube', None)` instead, which seems to work.

PR #1076, opened 2022-07-31
2 backend tests were failing when YOUTUBE_API_KEY was set (#1044)

glerzing:
Overriding YOUTUBE_API_KEY was not sufficient, because it does not cause the "youtube" variable in api_youtube.py to be recreated.


GresilleSiffle:
Currently rebasing on `main`.

GresilleSiffle (approved):

PR #1081, opened 2022-08-03
[back] feat: dynamic previews for video analysis page

lfaucon:
Not our target design yet. I'm just testing the easiness of creating our website preview with Python.
Seems to serve our purpose quite well. What do you think?


amatissart:
@lfaucon Injecting a dynamic value for meta tags with Nginx is quite straightforward. The only complication is to disable the pre-compression for html files, as the substitution cannot be applied on compressed inputs.
You can see an example on this branch: https://github.com/tournesol-app/tournesol/compare/main...og-meta-nginx

Then, it's possible to use a dynamic thumbnail whose URL would look like
https://api.tournesol.app/thumbnail/comparison?uidA=yt%3AGVJRZqI6h2k&uidB=yt%3AWP6T4MH0Dn4
or
http://api.tournesol.app/thumbnail/entities/yt:GuTgfnkILGs
(to follow the URL patterns used by the frontend)

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,50 @@
...
+import requests
+from io import BytesIO
+from django.http import HttpResponse, FileResponse
+from django.utils.decorators import method_decorator
+from drf_spectacular.utils import OpenApiTypes, extend_schema
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.permissions import AllowAny, IsAdminUser, IsAuthenticated
+from rest_framework.views import APIView
+
+POPPINS_FONT = ImageFont.truetype("../frontend/public/fonts/Poppins-Bold.ttf", 40)
```
amatissart, line 11: This relative path will not work on the dev-env (as the CI shows) or on deployed environments, as "frontend" and "backend" are assumed to run independently. On the dev-env, "frontend" and "backend" live in separate volumes, and on the VMs they are deployed in distinct folders: "/srv/tournesol-frontend" and "/srv/tournesol-backend".

If this font file becomes a dependency of the API, I would recommend to copy it somewhere into "backend".

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,50 @@
...
+    permission_classes = []
+
+    @extend_schema(
+        description="preview the ",
+        responses={200: OpenApiTypes.BINARY},
+    )
+    def get(self, request):
+        default_preview = open("../frontend/public/tournesol_screenshot_og.png", "rb")
+        response = FileResponse(default_preview, content_type="image/png")
+        return response
```
amatissart, line 24: The same for this image path. Either move the file to "backend" or use a HTTP redirection towards a frontend URL (if the frontend should remain responsible for this default preview image).

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    def get(self, request, uid):
+        response = HttpResponse(content_type="image/png")
+        tournesol_footer = Image.new("RGBA", (320, 60), (255, 200, 0, 255))
+        tournesol_footer_draw = ImageDraw.Draw(tournesol_footer)
+        tournesol_footer_draw.text((10, 0), "by", font=POPPINS_FONT, fill=(29, 26, 20, 255))
+        tournesol_footer_draw.text((80, 0), "Tournesol", font=POPPINS_FONT, fill=(29, 26, 20, 255))
+
+        # TODO double check that this should be always 320x180.
+        url=f"https://img.youtube.com/vi/{uid}/mqdefault.jpg"
```
amatissart, line 41: It would be safer to validate this `uid` first, for example by checking that the entity exists in the database.
Otherwise, it could be abused to forge arbitrary requests sent by our server.

(and fetching the `Entity` might also be useful in the future to include other metadata into the image)

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,50 @@
...
+        youtube_thumbnail = Image.open(BytesIO(requests.get(url).content)).convert("RGBA")
```
amatissart, line 42: Don't we need to handle errors here? I suppose a non-200 response would lead to invalid image data being used in the response, or causing an exception on `.convert()` ?

backend/tournesol/urls.py
```diff
@@ -191,4 +192,15 @@
         PollsCriteriaScoreDistributionView.as_view(),
         name="polls_score_distribution",
     ),
+    # Website Previews
+    path(
+        "preview/entities/yt:<str:uid>",
+        DynamicWebsitePreviewEntity.as_view(),
+        name="website_preview_antity",
```
amatissart, line 199: typo
```suggestion
        name="website_preview_entity",
```

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,50 @@
...
+from rest_framework.views import APIView
+
+POPPINS_FONT = ImageFont.truetype("../frontend/public/fonts/Poppins-Bold.ttf", 40)
+
+
+class DynamicWebsitePreview(APIView):
+    permission_classes = []
+
+    @extend_schema(
+        description="preview the ",
```
amatissart, line 18: the ... :eyes:

backend/tournesol/urls.py
lfaucon: Done

backend/tournesol/views/preview.py
lfaucon: I looked into it, and the response often has non-200 code (e.g. 304 or 443) while still containing the right content. Chose to log a warning in these cases

backend/tournesol/views/preview.py
lfaucon: Fixed

backend/tournesol/views/preview.py
lfaucon: I made sure that we use the `video_id` of our entities. Do you think this needs more validation?

backend/tournesol/views/preview.py
lfaucon: Done

backend/tournesol/views/preview.py
lfaucon: Copied the necessary files to `backend/tournesol/resources/`

Whe ndeploying I found it worked for starting the server but failed during the stage "Migrate Django database" becasue the font was evaluated when importing the file. I guess this is because this step is done from another folder? Circumvented this for now by avoidnig to evaluate the font when the file is imported.

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,95 @@
...
+                tournesol_footer_draw.text(
+                    (10, 25), "Tournesol Score: %.0f" % score, font=fnt, fill=(29, 26, 20, 255)
+                )
+
+            url = f"https://img.youtube.com/vi/{entity.video_id}/mqdefault.jpg"
+            thumbnail_response = requests.get(url)
+            if thumbnail_response.status != 200:
+                logger.warning(
+                    f"Fetching youtube thumbnail has non-200 status: {thumbnail_response.status}"
+                )
```
amatissart, line 72: you need `status_code` instead of `status`
```suggestion
            if thumbnail_response.status_code != 200:
                logger.warning(
                    f"Fetching youtube thumbnail has non-200 status: {thumbnail_response.status_code}"
                )
```

backend/tournesol/views/preview.py
amatissart: To make it more robust you can use `BASE_DIR` (from django settings) which points to the "backend" content.
E.g:
```py
fnt = ImageFont.truetype(BASE_DIR / "tournesol/resources/Poppins-Medium.ttf", 20)
```

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,95 @@
...
+                logger.warning(
+                    f"Fetching youtube thumbnail has non-200 status: {thumbnail_response.status_code}"
+                )
+                # Choosing not to raise an error here because the reponse often has a non-200
+                # status while containing the right content (e.g. 304, 443)
+                # raise ConnectionError
+            youtube_thumbnail = Image.open(BytesIO(thumbnail_response.content)).convert("RGBA")
+
+            # Merges the two images into one
+            preview_image = Image.new("RGBA", (320, 240), (255, 255, 255, 0))
```
amatissart, line 79: 320x240 doesn't seem to be a very suitable format for Twitter cards:
image

The image is displayed with a fixed ratio (~ 1.9), and the Tournesol score is not visible.

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,95 @@
...
+    @extend_schema(
+        description="Default website preview",
+        responses={200: OpenApiTypes.BINARY},
+    )
+    def get(self, request):
+        return DynamicWebsitePreviewDefault.default_preview()
+
+    @staticmethod
+    def default_preview():
+        default_preview = open("tournesol/resources/tournesol_screenshot_og.png", "rb")
```
amatissart, line 30: as discussed in a separate thread, `BASE_DIR` should be used to build relative path:
```suggestion
        default_preview = open(BASE_DIR / "tournesol/resources/tournesol_screenshot_og.png", "rb")
```

backend/tournesol/views/preview.py
```diff
@@ -0,0 +1,95 @@
...
+    permission_classes = []
+
+    @method_decorator(cache_page_no_i18n(3600 * 2))  # 2h cache
+    @extend_schema(
+        description="Website preview for entities page",
+        responses={200: OpenApiTypes.BINARY},
+    )
+    def get(self, request, uid):
+        fnt = ImageFont.truetype("tournesol/resources/Poppins-Medium.ttf", 20)
+        fnt_title = ImageFont.truetype("tournesol/resources/Poppins-Medium.ttf", 14)
```
amatissart, line 45: same here
```suggestion
        fnt = ImageFont.truetype(BASE_DIR / "tournesol/resources/Poppins-Medium.ttf", 20)
        fnt_title = ImageFont.truetype(BASE_DIR / "tournesol/resources/Poppins-Medium.ttf", 14)
```

backend/tournesol/views/preview.py
GresilleSiffle: Fixed by the review branch https://github.com/tournesol-app/tournesol/pull/1092

backend/tournesol/views/preview.py
GresilleSiffle: Fixed by the review branch https://github.com/tournesol-app/tournesol/pull/1092

GresilleSiffle:
I created a new PR based on this branch that target this branch to avoid making mistakes by editing directly this one.

See: https://github.com/tournesol-app/tournesol/pull/1092

I'm not sure about the remaining work:
- should we add the Tournesol score in the yellow footer?
- should we add the `n_contributors` / `n_comparisons` in the footer too?

aidanjungo:
> * should we add the Tournesol score in the yellow footer?
> * should we add the `n_contributors` / `n_comparisons` in the footer too?

For me the answer is yes, yes and yes :)  and I think ideally it should look like what we have on YouTube with the extension to be coherent.

image

GresilleSiffle:
Alright thanks @aidanjungo . I think we can aim for a bigger Tournesol score on the left side, and stack the number of comparisons / contributors on the right side.

lfaucon:
Because we really care about Twitter Cards, I changed the aspect ratio of our preview. So I also changed the layout and made a bit of mess in the code overall.

Screenshot 2022-08-23 at 20 54 03

We had a bug when a video had no score. This probably should have been caught by unit tests (blaming myself here).

TODO
- [ ] Add unit tests
- [ ] Demo during wednesday stand-up
- [ ] Deploy to prod

GresilleSiffle:
**to-do**
- [ ] size of score and ratings are different (score should be a bit bigger)
- [ ] properly handle negative score

lfaucon:
Here is my latest proposal:
tournesol_preview_yt_bbKHudT7dVk

I'd insist that we pick the best out of what we have now and don't invest more time into styline the preview. Instead, if we feel the need, I'll try to reach out to association members who know graphism/design to help us!


My votes:
- I have a strong preference for the white background with yellow borders against everything else.
- I have a very slight preference for the yellow background agains the blue

GresilleSiffle (approved):
:ok_hand: 


Aesthetic and subjective opinion: I have a strong preference for the blue and yellow backgrounds over the white one. The later appears to me rather generic and doesn't remind me of Tournesol. But it doesn't prevent me to approve : ) let's see how we feel by using the feature.

Issue #1059, opened 2022-07-19
[back] API endpoint to add & manage vouches


lfaucon:
4 endpoints are needed
- Create vouches between the logged in user and a target username
- Logged in user can delete one of his/her vouches
- Logged in user can list their given vouches (no pagination needed)
- Logged in user can list received vouches (no pagination needed)

PR #1083, opened 2022-08-09
[back] feat: add the vouchers API

sigmike:
Related issue: #1059 

Remaining to do:
- [x] Logged in user can delete one of his/her vouches
- [x] Logged in user can list their given vouches (no pagination needed)
- [x] Logged in user can list received vouches (no pagination needed)


backend/tournesol/serializers/voucher.py
```diff
@@ -0,0 +1,42 @@
...
+            "is_public",
+            "value",
+        ]
+
+        validators = [
+            UniqueTogetherValidator(
+                queryset=Voucher.objects.all(),
+                fields=["to", "by"],
+            )
+        ]
```
amatissart, line 29: To apply this constraint at the database level, it can be defined in the `Vouch` db model directly:
```py
class Voucher(models.Model):
    class Meta:
        unique_together = ["by", "to"]
```
similarly to what exists in the `RateLater` model for example: https://github.com/tournesol-app/tournesol/blob/5c0ce9b1f64375e686d36b85f4e4b4221631b237/backend/tournesol/models/rate_later.py#L38-L40

(a new [generated migration](https://docs.djangoproject.com/en/4.0/ref/django-admin/#makemigrations) will be required to apply the constraint in the database)

DRF `ModelSerializer` will be able to read the model definition and apply this constraint on validation too, so the `validators` list here can be removed.

sigmike: Ok indeed it works.

When I failed to make it return a proper error it was with this code:

```python
class Voucher(models.Model):
    class Meta:
        constraints = [ models.UniqueConstraint(fields=['by', 'to'], name="unique-voucher-by-to") ]
        indexes = [ models.Index(fields=['by', 'to']) ]
```

I guess in that case the `ModelSerializer` is not able to handle the constraint.

sigmike:
It should be complete now. A few notes:

- I used the routes `/vouchers/given/` and `/vouchers/received/` which may conflict with the delete route `/vouchers/{id}/` if the `id` can be any string, but I assumed no `id` will ever have the `given` or `received` value.
- The received vouchers list returns all the received vouchers. You may want to remove private vouchers here, but I think we said the public flag won't be used for now.
- I used 2 distinct serializers for the given and received vouchers because it seemed difficult to handle automatic `by` field while serializing it as a string (username). I think it works well with 2 serializers anyway.
- I didn't define `permission_classes` because we want the default one (`IsAuthenticated`) to apply. But since it's explicitly set in some other views you may want to define it here too. There are tests that will fail if the default is changed.

GresilleSiffle:
Hello

I'm going to review this PR.

**to-do**
- [x] rebase on `main`
- [ ] move the vouchers API tests to the vouch app

GresilleSiffle:
The current state of the PR is good for me.

There are still two limitations we should be aware of:

**(1)** the JSON representation of a given voucher is different than the JSON representation of a received voucher. The former has the field `to` and the later the field `by`. We can ease the front end implementation of a voucher component by always including the fields `to` and `by`, thus returning the same JSON for given and received vouchers.

This can be done in a second PR.

**(2)** the API doesn't provide an endpoint to deleted a received voucher, only to delete a given voucher.

GresilleSiffle (approved):

sigmike:
@GresilleSiffle 

> we could indeed remove all explicit occurrences of IsAuthenticated if it's the default permission of DRF.

I don't know if it's the default of DRF, but it's the default on Tournesol: https://github.com/tournesol-app/tournesol/blob/main/backend/settings/settings.py#L256

backend/vouch/tests/test_api_voucher.py
```diff
@@ -22,27 +24,24 @@ def setUp(self):
             "value": "1",
         }
 
-    def test_anonymous_cant_create(self):
+    def test_anon_401_create(self):
```
sigmike, line 28: @GresilleSiffle I fail to see how these method names are better. They look much more cryptic to me.

backend/vouch/tests/test_api_voucher.py
```diff
@@ -144,13 +144,24 @@ def test_auth_204_delete(self):
...
+        An authenticated user cannot delete a voucher given by a non-existing
+        user.
+        """
+        self.client.force_authenticate(user=self.user2)
+        response = self.client.delete(f"{self.voucher_base_url}unknown/", format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+        Voucher.objects.get(pk=self.voucher.pk)
+
     def test_auth_404_delete_someone_else_voucher(self):
```
sigmike, line 165: @GresilleSiffle this test probably doesn't make sense anymore (and it still uses the `pk`).

GresilleSiffle: Good catch @sigmike . I'm going to fix it in ithttps://github.com/tournesol-app/tournesol/pull/1091

backend/vouch/tests/test_api_voucher.py
GresilleSiffle: Yes it's very subjective indeed.

I sometimes do some experimentations with names to see if I'm able to navigate easily in the tests.

The original name `test_anonymous_cant_create` is nearly perfect, and very human readable. I started to use `anonymous` and `authenticated` few months ago to make the target of the test explicit.

But quickly these names became a bit long. I started then to replace `anonymous` by `anon`, `authenticated` by `auth` (sometimes `users` when the test is interacting with both). It saved few characters.

Then some other parts of the name caused me troubles. Sometimes we write `cant`, sometimes `must_not`, sometimes `should_not`, and it makes tests with the same intention look different.

To continue experimenting I finally decided to use the HTTP code directly in the test name, to avoid having to interpret "should" and "cant".

 Now I find these names more concise, and they allow me to quickly see in a test case which tests are missing. It's true that `test_anon_401_create` is not that explicit at the first glance, but unless it causes more problems that it solves I'd like to propose us to continue with it, and to see in few weeks how we feel with these names.

Issue #606, opened 2022-02-18
[back, front] refactor: Harmonize all API and Serializers (then:feat: Rate later list display information about the number of comparisons)


lfaucon:
As a user I want to see how many times I have compared the videos when I look at my rate later list so that I can easily chose which videos to remove from my rate later list.

Display the information similarly as in the card used on the comparison page.

This probably will be best implemented by updating the backend endpoint to include the extra info

GresilleSiffle:
Here is the JSON the API could return.

The new `poll_metadata` key contains all data related to the poll, like the total number of comparisons by all contributors.

The `user_metadata` key contains all data related to the logged user. Having different keys for collective and and personal data will allow the front end to: (1) know what kind of data are displayed ; (2) display both collective and personal data at the same time.

```json
{
  "entity": {
    "uid": "string",
    "type": "video",
    "metadata": {
      "additionalProp1": "string",
      "additionalProp2": "string",
      "additionalProp3": "string"
    }
  },
  "poll_metadata": {
    "n_comparisons": "number",
    "n_contributors": "number"
  },
  "user_metadata": {
    "n_comparisons": "number",
    "n_contributors": "number"
  }
}
```

lfaucon:
User metadata is also poll specific. The distinction is rather at the level or global/community/aggregated vs. user/personal ? Not poll vs. user.

lfaucon:
Apart from this, I really like the proposal to simplify the keys at the root of the entity JSON representations and allowing API routes to include the root entries or not depending on what is requested. 

I would even go further and impose fixed types for each of the root entries. For example by having one entry `global_metadata` and one `global_criteria_scores` to avoid changing the type of the global_metadata entry

lfaucon:
You have `n_contributors` in entry `user_metadata`. Is it a typo or something that I don't understand correctly?

amatissart:
Thanks @GresilleSiffle for the details, I think it's a step in the good direction.

> User metadata is also poll specific. The distinction is rather at the level or global/community/aggregated vs. user/personal ? Not poll vs. user.

I agree. "poll_metadata" and "user_metadata" could bring more confusion. For example, a "tournesol_score" (contrary to "name") is not a property of a `Poll`, it's part of an `EntityPollRating`. And keeping this distinction clear will also make the implementation of the serializers easier.

[...]

PR #1088, opened 2022-08-22
[ext] refactor: use the generic rate later API

GresilleSiffle:
**indirectly related to** #606

---

This PR makes the extension use the new generic rate later API instead of the deprecated one.


lfaucon (approved):
Works for me 👍

PR #1090, opened 2022-08-23
[front] fix: update OpenAPI schema

GresilleSiffle:
Could be merged after https://github.com/tournesol-app/tournesol/pull/1091 to include the new changes.

This PR adds the last API updates in the front end's schema definition.

As the old `RateLater` type is now `RateLaterLegacy` I updated the front end accordingly.


lfaucon (approved):
Works for me

Issue #1089, opened 2022-08-23
[back] fix: make JSON repr of given and received vouchers the same


GresilleSiffle:
See https://github.com/tournesol-app/tournesol/pull/1083#issuecomment-1222381050

The goal is to have in the front end the same schema for `GivenVoucher` and `ReceivedVoucher`. To see the schema, use the command `yarn update-schema` when the back end is running.

PR #1091, opened 2022-08-23
[back] fix: given and received vouchers have now the same JSON

GresilleSiffle:
**related to** #1089

### DIFF
--- a/backend/vouch/serializers.py
+++ b/backend/vouch/serializers.py
@@ -56,8 +56,8 @@ class ReadOnlyVoucherSerializer(ModelSerializer):
     Prefer `PublicReadOnlyVoucherSerializer` instead.
     """
 
-    by = serializers.CharField()
-    to = serializers.CharField()
+    by = serializers.CharField(read_only=True)
+    to = serializers.CharField(read_only=True)
 
     class Meta:
         model = Voucher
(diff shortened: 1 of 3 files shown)


backend/vouch/serializers.py
```diff
@@ -8,19 +7,27 @@
...
         model = Voucher
-
-        fields = [
-            "by",
-            "to",
-            "is_public",
-            "value",
-        ]
+        fields = VOUCHER_FIELDS
+        read_only_fields = ["by"]
```
amatissart, line 30: This `read_only_fields` has no effect, as `by` is defined explicitly as a (writable)  `CharField`.  
In particular, this is visible in the API docs: "by" is writable (even if any non-empty value would trigger a server error as a User object is expected to create the Vouch instance).

An alternative solution, with 2 distinct fields for read and write, could make the API schema cleaner:
```py
class GivenVoucherSerializer(ModelSerializer):
    _by = serializers.HiddenField(
        source="by", default=serializers.CurrentUserDefault(), write_only=True
    )
    by = serializers.CharField(read_only=True)
    to = serializers.CharField()

    class Meta:
        model = Voucher
        fields = VOUCHER_FIELDS + ["_by"]
```

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1091/commits/db441b13c030fee368bb4f8d9d784619376e2918.

amatissart (approved):

PR #1092, opened 2022-08-23
review: dynamic entity preview

GresilleSiffle:
**target branch** `dynamic_website_previews`

---

- [x] reworked the `DynamicWebsitePreviewEntity` get method to use the return early pattern
- [x] use `BASE_DIR` when opening static files
- [x] add `n_comparisons` and `n_contributors`
- [ ] add the Tournesol logo instead of string TS

### preview

capture


backend/tournesol/views/preview.py
```diff
@@ -8,12 +8,64 @@
...
+            f"{entity.rating_n_ratings} comparisons",
+            font=fnt_ratings,
+            fill=(29, 26, 20, 255),
+        )
+        tournesol_footer_draw.text(
+            ENTITY_N_CONTRIBUTORS_YX,
+            f"{entity.rating_n_contributors} contributors",
+            font=fnt_ratings,
+            fill=(29, 26, 20, 255),
+        )
```
lfaucon, line 65: We could split the number and text such that it doesn't get misaligned when the numbers don't have the same digits. This could be done with a right anchor on the count and left anchor on the word 

https://pillow.readthedocs.io/en/stable/handbook/text-anchors.html

lfaucon (approved):

PR #1094, opened 2022-08-25
[infra] sec: back end's admin is now accessible only from localhost

GresilleSiffle:
This PR aims to increase the security of our back end by limiting the access to its administration interface to HTTP requests coming from its own host.

Users having an account on the server can access the interface by using an SSH tunnel and a browser extension as described in the `README.md`.

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -98,6 +98,45 @@ You can do the same with another VM or a different username:
 source ./ansible/scripts/get-vm-secrets.sh "tournesol-vm" "jst"
 ```
 
+## Accessing the back end admin interface
+
+To protect the back end's admin interface, its access is restricted to
+HTTP requests coming from its own host.
+
+Users having an account on the server can access this interface by using an
+SSH tunnel.
+
+### Example: accessing the staging admin
+
+First, edit your own `/etc/hosts` or equivalent. This is required to make our
+requests reach the correct NGINX virtual host, and not the default one.
+
+```
+127.0.0.1    api.staging.tournesol.app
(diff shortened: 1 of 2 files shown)


infra/README.md
```diff
@@ -98,6 +98,45 @@ You can do the same with another VM or a different username:
...
+Users having an account on the server can access this interface by using an
+SSH tunnel.
+
+### Example: accessing the staging admin
+
+First, edit your own `/etc/hosts` or equivalent. This is required to make our
+requests reach the correct NGINX virtual host, and not the default one.
+
+```
+127.0.0.1    api.staging.tournesol.app
```
GresilleSiffle, line 115: This doesn't look sustainable on the long term, as resolving the back end to localhost will prevent us to use the front end...

jstainer-kleis: various solutions to these issues posted to #infrastructure Discord channel

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1094/commits/2e247199a4f96f22c359039c02ae98a8dc3a410f

GresilleSiffle:
**to-do** do a demo with the core team.

amatissart:
One thing to keep in mind: the admin url is currently used by the "external-urls-monitoring" service. Another public URL should be used to check that the API is healthy. For example https://api.tournesol.app/polls/videos/

GresilleSiffle:
We decided to use two factor authentication on the Django Admin / Grafana instead.

Issue #1087, opened 2022-08-19
[twitterbot] feat: Share Tournsol link instead of YouTube link


aidanjungo:
When #1080 will be merged and that we are happy with the preview on Twitter, we could replace the YouTube link shared by the TwitterBot by the Tournesol link pointing to the video.

PR #1097, opened 2022-08-29
[twitterbot] refactor: tweet Tournesol link instead of YT link

aidanjungo:
#1087

### DIFF
--- a/backend/twitterbot/tests/test_tournesolbot.py
+++ b/backend/twitterbot/tests/test_tournesolbot.py
@@ -117,17 +117,29 @@ class TestTournesolBot(TestCase):
         VideoCriteriaScoreFactory(
             entity=self.videos[8], criteria="largely_recommended", score=0.35
         )
-        VideoCriteriaScoreFactory(entity=self.videos[8], criteria="reliability", score=32)
-        VideoCriteriaScoreFactory(entity=self.videos[8], criteria="importance", score=52)
+        VideoCriteriaScoreFactory(
+            entity=self.videos[8], criteria="reliability", score=32
+        )
+        VideoCriteriaScoreFactory(
+            entity=self.videos[8], criteria="importance", score=52
+        )
         VideoCriteriaScoreFactory(entity=self.videos[8], criteria="engaging", score=38)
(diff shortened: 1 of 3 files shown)


backend/twitterbot/tests/test_tournesolbot.py
```diff
@@ -156,41 +168,41 @@ def test_prepare_tweet(self, mock_get_twitter_account_from_video_id):
...
-        )
+        self.videos[8].metadata[
+            "name"
+        ] = "Tournesol is great! But this title is way to long to fit in one tweet🌻 #tournesol"
 
         tweet_text_too_long = (
-            "Aujourd'hui, je recommande 'Tournesol is great! But this title is way to long to fit "
-            "in one tweet...' de @TournesolApp, comparée 77 fois sur #Tournesol🌻 par 28 "
+            "Aujourd'hui, je recommande 'Tournesol is great! But this title is way to long to...' "
+            "de @TournesolApp, comparée 77 fois sur #Tournesol🌻 par 28 "
```
amatissart, line 185: @aidanjungo Side note: this tweet text is currently limited deliberately to 274 chars. In practice, Twitter transforms URLs into shortened "t.co" links, and as a result the [length of a URL in a Tweet is always counted as 23 characters](https://developer.twitter.com/en/docs/counting-characters). 

So after the change introduced in this PR, there are ~ 15 additional characters that could be used safely to include longer titles.

amatissart (approved):

backend/twitterbot/tests/test_tournesolbot.py
aidanjungo: I'm not sure I understand we we would get the 15 additional characters, the YouTube link was not treated the same way by Twitter?

amatissart: It was treated the same way.   
But the previous link (`youtu.be/AAAAAAAAAAA`) was 20 chars long, and was counted as 23.
The new one (`tournesol.app/entities/yt:AAAAAAAAAAA`) is 37 chars long, and is now "shortened" to 23 chars. So with a naive truncation, some characters are lost whereas they could still be added safely.

GresilleSiffle (approved):

PR #1098, opened 2022-08-30
[front] refactor: use the new rate-later API

GresilleSiffle:
This PR makes the front end use the new generic rate-later API instead of the legacy one tightly coupled with videos.

(It doesn't make the `RateLater` page generic, directly usable for any poll, as it is out of the scope of this PR, it simply swaps the API)

**remaining to-do**
- [x] refactor the `RateLater` page
- [x] refactor `getVideoFromRateLaterListForComparison`
- [x] refactor `AddToRateLaterList`


lfaucon (approved):
LGTM

PR #1114, opened 2022-09-07
[back] refactor: delete the legacy rate-later API

GresilleSiffle:
**indirectly related to** #606 

---

Now that the browser extension has been published for Firefox and Chrome, it should be safe to delete the legacy and now unused video rate-later API.

The path is now open to finally start the feature mentioned in #606 :
- make the back end send information about the number of comparisons (the new JSON payload must be carefully designed to represent "a potentially compared entity viewed as an authenticated user")
- display this number in the front end

**to-do**
- [x] remove the last occurrences of `UsersService.usersMeVideoRateLaterList` in the front end


amatissart (approved):
In production logs I can see that the legacy API was still used 2 days ago (2022-09-12) to add videos to the rate-later list. Probably because the Firefox extension was not updated on all devices yet. So I would suggest to wait for a couple more days before deploying this change.

Otherwise, looks good to me :+1:

GresilleSiffle:
I confirm that the route has been used today too.

Issue #631, opened 2022-02-27
Create model to store score and stats related to (entity, poll)


amatissart:
Model `Entity` currently contains fields `rating_n_contributors`, `rating_n_ratings`, as well as `tournesol_score` introduced in #614.

These fields are actually specific to a given poll, and should be moved to a new model, with foreign keys to `Entity` and `Poll`.
Potential name: `EntityPollScore`

lfaucon:
My understanding of this issue is that it is not high priority because we only need this when we will have multiple Polls using the same entities. Please correct me if I understand this wrongly.

GresilleSiffle:
- [x] create the new model
- [ ] test the new model
- [ ] create the data migration / update the comparisons view to create or refresh the model after each comparison / update the serializers / update the API tests

lfaucon:
@bbosson can look at the PR #1149

PR #1115, opened 2022-09-08
[back] tests: test the model `EntityPollRating`

Mayeul-T:
**related to** #631 

---


GresilleSiffle (changes requested):
**to-do**

- [ ] rename the TestCase (it currently mention `RateLater`)
- [ ] make it clear that the first method is testing `n_comparisons`

Issue #1058, opened 2022-07-19
[front] UI to manage vouches


lfaucon:
- On a standalone vouching page accessible from top-right menu
- Should show vouches given and received (usernames displayed in [Chip](https://mui.com/material-ui/react-chip/#main-content)). For now, trust value is always 1.
- A text input allows to paste trusted friends's username for vouching them
- Given vouches can be deleted

lfaucon:
Quoting lenhoang:

> What does it mean exactly X vouches for Y? Does it mean that X believes that Y is a human and has an authentic behaviour on theplatform?
For now, I'd say yes, and this is the only thing one can be vouching for. Moreover X must believe that Y has a single active account on the platform. This should be clarified when proposing to vouch "I vouch that Y is a human, with an authentic behavior, and that Y does not actively use another account on Tournesol".

PR #1116, opened 2022-09-09
[front] feat: add the vouching UI

sigmike:
Related issue: #1058 

image

Remaining to do:

- [x] Fix the unrelated failing test (probably because of a package upgrade?)
- [x] Ability to delete a given voucher
- [ ] Change the text of the already existing voucher error
- [ ] Remove the `value` and `is_public` attributes in the voucher API
- [ ] Maybe refactor the voucher list components

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -459,6 +459,12 @@
     "proofOfVoteHelperText": "This code will be helpful to complete <2>our survey.</2>",
     "proofOfVote": "Proof of vote"
   },
+  "personalVouchers": {
+    "voucherCreated": "You successfuly vouched for \"{{username}}\"",
+    "introduction": "I vouch that this user is a human, with an authentic behavior, and that this user does not actively use another account on Tournesol:",
+    "usernameLabel": "Username",
+    "submitButton": "I vouch for this user"
+  },
   "myRateLaterListPage": {
     "title": "My rate-later list"
   },
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -463,7 +463,9 @@
     "voucherCreated": "You successfuly vouched for \"{{username}}\"",
     "introduction": "I vouch that this user is a human, with an authentic behavior, and that this user does not actively use another account on Tournesol:",
     "usernameLabel": "Username",
-    "submitButton": "I vouch for this user"
+    "submitButton": "I vouch for this user",
+    "noVoucherGiven": "None.",
+    "givenVouchersTitle": "Users you have vouched for"
   },
   "myRateLaterListPage": {
     "title": "My rate-later list"
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -150,7 +150,8 @@
   "loginButton": "Log in",
   "joinUsButton": "Join us",
   "personalMenu": {
-    "yourRecommendations": "My recommendations"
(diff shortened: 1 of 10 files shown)


GresilleSiffle:
### review

#### changes recommended before merging

**ux / ui **



#### minor concerns

#### future considerations

**ux / ui **

The new menu item `Vouch for other users` is self explanatory. As the related page allows to do several things in addition to vouch for someone, like see my given vouchers, see my received vouchers, it may be considered as the main page of the vouching system. How about a title like `Vouching system` / `Système de garant` ? It might be more friendly to keep `Vouch for other users`.

GresilleSiffle:
Thanks a lot @sigmike for this nice work :+1: 

As you said, there still a bit of work to do in the back end, we can do it after this PR:
- make the `value` always equal to 1 in the API
- do not expose the `is_public` in the API?
- transform the 400 Bad Request when vouching two times for the same user into a 409 Conflict to allow the front end to overload the original massage for a more human friendly one

One thing that prevented me to hit the accept button, is the result of `yarn run`, we can see some error logs in this job:
- https://github.com/tournesol-app/tournesol/actions/runs/3061192422/jobs/4940653911

Once this has been addressed we can merge.

```
      Warning: An update to PersonalVouchersProvider inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at PersonalVouchersProvider (/home/romain/_dev/pub.sources/tournesol/frontend/src/pages/personal/vouchers/context.tsx:27:3)
          at PersonalVouchersPage
          at SnackbarProvider (/home/romain/_dev/pub.sources/tournesol/frontend/node_modules/notistack/src/SnackbarProvider.tsx:30:9)
          at Component

      64 |     const loadGivenVouchers = async () => {
      65 |       const vouchers = await UsersService.usersMeVouchersGivenList();
    > 66 |       setValue((value) => ({ ...value, givenVouchers: vouchers }));
         |       ^
      67 |     };
      68 |     const loadReceivedVouchers = async () => {
      69 |       const vouchers = await UsersService.usersMeVouchersReceivedList();

      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:67:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:43:5)
      at warnIfNotCurrentlyActingUpdatesInDEV (node_modules/react-dom/cjs/react-dom.development.js:24064:9)
      at setValue (node_modules/react-dom/cjs/react-dom.development.js:16135:9)
      at loadGivenVouchers (src/pages/personal/vouchers/context.tsx:66:7)
```

GresilleSiffle:
Fixed: https://github.com/tournesol-app/tournesol/pull/1116/commits/7106a23c268f517f46544137282835660ff5890f

GresilleSiffle (approved):

sigmike:
@GresilleSiffle 

> Should we keep console.error? The errors should already be displayed in the UI.

I added the `console.error` to avoid unknown errors being swallowed.

All errors are caught in this process, not only API errors. I did that because we want to warn the user that something went wrong, whatever the cause. But the error notifier is (rightfully) not very verbose on unknown errors, so if something weird happens we developers can't know what really happened and we have to add the `console.error` to debug.

Maybe it's better to make the notifier do this `console.error` though.

Issue #1124, opened 2022-09-15
[back] feat: do not expose unused voucher fields


GresilleSiffle:
If we don't want users to change the public or private state of their vouchers, we can remove if from the API.

If we don't want users to change the value of their vouchers, we can remove the `value` field from the API, and make it equals to 1 by default. 

If we want to front end to be able to display friendly message when a user vouches two times for the same person, we can change the 400 Bad Request to a 409 Conflict. Read more here: https://github.com/tournesol-app/tournesol/pull/1116

PR #1125, opened 2022-09-16
[back] fix: improve error message on duplicate voucher

sigmike:
Related: https://github.com/tournesol-app/tournesol/pull/1116#issuecomment-1247718705 and https://github.com/tournesol-app/tournesol/issues/1124.

I was able to add a custom message without any change in the frontend. I think it's better that way.

I failed to generate the French translation though. On https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#localization-how-to-create-language-files I read I had to run `django-admin makemessages -l fr` but in dev-env I get this error:
```
CommandError: Unable to find a locale path to store translations for file manage.py. Make sure the 'locale' directory exists in an app or LOCALE_PATHS setting is set.
```

Remaining to do:

- [x] Add French translation

Please don't squash the commits since I added a unrelated very small improvements in tests in another commit.


amatissart:
> I failed to generate the French translation though. On https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#localization-how-to-create-language-files I read I had to run `django-admin makemessages -l fr` but in dev-env I get this error:
> 
> ```
> CommandError: Unable to find a locale path to store translations for file manage.py. Make sure the 'locale' directory exists in an app or LOCALE_PATHS setting is set.
> ```

`django-admin` does not load the settings specific to the current project, and will try to write new translation keys into a "locale" directory at the root of the project. It turns out that the default path for "locale" defined in the current settings is "core/locale":
https://github.com/tournesol-app/tournesol/blob/a2399e7c7401423192cae5e17571bd5e3e5163b8/backend/settings/settings.py#L160-L162

`python manage.py makemessages -l fr` will use these settings and execute correctly.  
(In practice, it seems that `python manage.py` can always be preferred to `django-admin`, which is only necessary when initializing a new project).

amatissart (approved):

PR #1127, opened 2022-09-16
[back][front] feat: do not expose unused voucher fields + clarify vouchers are public

### LINKED ISSUE #1124: [back] feat: do not expose unused voucher fields

amatissart:
Fixes #1124 

An additional message is at the top of the page to describe the vouching mechanism briefly, and inform the users that vouchers will be public. Maybe it takes too much space, and should be collapsible?

image


frontend/public/locales/fr/translation.json
```diff
@@ -475,7 +475,8 @@
     "noVoucherGiven": "Aucun.",
     "givenVouchersTitle": "Utilisateurs pour lesquels vous vous portez garant",
     "noVoucherReceived": "Aucun.",
-    "receivedVouchersTitle": "Utilisateurs qui se portent garant pour vous"
+    "receivedVouchersTitle": "Utilisateurs qui se portent garant pour vous",
+    "aboutVouchingMechanism": "En vous portant garant pour d'autres utilisateurs, vous leur permettez d'acquérir un droit de vote plus important sur Tournesol, y compris si leur adresse e-mail n'est pas considérée comme fiable. Sur Tournesol, la liste des utilisateurs pour lequel vous vous portez garant, ou qui se sont portés garant pour vous, est une information publique."
```
GresilleSiffle, line 479: > la liste des utilisateurs pour lequel vous vous portez garant, ou qui se sont portés garant pour vous

can be replaced by

> la liste des utilisateurs pour lequel vous vous portez garant, ou qui se portent garant pour vous

Mainly because we do not keep the history of vouching, so in fact we do not know who vouched for you, but rather who is currently vouching for you.

GresilleSiffle (approved):
Thank you @amatissart :ok_hand: 

Approved with a small comment on a french translation.

GresilleSiffle:
I've just rebased the branch to resolve the conflict. I'm going to merge soon.

Issue #1117, opened 2022-09-12
[back] feat: add the FAQ API


GresilleSiffle:
**sub-task of** #1067

---

The endpoint `GET /faq/`
- must return all questions
  - except those disabled
  - except those without answer
- must return the translations matching the HTTP Accept-Language header of the request

PR #1130, opened 2022-09-19
[back] feat: add the FAQ API

GresilleSiffle:
**related to** #1117 

**replace the PR** #1118

---

This PR is the second step of the FAQ API feature. The first step the creation of the models and can / should be reviewed first here: #1105

**to-do**
- [x] the API must not return questions without empty answer
- [x] add API tests
- [x] factorize common behaviours of `faq.models`
- [x] update the API schema in the front end

### preview

The front end will receive an ordered list of translated questions / answers. Disabled questions in the back end won't appear. 

```json
{
  "count": 3,
  "next": null,
  "previous": null,
  "results": [
    {
      "name": "i_dont_understand",
      "question": "Je n'ai pas compris le but du projet.",
      "answer": "Sauver le monde enfin, c'est pourtant simple."
    },
    {
      "name": "how_can_i_help",
      "question": "Comment puis-je aider ?",
      "answer": "Bla bli blou bla."
    },
    {
      "name": "i_dont_know_what_to_compare",
      "question": "Je ne sais pas quoi comparer",
      "answer": "Vous pouvez comparer ceci ou cela, ou ceci, mais aussi ça là bas, oui ça."
    }
  ]
}
```


backend/faq/models.py
```diff
@@ -28,6 +30,55 @@ class Meta:
...
+        Contrary to `self.get_text` this method consider the related instances
+        as already prefetched with `prefetch_related`, and use `.all()` instead
+        of `.get()` to avoid triggering any additional SQL query.
+        """
+        if lang is None:
+            lang = translation.get_language()
+
+        try:
+            locale = [loc for loc in getattr(self, related).all()
+                      if loc.language == lang][0]
```
lfaucon, line 63: Would this work?

```
locale = getattr(self, related).get(language=lang)
```

backend/faq/models.py
```diff
@@ -28,6 +30,55 @@ class Meta:
...
+                locale = [loc for loc in getattr(self, related).all()
+                          if loc.language == "en"][0]
+
+            except IndexError:
+                return self.name  # pylint: disable=no-member
+        return locale.text
+
+    def get_question_text_prefetch(self, lang=None):
+        """Return the translated text of the question."""
+        return self.get_text_prefetch(related="locales", lang=lang)
```
lfaucon, line 76: Reading this asymmetry where questions are called "locales" and answers are called "answers" makes me think that we may rename the main model "FAQEntry" which contains locales for questions and answers (very minor)

backend/faq/models.py
lfaucon: I guess it would raise an error when there are several, but the unique together property is set, so it should always work?

lfaucon (approved):
LGTM

backend/faq/models.py
GresilleSiffle: Good idea.

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1130/commits/f90ddf65b20a5648e84582c27f012f2df5b3c3bb

backend/faq/models.py
GresilleSiffle: Yes using `.get` will work but will trigger an extra SQL query. Using `.all` allows to take advantage of an eventual call to `.prefetch_related`, which is what we use in the view.

PR #1131, opened 2022-09-20
Update outdated voucher mocks in tests

sigmike:
And enable type check on the whole file so that in the future mocks still match the type they are replacing.

I think having an automatic check that the mocks match the type they are simulating is worth the slightly harder to read tests because of all the type annotations.


amatissart (approved):
Thanks for looking into this and making the unit tests more consistent with the latest API changes.

PR #1138, opened 2022-09-21
[back] fix: prevent a `Voucher` to be saved with the same `by` and `to`

GresilleSiffle:
It should not be possible for a user to vouch for him/herself.

The `clean()` method will ensure saving a model through the administration interface, of by code, will raise a proper `ValidationError` with a translated message.


amatissart (approved):
Great!

Side note for the future: after upgrading to Django 4.1, a similar behavior could be implemented using `Meta.constraints` in the model. With [the latest release](https://docs.djangoproject.com/en/4.1/releases/4.1/#validation-of-constraints) these constraints are checked during model validation (including on save from the admin forms). An additional benefit of this approach: the constraints are also enforced at the database level.

PR #1149, opened 2022-09-29
[back] refactor: sync `EntityPollRating` with the API and the ML train 

Mayeul-T:
**related** to #631

---

This PR synchronizes the `EntityPollRating`s with the comparison API and the ML train.

After each comparison, the `EntityPollRating`s of both videos are automatically updated.

During the ML train, the tournesol score is also saved in the `EntityPollRating`s.

The legacy fields of the Entity model `tournesol_score`, `rating_n_ratings`, `rating_n_contributors` are still used for now, and will be deleted in the next PR at the same time the serializers will be updated.


backend/tournesol/migrations/0051_migrate_ratings_to_its_own_model.py
```diff
@@ -0,0 +1,37 @@
...
+def create_entity_poll_rating(apps, poll, entity):
+    entity_pr_model = apps.get_model("tournesol", "EntityPollRating")
+    # `get_or_create` allows to run the migration back and forth
+    entity_pr_model.objects.get_or_create(
+        poll=poll,
+        entity=entity,
+        tournesol_score=entity.tournesol_score,
+        n_comparisons=entity.rating_n_ratings,
+        n_contributors=entity.rating_n_contributors,
+    )
```
amatissart, line 29: I think that the "get" check should use "poll" + "entity" only, just in case instances exist with outdated values.  
(That should not matter in production, where the table is empty)

```suggestion
    entity_pr_model.objects.get_or_create(
        poll=poll,
        entity=entity,
        defaults={
            "tournesol_score": entity.tournesol_score,
            "n_comparisons": entity.rating_n_ratings,
            "n_contributors": entity.rating_n_contributors,
        }
    )
```

backend/ml/outputs.py
```diff
@@ -74,14 +78,24 @@ def save_tournesol_scores(poll):
...
-            entity.tournesol_score = 10 * sum(
+            tournesol_score = 10 * sum(
                 criterion.score for criterion in entity.criteria_scores
             )
+
+        entity.tournesol_score = tournesol_score
         entities.append(entity)
 
+        if entity.single_poll_ratings:
+            entity_poll_rating = entity.single_poll_ratings[0]
```
amatissart, line 89: Note sure that's helpful, but we could use `.get()` to ensure that a single item is prefetched:
```suggestion
            entity_poll_rating = entity.single_poll_ratings.get()
```

amatissart (commented):
LGTM, except a minor comment in the migration

backend/ml/outputs.py
GresilleSiffle: At this point in the code `single_poll_ratings` is a Python list. Calling `.get()` will not work here.

amatissart (approved):

Issue #1071, opened 2022-07-24
feat: Include voting rights in public dataset


lfaucon:
To be done only when sufficiently advanced on the Vouching feature

TODO:
- Make sure it is clear to users that when they make public comparisons, an additional information will be added "voting right"
- The public dataset should be an archive containing the current csv file and addtional file with users & voting rights (possibly in the future more user information would be in this file too)
- The archive should also contain a short doc/readme file describing the content of the multiple files

lfaucon:
Where to display that voting rights are made public:
- On privacy policy: In paragraph "Donnée publiques", append "Si vous avez au moins une comparaison de videos publiques, Tournesol inclus dans le dataset public votre droit de vote calculé en fonction de votre email et des garants" 
- Display on the vouching page next to the voting right score

Readme doc:
- Includes "Voting right: The users without any trusted email or trusted vouches (voting rights of 0) can influence together at most 10% of the total collaborative evaluation of a video"

GresilleSiffle:
I move again this ticket to "in progress" to not forget to update the privacy policy.

sigmike:
This change in the privacy policy is being done in https://github.com/tournesol-app/tournesol/pull/1132.

sigmike:
The PR has been merged so I guess this issue can be closed.

There's still no mention of the trust score on the comparison page though, but I think it was decided to only mention it in the privacy policy. Maybe we could add a link to the privacy policy on the text "After submission, this comparison will be included in the public data"?

lfaucon:
As we were confused between trust scores and voting rights, I think that we have so far only included trust scores, but we would like to also add voting rights. We want both trust and voting rights anyway so the work done so far is very valuable. Thanks!

The voting right that we want to add are now computed and saved on the ContributorRatingCriteriaScore (since 31b3ac5a, merged yesterday). Could we iterate on the same steps as for including the trust score in the exports?

PR #1154, opened 2022-10-04
WIP: Include voting rights in public dataset

sigmike:
Related issue:  #1071.

I added a new export endpoint to get the public comparisons and the user voting rights in a zip file.

Remaining to do:
- [ ] Add to the zip file "a short doc/readme file describing the content of the multiple files"
- [ ] Make sure it is clear to users that when they make public comparisons, an additional information will be added "voting right"
- [ ] Update the link on the frontend to make it use the new endpoint
- [ ] Remove the API endpoint that downloads only the comparisons?

Please don't squash the commits.


GresilleSiffle:
I'm going to write here few notes.

Sometimes we use the terms public database, sometimes public dataset, maybe this PR is the good opportunity the use the most appropriate terms in the `/export/` APIs descriptions (and later in the front end). We can take care of this just before merging.

I didn't read the code yet, but I saw things we can improve right now:

**(a)** The name of the `.zip` file is `tournesol_export.zip`, I'd suggest this instead:
```
tournesol_datasets_{yyyy}_{mm}_{dd}_{hh}h{mm}.zip
```

We can also name the inner folder with the same name without the extension. With this, we can easily make the distinction between two exports and how fresh the data are.

**(b)** I'd suggest to slightly change some texts in the README.txt


Like the introduction:

```
This is the public database of Tournesol (https://tournesol.app/).
```

to:

```
This archive contains the public datasets of the Tournesol project.

All files are available under the terms of the Open Data Commons Attribution License (ODC-By) v1.0.
See: https://opendatacommons.org/licenses/by/odc_by_1.0_public_text.txt
```

Also this:
```
This file contains all the comparisons of public videos made by Tournesol's
users.
```

to:

```
This file contains all the public video comparisons made by Tournesol's users.
```

The videos are not public, it's the comparisons.

backend/tournesol/lib/public_dataset.py
```diff
@@ -60,3 +60,52 @@ def get_dataset(poll_name: str) -> QuerySet:
         """,
         {"poll_name": poll_name},
     )
+
+
+def get_user_dataset(poll_name: str) -> QuerySet:
```
GresilleSiffle, line 65:
```suggestion
def get_users_dataset(poll_name: str) -> QuerySet:
```

backend/tournesol/lib/public_dataset.py
```diff
@@ -60,3 +60,52 @@ def get_dataset(poll_name: str) -> QuerySet:
...
+    """
+    Retrieve users present in the public dataset and return a non-evaluated
+    Django `RawQuerySet`.
```
GresilleSiffle, line 68:
```suggestion
    Retrieve users with at least one public comparison and return a
    non-evaluated Django `RawQuerySet`.
```

backend/tournesol/views/exports.py
```diff
@@ -12,12 +12,33 @@
...
 from tournesol.entities.base import UID_DELIMITER
-from tournesol.lib.public_dataset import get_dataset
+from tournesol.lib.public_dataset import get_dataset, get_user_dataset
 from tournesol.models import Comparison, Poll
 from tournesol.serializers.comparison import ComparisonSerializer
 from tournesol.utils.cache import cache_page_no_i18n
 from tournesol.views.mixins.poll import PollScopedViewMixin
 
+README_CONTENT = """\
+This is the public database of Tournesol (https://tournesol.app/).
```
GresilleSiffle, line 22: I made few little suggestions about the content of `README_CONTENT` here: https://github.com/tournesol-app/tournesol/pull/1154#issuecomment-1277681110

GresilleSiffle (commented):
Thanks for the code, it works!

I'm going to check what we can do for the content of the `README.txt`. Honestly for a such a small file, it doesn't bother me to have it in the view.

GresilleSiffle:
Regarding the `README_CONTENT`, you can see how we currently use the following files in the view `preview.py`:

https://github.com/tournesol-app/tournesol/tree/main/backend/tournesol/resources

It should be good enough for this PR.

sigmike:
@GresilleSiffle 

> The videos are not public, it's the comparisons.

That's how @lfaucon suggested we say it in https://github.com/tournesol-app/tournesol/issues/1071#issuecomment-1268239920, and as I understand it's also how it's actually recorded in the database and prompted to the user (the public toggle is below each video). I made the change anyway

I also made the other suggested changes.

For the date and time in the filename I chose to use an ISO format: `tournesol_export_20221001T123456Z`. It's not the most readable one, but it's the only one that includes time and is compatible with filenames. Using `HHhMM` as you suggested felt weird because it's not a form that's used by English speaking countries.

backend/tournesol/views/exports.py
```diff
@@ -165,3 +187,35 @@ def get(self, request, *args, **kwargs):
...
         )
         return response
+
+
+class ExportPublicAllView(APIView):
+    """Export all the public data in a .zip file."""
+    throttle_scope = "api_export_comparisons"
+    permission_classes = [AllowAny]
+
+    @extend_schema(
```
amatissart, line 197: Let's define the same cache rule as the comparisons .csv:

```suggestion

    @method_decorator(cache_page_no_i18n(60 * 10))  # 10 minutes cache
    @extend_schema(
```

sigmike: Done in 3f8d589e1f1b508ce9127e4a8c9635224abd9fb9.

sigmike:
I renamed `voting_right` to `trust_score` (in the export only, not in the database).

GresilleSiffle (approved):

GresilleSiffle:
Merged, and unfortunately squashed by reflex =|

sigmike:
That's unfortunate. There was a few unrelated commits, and some commits had explanations I think would be relevant in the history. It's not a big deal though.

PR #1171, opened 2022-10-17
[dev-env] refactor: Update dataset and implement command to load public dataset

### LINKED ISSUE #876: Update database used in dev-env
## Problems

* common scenarios cannot be developed / tested due to missing data
* `tournesol_score` field is not populated
* e2e tests could fail because no recent data is available

## Potential solutions

1. Provide a script to import or generate recent data about videos / comparisons / scores
2. ??

amatissart:
A long-running branch that is finally in a clean state, I think :wink: 

Closes #876 

What's new?

#### `dev-env/dump-for-dev-env.sql.tgz` is updated with recent data

241 users (a sample of 10% of users present in the public dataset) + 2 test users (`user` and `user1`)
3776 videos, with complete metadata
14491 comparisons
and all related scores computed by Mehestan.

As a consequence, the new dump is significantly bigger: 12 MB instead of 2MB. It could be an issue if we plan to update this dump regularly (e.g all versions of this file would be downloaded when cloning the repository). Should we consider hosting these dumps outside of the repo?

#### A new management command `load_public_dataset`

This command downloads the public datasets of comparisons, and imports the data into the current database.
All related users, videos, ratings will be created if necessary.

It's used by the new option `./run-docker-compose.sh download` (as an alternative to the default behaviour that initializes the database from a dump committed in the repository). 

If a Youtube API key has been configured, video metadata will be fetched. The requests are executed from multiple threads 
As the googleapiclient is not thread-safe, a new client has to be initialized for each thread.  
> **Warning**
> The total number of queries to the Youtube API may be close to (or even reach) the daily quota allowed by the API, especially if no sampling is applied.

The commands support a `--user-sampling` option to import a given proportion of users. In any case, all core users will be imported, in addition to the sampled users.

####  Help message for `./run-docker-compose.sh`

```
> ./run-docker-compose.sh help

Usage: ./run-docker-compose.sh [command]

Commands:

  init (default)
  Initialize the development environment and load the default database.
  This will ask for confirmation before deleting any existing database.
    ./run-docker-compose.sh
    ./run-docker-compose.sh init

  download
  Download up-to-date public dataset and initialize dev-env with this data.
  Options for 'load_public_dataset' management command can be passed.
  WARNING: if a Youtube API key has been configured, a large amount of queries
  to the API may be required to fetch metadata about all videos.
    ./run-docker-compose.sh download
    ./run-docker-compose.sh download --user-sampling 0.1

  restart
  Recreate containers and restart all dev_env services using the existing database. Containers will be rebuilt if necessary.
    ./run-docker-compose.sh restart

  stop
  Stop dev containers. The database will be persisted in "db-data" folder.
    ./run-docker-compose.sh stop
```


aidanjungo (approved):
LGTM, but it is probably better than someone else read de code to be sure, I could have missed something.
I tested the dev-env on both Linux and MacOS (it work works at the third try, there was no error but the front was not finishing to build...) and it seems to works great on both now.
Side question, when we get the vouch in the public dataset, should we also "add" them to the public user?

GresilleSiffle:
Hello, I'm currently reviewing.

GresilleSiffle:
Thank you Adrien for this PR :+1: :+1: :+1: 

I was wondering, should the end-to-end tests pass after having initialized the environment with `./run-docker-compose.sh download`?

With only `./run-docker-compose.sh` (without the download option), the tests pass, because the environment is fixed.

With the `download` option, the test may fail depending on the downloaded data. If there is no API key the tests relying on videos metadata always fail. If there is an API key some tests can still fail (one test is expected no recommendation when filtering the results by "uploaded a day ago").

In my opinion it is expected that the tests may fail with the `download` option. Also I expect the `tests`'s README.md to explicitly say that the environment should be initialized without this option.

Currently the README.md says as a requirement: 
> A local instance of Tournesol, as provided by `dev-env` 

We could say instead:
 > A local instance of Tournesol, as provided by `dev-env`, initialized with its embedded local data

GresilleSiffle (approved):

GresilleSiffle:
Feel free to merge when you are ready @amatissart :ok_hand:

GresilleSiffle:
It works well on my side :+1:

GresilleSiffle (approved):

PR #1179, opened 2022-10-20
[back] refactor: `voting_right` is now `trust_score`

GresilleSiffle:
This PR renames the `User.voting_right` field into `User.trust_score`.

@lfaucon @amatissart should we also update the following names:
- `vouch.trust_algo.MIN_PRETRUST_VOTING_RIGHT`
- `vouch.trust_algo.compute_voting_rights`
- and all var names in this file?

Should we also replace the mention of voting rights in the related tests?


amatissart:
> @lfaucon @amatissart should we also update the following names:
> 
>     * `vouch.trust_algo.MIN_PRETRUST_VOTING_RIGHT`
> 
>     * `vouch.trust_algo.compute_voting_rights`
> 
>     * and all var names in this file?
> 
> 
> Should we also replace the mention of voting rights in the related tests?

Yes, I think all mentions of "voting_right" in vouch/ should be renamed. As far as I understand, the point of this change is to make voting rights completly out of the scope of the Eigentrust algorithm and vouching mechanism. A "voting_right" is a concept that will be introduced when computing entity scores, and and is no longer related to a user directly.

backend/vouch/trust_algo.py
```diff
@@ -9,39 +9,42 @@
 
 logger = logging.getLogger(__name__)
 
-# In this algorithm, we leverage pretrust (e.g., based on email domains) and vouching
-# to securely assign voting rights to a wider set of contributors.
-# The algorithm inputs pretrust status and a vouching directed graph.
+# In this algorithm, we leverage pre-trust (e.g., based on email domains) and
```
GresilleSiffle, line 12: Here `pre-trust` refers to pre-trusted users or to something else?

backend/vouch/trust_algo.py
```diff
@@ -9,39 +9,42 @@
...
 
-# Voting rights are computed iteratively, which yields an approximate solution.
+# Trust scores are computed iteratively, which yields an approximate solution.
 APPROXIMATION_ERROR = 1e-8
 
-# In our model, users that vouch for few (if any) implicitly vouch for pre-trusted users.
-# IMPLICIT_PRETRUST_VOUCH is the implicit amount of vouch given to pretrusted,
-# as opposed to the explicit vouches, each of which is of unit value.
+# In our model, users that vouch for few (if any) implicitly vouch for
+# pre-trusted users. IMPLICIT_PRETRUST_VOUCH is the implicit amount of vouch
```
GresilleSiffle, line 25: Could `amount of vouch` be replaced by `amount of trust`? As there are several concept in this algorithm, the less we use synonyms, the more users with no prior knowledge will understand.

Here are all concepts I found in this file:
- pre-trust (and pre-trust status)
- pre-trusted users
- trust (and trust weight)
- trust score
- vouch (and amount of vouch, IMPLICIT_PRETRUST_VOUCH)
- vouching
- vouchee
- voucher

sigmike: "Vouch" doesn't really exist as a noun (https://en.wiktionary.org/wiki/vouch), even if we can easily understand the meaning in the code. That's why I didn't use it in the frontend. And "voucher" doesn't mean exactly what we want here either (https://en.wiktionary.org/wiki/voucher#English). It looks like there's no noun for the concept we want. But in the code we can probably define one like we define other abstractions.

backend/faq/migrations/0004_alter_faqentry_options.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.0.8 on 2022-10-20 14:36
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
lfaucon, line 6: Why do we have this change here? Was this migration simply forgotten in a previous change?

backend/vouch/trust_algo.py
lfaucon: I would propose to rephrase "pre-trust" to "pre-vouching". These are trust scores that we use to initialize (and more) the EigenTrust algorithm (vouching)

backend/vouch/trust_algo.py
lfaucon: X_X 

This is indeed quite complicated with all these concepts. "vouch" and "trust" are not synonyms and cannot be changed here.

lfaucon (commented):
I'll read through once more to make sure it is all well said and if we could rephrase things to be more clear

backend/core/migrations/0011_user_votingright_isnow_trustscore.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0010_alter_user_voting_right"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
baptou12, line 13: Won't the data be lost if we remove and recreate in the same migration?

backend/vouch/trust_algo.py
GresilleSiffle: OK I see, thank you. If these are trust scores used to initialized the algorithm how about pre-trust scores? or initial trust? (or initial network trust?)

I'm OK with the current formulation in fact, I wanted to be sure that it was a different concept than pre-trust users. If no it was the good moment to change the comment.

backend/faq/migrations/0004_alter_faqentry_options.py
GresilleSiffle: Exactly

backend/core/migrations/0011_user_votingright_isnow_trustscore.py
GresilleSiffle: Yes, but we currently have no data in production. If I remember correctly I can rename the field instead of deleting it. I'll do it.

backend/vouch/trust_algo.py
GresilleSiffle: I see. They may be an unavoidable complexity related to the feature itself, in this case it can be expected and acceptable.

What we can do, is to define in the `trust_algo` module docstring, all key concepts used by the algorithm, like pre-trust, pre-trusted users, trusted users, post-trust, trust score (voting rights if it's part of the same algorithm).

amatissart: The "amount of vouch" (or "vouch value") is a value assigned to an _edge_ of the vouch network, i.e to a pair of user: "vouching user" -> "vouchee" (user receiving the voucher). The explicit vouchers, created by users on the platform are currently assigned the default "value" 1 in the database, but the algorithm completes the network with additional links with smaller values (hence "implicit vouch").

The "trust" values are assigned to the _nodes_ of the vouch network, i.e users.  Pretrusts/posttrusts/and final trust scores are trust values, defined at different steps of the "trust algo".

"Pre-trust" is the trust value assigned to a user before the vouch network has been taken into account. So "pre-vouching" could fit that definition. However to avoid any confusion with vouch values, I would prefer to rename them to "initial_trust", "input_trust" or "apriori_trust"

"Posttrusts" is the output of the Eigen-trust algorithm, after vouchers have been taken into account to update trust values. These values need to be rescaled and clipped (in `compute_trust_scores()`) before they can be stored as the final "trust_score" for each user.

backend/core/models/user.py
```diff
@@ -194,11 +194,11 @@ class User(AbstractUser):
         validators=[validate_avatar],
         null=True,
     )
-    voting_right = models.FloatField(
+    trust_score = models.FloatField(
         null=True,
         blank=True,
         default=None,
-        help_text="The voting right assigned to the user based on the vouching mechanism.",
+        help_text="The trust score assigned to the user based on the vouching mechanism.",
```
lfaucon, line 201: This would be more correct

```suggestion
        help_text="The trust score assigned to the user based on trusted domains and the vouching mechanism.",
```

backend/tournesol/tests/test_exports.py
```diff
@@ -193,32 +195,35 @@ def test_not_authenticated_can_download_all_exports(self):
...
-            with zip_file.open(root + '/comparisons.csv', 'r') as file:
+            with zip_file.open(root + "/comparisons.csv", "r") as file:
                 content = file.read()
                 expected_content = self.client.get("/exports/comparisons/").content
                 self.assertEqual(content, expected_content)
 
-            with zip_file.open(root + '/users.csv', 'r') as file:
-                content = file.read().decode('utf-8')
+            with zip_file.open(root + "/users.csv", "r") as file:
+                content = file.read().decode("utf-8")
```
lfaucon, line 226: Want to enforce something (recommended editor settings / pre-commit hooks / black --check in CI / ...) to avoid this kind of diffs? It makes review slightly harder and has a small chance that a reviewer miss a relevant change hidden among syntax changes.

lfaucon (approved):
I trust this change, no need to vote on it ;)

backend/tournesol/tests/test_exports.py
lfaucon: Relevant bit from our readme about the pre-commit hooks:
```
## F.A.Q.

**Why are the code quality tools not automatically triggered locally with the
help of the Git pre-commit hook?**

We try to use the [pre-commit][pre-commit] framework, but unfortunately it's
not adapted to work smoothly within our mono-repository
configuration [[faq-1][faq-1]].

Each time we found a solution a new issue appeared. In the end it requires much
more efforts to be configured and used rather than simply running the code
quality checks manually.

As we do not want to make the developers to have complex and error-prone local
setups, we decided to add the code quality checks only in the CI for now.
```

Issue #1174, opened 2022-10-19
[back] refactor: define the new default ordering of the ratngs API list


GresilleSiffle:
Currently the rated entities are displayed in the front end thanks to the ratings API.

Ratings can be ordered by `last_compared_at`, `n_comparisons` or by the default ordering which is the entity publication date.

**question 1**

What should be the default ordering? As we have been asked few times on Discord to order by comparison date, the `last_compared_at` seems to be a good candidate.

**question 2**

Do we want to add the parameter publication date to the currently available ordering parameters? If yes we need to decide if all entity types should have a publication date, or if each entity type can define additional ordering parameters in addition to the generic `last_compared_at` and `n_comparisons`.

PR #1180, opened 2022-10-24
[back] feat: ratings are now order-able by some entity metadata fields

GresilleSiffle:
**related to** #1174 

---

This PR allows the results of the ratings API to be ordered by `duration` and `publication_date`.

As requested by some users, the default order has changed and is now `-last_compared_at` instead of the video metadata `publication_date`. This change will automatically be visible in the front end when the API will be updated.

Following the "never trust user input" principle, the available metadata fields that can be used to order a entity list through the ratings API are defined in a fixed allow list, per entity type. A `video` can be ordered by `duration`, a candidate no.

### DIFF
--- a/backend/tournesol/entities/base.py
+++ b/backend/tournesol/entities/base.py
@@ -1,6 +1,6 @@
 import logging
 from abc import ABC, abstractmethod
-from typing import Dict, Iterable, List, Type
+from typing import Dict, Iterable, List, Optional, Type
 
 from django.contrib.postgres.search import SearchQuery, SearchRank
 from django.db.models import F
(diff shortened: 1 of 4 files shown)


backend/tournesol/entities/base.py
```diff
@@ -186,6 +185,14 @@ def filter_date_lte(cls, qs, max_date):
     def filter_date_gte(cls, qs, min_date):
         return qs.filter(add_time__gte=min_date)
 
+    @classmethod
+    def get_allowed_meta_order_fields(cls) -> Optional[List[str]]:
+        """
+        Return a list of metadata fields that can be used to order an entity
+        list.
+        """
+        return None
```
amatissart, line 194: minor: would it make sense to use the same return type `List[str]` here and in child classes, and return `[]` by default?

amatissart (approved):

backend/tournesol/entities/base.py
GresilleSiffle: Sure, it will be more easy to read and understand.

PR #1181, opened 2022-10-25
[front] feat: add an order by input in the ratings page

GresilleSiffle:
**follow the work of** #1174 

---

This PR adds an orderBy input in the front end's My rated elements page.

This input contains two generic values available for all entity type: `last_compared_at` and `n_comparisons`. On the poll `videos` two additional parameters are available: `duration` and `publication_date`.

The maximum width of the ratings page has been increased to match the recommendations page width, and to have more space to display the new input.

At first I wanted to create full and exhaustive tests for each ordering parameter. I then realized the front end only displays the results returned by the back end without altering the order. While the back end's tests doesn't 100% ensure the front end will work, the absence of complexity in the front end makes me feel that it might not be worth to add right now and maintain extra e2e tests, and extra CI duration to test the correctness front end's display. My proposition is to rely on the back end for now, and add the needed e2e tests the day this feature will fail.

**to-do**
- [x] add entity type specific parameters
   - [x] `duration` and `publication_date` for video 
- [x] update the e2e tests
- [x] rework the fr translation (asc. vs. crois / desc. vs. décrois.)

### preview

capture

### DIFF
--- a/frontend/src/features/ratings/RatingOrderByInput.tsx
+++ b/frontend/src/features/ratings/RatingOrderByInput.tsx
@@ -42,7 +42,6 @@ function RatingOrderByInput(props: FilterProps) {
           onChange={handleChange}
           inputProps={{ 'data-testid': 'order-by-metadata-input' }}
         >
-          <MenuItem value="">--</MenuItem>
           <MenuItem value="last_compared_at">
             {t('ratingOrderByInput.lastComparisonDate')}&nbsp;
             {t('ratingOrderByInput.(asc)')}
--- a/frontend/src/features/ratings/RatingOrderByInput.tsx
+++ b/frontend/src/features/ratings/RatingOrderByInput.tsx
@@ -35,6 +35,7 @@ function RatingOrderByInput(props: FilterProps) {
           label={t('ratingOrderByInput.order')}
           value={props.value}
           onChange={handleChange}
+          inputProps={{ 'data-testid': 'order-by-metadata-input' }}
         >
           <MenuItem value="">--</MenuItem>
           <MenuItem value="last_compared_at">
--- a/frontend/src/features/ratings/RatingOrderByInput.tsx
+++ b/frontend/src/features/ratings/RatingOrderByInput.tsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import {
+  Box,
   FormControl,
   InputLabel,
+  ListItemText,
   MenuItem,
   Select,
   SelectChangeEvent,
@@ -30,6 +32,32 @@ function RatingOrderByInput(props: FilterProps) {
     props.onChange(event.target.value);
   };
 
+  const genericItems = useMemo(
+    () => [
+      {
+        label: t('ratingOrderByInput.lastComparisonDate'),
+        value: 'last_compared_at',
+        icon: CallMadeIcon,
+      },
+      {
+        label: t('ratingOrderByInput.lastComparisonDate'),
+        value: '-last_compared_at',
+        icon: CallReceivedIcon,
+      },
+      {
+        label: t('ratingOrderByInput.lastComparisonDate'),
+        value: 'n_comparisons',
+        icon: CallMadeIcon,
+      },
+      {
+        label: t('ratingOrderByInput.lastComparisonDate'),
+        value: '-n_comparisons',
+        icon: CallReceivedIcon,
+      },
+    ],
+    [t]
+  );
+
   return (
     <TitledSection title={t('ratingOrderByInput.orderBy')}>
       <FormControl fullWidth size="small">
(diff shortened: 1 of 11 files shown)


frontend/src/features/ratings/RatingOrderByInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+          labelId="order-by-metadata-label"
+          label={t('ratingOrderByInput.order')}
+          value={props.value}
+          onChange={handleChange}
+          inputProps={{ 'data-testid': 'order-by-metadata-input' }}
+        >
+          <MenuItem value="">--</MenuItem>
+          <MenuItem value="last_compared_at">
+            {t('ratingOrderByInput.lastComparisonDate')}&nbsp;
+            {t('ratingOrderByInput.(asc)')}
```
Mayeul-T, line 48: I don't known if we want do that in this PR but I think a generic Select should be a good idea to simplify implementation. We should create a Select wich be implemented like this:

```
<GenericSelect 
    items=[{value: '', label: '--'}, {value: 'last_compared_at', label: <></>}]
/>
```

Example of simple implementation of this generic Select

```
export default function GenericSelect({items, array, label: string}){
  return <FormControl ...>
    <InputLabel>{labe}</InputLabel>
    <Select ...>
      {items.map(item => <MenuItem value={value}>{label}</MenuItem>)}
    </Select>
    </FormControl>
}
```

GresilleSiffle: As discussed with @Mayeul-T we won't create a generic component in this PR.

frontend/src/features/ratings/RatingsFilter.tsx
```diff
@@ -22,14 +24,20 @@ function RatingsFilter() {
...
               value={filterParams.get('isPublic') ?? ''}
               onChange={(value) => setFilter('isPublic', value)}
             />
           </Grid>
-          <Grid item xs={12} sm={6} md={5}>
+          <Grid item xs={12} sm={6} md={5} lg={4}>
+            <RatingOrderByInput
+              value={filterParams.get('orderBy') ?? ''}
+              onChange={(value) => setFilter('orderBy', value)}
+            />
```
amatissart, line 38: 2 small things about the behavior of this input:
(They are not blocking, and may be addressed separately if considered relevant.)

1. By default when accessing the /ratings page, no value for `orderBy` is selected. The order will depend on the default order implemented in the backend. I think it would be a slightly better UX to have a selected value at all time, controlled by the frontend. It may require to extend the implementation of `useListFilter()` to pass a default value when none is present in the URL.

2. The order of items with `?orderBy=-last_compared_at` may be different from the default order (when no explicit orderBy is defined). That's because the backend implementation uses the primary key to break ties by default:
https://github.com/tournesol-app/tournesol/blob/6eda6020a9cb56711c23631f7e82efc4606327de/backend/tournesol/views/ratings.py#L28
I think it would make sense to append `"-pk"` in the backend implementation when ordering entities, to keep a deterministic order in all cases.

amatissart (approved):

glerzing:
Is it possible to have arrows that go up or down instead of having "crois." and "decrois." ?

frontend/src/features/ratings/RatingsFilter.tsx
GresilleSiffle: Thank you, I'll try to fix both points in this PR.

Regarding the point 1. I've been able to add a default filter value by updating the `RatingsFilter`. Do you think it's better to extend `useListFilter()` instead? I currently don't know if the hook `useListFilter()` is a good place to handle the default values, mainly because my lack of knowledge about React's hooks.

See: https://github.com/tournesol-app/tournesol/pull/1181/commits/b4757890b3927f35022ba37bb8b5715fd6907152

This commit also includes the point 2.

GresilleSiffle: I finally fixed the point 1 as suggested.

https://github.com/tournesol-app/tournesol/pull/1181/commits/c93faf66ec808dfa0e3e0135bfe74d042c149fd9

GresilleSiffle:
Hello @amatissart 

I extended `useListFilter()` to make it able to manage default value (here: https://github.com/tournesol-app/tournesol/pull/1181/commits/c93faf66ec808dfa0e3e0135bfe74d042c149fd9), and I simplified the component JSX by using a loop instead of hard coding every item (here: https://github.com/tournesol-app/tournesol/pull/1181/commits/717eab2a4f892c0581d3a0bada061b6a58ac61fa).

What do you think? I also used `useMemo` to define the items that shouldn't change between rendering, I'm not sure if it's a good practice for such a list or if it's not that much appropriate here.

frontend/src/features/ratings/RatingOrderByInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+      },
+      {
+        label: t('ratingOrderByInput.lastComparisonDate'),
+        value: 'n_comparisons',
+        icon: CallMadeIcon,
+      },
+      {
+        label: t('ratingOrderByInput.lastComparisonDate'),
+        value: '-n_comparisons',
+        icon: CallReceivedIcon,
```
amatissart, line 55: The same label is used 4 times.

frontend/src/features/ratings/RatingsFilter.tsx
```diff
@@ -4,13 +4,24 @@ import { Box, Collapse, Grid } from '@mui/material';
...
+  defaultFilters: Array<{ name: string; value: string }>;
+}) {
   const { t } = useTranslation();
   const [expanded, setExpanded] = useState(false);
-  const [filterParams, setFilter] = useListFilter();
+  const [filterParams, setFilter] = useListFilter({ defaults: defaultFilters });
+
+  const defaultFilterOrderBy = defaultFilters
+    .filter((param) => param.name === 'orderBy' && param.value != null)
+    .pop() ?? { value: '' };
```
amatissart, line 24: This is unnecessary, isn'it ? `filterParams.get('orderBy') ?? ""` can be used directly.

```suggestion
```

GresilleSiffle: Indeed, fixed by https://github.com/tournesol-app/tournesol/pull/1181/commits/3be9baf39d85d851ae5e95a0ebb80981be6ef2d9

frontend/src/features/ratings/RatingOrderByInput.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1181/commits/3be9baf39d85d851ae5e95a0ebb80981be6ef2d9

PR #1183, opened 2022-10-26
[front] refactor: delete the `ContributeSection` from the videos home

GresilleSiffle:
This section was redundant with the recent `ComparisonSection`.

This PR also:

* Increases the vertical padding of each section of the home page from 32 pixels to 48. This allows the home page to feel less "squashed" and more "modern", especially on mobile. I also tested 64 pixels, which is nice too but I'd prefer to wait for the complete home page refactor before proposing 64. 

* Updates the `AlternatingBackgroundColorSectionList` component to use the app palette instead of defining hard-coded custom colors. Also update this component to be able to choose the color of its last item displayed.

* Adds a title `ComparisonSection` has now a proper title.

### Preview

The important part are the new contribute title, the color of the last sections (stats and polls) and the vertical padding of each section (which is a bit subtle on the image).

screencapture-localhost-3000-2022-10-26-18_16_32

### DIFF
--- a/frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
+++ b/frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
@@ -56,7 +56,19 @@ const AlternatingBackgroundColorSectionList = ({
   lastItemIsPrimary?: boolean;
 }) => {
   const theme = useTheme();
-  const lastItem = children.at(-1);
+
+  // children.at(-1) didn't work here...
+  const lastItem = children.slice(children.length - 1);
+
+  const nextColor = (idx: number): string => {
+    return idx % 2 == 0 ? secondaryColor : theme.palette.text.primary;
+  };
+
+  const nextBackground = (idx: number): string => {
+    return idx % 2 == 0
+      ? secondaryBackground
+      : theme.palette.background.primary;
+  };
 
   return (
     <Grid container width="100%">
@@ -66,10 +78,8 @@ const AlternatingBackgroundColorSectionList = ({
           key={i}
           child={child}
           verticalPadding={i == 0 ? firstElemVerticalPadding : verticalPadding}
-          color={i % 2 == 0 ? secondaryColor : theme.palette.text.primary}
-          background={
-            i % 2 == 0 ? secondaryBackground : theme.palette.background.primary
-          }
+          color={nextColor(i)}
+          background={nextBackground(i)}
         />
       ))}
 
--- a/frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
+++ b/frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
@@ -60,11 +60,11 @@ const AlternatingBackgroundColorSectionList = ({
   // children.at(-1) didn't work here...
   const lastItem = children.slice(children.length - 1);
 
-  const nextColor = (idx: number): string => {
+  const colorAtIndex = (idx: number): string => {
     return idx % 2 == 0 ? secondaryColor : theme.palette.text.primary;
   };
 
-  const nextBackground = (idx: number): string => {
+  const backgroundColorAtIndex = (idx: number): string => {
     return idx % 2 == 0
       ? secondaryBackground
       : theme.palette.background.primary;
@@ -78,8 +78,8 @@ const AlternatingBackgroundColorSectionList = ({
           key={i}
           child={child}
           padding={i == 0 ? firstItemPadding : itemPadding}
-          color={nextColor(i)}
-          background={nextBackground(i)}
+          color={colorAtIndex(i)}
+          background={backgroundColorAtIndex(i)}
         />
       ))}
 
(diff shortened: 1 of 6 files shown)


frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
```diff
@@ -2,45 +2,105 @@ import React from 'react';
...
+    <Grid
+      item
+      xs={12}
+      sx={{
+        display: 'flex',
+        justifyContent: 'center',
+        padding: verticalPadding,
+        [theme.breakpoints.down('md')]: {
+          padding: `${verticalPadding} 16px ${verticalPadding} 16px`,
+        },
```
Mayeul-T, line 28: You can also write:

`
{
p: verticalPadding,
px: {xs: '16px', md: verticalPadding} 
}`

I don't know if verticalPadding is a right naming because it defined also horizontal padding if you display the desktop version

frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
```diff
@@ -2,45 +2,105 @@ import React from 'react';
...
+            background={
+              lastItemIsPrimary
+                ? theme.palette.background.primary
+                : i % 2 == 0
+                ? secondaryBackground
+                : theme.palette.background.primary
+            }
+          />
+        ) : (
+          <SectionGridItem
```
Mayeul-T, line 88: This condition is very long and I think not really readable.

You should separate more the element with something like
```
{
   children.slice(0, 1).map(childre, => <>...</>)
}
{
   <>Display last item using "children.at(-1)"</>
}
```

frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
```diff
@@ -2,45 +2,105 @@ import React from 'react';
...
+            child={child}
+            verticalPadding={
+              i == 0 ? firstElemVerticalPadding : verticalPadding
+            }
+            color={i % 2 == 0 ? secondaryColor : theme.palette.text.primary}
+            background={
+              i % 2 == 0
+                ? secondaryBackground
+                : theme.palette.background.primary
+            }
```
Mayeul-T, line 99: You repeat this ternary expression very often in this file. You can write a function to do that no ?

frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
GresilleSiffle: Thank you, fixed by https://github.com/tournesol-app/tournesol/pull/1183/commits/c238f4591d3121ff50dc14cc506d536d5b14d196

frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
GresilleSiffle: Sure, I made the logic more explicit by creating two functions. See https://github.com/tournesol-app/tournesol/pull/1183/commits/5f37b36e3022f7e76b358ea9718073ab453d7448

frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1183/commits/41025b059d24bda944d1fcfd6f8cb837d6ab35f8

Mayeul-T (approved):

frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
```diff
@@ -2,45 +2,107 @@ import React from 'react';
...
+  // If true the last item will use the primary color regardless of its
+  // position in the list.
+  lastItemIsPrimary?: boolean;
 }) => {
   const theme = useTheme();
 
+  // children.at(-1) didn't work here...
+  const lastItem = children.slice(children.length - 1);
+
+  const nextColor = (idx: number): string => {
```
lfaucon, line 63: minor: slightly better names: `colorAtIndex(idx: number)` and `backgroundColorAtIndex(idx: number)`

frontend/src/pages/home/AlternatingBackgroundColorSectionList.tsx
```diff
@@ -2,45 +2,107 @@ import React from 'react';
...
+      {/* display the last item */}
+      <SectionGridItem
+        child={lastItem}
+        padding={itemPadding}
+        color={
+          lastItemIsPrimary
+            ? theme.palette.text.primary
+            : children.length % 2 != 0
+            ? secondaryColor
+            : theme.palette.text.primary
```
lfaucon, line 95: Could this be moved to the `nextColor` function and then not require to make a special case in the map on line 76 ?

lfaucon (approved):
Happy with the interface change. Only suggesting a minor code simplification :)

Issue #1176, opened 2022-10-20
[back] Compute per video contributor voting rights


lfaucon:
Voting rights are computed based on the trust scores coming from Email certificatino and Eignentrust. Contributors have a varying voting right for each video based on the number of contributors who have rated the video and their trust scores 

We simply estimate an amount of over trust which can be safely distributed to users with lowest trust scores based on estimating a minimum voting right.

lfaucon:
> Does the first mention of trust score refers to voting right?

Yes! Thanks. I fixed the description

PR #1186, opened 2022-10-30
[back] feat: Add algorithm to compute voting rights

lfaucon:
Related to #1176 

This is the first part of the required implementation. Next steps will be to include this algorithm in the ML routine such that it reads and write data from/to the database.

### DIFF
--- a/backend/vouch/voting_rights.py
+++ b/backend/vouch/voting_rights.py
@@ -1,5 +1,4 @@
 import numpy as np
-from typing import List
 
 # On a given entity, an amount of extra voting right is given to users with lowest trust scores.
 # This amount is call overtrust and we want it to be low compared to regular trusted voting rights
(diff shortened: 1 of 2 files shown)


backend/vouch/voting_rights.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
amatissart, line 1: Maybe this algorithm should be moved to backend/ml/ ? It's not directly related to the vouching mechanism.

amatissart (dismissed):
As discussed, we can merge this first version, and integrate the privacy penalty separately.

backend/vouch/voting_rights.py
lfaucon: yeah, I hesitated. I'm thinking that `ml` is quite a bad name.

GresilleSiffle: I also think `ml` is not appropriate anymore. It could be a good move to rename it one day.

Few thoughts:

`ml` is very very generic, we could have several Django app using ML algorithms in the back end, in this case they will have names based on their purpose and not based on what they are. Using `ml` for a Django app is a bit like naming a variable `integer` or `string`.

amatissart (approved):

Issue #1185, opened 2022-10-30
spike: Review, propose improvements to ML unit tests


lfaucon:
Acceptance:
- Get an overview of existing tests for Tournesol ML algorithms (vouch & scores)
- Share overview with the team to spread knowledge
- Make a proposal for ways to improve testing
- Agree with team on the proposal
- Create tickets as approapriate

PR #1187, opened 2022-10-30
[back] fix: update `QrUnc` in ml primitives, as defined in reference paper

amatissart:
Updates the algorithm computing the quadratically regularized uncertainty `QrUnc`. 

From **Section C.4** in supplementary material, the individual contributions to the second derivative used in the computation should be capped to 1/W, instead of 1.
> 

image

### DIFF
--- a/backend/ml/mehestan/primitives.py
+++ b/backend/ml/mehestan/primitives.py
@@ -83,8 +83,9 @@ def QrUnc(
         qr_med = QrMed(W, w, x, delta)
     qr_dev = QrDev(W, default_dev, w, x, delta, qr_med=qr_med)
     delta_2 = delta ** 2
+    bound = np.inf if W <= 0 else 1 / W
     h = W + np.sum(
-        w * np.minimum(1, delta_2 * (delta_2 + (x - qr_med) ** 2) ** (-3 / 2))
+        w * np.minimum(bound, delta_2 * (delta_2 + (x - qr_med) ** 2) ** (-3 / 2))
     )
 
     if h <= W:
--- a/backend/ml/tests/test_mehestan_primitives.py
+++ b/backend/ml/tests/test_mehestan_primitives.py
@@ -1,0 +1,66 @@
+import numpy as np
+from django.test import TestCase
+
+from ml.mehestan.primitives import QrMed, QrUnc
+
+
+class QrMedTest(TestCase):
+    def test_qrmed(self):
+        W = 2
+        weight = 1
+        self.assertAlmostEqual(
+            QrMed(
+                W,
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            0.5, # influence of the median contributor is bounded by weight / W
+            places=3
+        )
+
+    def test_qrmed_with_W_equals_zero(self):
+        W = 0
+        weight = 1
+        self.assertAlmostEqual(
+            QrMed(
+                W,
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            1.0,
+            places=3
+        )
+
+
+class QrUncTest(TestCase):
+    def test_qrunc(self):
+        W = 2
+        weight = 1
+        self.assertAlmostEqual(
+            QrUnc(
+                W,
+                default_dev=0,
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            0.506,
+            places=3,
+        )
+
+    def test_qrunc_with_W_equals_zero(self):
+        W = 0
+        weight = 1
+        self.assertAlmostEqual(
+            QrUnc(
+                W,
+                default_dev=0,
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            0.033,
+            places=3,
+        )
--- a/backend/tournesol/tests/test_ml_run.py
+++ b/backend/tournesol/tests/test_ml_run.py
@@ -84,6 +84,20 @@ class TestMlTrain(TransactionTestCase):
         self.assertEqual(scores_mode_default.filter(poll=Poll.default_poll()).count(), 8)
 
 
+    def test_ml_run_with_video_having_score_zero(self):
+        video = VideoFactory()
+        ComparisonCriteriaScoreFactory(
+            comparison__entity_1=video,
+            score=0,
+            criteria="largely_recommended",
+        )
+
+        self.assertEqual(video.tournesol_score, None)
+        call_command("ml_train")
+        video.refresh_from_db()
+        self.assertAlmostEqual(video.tournesol_score, 0.0)
+
+
     def test_individual_scaling_are_computed(self):
         # User 1 will belong to supertrusted users (as staff member)
         user1 = UserFactory(email="user@verified.test", is_staff=True)


backend/ml/mehestan/primitives.py
```diff
@@ -83,8 +83,9 @@ def QrUnc(
         qr_med = QrMed(W, w, x, delta)
     qr_dev = QrDev(W, default_dev, w, x, delta, qr_med=qr_med)
     delta_2 = delta ** 2
+    bound = np.inf if W <= 0 else 1 / W
     h = W + np.sum(
-        w * np.minimum(1, delta_2 * (delta_2 + (x - qr_med) ** 2) ** (-3 / 2))
+        w * np.minimum(bound, delta_2 * (delta_2 + (x - qr_med) ** 2) ** (-3 / 2))
     )
```
amatissart, line 89: Due to the definition of "Collaborative preference scaling factor of non-scaling-calibration contributors", `QrUnc` may be called with `W = 0`. 
For instance, this is the case for entities where all individual scores are zero. The uncertainty about the scaling factor is computed using $W = 8W||\theta_n||\_\infty$ where $||\theta_n||\_\infty = 0$

@lenhoanglnh What do you think? Is it correct to extend the definition of `QrUnc` to W=0, or should this case be handle explicitly in the scaling algorithm?

backend/ml/tests/test_mehestan_primitives.py
```diff
@@ -0,0 +1,66 @@
...
+        W = 2
+        weight = 1
+        self.assertAlmostEqual(
+            QrMed(
+                W,
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            0.5, # influence of the median contributor is bounded by weight / W
```
glerzing, line 18: I don't understand why this works. From my understanding of QrMed, it should be somewhere between 0 and 0.5, but not exactly 0.5. It is not like for BrMean where we have a clipping.

amatissart: It is, the exact value is about 0.4999990 (i.e not a rounding error). It illustrates how with low enough uncertainties in `delta`, the result will be arbitrary close to the theoretical bound for individual contributions. And I thought that could be a useful example to show how QrMed behaves like a "regularized median". Would it be less confusing to use more decimals in the test?

I suppose a good task for #1185 could be to define more tests cases to explain the impact of each parameter, and ideally find examples where the expected result would seem relatively intuitive.

glerzing: I just understood QrMed badly, you test looks good.

backend/ml/tests/test_mehestan_primitives.py
```diff
@@ -0,0 +1,66 @@
...
+            QrUnc(
+                W,
+                default_dev=0,
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            0.033,
+            places=3,
+        )
```
glerzing, line 66: We could also check that given an empty data set, it returns default_dev.

backend/ml/tests/test_mehestan_primitives.py
```diff
@@ -0,0 +1,66 @@
...
+        weight = 1
+        self.assertAlmostEqual(
+            QrUnc(
+                W,
+                default_dev=0,
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            0.033,
```
glerzing, line 64: I'm not sure how to interpret the fact that QrUnc is much lower when W is equal to 0.

glerzing: @lenhoanglnh QrUnc looks like it gives weird results depending on W (other variables are here the same as in Adrien's test). I don't think the problem comes from the python calculation, I think it's the formula that is not perfect.

image

glerzing: Here is when W = 2 and the weights vary :
image

Here is when W = 2 and each score is increased by x (scores = [-10 + x, 1 + x, 10 + x]) :
image

This is very weird

glerzing: and for the uncertainty

image

glerzing: QrUnc(default_dev) looks almost good

image

glerzing: Sorry Adrien, I forgot this PR. What you made looks correct to me. But the behaviour of QrUnc looks very weird and I think it's well worth investigating, notably whether you find the same patterns of curves. Not necessarily in this PR though, so I can validate if you want to do it separately.

glerzing: I'll investigate it in-depth this week, please let me a few days.

glerzing: I continue to investigate, but one thing that may be interesting would be to use the LogSumExp smooth minimum instead of the SoftMinMean.

One reason is that the SoftMinMean isn't really consistent, you can have x1 > x2 and yet SoftMinMean(x1, y) < SoftMinMean(x2, y). Moreover, (I didn't understand why, but) in the paper we say that both elements of SoftMeanMin "represent upper bounds on uncertainty" obtained with SoftMeanMin. So we want SoftMeanMin(x, y) < x and SoftMeanMin(x, y) < y. But this is not true, as shown in the following figure, it's the contrary.

The magenta curve represents the min and the red curves the SoftMinMean.

image

The LogSumExp looks like it fits more what we are trying to do :

image

Here is the code to generate and modify the images :
https://colab.research.google.com/drive/1QcNQXFxGcu9PnLPIhyGafCH6QyNx936t?usp=sharing

backend/ml/mehestan/primitives.py
```diff
@@ -83,8 +83,9 @@ def QrUnc(
...
 
     if h <= W:
```
glerzing, line 91:
```suggestion
    if h == W:
```

backend/ml/tests/test_mehestan_primitives.py
amatissart: That's very nice you can look deeper into this. I'm not really familiar with the initial motivation for using SoftminMax, but I would be curious to see the impact on "QrUnc" if LogSumExp would be use instead.

What do you think of opening a new issue to track the potential changes to apply on `QrUnc` and discuss this?

If there is no objection, I would like to merge this PR as it is. Even if it does not solve everything, it would bring consistency between the paper and what is used on production, and that would useful as a baseline. What do you think?

glerzing: Sure. And I'm starting to think that the fact that the curves are weird probably doesn't reveal a problem. I haven't finished to investigate, but it looks like when there is too little voting rights, there can be some clipping that causes steep irregularities.

I still think using LogSumExp is an improvement and opened the issue #1232 , but it shouldn't have a major effect on the weirdness of the curves.

glerzing (approved):

backend/ml/tests/test_mehestan_primitives.py
```diff
@@ -0,0 +1,66 @@
...
+                w=weight,
+                x=np.array([-10.0, 1.0, 10.0]),
+                delta=np.array([1e-3, 1e-3, 1e-3]),
+            ),
+            1.0,
+            places=3
+        )
+
+
+class QrUncTest(TestCase):
```
lfaucon, line 37: Are these tests created by running the algorithm first and then writing the test with the output value? If not, could we have comments that explain how the values are obtained, like you did for `QrMedTest.test_qrmed`?

I wish we would somehow be able to see the properties of QrUnc in the test suite (e.g. "always positive" or "always between zero and the true median" tested with multiple random inputs).

I say this while barely understanding QrUnc. Happy to be taught everything.

lfaucon:
Thanks for this work @amatissart @glerzing 🌻

PR #1194, opened 2022-11-10
[front] chore: upgrade dependencies, including mimimatch

amatissart:
Upgraded via `yarn upgrade`

An explicit resolution about "@types/react" had to be specified to avoid mixing multiple versions of React, which would break the TypeScript build.

This should also resolve the alert about a ReDos vulnerability in "minimatch".

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -96,5 +96,8 @@
         "name": "Français"
       }
     ]
+  },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


GresilleSiffle (approved):
Everything looks good to me.

I manually tested the tutorial as it's not part of our e2e tests.

PR #1195, opened 2022-11-10
[analytics] fix: bad command in the `README.md`

GresilleSiffle:
This PR fixes the `README.py` of the analytics folder.

As the title is Tournesol Data Vizualisation, the main script is now `tournesol_dataviz.py` instead of `Tournesol_analytics.py` (the starting upper case was not required).


aidanjungo:
The name `Tournesol_analytics.py` seem a bit strange but is was like that for a good reason. In Streamlit, this name is use as the name of the main page and the name display on the browser tab.

aidanjungo:
So we could keep the change, but the name sould be I think `Tournesol_Dataviz.py`

GresilleSiffle:
Oops sorry, thanks for your vigilance ; )

It could be nice if Streamlit gives more control on the UI, so that we can keep standard script names. Feel free to add the uppercase again later, sorry for the extra work.

PR #1197, opened 2022-11-11
[front] fix: recommendations could be inconsistent with filters due to concurrent requests to API

amatissart:


frontend/src/pages/recommendations/RecommendationPage.tsx
```diff
@@ -93,6 +82,7 @@ function RecommendationsPage() {
     }
     locationSearchRef.current = location.search;
 
+    const searchParams = new URLSearchParams(location.search);
```
GresilleSiffle, line 85: This line looks like a duplicate of the line 64. Is it required?

GresilleSiffle (approved):
Thank you :+1: 

I made a comment about a possible duplicated line.

GresilleSiffle:
I realized after approving that this PR was still marked as draft. Feel free to delete the review if more work is expected.

frontend/src/pages/recommendations/RecommendationPage.tsx
amatissart: It's not strictly required. But removing it would trigger a warning about a missing dependency for this effect. In the previous version this was avoided by creating this `searchParams` with `useMemo()`, but that does not seem correct either: `searchParams` is mutable (e.g via `searchParams.set("key", "value")`), so memoizing it based on `location.search` could introduce subtle bugs.

That's why I preferred to recreate it explicitly, as it's quite cheap. I will add a comment though.

amatissart:
> I realized after approving that this PR was still marked as draft. Feel free to delete the review if more work is expected.

This implementation is done, but I'd like to add more context and justification for these changes.

tests/cypress/integration/frontend/recommendationsPage.ts
```diff
@@ -66,7 +66,7 @@ describe('Recommendations page', () => {
           cy.contains('Filters', {matchCase: false}).click();
 
           cy.contains('A year ago', {matchCase: false}).should('be.visible');
-          cy.get('input[type=checkbox][name=""]').check();
+          cy.contains('All time', {matchCase: false}).click();
```
amatissart, line 69: For some reason clicking on this input element would fail because the element was partially hidden :shrug:

PR #1198, opened 2022-11-13
[back] Round trust scores and display in % in admin

aidanjungo:


lfaucon (approved):
LGTM

backend/core/admin.py
```diff
@@ -108,6 +110,10 @@ def get_fieldsets(self, request, obj=None) -> List[Tuple]:
     def is_trusted(self, instance):
         return instance.is_trusted
 
+    @admin.display(ordering="-trust_score", description="Trust score")
+    def get_trust_score(self, instance):
+        return f"{instance.trust_score:.1%}"
```
amatissart, line 115: This would cause a `TypeError` if the trust_score is None (default value for new users before the algo has run).

```suggestion
        if trust_score is None:
            return ""
        return f"{instance.trust_score:.1%}"
```

amatissart (changes requested):

amatissart (approved):

Issue #1157, opened 2022-10-05
[front] feat: round all scores


aidanjungo:
Score display on the graphs of scores and on criteria tooltip should be rounded to the unit. The extra digit is an heritage from past version, it did not bring any interesting information and it overloads the display.

<img width="344" alt="image" src="https://user-images.githubusercontent.com/53896963/194036553-723849f6-deb6-48cc-afc7-71efd90c5e55.png">

<img width="510" alt="image" src="https://user-images.githubusercontent.com/53896963/194036804-3c4c8495-7ba9-4b61-abcb-5af3efc909a3.png">

GresilleSiffle:
We want to round the scores displayed in the graph and the ones displayed in the card tooltip right?

aidanjungo:
yes, it's not what I wrote in the issue?

GresilleSiffle:
Sure, I read too fast sorry : p

riteshsoni123:
Hello @aidanjungo I would like to solve this issue. Can you please assign this to me?

GresilleSiffle:
Hello @riteshsoni123 

I just assigned the issue to you. Thanks for helping us : ) In case of need, do not hesitate to open a draft PR with your questions or simply join our Discord to talk with developers: https://discord.gg/WvcSG55Bf3

PR #1199, opened 2022-11-13
[front] fix: round the criteria scores to the nearest integer

lfaucon:
**related to** #1157

Changes:
* Scores on CriteriaBarChart and tooltips are rounded to the nearest integer
* Removed the `displayScore`methods as they were duplicated and felt unnecessary


aidanjungo (approved):
LGTM

GresilleSiffle (approved):
Good, ready to merge :ok_hand: 

The benefit I see of the `displayScore` function, is that developers don't have to think about the precision of the scores. I'd even say they are invited to not choose a different precision than the one currently implemented.

I'm ok with the latest changes, and I was also ok with the previous usage of `displayScore` (except the duplicated function).

PR #1204, opened 2022-11-15
[front] trans: use trust score instead of voting right in vouching page

GresilleSiffle:
As suggested by @sigmike , I propose to use this PR to update the translations in order to use trust score instead of voting right when it's appropriate.

Here are the translations we may want to update:
- **(wont fix, it's the correct terms)** https://github.com/tournesol-app/tournesol/blob/main/frontend/public/locales/en/translation.json#L239

Proposition for the key `howToChangeIt`:

Original
> You can increase it by using a certified email and when other users vouch for you.

Proposition
> You can gain more trust by using an email from a trusted domain or when other users vouch for you.

As discussed with Adrien, trusted domain is more accurate.
 
Additional proposition for the key `explanation`:

Original
> Your trust score changes how your comparisons count in Tournesol's recommendations.

Proposition
> Your trust score determines how much your comparisons influence the Tournesol's recommendations.

I think using the word influence is more explicit than count.

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -511,14 +511,14 @@
       "high": "High",
       "title": "Trust score",
       "description": {
-        "explanation": "Your trust score changes how your comparisons count in Tournesol's recommendations.",
-        "howToChangeIt": "You can increase it by using a certified email and when other users vouch for you.",
+        "explanation": "Your trust score determines how much your comparisons influence the Tournesol's recommendations.",
+        "howToChangeIt": "You can gain more trust by using an email from a trusted domain or when other users vouch of you.",
         "includedInPublicDatabase": "Your trust score is included the public database if you have at least one public comparison."
       },
-      "isTrusted": "Your email address is certified",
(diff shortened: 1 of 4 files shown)


GresilleSiffle:
I also made the trust meaning more visual by using a different color depending on the score.

Unknown and low scores use the `text.secondary` color.

capture

The medium scores use the app `secondary` color.

medium

And finally the high scores use the default success color of Material (which is not that beautiful in fact, but at least coherent with the check boxes displayed under the trust).

hight

amatissart:
> Also, until now we have used the terms trusted emails instead of certified emails, but the current vouching page says certified (terms introduced by the last PR I merged). Which terms should we use to stay coherent?

I think it would be more correct and clearer to use "trusted" everywhere, to be consistent with the page https://tournesol.app/about/trusted_domains where we refer to "trusted domains" / "domaines fiables".

GresilleSiffle:
I also simplified few french translations to make them less verbose, so they can fit in one line on small screens.

amatissart (approved):

PR #1208, opened 2022-11-16
[front] fix: reset scroll position on page change

amatissart:
When navigating to another page via a React Router link, the scroll position persisted to the previous position. Especially visible when clicking on "see more" in the Recommendations section on the home page.

The implementation is similar to the approach described on https://v5.reactrouter.com/web/guides/scroll-restoration/scroll-to-top


GresilleSiffle (approved):
Thank you.

PR #1220, opened 2022-11-28
WIP / [front] feat: add an hardcoded research banner

GresilleSiffle:


browser-extension/src/addTournesolRecommendations.js
```diff
@@ -5,6 +5,11 @@
 
 const videosPerRow = 4;
 const rowsWhenExpanded = 3;
+
+// These are placeholder values that will be updated.
+const TS_BANNER_DATE_START = new Date('2024-01-01T00:00:00Z');
+const TS_BANNER_DATE_END = new Date('2024-06-01T00:00:00Z');
```
Mayeul-T, line 11: You can write "TODO" because these values should be hardcoded and most IDE detect TODO word. Without that, we will forget this comment

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1220/commits/408d66015e44d1f24786dbb0e79c1f1d58d8fad5

amatissart:
LGTM  
Let's just confirm the date period when the banner should appear.

minor suggestions:
* add a "close" button on the extension banner. Its state would be persisted in the extension storage. 
* Close the banner after a click on "join"

browser-extension/src/addTournesolRecommendations.js
lfaucon, line 1: Question / Not for this PR / Should we try to modularize the code of the extension, for example `addTournesolRecommendations` could import `buildStudyBanner` from another file and use it in one line?

lfaucon (commented):
Looks fantastic! I'm already trying to think of the hundreds of things that we want to announce this way. Thanks a lot

GresilleSiffle:
> Looks fantastic! I'm already trying to think of the hundreds of things that we want to announce this way. Thanks a lot

Maybe we would like to announce the future /banners/ API that allows us to talk directly with the users ah ah : DD Honestly I think it can be a really nice step forward in our way of communicating with the community.

browser-extension/src/addTournesolRecommendations.js
GresilleSiffle: Yes that would be great. For now there is no good reason to keep the banner code in the `addTournesolRecommendations` file. I directly updated this file to quickly have a working banner, and thinking the code will be deleted soon, but it seems I have time now : )

GresilleSiffle: It looks like importing JavaScript modules from a content script is not as straightforward as expected.

I didn't investigate all methods yet, but I should either dynamically import the modules and their functions one by one, which looks a bit hacky, or completely rework how the file `addTournesolRecommendations.js` works (there may be other possibilities I haven't found yet).

**solution 1**

```javascript
// addTournesolRecommendations.js
let isNavigatorLang;

(async () => {
    const {
      isNavigatorLang: isNavigatorLangImported,
    } = await import(chrome.extension.getURL("/lib/i18n.js"));

   isNavigatorLangImported = isNavigatorLang;
})();

// isNavigatorLang is now available in the rest of the file.
```

**solution 2**

Move the whole content of `addTournesolRecommendations.js` in a JavaScript module that will be able to import other modules with the regular import from syntax.

See the example here: https://github.com/otiai10/chrome-extension-es6-import

If I do this in this PR it might be in conflict with the PR https://github.com/tournesol-app/tournesol/pull/1436. It seems better to wait a bit before adding these changes.

GresilleSiffle:
Hello @amatissart @lfaucon 

I added the possibility to hide and to display the banner on demand. The banner now includes a close button, and an additional action next to refresh and expand to re-display the banner.

I'm going to add a condition to display the campaign button only if the current date is between the start / end dates of the study.

capture

I decided to not automatically close the banner after a click on the Join button. I'm worried about users clicking on Join, not completing the form, and then losing track of the banner. I prefer the let the users close the banner manually by using the close button. What do you think?

This PR will probably be in conflict with https://github.com/tournesol-app/tournesol/pull/1436 . Depending on which one we choose to merge first we will either delay the research study or the search feature.

GresilleSiffle:
**to-do**
- [ ] consider the banner hidden by default to avoid a blink effect ; display the banner after the local storage property `displayBanner` has been checked

browser-extension/src/addTournesolRecommendations.js
```diff
@@ -123,6 +123,25 @@ const getLocalizedActionError = () => {
   );
 };
 
+const displayElement = (element) => {
+  const classes = element.className.split(' ');
+
+  if (!classes.includes('displayed')) {
+    classes.push('displayed');
+    element.className = classes.join(' ');
+  }
```
amatissart, line 132: [`classList` ](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList )is what we need: 
Can be simplified into
```suggestion
  element.classList.add('displayed');
```

browser-extension/src/addTournesolRecommendations.js
```diff
@@ -123,6 +123,25 @@ const getLocalizedActionError = () => {
...
+};
+
+const hideElement = (element) => {
+  const classes = element.className.split(' ');
+  const index = classes.indexOf('displayed');
+
+  if (index > -1) {
+    classes.splice(index, 1);
+    element.className = classes.join(' ');
+  }
```
amatissart, line 142: the same here
```suggestion
  element.classList.remove('displayed');
```

browser-extension/src/addTournesolRecommendations.js
```diff
@@ -134,21 +153,34 @@ const getLocalizedActionError = () => {
 const createBanner = () => {
   const banner = document.createElement('div');
   banner.id = 'tournesol_banner';
+  banner.className = 'tournesol_banner';
+
+  // Only display the banner if the user didn't explicitly close it.
+  chrome.storage.local.get('displayBanner', (items) => {
```
amatissart, line 159: If we plan to reuse this mechanism for other banners, a more specific key should be used.
Something like `"displayBannerStudy2023"` ?

browser-extension/src/addTournesolRecommendations.js
```diff
@@ -42,6 +52,241 @@ const getParentComponent = () => {
...
+
+  return false;
+};
+
+const getLocalizedBannerText = () => {
+  if (isNavigatorLang('fr')) {
+    return (
+      'Le projet Tournesol est-il vraiment efficace? Nous étudions' +
+      " actuellement l'impact de notre extension navigateur sur les" +
+      " habitudes d'utilisation de YouTube. Rejoingnez notre étude" +
```
amatissart, line 85: the typo is still present in the extension
```suggestion
      " habitudes d'utilisation de YouTube. Rejoignez notre étude" +
```

browser-extension/src/addTournesolRecommendations.js
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1220/commits/8f5fd25d2a2bc6dbcce3a6b7a8eb6c3461251e53

amatissart (approved):

Issue #1161, opened 2022-10-05
[back] feat: add the possibility to preview a comparison


GresilleSiffle:
### Context

Currently, sharing the link of the front end analysis page of video in an application supporting the Open Graph Protocol displays a nice preview of the entity.

Example, sharing this link https://tournesol.app/entities/yt:1t6hoJNXugM produces this image:

![0ed2b6bf-d579-4e35-90d9-cdc4d0fd8d91](https://user-images.githubusercontent.com/39056254/194068405-85fed176-d5bc-4f5c-8b87-b43b3b60dd46.png)

(You can try it out directly on the production API https://api.tournesol.app/docs/#/preview/preview_entities_retrieve simply use the YouTube video id you want prefixed by `yt:`)

This is done thanks to a back end's route. We would like to create a second endpoint able to return the visual representation of a comparison.

### Expected

...

PR #1226, opened 2022-12-02
[back] feat: add the preview of videos comparisons

sigmike:
Related issue: #1161

Changes:
- Removed a comment to disable an already disabled pylint warning
- Implemented a dichotomic search for text truncation
- Continued the comparison preview
- Made the video preview 2 times bigger

Current result:
image


backend/tournesol/views/preview.py
```diff
@@ -260,14 +313,339 @@ def get(self, request, uid):
...
+            size=logo_size,
+            position=(base_size[0] // 2 - logo_size // 2, entity_description_top),
+        )
+
+        available_width = (
+            base_size[0] // 2
+            - logo_size // 2
+            - entity_descriptions_horizontal_margin * 2
+        )
+        entity_description_height = self.draw_entity_description(
```
GresilleSiffle, line 420: Should we avoid an assignation here, if the name `entity_description_height` is re-assigned another value just after?

backend/tournesol/views/preview.py
```diff
@@ -260,14 +313,339 @@ def get(self, request, uid):
...
+        logo_size = 32
+        margin_above_slider = base_margin * 3
+        margin_below_slider = base_margin
+        vs_band_width = 34
+        vs_tilt_in_pixels = 10
+        text_below_slider = "Which one should be largely recommended?"
+
+        final = self.generate_image(base_size, color=COLOR_WHITE_BACKGROUND)
+
+        thumbnail_size = self.draw_thumbnail(
```
GresilleSiffle, line 369: Should we avoid an assignation here, if the name `thumbnail_size` is re-assigned another value just after?

backend/tournesol/views/preview.py
```diff
@@ -260,14 +313,339 @@ def get(self, request, uid):
...
+
+class DynamicWebsitePreviewComparison(BasePreviewAPIView, APIView):
+    """
+    Return the preview of the Tournesol front end's comparison page.
+    """
+
+    permission_classes = []
+
+    # TODO: set the same cache value as DynamicWebsitePreviewEntity
+    @method_decorator(cache_page_no_i18n(0 * 2))
```
GresilleSiffle, line 624: This file has very slightly evolved on `main` since the beginning of this branch.

Now we have a constant `CACHE_ENTITY_PREVIEW` that we use to define cache of an entity preview. I think we should use this constant too for the comparison preview.

We can merge `main` into this branch, or update this value in the following PR.

GresilleSiffle (approved):
Thanks for the work :ok_hand: the result is very nice.

I made few minor comments, they shouldn't be blocking.

backend/tournesol/views/preview.py
sigmike: I hesitated but I thought it was better to make the 2 instructions identical to make it clear we're doing almost the same thing twice.

It may be better to get the 2 values in different variables and use the the maximum of them. I didn't bother because they return the same value right now and it's likely we will change this section if the sizes ever differ in future changes.

But I agree it's weird and it makes you wonder about the validity of the code so I changed that in 23e421493818307919edeedf40e3b62828de6714.

backend/tournesol/views/preview.py
sigmike: Changed in 23e421493818307919edeedf40e3b62828de6714 too.

backend/tournesol/views/preview.py
sigmike: I merged main and changed this value in c8610e621d8d817a05afd45844e557a45971659f.

backend/tournesol/urls.py
```diff
@@ -185,13 +189,18 @@
         name="polls_score_distribution",
     ),
     # Website Previews
+    path(
+        "preview/comparisons/<str:uid_a>/<str:uid_b>",
```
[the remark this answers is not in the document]
GresilleSiffle: It's true, once again I forgot about it.

The path `<str:uid_a>/<str:uid_b>/` should then be removed from the URL.

What is the expected behaviour if one of the parameter is missing? We could return the default preview.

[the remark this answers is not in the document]
sigmike: I made the change in 815dcb4e61a03baa44865c1a33350432b7121417. I also changed the path because it didn't match the frontend URL either.

Issue #1230, opened 2022-12-07
[back] fix: some thumbnails are not available in `maxres` resolution


GresilleSiffle:
### context

When requesting some comparison previews we saw several videos with a placeholder YT thumbnail instead of their actual thumbnail.

It seems that the requested `maxres` thumbnails in the views are not always available.

- [OK] https://img.youtube.com/vi/nffV2ZuEy_M/mqdefault.jpg
- [OK] https://img.youtube.com/vi/nffV2ZuEy_M/maxresdefault.jpg
- [OK] https://img.youtube.com/vi/vFQYE3aTPLI/mqdefault.jpg
- [placeholder] https://img.youtube.com/vi/vFQYE3aTPLI/maxresdefault.jpg

### proposition

Let's find a solution.

sigmike:
Note that the "hq" and "standard" versions do not have the same aspect ratio, and seem to always have black top and bottom borders. So we must probably use the "mq" version when "maxres" is not available. Or maybe we can crop them to remove the black borders.

If a quick fix is needed you can just switch back to "mq" quality.

PR #1231, opened 2022-12-07
[back] fix: temporarily downgrade the previews quality

GresilleSiffle:
**related to** #1230

---

The images returned by the preview API could be broken if one of the two videos didn't have a `maxres` thumbnail. For now, only `mq` will be use.

In the future, we could check if both videos have `maxres` thumbnails to determine the preview quality.


sigmike (approved):

sigmike:
> In the future, we could check if both videos have maxres thumbnails to determine the preview quality.

The two thumbnails can have a different resolution. They must have the right aspect ratio though.

PR #1233, opened 2022-12-09
`maxres` previews with fallback

sigmike:
Related issue: #1230

For both previews the `maxres` quality is tried first. If it returns a 404 placeholder we try the cropped `hq` version, and if it still fails we return the `mq` thumbnail. If the `mq` thumbnail is still a placeholder we crop it to return the expected aspect ratio.

It may be better to fetch the list of available thumbnails with the YouTube API, but since the `maxres` version seems to exist on most videos this implementation doesn't add an extra request most of the time.

I'm not sure if a test is relevant here. Writing a good one seems to be difficult.

### DIFF
--- a/backend/tournesol/views/preview.py
+++ b/backend/tournesol/views/preview.py
@@ -82,7 +82,9 @@ class BasePreviewAPIView(APIView):
             .resize(size)
         )
 
-    def get_yt_thumbnail(self, entity: Entity, quality="mq") -> Image:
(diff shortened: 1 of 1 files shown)


sigmike:
When a thumbnail doesn't exist the placeholder is very low resolution and looks pretty bad when resized in the previews. We may prefer a custom one, or pad it instead of resizing it.

PR #1251, opened 2022-12-19
[back] feat: the proof of vote API is now a more generic proof API

GresilleSiffle:
This PR enhances the previous proof of vote API to make it more generic and able to provide different kind of proof.

This need comes from our future study about the browser extension. Previously a proof was only provided for users having at least one comparison. This behaviour still exists but the API is now also able to provide a proof of "activated account" for users without comparison.

### details

The proof view now accepts a keyword query parameter identifying the requested proof. It is used as an input by the signing algorithm, making the API able to provide a proof per poll per user per keyword. A keyword can be any string.

For instance, the `presidentielle2022` uses the keyword `proof_of_vote` to generate for each user a proof of participation.

The poll `videos` can now give to each user a proof of "activated account" by using an empty keyword, or the standard `activated` string.

We can use any keyword to generate a proof that ensures a user meets a specific set of expectations, as long as those expectations are coded in the back end.

**to-do: back end**

- [x] `ProofOfVote` API is now `Proof`
  - [x] the default behaviour is a proof of user account
  - [x] add the `keyword` query parameter
  - [x] update the API doc
  - [x] add more tests

**to-do: front end**
- [x] update the API schema
- [x] update the the `<ProofOfVote>` component accordingly

### future

I'll rework the export proof API so that it can work with the keyword parameter. 

We may want to use a model to store the allowed keywords in the database. This will help to keep a list of the different kind of proof that have been issued, the administrators won't have to guess or to remind the keywords. We will also be able to activate / deactivate a specific kind of proof, as we may not want to issue a proof of participation to the research study about the browser extension in 2024 if the study ended in 2023. This will also make Tournesol more usable, as some proof of participation are very specific and shouldn't be commited or hard-coded if possible (like `research_study_2023_browser_ext`).

### DIFF
--- a/backend/tournesol/models/poll.py
+++ b/backend/tournesol/models/poll.py
@@ -83,15 +83,12 @@ class Poll(models.Model):
     def __str__(self) -> str:
         return f'Poll "{self.name}"'
 
-    def get_proof_of_vote(self, user_id: int):
+    def get_user_proof(self, user_id: int, keyword: str):
         """
-        Returns the user_id signed with a signature,
-        derived from the Django SECRET_KEY and the current poll name.
-
-        Should only be provided for the user after they submitted
-        at least 1 comparison in the current poll.
+        Return the user_id signed with a signature, derived from the current
+        poll name and the specified keyword.
         """
-        signer = Signer(salt=f"proof_of_vote:{self.name}")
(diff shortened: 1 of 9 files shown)


GresilleSiffle:
**to-do**
- [ ] fix the export proof of vote API

amatissart (approved):

PR #1252, opened 2022-12-20
[infra] fix: VM provisioning

amatissart:
* "Populate Django DB" task was broken now that the dev-env uses a public database embedeed in a Docker image, et no dump is available in the repository.  
I think it can be removed. If needed, the database can be populated by either loading a backup (that should be tested regularly anyway) or by using the management command "load_public_dataset".

* Workaround a bug in Granafa when initializing admin password

* Update Debian base image version

### DIFF
--- a/infra/ansible/roles/django/handlers/main.yml
+++ b/infra/ansible/roles/django/handlers/main.yml
@@ -45,14 +45,6 @@
   become: true
   become_user: gunicorn
 
-- name: Populate Django DB from Public Dataset
-  shell:
-    cmd: cd /tmp && tar xvf /srv/tournesol-backend/scripts/dataset-import/dump-for-migrations-core-0004-tournesol-0007.sql.tgz && psql -1 -d tournesol < dump.sql && rm dump.sql
-    executable: /usr/bin/bash
-  become: yes
-  become_user: postgres
-  when: populate_django_db_from_public_dataset == True
-
 - name: Create Swagger UI OAuth application in Django database
   shell:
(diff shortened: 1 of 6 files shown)


infra/ansible/roles/django/handlers/main.yml
```diff
@@ -45,14 +45,6 @@
   become: true
   become_user: gunicorn
 
-- name: Populate Django DB from Public Dataset
```
GresilleSiffle, line 48: I got an error while trying this PR, and I realized that this task is still notified by another task in the django recipe. We can remove it from the notify list.



```yaml
# file: `infra/ansible/roles/django/tasks/main.yml`

- name: Copy Django OIDC RSA private key
  template:
    src: django_oidc_rsa_private_key.j2
    dest: /root/django_oidc_rsa_private_key
    mode: u=rw,g=,o=
  notify:
    - Populate Django DB from Public Dataset
    - Create Swagger UI OAuth application in Django database
 ```

**Review by GresilleSiffle — COMMENTED** [review]
Looks good to me :ok_hand: 

There is still a little bug in the Django recipe we should fix before merging.

On my side I also have an error in the monitoring tasks `Create Grafana dashboards`, but I don't know if it's related to this PR (I think no).

```
HTTP Error 401: Unauthorized', 'body': b'{\"message\":\"invalid username or password\",\"traceID\":\"\"}\\n
```

GresilleSiffle:
> On my side I also have an error in the monitoring tasks Create Grafana dashboards, but I don't know if it's related to this PR (I think no).

I found the issue, I've been able to run all monitoring tasks without problem. The Grafana admin password used during the deployment was different than the one defined in the VM. I manually fix the issue by exporting `GRAFANA_ADMIN_PASSWORD`.

GresilleSiffle (approved):

infra/ansible/roles/django/handlers/main.yml
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1252/commits/597cc79643acd5a3d7fd71790f5368ba8fe20fee

GresilleSiffle (approved):

PR #1256, opened 2022-12-21
[back] feat: add individual criteria scores to public export

sigmike:
Related issue: #1071

### DIFF
--- a/backend/tournesol/lib/public_dataset.py
+++ b/backend/tournesol/lib/public_dataset.py
@@ -111,3 +111,29 @@ def get_users_dataset(poll_name: str) -> QuerySet:
         """,
         {"poll_name": poll_name},
     )
+
+
+def get_public_contributor_rating_criteria_scores_dataset(poll_name: str) -> QuerySet:
+    """
+    Retrieve ContributorRatingCriteriaScores made on public comparisons and return a
+    non-evaluated Django `RawQuerySet`.
+    """
+    from tournesol.models import (  # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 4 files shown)


GresilleSiffle:
Thank you @sigmike , this will be very useful for the future data analysis :+1: 

I had an idea while reading the generated SQL query, how about renaming the `voting_rights.csv` file into `individual_criteria_scores.csv` and include in addition to the voting rights the fields `score` and `uncertainty` (and maybe more)?

Motivations: the computed voting rights and the computed individual scores are saved in the same database table, and can fit on the same CSV line. Instead of doing another PR to add the individual scores we could do this right now in this PR.

In the future our target could be to have the following content in dataset archive:
- `individual_criteria_scores.csv`  (output of the algorithm, could replace the current voting rights CSV)
- `collective_criteria_scores.csv` (output of the algorithm, doesn't not exist in the archive yet, out of scope of this PR) 
- `comparisons.csv` (input of the algorithm, already exists)
- `users.csv` (input of the algorithm, already exists)

What do you think @amatissart @aidanjungo @lfaucon @sigmike ?

lfaucon:
- [ ] Update ReadMe of the export (file `export_score.txt`)
- [ ] Rename file `individual_criteria_scores.csv`
- [ ] Include score column in `individual_criteria_scores.csv`

sigmike:
I added the score and renamed the csv. I think it's ready to update the readme and merge, @GresilleSiffle or @lfaucon.

GresilleSiffle (approved):

Issue #209, opened 2021-09-29
docs: Logs on Grafana


lfaucon:
Acceptance criteria:
- Provide a short documentation Grafana functionalities, including which logs are stored and available
- Ensure all logs are available on grafana, if not, document which are missing and why

GresilleSiffle:
**to-do**
- [ ] document how to access logs in case of problem
- [ ] document how to access Grafana, explain what kind of problem it can solve

aidanjungo:
closed by https://github.com/tournesol-app/tournesol/pull/1264

PR #1264, opened 2022-12-29
[infra] update README with instructions to explore logs in Grafana

amatissart:
Related to #209


GresilleSiffle:
That's great, thank you @amatissart !

I wasn't able to read the log of the unit `ml-train.service` on staging and production, every other queries worked but when I use the `unit` filter I'm not able to get any log.

```
{unit="ml-train.service"}
```

GresilleSiffle:
I found the date range buttons, they solve my problem.

GresilleSiffle (approved):

PR #1265, opened 2022-12-29
[back] feat: restore the individual scores export

GresilleSiffle:
I manually re-wrote the SQL query to have more control on it.

Now only the required columns are fetched from the database, which significantly reduce the amount of data read from the disk. The previous query used to fetch all columns of all involved tables (including hashed passwords) even if there were not used by the view.

The cost of this new query (in PostgreSQL arbitrary unit) is `48,233.24` contrary to the previous one which was `111,985.19`. The execution time is more than three time faster and could be even further reduced by sorting the results only by usernames (which can greatly increase the sorting effort required by people using our dataset).

Finally the view doesn't go from one table to another using the ORM, preventing any unexpected additional database queries.

### DIFF
--- a/backend/tournesol/lib/public_dataset.py
+++ b/backend/tournesol/lib/public_dataset.py
@@ -144,7 +144,7 @@ def get_individual_criteria_scores_data(poll_name: str) -> QuerySet:
 
         FROM core_user
 
-        LEFT OUTER JOIN tournesol_contributorrating
+        JOIN tournesol_contributorrating
           ON tournesol_contributorrating.user_id = core_user.id
 
         JOIN tournesol_poll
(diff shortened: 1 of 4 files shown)


backend/tournesol/lib/public_dataset.py
```diff
@@ -124,26 +124,50 @@ def get_individual_criteria_scores_data(poll_name: str) -> QuerySet:
...
+            tournesol_contributorratingcriteriascore.id,
+            core_user.username,
+            tournesol_entity.uid,
+            tournesol_contributorratingcriteriascore.criteria,
+            tournesol_contributorratingcriteriascore.score,
+            tournesol_contributorratingcriteriascore.voting_right
+
+        FROM core_user
+
+        LEFT OUTER JOIN tournesol_contributorrating
```
amatissart, line 147: Out of curiosity, any idea why an outer join is preferred here?  
These row values are required to be present in order to match the following joins, so an inner join should be equivalent here, no?

GresilleSiffle: I'm going to double-check by running the query again. I might have instinctively used an outer join even if it was not required.

GresilleSiffle: I confirm the `OUTER JOIN` has no effect, it's a mistake.

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1265/commits/18b7d67ee688ba33926a38e0f7531470f101274a

amatissart (approved):

PR #1269, opened 2023-01-02
[infra] fix: ensure that discord alert script runs with necessary permissions in Systemd services

amatissart:
When introducing `post-on-discord.sh`  in #979, the script permissions have been restricted as it requires to be run in the Django virtualenv. 

One of the services relying on this script does not use the user "root" or "gunicorn" and needs extra permissions to trigger the alert (when the service fails).


lfaucon (approved):

lfaucon:
How did you discover the issue? Did the back-up fail for any reason?

amatissart:
> How did you discover the issue? Did the back-up fail for any reason?

no problem with the backups. That's just something I realized recently after reviewing #1261

Issue #1247, opened 2022-12-15
Make `/unconnected_entities/`  return results sorted by desc distance if the graph is connected


amatissart:
This is to avoid empty results on the tab "To Connect" when the user comparisons graph is connected.

Maybe this should be implemented as an option on the API, to avoid misleading usage of this route `/users/me/unconnected_entities/...`

PR #1270, opened 2023-01-02
[back] feat: enable non-strict option in unconnected entities API route

lfaucon:
This is a first step for #1247. Next step would be to use this option in the frontend entity selector, but I prefer making this change once the backend route is approved.

### DIFF
--- a/backend/tournesol/views/unconnected_entities.py
+++ b/backend/tournesol/views/unconnected_entities.py
@@ -1,18 +1,38 @@
 """
 API endpoints interacting with unconnected entities.
 """
+import math
 from collections import defaultdict, deque
 
-from django.db.models import Case, Func, OuterRef, Q, Subquery, When
 from django.shortcuts import get_object_or_404
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
 from rest_framework import generics
+from rest_framework.pagination import LimitOffsetPagination
 from rest_framework.permissions import IsAuthenticated
 
 from tournesol.models import Comparison, Entity
 from tournesol.serializers.entity import EntityNoExtraFieldSerializer
 from tournesol.views.mixins.poll import PollScopedViewMixin
+
+
+# We override the pagination filtering for our queryset for performance reasons. We decided to
+# already filter and sort the queryset in python directly because the annotation and sorting
+# operations were inneficient when done on the database.
+class AlreadyFilteredQuerySetLimitOffsetPagination(LimitOffsetPagination):
+
+    hacked_count_updated_by_view = -1
+
+    def paginate_queryset(self, queryset, request, view=None):
+        # Here the parameter queryset is instead the list of ids returned by
+        # UnconnectedEntitiesView.get_queryset. This is done this way for performance.
+        paginated_entity_ids = super().paginate_queryset(queryset, request, view)
+        entity_by_id = {e.id: e for e in Entity.objects.filter(id__in=paginated_entity_ids)}
+        entities = [entity_by_id[e_id] for e_id in queryset if e_id in entity_by_id]
+        return entities
+
+    def get_count(self, queryset):
+        return self.hacked_count_updated_by_view
 
 
 @extend_schema_view(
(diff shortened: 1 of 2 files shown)


backend/tournesol/views/unconnected_entities.py
```diff
@@ -75,7 +87,8 @@ def get_related_entities(entity_id):
         )
 
         return (
-            Entity.objects.filter(id__in=user_all_entities - user_related_entities)
+            Entity.objects.filter(id__in=entity_ids_filter)
+            .annotate(distance=Case(*distance_annotation_whens, default=1+len(neighbors)))
             .annotate(n_comparisons=Subquery(comparison_counts))
-            .order_by("n_comparisons")
+            .order_by("-distance", "n_comparisons")
```
amatissart, line 93: The performance of this queryset seems to be okay actually. The bad performance we observed was probably due to heavy/unoptimized test setup :shrug: 

Locally on the dev-env with 1277 compared videos, I get a response from the API in ~200ms.

To be extra cautious, we could filter this queryset to keep (e.g) only the 1000 entities with the largest distance. That would prevent the SQL query from becoming very large (and possibly very slow) for some users in the future, without the need to re-implement the pagination.

backend/tournesol/views/unconnected_entities.py
```diff
@@ -32,38 +46,36 @@ class UnconnectedEntitiesView(PollScopedViewMixin, generics.ListAPIView):
...
+                    distances[next_node] = distance+1
+                    bfs_queue.append((next_node, distance+1))
 
-        user_related_entities = get_related_entities(source_node.id)
-        user_all_entities = set()
+        # The entities filtered are the entities which don't have a distance
+        # because they are unreachable if the strict option has been left to
+        # True (the default). Otherwise all the entities except the source are
+        # selected.
+        strict = self.request.GET.get("strict") != "false"
```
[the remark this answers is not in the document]
lfaucon: Done in new version

backend/tournesol/views/unconnected_entities.py
```diff
@@ -18,7 +19,20 @@
     get=extend_schema(
         description="List unconnected entities of the current user from a target entity"
                     " and the user's graph of comparisons (entities are vertices and"
-                    " comparisons are edges)."
+                    " comparisons are edges).",
+        parameters=[
+            OpenApiParameter(
+                "strict",
+                type=OpenApiTypes.STR,
```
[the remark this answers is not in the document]
lfaucon: Thanks! Changed in new version

backend/tournesol/views/unconnected_entities.py
lfaucon: I went on an optimized it anyway. Customizing the Paginator class was not too difficult. I also added tests that cheeck that the API call takes < 1s.

lfaucon: Even the 100'000 videos with ~200'000 comparisons passes the test for < 1s api call (although the setup for the test takes 10 minutes (that's surprisingly slow...)

backend/tournesol/views/unconnected_entities.py
```diff
@@ -28,54 +58,40 @@ class UnconnectedEntitiesView(PollScopedViewMixin, generics.ListAPIView):
...
+        # True (the default). Otherwise all the entities except the source are
+        # selected.
+        strict = self.request.query_params.get("strict") != "false"
+        entity_ids = set(e_id for e_id in neighbors if (
+            (e_id not in distances) if strict else (distances.get(e_id, 2) > 1)
+        ))
+        sorted_entity_ids = sorted(
+            entity_ids,
+            # Sorting first by distance and secondly by number of comparisons
+            key=lambda x: (-distances.get(x, math.inf), len(neighbors.get(x, [])))
```
amatissart, line 95: To ensure the sorted list is deterministic when requesting multiple pages:
```suggestion
            key=lambda x: (-distances.get(x, math.inf), len(neighbors.get(x, []), x))
```

amatissart (approved):

backend/tournesol/views/unconnected_entities.py
lfaucon: Fixed. Thanks!

PR #1278, opened 2023-01-05
Auto fill empty entity selectors in comparison

sigmike:
This PR automatically loads entities on the comparison page when they are not already set.

Remaining to do:
- [x] Make sure the comparison tutorial is not affected
- [x] Fix the end-to-end test
- [x] Allow clearing the input? (currently it loads a new video)


lfaucon:
If it's complicated, no need to allow clearing the input. The only downside is that it is rather weird to see a field that the user tries to empty getting filled in again automatically. Apart from this, I don't think we have any reason to let this field be empty, do we?

sigmike:
I thought I fixed the end-to-end test but they failed on a second run. I'm not sure why. They always pass on my system.

@lfaucon I fixed the copy/paste problem you had: nothing is loaded anymore when you clear the input.

sigmike:
I ran the tests 3 times and they passed each time, so it should be good now.

frontend/src/features/entity_selector/AutoEntityButton.tsx
```diff
@@ -20,22 +21,43 @@ const AutoEntityButton = ({
...
+    };
+  }, []);
+
+  const previousUidRef = useRef<string | null | undefined>(undefined);
+  useEffect(() => {
+    // We only want to autofill when the currentUid changes, not when the other dependencies change
+    if (currentUid === previousUidRef.current) return;
+    previousUidRef.current = currentUid;
+
+    if (autoFill && currentUid === null) askNewVideo();
```
lfaucon, line 59: The only problem I found is that it may place twice the same video:
Screenshot 2023-01-30 at 23 03 32

It occurs particularly often with a single (or few) video in the rate later list (this maybe be often the case for users)

Could it be improved with the following behaviour: The component should keep track that the incoming selected uid comes from the autofill mechanism. If it happens to be equal to the changed property "other_id" then autofill is re-triggered?

lfaucon (changes requested):
Thanks! It's nearly perfect now

frontend/src/features/entity_selector/AutoEntityButton.tsx
```diff
@@ -20,22 +21,43 @@ const AutoEntityButton = ({
...
+  useEffect(() => {
+    mountedRef.current = true;
+    return () => {
+      mountedRef.current = false;
+    };
+  }, []);
+
+  const previousUidRef = useRef<string | null | undefined>(undefined);
+  useEffect(() => {
+    // We only want to autofill when the currentUid changes, not when the other dependencies change
```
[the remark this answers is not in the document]
sigmike: No we also want the auto fill to work when the `currentUid` is changed to `null` because that's what happens when you already are on the comparison page with videos and you click again on the "Compare" link. It may be possible to force a unmount/mount in a parent but it's probably not trivial given the current code.

I'll try to make that clearer in the comments.

frontend/src/features/entity_selector/AutoEntityButton.tsx
sigmike: Yes we can probably do that by checking the updated `otherUid` after the `newVideoId` is received (only when `askNewVideo` was called from the auto fill). We will probably need another ref for that because `askNewVideo` can't access the props updated after it was called.

I think it would still be possible to have identical ids loaded when both responses arrives at the exact same time and the `otherUid` has not yet been propagated by a new render. But I think it should be rare or even impossible (it depends on the internals of React).

Another solution could be to move this autofill mechanism up to a component having both uids and having a special process to request the ids when both are null.

I'll try the `otherUid` method.

frontend/src/features/entity_selector/AutoEntityButton.tsx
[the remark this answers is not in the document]
sigmike: I did that in b24912325a845c39b15a77410ca96a113cfa363d.

frontend/src/features/entity_selector/AutoEntityButton.tsx
sigmike: Done in 89b60737589832c92c2f89621f556f6a2d1d6bf2.

sigmike:
The auto fill broke the back button when we click on Compare. The back button cleared the input and triggered a new auto fill that would add a new entry to the history, so moving to the previous page we visited was difficult.

I changed the behavior when an ID is changed on the comparison page: it replaces the history entry instead of adding a new one. It fixes the issue but it also changes the behavior when we change the ID manually (by clicking on "Auto" or by typing). It believe it should have been like that anyway. Do you agree?

I did that in 9911931ef3ff264395b8f0590359e5b4fe628376.

lfaucon:
> The auto fill broke the back button when we click on Compare. The back button cleared the input and triggered a new auto fill that would add a new entry to the history, so moving to the previous page we visited was difficult.
> 
> I changed the behavior when an ID is changed on the comparison page: it replaces the history entry instead of adding a new one. It fixes the issue but it also changes the behavior when we change the ID manually (by clicking on "Auto" or by typing). It believe it should have been like that anyway. Do you agree?
> 
> I did that in [9911931](https://github.com/tournesol-app/tournesol/commit/9911931ef3ff264395b8f0590359e5b4fe628376).

Good catch! It's typically something we miss regularly. I'm fine with the change :+1:

lfaucon (approved):
@GresilleSiffle can you double check and ship it?

GresilleSiffle:
Yes. I'm currently testing this branch.

GresilleSiffle (approved):
Looks good to me :+1:

Issue #1274, opened 2023-01-03
Rework comparison flow on mobile


amatissart:
Suggestions that have been made to make the UI easier to use: 
 * scroll criteria separately from videos
 * compare criteria one by one

amatissart:
Suggestions made by @aidanjungo on Discord

![image](https://github.com/tournesol-app/tournesol/assets/4726554/7618bf8c-c1db-4078-a734-9fc21a76702c)

* Keep thumbnails at the top of the screen when scrolling
* Move "Auto" buttons at the bottom of videos
* Evaluate criteria one by one
* Remove sliders and use scale with 3 or 5 buttons instead
* Swipe up and down to navigate between criteria

GresilleSiffle:
Related issues:
- https://github.com/tournesol-app/tournesol/issues/1931

Issue #1240, opened 2022-12-14
[front] style: making comparison on mobile devices less of a painful experience


GresilleSiffle:
### Context

The main problem identified is the scrolling behaviour. When a user tries to scroll the sliding movement will often be grabbed by the slider inputs instead of being interpreted as a page scroll. In those cases the slider handle will move, potentially provoking the loss of an already selected rating.

We may discover other challenges after having addressed this one.

### Proposition

This issue can be used to identify potential "quick wins" that will fix or at least greatly ease the comparison on mobile devices. 

Few random ideas:
- See if there are options in the MUI sliders documentation that can help us.
- Increase the gap between sliders on mobile, to increase the chance of touching the background instead of an input field (this will probably be not enough to fix the problem).
- Display a visible vertical scroll bar? (Just in case nothing else worked, we could precisely display the scroll-able surface).
- Any other ideas are welcome.

sigmike:
> See if there are options in the MUI sliders documentation that can help us.

I didn't find any option that would help.

I found a possibly related npm package: https://www.npmjs.com/package/mui-scrollable-slider-hook. I tried it but it doesn't seem to solve the problem. It was made by the author of an anwser in https://stackoverflow.com/questions/64776560/scroll-events-unintentionally-changing-material-ui-slider-values.

I also found this issue that seems related: https://github.com/mui/material-ui/issues/20990. The behaviors described in [this comment](https://github.com/mui/material-ui/issues/20990#issuecomment-627246200) sound good, but it looks like they haven't been implemented yet.

GresilleSiffle:
A more robust fix will be implemented by https://github.com/tournesol-app/tournesol/issues/1274

If https://github.com/tournesol-app/tournesol/issues/1274 is addressed with a satisfying solution, this issue may be closed.

lfaucon:
Improvements to the scrolling improved a lot the situation (see #1279). Closing this ticket but keeping #1274 for later.

PR #1279, opened 2023-01-07
[front] fix: disable pointer events on slider rails on touchscreens

amatissart:
Following the discussion in #1240, I realized this could be a quick fix, before the future comparison redesign on mobile (#1274)

This change would make the slider handle the only interactive part on sliders when using a touch screen (or any pointing device of limited accuracy covered by `pointer: coarse`).

Note that you need to emulate a mobile device to test this change on a desktop browser.

[Capture vidéo du 2023-01-07 11-42-52.webm](https://user-images.githubusercontent.com/4726554/211146921-7d9335d5-91d5-4043-b234-3111e34ed484.webm)

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -55,6 +55,10 @@ html.embedded .SnackbarContainer-root.belowTopBar {
   }
   .MuiSlider-root .MuiSlider-thumb {
     pointer-events: auto;
+    /* `touch-action: none` is used internally by MUI on Slider to disable
+    scrolling and zooming on touch events. But it conflicts with
+    `pointer-events: none` on Firefox. So it's required here too. */
+    touch-action: none;
   }
 }
 
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -45,4 +45,17 @@ html.embedded .SnackbarContainer-root.belowTopBar {
   top: 4px;
 }
 
+
+/* pkg: @mui/material */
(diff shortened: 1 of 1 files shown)


sigmike:
I tested on Firefox mobile and it works well but there are some side effects:
- we can't "click" on the bar anymore to move the slider to that position, we have to move the handle
- when we touch the handle and move diagonally it scrolls in addition to moving the handle

The change still looks better than the current situation but I don't use the site on mobile so I really don't know.

amatissart:
@sigmike thanks for testing!

>    * we can't "click" on the bar anymore to move the slider to that position, we have to move the handle

Yes, definitely not ideal, but that's expected due to the change.

>    * when we touch the handle and move diagonally it scrolls in addition to moving the handle

That's more surprising, and the behavior on Firefox and Chrome seems to differ. The `SliderRoot` uses `touch-action: none` to prevent scrolling. And it apparently conflicts with `pointer-events` in this specific situation on Firefox. Repeating `touch-action: none` on the inner `.MuiSlider-thumb` fixes this. I will add the change.

lfaucon (approved):
Nice! Thanks

I'm surprised that this change cannot be done with arguments/props on the Slider component? 

Off-topic: I looked again at the MUI Slider and I found this one: https://mui.com/material-ui/react-slider/#minimum-distance which I had not seen before. Would be my proposition for enable to express uncertainty with the sliders.

PR #1282, opened 2023-01-10
[dev-env] fix: postgres container could be healthy before the database is ready

amatissart:
`pg_isready` is not enough to check that the "database" itself has been created and is ready:
https://dba.stackexchange.com/a/298647


lfaucon (changes requested):
Tested twice on my machine and it fails with the same issue that we have seen together tonight.

lfaucon (approved):
Latest changes work for me. Thanks!

PR #1285, opened 2023-01-16
[infra] fix: update the grafana rapository URL

GresilleSiffle:
The Grafana packages are served from a different apt repository and our Ansible configuration must be updated. See the note in the [grafana documentation](https://grafana.com/docs/grafana/latest/setup-grafana/installation/debian/#repository-migration-november-8th-2022).

After this change is merged, and before deploying on your local VM,  the previous Grafana apt repository configuration must be manually removed from `/etc/apt/sources.list.d/`. The staging and production servers are already updated.


lfaucon (approved):
Thanks for the fix

PR #1286, opened 2023-01-17
[front] fix: use Playlist YouTube instead of YT Playlist in french

GresilleSiffle:
This PR changes `YouTube Playlist` for `Playlist YouTube` in the french translation. It sounds more natural to have the main noun first in a mot-valise.

We can also think about the `Twitter Bot`. Should we use `Bot Twitter` in french? In my personal opinion yes, to be coherent with the YouTube Playlist. The contraction `Bot` (from the French and English robot) being more common than `agent` or `robogiciel` I think `Bot Twitter` is an acceptable translation.


aidanjungo:
For the bot I would say we can let it like that, it is also how we call in on twitter.

aidanjungo (approved):

Issue #1293, opened 2023-01-20
[back] feat: create an API to interact with the user settings


GresilleSiffle:
### technical specifications

The users' settings are JSON objects representing the users' preferences. Each key is a setting used by the back end or the front to modify certain of their behaviours. A user has exactly one set of personal settings.

Example:
```json
{
  "frontend.alpha_ui": false,
  "frontend.beta_ui":  true,
  "rate_later.auto_remove": 4
}
``` 

(The namespaces are here to make the settings more understandable, some of them are front end only, some of them might be related to a specific part of the API, etc.)

**Q1 Where to store the settings?**

The settings can be either saved as a new `JSONField` in the currnet `User` model, or in a related one-to-one model. For performance reason I propose to directly alter the current `User` model. It may also ease retrieving the settings during the log in process.

This field could could be `blank=True`, `null=False` and `default=dict`.

**Q2 Where to store the settings serializers and API?**

Short answer: saving everything in `core` is not perfect but looks simple and good enough.

The API and its related tests could go in the folder `core`, because its behaviour will be rather generic whatever the content of the settings: receiving a JSON input, using a serializer to validate the data and saving them, etc.

Concerning the serializer(s) it's a bit different. I think it depends on how independent and re-usable we want the API and the front end to be. It depends on how easy we want for a fork of Tournesol to re-use the back with a completely new UI with its own set of user settings.

Some settings will be used only by the front end, making them specific to the UI and useless for our API. If we want to avoid writing the Tournesol front end's specific keys in the supposedly generic folder `core` of our API, maybe we would like to move both the API definition and its serializer(s) elsewhere, like in the `tournesol` folder. Nevertheless this is also a very imperfect solution. Without a proper way to configure the settings in the back end without hard coding them in, it seems like choosing the right folder to store the specific serializer(s) just moves the problem from one place to another.

**to-do**

Create the routes
- [x] `GET /users/me/settings`
- [x] `PUT /users/me/settings`
  - replace the complete user settings by the new JSON provided 
- [x] `PATCH /users/me/settings`
  - replace only the provided keys

Check how the API can return the user settings after a login (it shouldn't be complex if the settings are a field of the `User` model).

GresilleSiffle:
Here is the current path I'm following. The objective is to:
- allow the users to have different settings per poll
- allow the polls to have their own specific settings
- allow the platform to have global poll-agnostic settings

```json
{
  "common": {
    "email_notification__platform_updates": true,
[...]

PR #1296, opened 2023-01-23
WIP / [back] feat: add user preferences in the User model

GresilleSiffle:
**related to** #1293

---


backend/core/models/user.py
```diff
@@ -202,6 +202,13 @@ class User(AbstractUser):
                   " domains and the vouching mechanism.",
     )
 
+    settings = models.JSONField(
+        null=False,
+        blank=True,
+        default=dict,
+        help_text=_("The user' preferences.")
+    )
```
amatissart, line 210: What are the advantages of using a `JSONField` to store user settings? I haven't thought too much about this, but I feel like individual fields (maybe even in a separate model) would be easier to maintain. For example, when a new setting is required, its default value would be tracked automatically by Django migrations.

GresilleSiffle: Yes it's true that a `JSONField` forces us to write manual data migrations if we want to set a default value for a new setting. With one model field per setting it's automatic.

Nevertheless I considered the `JSONField` because I think it can help us to easily solve several "complex" problems.

The main problem: some settings can be poll specific, and not all polls have the same settings. If I create a single table to store the settings, let's say with a poll foreign key to not force every poll to behave the same way, how do we store settings that are specific to only the `videos` and not to the other? (I think this problem has a precise name, is it polymorphic table?)

Let's imagine a setting `recommendations__max_duration` that defines the maximum duration of videos displayed on the recommendations page. I see two ways to add this setting:

**(a)** if I add a column `recommendations__max_duration` in the dedicated settings table, all records will have this column, even those that don't have this concept of max duration. The value of this column will probably be null for those records, and the table will grow "horizontally" each time we add a new setting. I personally don't like tables like this because it implies that records can store values that have no meaning for them ; and because the records don't represent a clear concept anymore, but rather an aggregation or possibly unrelated ones.

**(b)** I could add instead a new column named `specific` to store a JSON of all settings exclusively related to the poll. This hybrid solution has some settings stored in columns with regular data type and some settings stored in a `JSONField`. It's better than the solution **(a)** but I think it adds extra work without bringing more comfort: we need to do SQL joins to get the settings when we have a user instance, and depending on the poll we need to use a specific serializer to read the settings.

The advantage of a single `JSONField` :
- less code
- no SQL join is required, the settings are available each time a user instance is available
- the polls can have specific settings, adding new one doesn't require to update a table shape
- some settings can be poll agnostic, like `email_notification__platform_updates`

And the main disadvantage is that we will have to manually manage the default values of settings.

amatissart: Ok, if we need to store settings that are specific to a given poll, a JSONField is indeed the best way.

amatissart (approved):

PR #1311, opened 2023-01-26
[back] feat: Generate and tweet montly Top contributor figure

aidanjungo:
- [x] modif exiting test
- [x] add test for `tweet_top_contributor_graph`
- [x] add a django management command


GresilleSiffle:
Before merging:

We could move the Tournesol logo in a resources folder in the `twitterbot`. We already have a logo in the `tournesol` app, should we re-use it?

We should precisely set the `matplotlib` version in the requirements to keep our build deterministic

aidanjungo:
> We could move the Tournesol logo in a resources folder in the `twitterbot`. We already have a logo in the `tournesol` app, should we re-use it?

Yes it is in my TODO to move it, but I think i added it because we did not have it yet with the good format. I will check.


> We should precisely set the `matplotlib` version in the requirements to keep our build deterministic

Done

backend/twitterbot/settings.py
```diff
@@ -27,6 +27,28 @@
     ),
 }
 
+# Monthly top contributor text template
+graph_title_text_template = {
+    "en": ("Who compared the most on Tournesol in {month_name} {year}"),
+    "fr": ("Qui a le plus comparé sur Tournesol en {month_name} {year}"),
```
amatissart, line 33:
```suggestion
    "en": ("Who compared the most on Tournesol in {month_name} {year}?"),
    "fr": ("Qui a le plus comparé sur Tournesol en {month_name} {year} ?"),
```

backend/twitterbot/management/commands/tweet_top_contribors.py
```diff
@@ -0,0 +1,29 @@
...
+    help = "Runs Tournesol Twitter Bot to tweet Top contributor figure"
+
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            "-n",
+            "--bot-name",
+            type=str,
+            help="Name of the bot '@TournesolBot' or '@TournesolBotFR'",
+        )
```
amatissart, line 16: Would make sense to make this argument required

```suggestion
        parser.add_argument(
            "-n",
            "--bot-name",
            type=str,
            required=True,
            help="Name of the bot '@TournesolBot' or '@TournesolBotFR'",
        )
```

amatissart (approved):

Issue #1229, opened 2022-12-07
[back] feat: display the video duration in entity preview and comparison preview


GresilleSiffle:
### context

The duration is an important metric that invites (or prevents) users to watch a video.

Thank to the dynamic preview, when we share a entity or a comparison link we are already able to see the thumbnails and the videos' title, but not the duration yet.

The same way we display the duration in the front end, we could display the duration in the dynamic previews.

![capture](https://user-images.githubusercontent.com/39056254/206126158-7b376563-7805-4320-aa86-3b8e399ecaf9.png)

(It could be nice to copy the front end's style, back ground color and transparency, polygon size, etc.)

tbrnd:
hi, is someone working on this? if not I'd like to implement it!

aidanjungo:
Hello, yes you can take care this issue, I assign you to it 🙂

PR #1312, opened 2023-01-27
[back] feat: add video duration in entity preview

tbrnd:
Hi everyone, 

started working on my first contribution with #1229, currently WIP so do not pay attention to the code itself.

Here are the results I get with two random videos (<60 minutes and >60 minutes):

image
image

I tried to mimic the front-end as much as I can, for the font, I opted for `Roboto-Bold` since the backend does not have a native `arial sans-serif` [as in the front-end](https://github.com/tournesol-app/tournesol/blob/2512e03efe3b2dc3e2efc6191885235dadbdf536/frontend/src/components/entity/EntityImagery.tsx#L43).

(I also noticed that `Poppins` the font for the social preview does not support unicode) 

Comments/suggestions are welcome, is there anything else I should consider before moving forward?


GresilleSiffle:
Hello, nice to see you here @tbrnd 

Ping us when your work is ready :+1:

tbrnd:
> Hello, nice to see you here @tbrnd
> 
> Ping us when your work is ready 👍

Great, just wanted to make sure I was on the right track feature-wise.

Also, would you also expect the duration on the comparison preview or that would be too crowded?

image

aidanjungo:
Thank you it looks very nice on the preview. We think if is maybe not useful to have the duration on the comparaison preview.

lfaucon:
> Also, would you also expect the duration on the comparison preview or that would be too crowded?

If it takes time, we could do this in a separate PR or later. The most important is as you did the video preview! It's good to also have it for hte comparison preview. One idea would be to show the duration in the bottom-left corner for the left video to avoid overlaping with the center slanted yellow rectangle.

Thanks for this feature 👍

tbrnd:
hi everyone, I refactored/linted/cleaned, looking for your feedback!

backend/tournesol/views/preview.py
```diff
@@ -40,7 +43,9 @@
 COLOR_YELLOW_BACKGROUND = (255, 200, 0, 16)
 COLOR_WHITE_BACKGROUND = (255, 250, 230, 255)
 COLOR_BROWN_FONT = (29, 26, 20, 255)
+COLOR_WHITE_FONT = (255, 255, 255, 255)
 COLOR_NEGATIVE_SCORE = (128, 128, 128, 248)
+COLOR_DURATION_RECTANGLE = (0, 0, 0, 128)
```
lfaucon, line 48: The background used on YouTube is `rgba(0, 0, 0, 0.8)`. I think it is the same as `(0,0,0,201)` ? Which should be darker

backend/tournesol/views/preview.py
```diff
@@ -329,6 +337,30 @@ def _draw_logo(self, image: Image, entity: Entity, upscale_ratio: int):
...
+    def _draw_duration(self, image: Image, entity: Entity, fnt_config, upscale_ratio: int):
+        """
+        Draw the duration on the preview.
+        """
+
+        duration = entity.metadata.get("duration", 0)
+        minutes, seconds = divmod(duration, 60)
+        hours, minutes = divmod(minutes, 60)
+
+        overlay = Image.new('RGBA', ENTITY_DURATION_WH, COLOR_DURATION_RECTANGLE)
```
lfaucon, line 349: I find that the grey box is sometimes a bit too large.

I think it should be possible to adjust the width of the overlay dynamically based on the text length (See https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.textlength) ?

backend/tournesol/views/preview.py
tbrnd: ok! for my understanding, where do you get this value `0.8`? My reference was this https://github.com/tournesol-app/tournesol/blob/2512e03efe3b2dc3e2efc6191885235dadbdf536/frontend/src/components/entity/EntityImagery.tsx#L41 but I may not be in the right code.

tbrnd:
@lfaucon thanks for the comments, I actually ended up rewriting the management of position/size to dynamically adapt the duration overlay in function of its text size. This will be useful if we later want to increase/decrease its size.

Here are some samples:

image

image

Feedback is welcome!

backend/tournesol/views/preview.py
sigmike: `0.8` is the value used on YouTube, but the design is quite different:

image

We should probably use the same style as the frontend, and maybe later update both to match YouTube's style.

backend/tournesol/views/preview.py
```diff
@@ -356,13 +405,16 @@ def get(self, request, uid):
         except ConnectionError:
             return self.default_preview()
 
-        youtube_thumbnail = youtube_thumbnail.resize(
-            numpy.multiply((320, 180), upscale_ratio)
-        )
+        # (width, height, left, top)
```
sigmike, line 408: Is it common to store bounding boxes like that? I would have put the position first. But anyway I think using a bbox to store these value makes the code more difficult to read an maintain. You always have to remember where each value is in the tuple, and expressions like `youtube_thumbnail_bbox[2:4]` are not easy to interpret for us humans. I would rather have a position and a size.

And if we keep a bounding box I would rather use Pillow's convention which seems to be `[left, top, right, bottom]` ([[1]](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.textbbox), [[2]](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.rectangle)) since that's the tool we use to draw things.

sigmike (approved):
I made some suggestions but it works as it is.

lfaucon (approved):

tbrnd:
@sigmike thanks for the feedback, I'll address these in a new PR, especially for the third item which is challenging (I haven't figured out how to retrieve a correct height value with PIL yet)

sigmike:
> (I haven't figured out how to retrieve a correct height value with PIL yet)

@tbrnd [textbbox](https://pillow.readthedocs.io/en/stable/reference/ImageDraw.html#PIL.ImageDraw.ImageDraw.textbbox) may work here. Otherwise using `font_height` is good enough I think.

PR #1317, opened 2023-01-28
clean-up: remove unused fields on users

lfaucon:
Yearly clean-up. The data saved on these fields is incorrect and we don't have any plan to collect it any way.


backend/core/models/user.py
```diff
@@ -34,166 +34,13 @@ class User(AbstractUser):
     LOGIN_FIELDS = ("username", "email")
 
     email = models.EmailField(_("email address"), unique=True)
-    is_demo = models.BooleanField(default=False, help_text="Is a demo account?")
```
lfaucon, line 37: Maybe we could delete all these accounts before removing the field?

lfaucon: Note, that I'm not sure the field is set to True for any account and it may be set to True mistakenly.

amatissart: I confirm it's not used in production data

amatissart (approved):

PR #1320, opened 2023-01-29
[front] feat: set 'autocomplete' attr on login and signup forms

amatissart:
I am not sure about how to test this properly. On my environment, it seems that browsers suggest auto-completion options automatically. The browsers may also behave differently on "localhost". 

@lfaucon Could you have a look with your browser?


lfaucon (approved):
Works for me. Thanks!

Issue #1299, opened 2023-01-24
[back] Comparison export should include week-timestamp


lfaucon:
We want to include timestamp at the granularity of weeks such that it reveals no sensitive user data into the comparison export.

- [x] make sure our privacy policy is ok with this
- [x] The dataset should not contain data from the current week
- [x] The dataset should be sorted by username>timestamp

tbrnd:
hi @lfaucon couple of questions:

- for the privacy policy, I do not see any conflict with https://tournesol.app/about/privacy_policy but feedback is welcome
- there are two timestamps recorded: datetime_add and datetime_lastedit, which one do you expect?
- can you clarify the last item? what's the difference between week and timestamp?

lfaucon:
> * for the privacy policy, I do not see any conflict with https://tournesol.app/about/privacy_policy but feedback is welcome

All good ☺️ thanks 

> * there are two timestamps recorded: datetime_add and datetime_lastedit, which one do you expect?

Let's use the `datetime_add`, maybe also mention it in the document attached with the dataset.

> * can you clarify the last item? what's the difference between week and timestamp?

`week` is the Monday at 00:00 before the comparison was done, but we still want to be sorting by the true timestamp even if it is not included in the export.

tbrnd:
@lfaucon I think we can close this one!

lfaucon:
Yes! Thanks for your work

PR #1324, opened 2023-02-01
[back] feat: add week_date in public export

tbrnd:
Opened this PR to address #1299


backend/tournesol/lib/public_dataset.py
```diff
@@ -26,7 +26,8 @@ def get_comparisons_data(poll_name: str) -> QuerySet:
             entity_2.uid AS uid_b,
             comparisoncriteriascore.criteria,
             comparisoncriteriascore.weight,
-            comparisoncriteriascore.score
+            comparisoncriteriascore.score,
+            EXTRACT('epoch' FROM DATE_TRUNC('week', datetime_add)) AS week_timestamp
```
tbrnd, line 30: @lfaucon is it what you expect? for a video I uploaded on my local setup this week, it returns `1674432000.0` which stands for `Monday, January 23, 2023 12:00:00 AM`

GresilleSiffle: I think we can aim for a more human readable value.

As defined by [ISO 8601](https://fr.wikipedia.org/wiki/ISO_8601) we can represent the date by the string `yyyy-mm-dd`.

The following line should do the trick:

```sql
DATE(datetime_add) AS week

-- or
DATE(datetime_add) AS comparison_week
```

lfaucon: > @lfaucon is it what you expect? for a video I uploaded on my local setup this week, it returns `1674432000.0` which stands for `Monday, January 23, 2023 12:00:00 AM`

Weird, right now is the week of Monday Jan 30, 2023 right?

tbrnd: my bad, the comparisons were done last week and not this week so it makes sense

@GresilleSiffle don't we want to hide the day to preserve privacy? Let's say I add a comparison today (2023-02-01), we would expect `2023-01-30` in the export, right?

lfaucon: > @GresilleSiffle don't we want to hide the day to preserve privacy? Let's say I add a comparison today (2023-02-01), we would expect `2023-01-30` in the export, right?

Correct

GresilleSiffle: Sorry I didn't express myself correctly =/

I thought that using `DATE` could help to convert a timestamp to a human readable date, and yes using `datetime_add` as argument is not correct as it will leak the real comparison date.

tbrnd:
> @GresilleSiffle Should we add new tests for this route?
> 
> New behaviour to be tested:
> 
> * Exclusion of most recent comparisons
> * Ordering
> * Values in the new `week_date` column

This is something I wanted to ask, I haven't found any unit testing related to this section of the code except [this throttling test](https://github.com/tournesol-app/tournesol/blob/3521ec43bc30c2faf8039c287ccc8f5ecb0e6dde/backend/tournesol/tests/test_throttling.py#L26), so yes, as the export is getting more complex, I think I'll create unit tests for these behaviours.

lfaucon (approved):

backend/tournesol/tests/test_exports.py
```diff
@@ -434,3 +445,55 @@ def test_all_export_sorts_by_username(self):
...
+            self.add_comparison(user=last_user, is_public=True)
+
+        # comparisons#2, the same week on a Sunday
+        with MockNow.Context(datetime(2023, 1, 22, tzinfo=timezone.utc)) as mock_date_2:
+            self.add_comparison(user=first_user, is_public=True)
+            self.add_comparison(user=last_user, is_public=True)
+
+        # comparisons#3. random date in current week
+        random_date_this_week = datetime.now()
+        random_date_this_week = random_date_this_week - timedelta(random_date_this_week.weekday()) + timedelta(random.randint(0, 6))
```
aidanjungo, line 465: I think as you mock a Sunday as now it simplify a bit and you could use the function `time_ahead` that I added recently, just need to import:

```python
from core.utils.time import time_ahead
```

```suggestion
        with MockNow.Context(datetime(2023, 1, 22, tzinfo=timezone.utc)) as mock_date_2:
            self.add_comparison(user=first_user, is_public=True)
            self.add_comparison(user=last_user, is_public=True)

        # comparisons#3. random date in current week
        random_date_this_week = time_ahead(day=random.randint(0, 6))
```

In fact I am not just MockNow will work in this case...

backend/tournesol/tests/test_exports.py
```diff
@@ -434,3 +445,55 @@ def test_all_export_sorts_by_username(self):
...
+
+        response = self.client.get("/exports/all/")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.headers["Content-Type"], "application/zip")
+        root = self.extract_export_root(response)
+
+        zip_content = io.BytesIO(response.content)
+
+        with zipfile.ZipFile(zip_content, "r") as zip_file:
+            with zip_file.open(root + "/comparisons.csv", "r") as file:
```
aidanjungo, line 479: Maybe consider using Pathlib

backend/tournesol/tests/utils/mock_now.py
```diff
@@ -0,0 +1,30 @@
+from contextlib import contextmanager
```
aidanjungo, line 1: should it be in a more general directory? @GresilleSiffle what do you think?
Should we have a test for this function?

backend/tournesol/tests/test_exports.py
tbrnd: @aidanjungo I simply copied-pasted from another test 😅

backend/tournesol/tests/test_exports.py
tbrnd: @aidanjungo yes I intentionally create two comparisons one on Monday and one on Sunday and make sure both are "truncated" within the same week.

Your `time_ahead` function is relevant, I'll give it a try!

backend/tournesol/tests/test_exports.py
lfaucon: I'm touching these same tests in another PR. Best to leave it like this for now I think

aidanjungo: Yes I will do another PR to change that everywhere

backend/tournesol/views/exports.py
```diff
@@ -31,7 +31,7 @@ def write_logged_user_comparisons_file(request, write_target):
 
     Comparisons from all polls are included.
     """
-    fieldnames = ["video_a", "video_b", "criteria", "weight", "score"]
+    fieldnames = ["video_a", "video_b", "criteria", "weight", "score", "week_date"]
```
amatissart, line 34: This new column "week_date" is currently empty in the personal export. It should either be removed, or implemented below.

I don't think this has been specified yet. So I would remove it for now. And as there is no privacy concern in this personal dataset, so the exact datetime could be used here, in a future PR.

backend/tournesol/lib/public_dataset.py
```diff
@@ -69,6 +70,10 @@ def get_comparisons_data(poll_name: str) -> QuerySet:
           -- keep only public ratings
           AND rating_1.is_public = true
           AND rating_2.is_public = true
+          -- excluse current week comparisons
+          AND datetime_add < DATE_TRUNC('week', now())
```
amatissart, line 74: This query looks good to me.

Let's just keep in mind that the dataset is currently scheduled to be generated on Sundays at 23:55 UTC. 
https://github.com/tournesol-app/tournesol/blob/56f19a59bfcecd54865c5d3d060fa3efd846c6e8/infra/ansible/inventory.yml#L193

I think that @GresilleSiffle's intention was to not include any comparison from the new week, while avoiding a conflict with the backup script running on 00:00. Now that the SQL query excludes comparisons submitted in the current week, it should be okay to generate the dataset on Mondays, let's say at 00:02.

(we can address the change in inventory.yaml separately)

backend/tournesol/views/exports.py
tbrnd: done, thanks!

backend/tournesol/lib/public_dataset.py
tbrnd: makes sense, changed the yaml in my latest commit https://github.com/tournesol-app/tournesol/pull/1324/commits/16097207e2ad5054f858c12ee5f5ecbffc55c7c4 let me know if this is how you want to proceed

amatissart (approved):

backend/tournesol/lib/public_dataset.py
GresilleSiffle: > I think that @GresilleSiffle's intention was to not include any comparison from the new week, while avoiding a conflict with the backup script running on 00:00. Now that the SQL query excludes comparisons submitted in the current week, it should be okay to generate the dataset on Mondays, let's say at 00:02.

Exactly, scheduling the dataset creation before Monday was a way to workaround the absence of filter by date in the public dataset lib.

I thought about improving the lib to make the functions able to receive the `until` date as a parameter, to make it more flexible, but excluding the comparison directly in SQL is also good for me.

backend/tournesol/tests/utils/mock_now.py
GresilleSiffle: > should it be in a more general directory? @GresilleSiffle what do you think?

Hard to say. It looks good to me here, in the `tests` folder of the main Django app, the same way it will look good in the `tests` folder of the core app.

If we want the `tests` folders to only contain tests we could choose to move this class to `tournesol.lib.test` (singular, to not be confused with a folder containing actual tests). This file is very small so I don't mind thinking about the architecture after the merge.

> Should we have a test for this function?

I'm always in favor of adding unit tests in the back end : ) It's ok for me if we add tests after the merge. We can do it together if you want @aidanjungo

Issue #1294, opened 2023-01-20
[back] refactor: add a `max` parameter to `rate_later__auto_remove`


GresilleSiffle:
**blocked by** #1294

The method `rate_later__auto_remove` must accept a new parameter `max` to replace the currently hard coded value `4` in:

```python
if n_comparisons >= 4:
    RateLater.objects.filter(
        poll=poll, user=user, entity=self
    ).delete()
```

The views calling `rate_later__auto_remove` must fill `max` according to the user settings. They should already have access to the logged user model with all its fields.

PR #1325, opened 2023-02-01
[back] feat: the rate-later auto remove is now connected to the user settings

GresilleSiffle:
**related to** #1294 

---

This PR makes the comparisons API use the `rate_later__auto_remove` setting.

If the setting is not defined in the requested poll, the default value defined in the `RateLater` model is used, currently `4`. If the setting is defined, its value is used instead.

The new tests of the comparisons API ensure the API call `auto_remove_from_rate_later` with the user setting or the default value in case of need.

The new tests of the rate-later API ensure the `Entity.auto_remove_from_rate_later` behaves as expected (note: in the future these tests could be moved in the entity unit tests, as they test a behaviour of the `Entity` model, not a behaviour of the rate-later API).

**to-do**
- [x] connect the auto remove feature to the user settings
- [x] define the default auto remove setting as a constant
- [x] update the tests
- [x] add the missing tests

**next steps**

- check if the settings are returned after an authentication
- ensure they are saved in the front end's redux store
- create the update form in the front end

### DIFF
--- a/backend/tournesol/models/entity.py
+++ b/backend/tournesol/models/entity.py
@@ -24,7 +24,7 @@ from tournesol.entities import ENTITY_TYPE_CHOICES, ENTITY_TYPE_NAME_TO_CLASS
 from tournesol.entities.base import UID_DELIMITER, EntityType
 from tournesol.entities.video import TYPE_VIDEO, YOUTUBE_UID_NAMESPACE
 from tournesol.models.entity_score import EntityCriteriaScore, ScoreMode
-from tournesol.models.rate_later import RateLater
+from tournesol.models.rate_later import RATE_LATER_AUTO_REMOVE_DEFAULT, RateLater
 from tournesol.serializers.metadata import VideoMetadata
 from tournesol.utils.constants import MEHESTAN_MAX_SCALED_SCORE
 from tournesol.utils.video_language import (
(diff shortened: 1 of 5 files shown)


backend/tournesol/models/entity.py
```diff
@@ -200,10 +200,10 @@ def update_n_ratings(self):
         )
         self.save(update_fields=["rating_n_ratings", "rating_n_contributors"])
 
-    def auto_remove_from_rate_later(self, poll, user) -> None:
+    def auto_remove_from_rate_later(self, poll, user, max_) -> None:
```
lfaucon, line 203: `max_` is not a very helpful name. Why not `auto_remove_threshold`?

lfaucon (approved):
Looking forward to lower my setting to 3 and empty my rate later list

backend/tournesol/models/entity.py
GresilleSiffle: > max_ is not a very helpful name. Why not auto_remove_threshold?

`auto_remove_threshold` and `threshold` are indeed better, thanks for the suggestion.

> Second small question: Why reading the settings value outside of the method and passing it as argument rather than reading it from inside the method?

It's an habit, I try to write more or less pure methods to make them easily testable. But in our case, there is no benefit of using any additional parameter indeed.

lfaucon: I'm fine with any decision to that regard. Already approved the change 👍  Thanks!

GresilleSiffle: Alright, I took the time to move the logic from the view to the model  to follow the Django's "big models small views" mindset. It also simplifies the tests as now they are gatherer in only one file.

PR #1227, opened 2022-12-03
[back] fix: reduce the risk of DB deadlock during batch operations

GresilleSiffle:

PR #1337, opened 2023-02-04
[back] Fix deadlock between comparison and ml

amatissart:
## Problem

A deadlock was recently observed at the end of the "ml_train" job. It's similar to the issue addressed in #1227 
And I think I have now realized what exactly is the root cause for this situation. When creating a comparison at the moment when tournesol scores are being updated, there may exist 2 concurrent transactions which try to update the same Entity objects in a different order:

### Deadlock scenario

```
      TRANSACTION 1 (ml)                   TRANSACTION 2 (api)
    save_tournesol_scores()                 perform_create()
        ----------------                    ---------------


+-------------------------------+                  |
|    Update tournesol_score     |                  |
|         Entity 1000           |                  |
|   (Locked by Transaction 1)   |                  |
|                               |                  |
+-------------+-----------------+                  v
              |
              |                    +-------------------------------+
              |                    |      Update n_ratings         |
              |                    |          Entity 2000          |
              |                    |    (Locked by Transaction 2)  |
              .                    |                               |
              .                    +-------------------------------+
              .                                   |
              .                                   |
              .                                   v
              .                    +-------------------------------+
              |                    |      Update n_ratings         |
              |                    |       Entity 1000             |
              |                    |                               |
              |                    |   (waiting for Transaction 1  |
              |                    |       to complete...)         |
              v                    |                               |
                                   +-------------------------------+
+--------------------------------+
|   Update tournesol_score       |                |
|     Entity 2000                |                |
|                                |                |
|    (waiting for Transaction 2  |                |
|     to complete ....)          |                |
+--------------------------------+                |
              |                                   |
              |                                   |
              |
+-------------------------------------------------------------------+
|                             DEADLOCK                              |
+-------------------------------------------------------------------+
```

## Suggested changes

1. Tournesol scores should not be updated from a single long-running transaction.  
Even when using `batch_size`, the whole `bulk_update()` runs [inside a single transaction](https://github.com/django/django/blob/ca891ef44b27eb93890846615e0121037e6c8a35/django/db/models/query.py#L640-L642). Let's use manual batches to run smaller distinct transactions, to avoid locking too many entities at the same time.

2. When POSTing a new comparison the transaction on `perform_create()` is redundant and can be removed.  
It was motivated in #315 to ensure that related ContributorRatings are created in the same transaction. Ratings are now created beforehand (and checked by "ml-train"), and the `ComparisonSerializer` also creates its own transaction to save all criteria scores safely. So the top transaction can be removed, in order to execute `update_n_ratings()` for both entities in distinct transactions.

### DIFF
--- a/backend/ml/outputs.py
+++ b/backend/ml/outputs.py
@@ -1,3 +1,4 @@
+from itertools import islice
 from typing import Iterable, Optional, Union
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


lfaucon (approved):
Looks good to me, but I prefer letting @GresilleSiffle also leave a review.

GresilleSiffle (approved):

Issue #1329, opened 2023-02-02
[infra] feat: create an ansible recipe to set up a self-hosted instance of plausible


GresilleSiffle:
**This issue continues the work initiated by** #1111

---

The goal is to updated our Ansible recipes to be able to automatically deploy and configure a Plausible Analytics instance. The deployment of the instance should happen only if it's explicitly configured in the Ansible inventory.

This issue can be divided in several issues to ease the development. The docker recipe could be developed first and the Plausible Analytics one after.

**(a)** The self-hosting documentation of Plausible Analytics describes how to install the software with docker:

- https://docs.docker.com/engine/install/debian/

_create a new Ansible recipe that will_
- [x] install the docker apt repository
- [x] install the requirements
- [x] install docker and the docker compose plugin
- [x] check the security good practices
  - tracked by https://github.com/tournesol-app/tournesol/issues/1499

**(b)** The plausible install documentation:

- https://plausible.io/docs/self-hosting

_create a new Ansible recipe that will_
- [x] create a new systemd unit that will run `docker compose up`
  - [x] by default, make all created containers send logs to the `systemd` journal
    - see: https://docs.docker.com/config/containers/logging/configure/
  - [x] check if the applications running in the containers have additional log files
- [x] create a new NGINX server configuration to act as a reverse proxy of Plausible Analytics
- [x] download the [hosting](https://github.com/plausible/hosting) repository in `/opt/plausible-analytics`
- [x] configure `BASE_URL`
- [x] configure `SECRET_KEY_BASE` (this will be a new secret handled by our script dedicated to secrets)
- [x] turn off the registration `DISABLE_REGISTRATION=true`

**(c)** The front ends must be configured to send data to their related Plausible Analytics instance:
- [x] the recipe must be applied on the local vm, staging and production
- [x] install [plausible-tracker](https://www.npmjs.com/package/plausible-tracker)
- [x] configure the CSP to allow plausible-tracker to work
- [x] track all page views
- [x] track custom goals related to the tutorial
- [x] define with the team the other custom goals (see [this comment](https://github.com/tournesol-app/tournesol/issues/1111#issuecomment-1240745283))
  - tracked by https://github.com/tournesol-app/tournesol/issues/1498 

**(d)** Define the Plausible Analytics git repository update policy

- [x] do it
  - see the wiki: https://github.com/tournesol-app/plausible-self-hosting/wiki 

**(e)** Define the Plausible Analytics container update policy

- [x] do it
[...]

Issue #1111, opened 2022-09-06
Investigate open source web analytics software 


GresilleSiffle:
WIP I'm still gathering information...

What we could look in each software:
- what are the features? visitors / heatmaps / etc.
- how about privacy?
- what is the maintenance cost of self-hosting? is it lightweight tool or a mastodon?
- how actively it is maintained?

**TL;DR**

Matomo is a mature Google Analytics-like tool. It comes with a lof of features and a relative complexity. Visitors consent is required with default configuration.

Plausible is new, lightweight and has a strong focus on simplicity, speed and privacy. No visitor consent is required to use it.

Open Web Analytics, to complete...

**I recommend to test as many solutions as possible when the team behind the projects propose a free trial**

---

### Matomo

It looks like one of the oldest open source analytics tools, initially published in 2007. It as now a lot of features and the team behind it propose professional services like hosted instances and training. 

#### Matomo and privacy

- https://matomo.org/privacy/?menu

The French CNIL organization states that Matomo version 4 can be configured to gather only a limited amount of data, and if configured properly can be exempted of asking the explicit consent to the website visitors (see https://www.cnil.fr/fr/cookies-et-autres-traceurs/regles/cookies-solutions-pour-les-outils-de-mesure-daudience). 

#### More info

Description from GitHub

> Matomo is the leading Free/Libre open analytics platform.

> Matomo is a full-featured PHP MySQL software program that you download and install on your own webserver. At the end of the five-minute installation process, you will be given a JavaScript code. Simply copy and paste this tag on websites you wish to track and access your analytics reports in real-time.

> Matomo aims to be a Free software alternative to Google Analytics and is already used on more than 1,400,000 websites. Privacy is built-in!

- website: https://matomo.org/
- features: https://matomo.org/product-features/
- demo: https://demo.matomo.org/index.php

First version published in 2007, last version 4.11.0 the 3rd of august 2022.

- license: GPL-3.0-or-later
- source code: https://github.com/matomo-org/matomo
- state: **actively maintained** dozen of commits every month (28,636 commits / 16.8k stars)
- self-hosted: **yes** - https://matomo.org/faq/on-premise/installing-matomo/
- saas: **paid** - https://matomo.org/pricing/
  - starting at 19€ per month, for a monthly traffic of 50,000 hits
  - data are stored in Europe
  - support by e-mail
  - 30 websites / 30 team members (unlimited with the self-hosted)
  - GDPR manager included (also available in the self-hosted)
[...]

PR #1340, opened 2023-02-07
[infra] feat: add Plausible Analytics

GresilleSiffle:
**related to** #1329
 
---

### description

This PR adds two new Ansible recipes to automatically install and configure a docker-based self-hosted instance of Plausible Analytics on staging, production, and on the local VM.

The first recipe is independent and is responsible of installing and configuring Docker. The second one set up Plausible Analytics and depends on the first one and on other existing recipes (like `nginx` and `repository`).

The PR also configure the front end to send data to the deployed analytics server. Three custom events have been configured for now:
- signup
- tutorial (step by step progression)
- account deleted

---

This PR addresses most of all tasks described in the issue #1329, here are the details.

### Tasks addressed

**(a) new Ansible tasks to install and configure Docker**
- [x] install the docker apt repository
- [x] install the requirements
- [x] install docker and the docker compose plugin
- [x] configure the default log driver

**(b) new Ansible tasks to install Plausible Analytics**
- [x] create a new systemd unit that will run `docker compose up`
- [x] create a new NGINX server configuration to reverse proxy Plausible Analytics
- [x] download the [hosting](https://github.com/plausible/hosting) repository in `/opt/plausible-analytics`
- [x] configure `BASE_URL`
- [x] configure `SECRET_KEY_BASE` (this will be a new secret handled by our script dedicated to secrets)
- [x] turn off the registration `DISABLE_REGISTRATION=true`

**(c) configure the front end**
- [x] install [plausible-tracker](https://www.npmjs.com/package/plausible-tracker)
- [x] configure the CSP to allow plausible-tracker to work
- [x] track all page views
- [x] track custom goals related to the tutorial

### Remaining tasks

**In the Plausible Analytics recipe**
- check if the applications running in the containers have additional log files
  - /var/log/clickhouse-server/clickhouse-server.log
  - /var/log/clickhouse-server/clickhouse-server.err.log
  - mount this files on the host
  - configure the log rotation 
-  define the Plausible Analytics git repository update policy
 - define the Plausible Analytics container update policy

**In the front end**
- define with the team the other custom goals (see [this comment](https://github.com/tournesol-app/tournesol/issues/1111#issuecomment-1240745283))
  - dataset download
  - donation page visit
  - click on uTip / PayPal
  - etc. 

### issues to create

- [x] systemd services using docker should be restarted when the docker configuration changes in /etc 
  - tracked by #1347 
- [x] changing the recommendations filters generates 2 page views (instead of 1) 
- [ ] should we plan to have a backup/restore script? (see https://github.com/plausible/analytics/discussions/1132)
- [ ] add plausible analytics to the list of monitored URLs?


frontend/src/App.tsx
```diff
@@ -70,6 +71,13 @@ function App() {
     initializeOpenAPI(loginState, i18n);
   }, [loginState, i18n]);
 
+  // Plausible will not make HTTP requests to track events when
+  // `localtion.hostname` is "localhost".
+  const plausible = Plausible({
+    apiHost: process.env.REACT_APP_WEBSITE_ANALYTICS_URL,
+  });
+  plausible.enableAutoPageviews();
```
amatissart, line 79: What if `process.env.REACT_APP_WEBSITE_ANALYTICS_URL` is empty or not defined ?  
I would find clearer to wrap this initialization into a condition to check that the configuration is present.

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
```diff
@@ -53,6 +57,9 @@ const ComparisonSeries = ({
   const location = useLocation();
 
   const { name: pollName } = useCurrentPoll();
+  const { trackEvent } = Plausible({
+    apiHost: process.env.REACT_APP_WEBSITE_ANALYTICS_URL,
+  });
```
amatissart, line 62: Don't you think it would make sense to initialize the Plausible client globally, and export it from a dedicated file?

Even if this initialization has no side-effect (I suppose?), it would make the maintainance easier to configure it from a single place.

infra/ansible/roles/repository/tasks/main.yml
```diff
@@ -18,3 +18,11 @@
     version: "{{git_reference}}"
   become: yes
   become_user: git
+
+- name: Checkout Plausible Analytics self-hosting repository
+  git:
+    repo: 'https://github.com/tournesol-app/plausible-self-hosting'
+    dest: /home/git/plausible_hosting
+    version: tournesol-deployment
```
amatissart, line 26: regarding the remaining tasks:
> define the Plausible Analytics git repository update policy
 define the Plausible Analytics container update policy

maybe a git **tag** could be used here, to make any future update more explicit

infra/ansible/roles/plausible_analytics/templates/tournesol-website-analytics.service.j2
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=The Tournesol website analytics
+PartOf=docker.service
```
amatissart, line 3: Isn't this `PartOf` sufficient to solve the issue described in https://github.com/tournesol-app/tournesol/issues/1347 ?

According to Systemd documentation
> **PartOf=** Configures dependencies similar to Requires=, but limited to stopping and restarting of units. When systemd stops or restarts the units listed here, the action is propagated to this unit.

amatissart (commented):
Minor comments, but overall it looks good to me.

I still need to test this deployment on my local VM though.

infra/ansible/roles/plausible_analytics/templates/tournesol-website-analytics.service.j2
GresilleSiffle: Good catch, I haven't tried yet.

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
amatissart: After some tests on a local VM, it seems that initializing `Plausible` several times does have a side-effect: duplicate event handlers will be defined, leading to multiple "pageviews" events being sent to plausible server.

So we need to be sure this client is created only once, outside of the React lifecycle.

amatissart:
Deploying this branch on a local VM worked flawlessly :+1: 

Besides previous comments, 2 things to keep in mind when deploying to staging/production:
- the new "plausible_analytics_secret_key" is generated on provisioning in "generate_secrets.sh". To deploy the service on existing environments, the file `/root/plausible_analytics_secret_key` needs to be initialized manually (unless @GresilleSiffle has another approach in mind ?).
- deploying on staging/production requires a new TLS certificate to be generated. This is not tested when deploying in a local VM and it may cause a temporary downtime due to how `certbot` is configured. To check carefully on staging.

GresilleSiffle:
Thank you for the review. I still have to rework how the Plausible client is initialized and how the new service behaves when the Docker service is restarted. 

> the new "plausible_analytics_secret_key" is generated on provisioning in "generate_secrets.sh". To deploy the service on existing environments, the file /root/plausible_analytics_secret_key needs to be initialized manually (unless @GresilleSiffle has another approach in mind ?).

Yes I thought about initializing the secret manually before deploying.

> deploying on staging/production requires a new TLS certificate to be generated. This is not tested when deploying in a local VM and it may cause a temporary downtime due to how certbot is configured. To check carefully on staging.

Alright :ok_hand:

frontend/src/features/comparisonSeries/ComparisonSeries.tsx
GresilleSiffle: I created a custom hook in https://github.com/tournesol-app/tournesol/pull/1340/commits/4487a0e132a3bc39f5e1f2b7f28459c2d8695c0e, I'm going to try if it's enough to prevent multiple initialization.

GresilleSiffle: It seems to trigger even more page views.

GresilleSiffle: I think I've fixed the major problems with:
- https://github.com/tournesol-app/tournesol/pull/1340/commits/315b7af384d0c9bb34e25cab207c4b25ad7dcbf9
- https://github.com/tournesol-app/tournesol/pull/1340/commits/c858ee1bfad79b75e9d7d1e545278b0078e1a309

I'm trying to see if the page views counter is now reliable : )

frontend/src/App.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1340/commits/e2d4dc05ea9be7d63636029aff7f3be0f498311e.

`enableAutoPageviews` and `trackEvent` are now wrapped inside a `if (client)` condition. The variable `client` is defined only when `REACT_APP_WEBSITE_ANALYTICS_URL` is defined and not empty.

infra/ansible/roles/plausible_analytics/templates/tournesol-website-analytics.service.j2
GresilleSiffle: Restarting the docker service automatically restarts tournesol-website-analytics.service :+1:

GresilleSiffle:
Only two issues remaining:
- using a tag to deploy `tournesol-app/plausible-self-hosting`
- nginx returns an error 502 after a VM reboot when trying to access Plausible, it serves the website correctly if I run `systemctl restart nginx`

infra/ansible/roles/repository/tasks/main.yml
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1340/commits/fd0b65a8bc4bcf8534a92c90f71c51fb3ad2e61d

amatissart (approved):

PR #1346, opened 2023-02-09
Fix extension's iframe event listener not removed

sigmike:
I think it was a typo made when the the code was initially introduced in 4185735c06269a5ae6f3aafef022f91a604d75f7.


amatissart (approved):

GresilleSiffle (approved):
Thanks for the fix @sigmike.

PR #1350, opened 2023-02-10
[dataviz] Add a video podium and more filters in the sidebar

aidanjungo:
To be merged after #1349 

New podium section with filter from the sidebar:

image

### DIFF
--- a/data-visualization/pages/02_Videos_and_channels.py
+++ b/data-visualization/pages/02_Videos_and_channels.py
@@ -32,6 +32,16 @@ def add_sidebar_select_channels():
     if len(selected_uploaders):
         df = df[df["uploader"].isin(selected_uploaders)]
 
+    all_languages = df["language"].unique()
+    selected_languages = st.sidebar.multiselect("Language", all_languages, [])
+    if len(selected_languages):
+        df = df[df["language"].isin(selected_languages)]
+
+    min_contributor = st.sidebar.number_input(
+        "Minimum number of contributors to be included", value=3, min_value=1
+    )
+    df = df[df["rating_n_contributors"] >= min_contributor]
+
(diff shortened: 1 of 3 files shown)


data-visualization/pages/02_Videos_and_channels.py
```diff
@@ -63,13 +111,11 @@ def add_expander_avg_values():
             st.warning(MSG_NO_DATA)
             return
 
-        col1, col2 = st.columns(2)
-        with col1:
+        st.markdown("Criteria to be included:")
```
GresilleSiffle, line 114: This text is properly displayed but I don't understand exactly its purpose.

capture

When I read criteria I first thought about the rated criteria but it's something different right? If this text is a title to notify the presence of inputs that can filter the results, I think it can be safely deleted. The input looks explicit enough, and the UI will be more clear with less text here.

GresilleSiffle (approved):
It works well : )

I made a minor comment about a text that was a bit confusing to me.

I also saw some warnings in the Streamlit logs that are unrelated to this PR. I'll open another PR to fix them.

```
`label` got an empty value. This is discouraged for accessibility reasons and may be disallowed in the future by raising an exception. Please provide a non-empty label and hide it with label_visibility if needed.
```

data-visualization/pages/02_Videos_and_channels.py
aidanjungo: Yes, at some point I put several criteria here, but then I move them in the sidebar so I can remove that

aidanjungo: I'll fix the label in another PR, don't lose to much time on dataviz, I think it is really not high priority.

PR #1352, opened 2023-02-11
[back] ml: rename 'supertrusted' into 'scaling-calibration' users, sync definition with latest paper

### LINKED ISSUE #1351: Update the Mehestan implementation of scaling contributors
Mehestan does not use "supertrusted users" anymore, which used to be members of the core team. 

Now it is called "scaling contributors". There are up to 100 scaling contributors which are, among the contributors who have a trust score higher than 0.1, those who have the highest number of comparisons.

There are 3 problems in the implementation :
- the term "supertrusted" is still used to refer to scaling contributors, which leads to confusions
- `ratings_properties` still only picks the 6 contributors that have the highest count of comparisons. It should pick the 100 contributors with a trust score higher than 0.1 and the highest count of comparisons.
- `get_supertrusted_users` is filtering on pretrusted users instead of users with a trust score higher than 0.1. And it does not look up-to-date anyway.

amatissart:
Fixes #1351 

Note: a few references to "supertrusted" users are still present in the suggestion provider, which uses a subset of comparisons from the staff team to initialize its algorithm. (Not related to Mehestan)

### DIFF
--- a/backend/ml/inputs.py
+++ b/backend/ml/inputs.py
@@ -57,7 +57,7 @@ class MlInputFromPublicDataset(MlInput):
         if isinstance(dataset_zip, str) and (
             dataset_zip.startswith("http://") or dataset_zip.startswith("https://")
         ):
-            dataset_zip, _headers = urlretrieve(dataset_zip)
+            dataset_zip, _headers = urlretrieve(dataset_zip)  # nosec B310
 
         with zipfile.ZipFile(dataset_zip) as zip_file:
             zip_root_dir = next(path for path in zipfile.Path(zip_file).iterdir() if path.is_dir())
--- a/backend/ml/inputs.py
+++ b/backend/ml/inputs.py
@@ -1,6 +1,8 @@
+import zipfile
 from abc import ABC, abstractmethod
 from functools import cached_property
 from typing import Optional
+from urllib.request import urlretrieve
 
 import pandas as pd
 from django.db.models import Case, F, QuerySet, When
@@ -11,10 +13,13 @@ from tournesol.models import ComparisonCriteriaScore, ContributorRating, Contrib
 
 
 class MlInput(ABC):
+    SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE = 20
+    SCALING_CALIBRATION_MIN_TRUST_SCORE = 0.1
+    MAX_SCALING_CALIBRATION_USERS = 100
+
     @abstractmethod
     def get_comparisons(
         self,
-        trusted_only=False,
         criteria: Optional[str] = None,
         user_id: Optional[int] = None,
     ) -> pd.DataFrame:
@@ -48,45 +53,59 @@ class MlInput(ABC):
 
 
 class MlInputFromPublicDataset(MlInput):
-    def __init__(self, csv_file):
-        self.public_dataset = pd.read_csv(csv_file)
-        self.public_dataset.rename(
-            {"video_a": "entity_a", "video_b": "entity_b"}, axis=1, inplace=True
-        )
-        self.public_dataset["user_id"], self.user_indices = self.public_dataset[
-            "public_username"
-        ].factorize()
+    def __init__(self, zip_path: str):
+        if zip_path.startswith("http://") or zip_path.startswith("https://"):
+            zip_path, _headers = urlretrieve(zip_path)
 
-    def get_comparisons(self, trusted_only=False, criteria=None, user_id=None) -> pd.DataFrame:
-        dtf = self.public_dataset.copy(deep=False)
+        with zipfile.ZipFile(zip_path) as zip_file:
+            zip_root_dir = next(path for path in zipfile.Path(zip_file).iterdir() if path.is_dir())
+
+            with (zip_root_dir / "comparisons.csv").open() as comparison_file:
+                self.comparisons = pd.read_csv(comparison_file)
+                self.comparisons.rename(
+                    {"video_a": "entity_a", "video_b": "entity_b"}, axis=1, inplace=True
+                )
+
+            with (zip_root_dir / "users.csv").open() as users_file:
+                self.users = pd.read_csv(users_file)
+                self.users.index.name = "user_id"
+
+            username_to_user_id = pd.Series(
+                data=self.users.index, index=self.users["public_username"]
+            )
+            self.comparisons = self.comparisons.join(username_to_user_id, on="public_username")
+
+    def get_comparisons(self, criteria=None, user_id=None) -> pd.DataFrame:
+        dtf = self.comparisons.copy(deep=False)
         if criteria is not None:
             dtf = dtf[dtf.criteria == criteria]
         if user_id is not None:
             dtf = dtf[dtf.user_id == user_id]
+        dtf["weight"] = 1
         return dtf[["user_id", "entity_a", "entity_b", "criteria", "score", "weight"]]
 
     @cached_property
     def ratings_properties(self):
         user_entities_pairs = pd.Series(
             iter(
-                set(self.public_dataset.groupby(["user_id", "entity_a"]).indices.keys())
-                | set(self.public_dataset.groupby(["user_id", "entity_b"]).indices.keys())
+                set(self.comparisons.groupby(["user_id", "entity_a"]).indices.keys())
+                | set(self.comparisons.groupby(["user_id", "entity_b"]).indices.keys())
             )
         )
         dtf = pd.DataFrame([*user_entities_pairs], columns=["user_id", "entity_id"])
         dtf["is_public"] = True
-        top_users = dtf.value_counts("user_id").index[:6]
-        dtf["is_scaling_calibration_user"] = dtf["user_id"].isin(top_users)
-        # TODO read trust_scores from the public dataset
-        dtf["trust_score"] = dtf["is_scaling_calibration_user"] * 0.8
+        dtf["trust_score"] = dtf["user_id"].map(self.users["trust_score"])
+        scaling_calibration_user_ids = (
+            dtf[dtf.trust_score > self.SCALING_CALIBRATION_MIN_TRUST_SCORE]["user_id"]
+            .value_counts(sort=True)[: self.MAX_SCALING_CALIBRATION_USERS]
+            .loc[lambda count: count >= self.SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE]
+            .index
+        )
+        dtf["is_scaling_calibration_user"] = dtf["user_id"].isin(scaling_calibration_user_ids)
         return dtf
 
 
 class MlInputFromDb(MlInput):
-    SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE = 20
-    SCALING_CALIBRATION_MIN_TRUST_SCORE = 0.1
-    MAX_SCALING_CALIBRATION_USERS = 100
-
     def __init__(self, poll_name: str):
         self.poll_name = poll_name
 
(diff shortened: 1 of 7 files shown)


backend/ml/inputs.py
```diff
@@ -69,35 +67,33 @@ def get_comparisons(
...
                 set(self.public_dataset.groupby(["user_id", "entity_a"]).indices.keys())
-                | set(
-                    self.public_dataset.groupby(["user_id", "entity_b"]).indices.keys()
-                )
+                | set(self.public_dataset.groupby(["user_id", "entity_b"]).indices.keys())
             )
         )
         dtf = pd.DataFrame([*user_entities_pairs], columns=["user_id", "entity_id"])
         dtf["is_public"] = True
         top_users = dtf.value_counts("user_id").index[:6]
```
glerzing, line 78: Why don't we set `top_users = dtf.value_counts("user_id").index[:MAX_SCALING_CALIBRATION_USERS]` ?

glerzing: Actually, the top_users also need I guess a trust score higher than 0.1.

backend/ml/inputs.py
```diff
@@ -69,35 +67,33 @@ def get_comparisons(
...
-        dtf["is_trusted"] = dtf["is_supertrusted"] = dtf["user_id"].isin(top_users)
+        dtf["is_scaling_calibration_user"] = dtf["user_id"].isin(top_users)
+        # TODO read trust_scores from the public dataset
+        dtf["trust_score"] = dtf["is_scaling_calibration_user"] * 0.8
```
glerzing, line 81: Shouldn't we just retrieve directly the trust score, from the database or the dataset, instead of multiplying this by 0.8 ?

glerzing: Otherwise it ignores the vouching.

backend/ml/inputs.py
```diff
@@ -114,29 +110,22 @@ def get_supertrusted_users(self) -> QuerySet[User]:
                 (self.poll_name,),
             )
         )
-        if n_alternatives <= self.SUPERTRUSTED_MIN_ENTITIES_TO_COMPARE:
-            # The number of alternatives is low enough to consider as supertrusted
```
glerzing, line 118: I'm not convinced that this "if" and the code inside are useful. I would have probably also removed the notion  of `SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE` unless we need it for debugging or automated tests (the 100 top contributors already have more than 20 comparisons).

glerzing (commented):
I let you check my comments on inputs.py.

backend/ml/inputs.py
amatissart: It's definitely not required for the poll "videos". This mechanism was introduced for the poll "presidentielles2022" where calibrating the scaling on top contributors would not make sense.

I think it's nice to keep this possibility for potential future polls, and for tests purposes as well. On small test cases it gives a clear-cut definition for scaling calibration users (i.e any trusted user who compared all available entities).

glerzing: I'm not sure why it would not make sense, but ok if there is a good reason.
Can you check the other comments ?

backend/ml/inputs.py
amatissart: I updated `MlInputFromPublicDataset` to use the same parameters for scaling calibration. Hopefully, it will make things clearer, even if `MlInputFromPublicDataset` is not used for now.

backend/ml/inputs.py
amatissart: My point was:  in polls with a small number of entities, we can expect that a large proportion of users will have compared the exact same number of entities. So the set of 100 scaling calibration users is not properly defined, and that's why at the time we opted for selecting all contributors who compared the entire set of entities.

backend/ml/inputs.py
```diff
@@ -41,63 +46,75 @@ def ratings_properties(self) -> pd.DataFrame:
...
             * `is_public`: bool
-            * `is_trusted`: bool
-            * `is_supertrusted`: bool
+            * `is_scaling_calibration_user`: bool
+            * `trust_score`: float
         """
         raise NotImplementedError
 
 
 class MlInputFromPublicDataset(MlInput):
```
glerzing, line 55: It's not clear to me what's the use case for MlInputFromPublicDataset.

Shouldn't we have more of a standalone interface that also downloads the dataset and unzips it without having to find the right URL ? Or is MlInputFromPublicDataset just the low-level component and we are going in the future to add a user-friendly standalone script that will use it ? Will MlInputFromPublicDataset have any use for people that want to work with notebooks ?

backend/ml/inputs.py
```diff
@@ -41,63 +46,75 @@ def ratings_properties(self) -> pd.DataFrame:
...
+            )
+            self.comparisons = self.comparisons.join(username_to_user_id, on="public_username")
+
+    def get_comparisons(self, criteria=None, user_id=None) -> pd.DataFrame:
+        dtf = self.comparisons.copy(deep=False)
         if criteria is not None:
             dtf = dtf[dtf.criteria == criteria]
         if user_id is not None:
             dtf = dtf[dtf.user_id == user_id]
+        dtf["weight"] = 1
```
glerzing, line 86: I understand why users have a trust_score, and why user ratings have weights. But not why comparisons have weights. Should we keep this concept ?

backend/ml/inputs.py
```diff
@@ -41,63 +46,75 @@ def ratings_properties(self) -> pd.DataFrame:
...
         )
         dtf = pd.DataFrame([*user_entities_pairs], columns=["user_id", "entity_id"])
         dtf["is_public"] = True
-        top_users = dtf.value_counts("user_id").index[:6]
-        dtf["is_trusted"] = dtf["is_supertrusted"] = dtf["user_id"].isin(top_users)
+        dtf["trust_score"] = dtf["user_id"].map(self.users["trust_score"])
+        scaling_calibration_user_ids = (
+            dtf[dtf.trust_score > self.SCALING_CALIBRATION_MIN_TRUST_SCORE]["user_id"]
+            .value_counts(sort=True)[: self.MAX_SCALING_CALIBRATION_USERS]
+            .loc[lambda count: count >= self.SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE]
```
glerzing, line 103: I'm not sure this ".loc" does what we want.

For videos, the constraint SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE isn't effective. Because there is at least one contributor that compared more than SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE entities (cf. get_scaling_calibration_users).

And for the elections poll, we don't have SCALING_CALIBRATION_MIN_ENTITIES_TO_COMPARE entities, so this would return 0 scaling contributors.

backend/ml/inputs.py
```diff
@@ -41,63 +46,75 @@ def ratings_properties(self) -> pd.DataFrame:
...
+        if isinstance(dataset_zip, str) and (
+            dataset_zip.startswith("http://") or dataset_zip.startswith("https://")
+        ):
+            dataset_zip, _headers = urlretrieve(dataset_zip)  # nosec B310
+
+        with zipfile.ZipFile(dataset_zip) as zip_file:
+            zip_root_dir = next(path for path in zipfile.Path(zip_file).iterdir() if path.is_dir())
+
+            with (zip_root_dir / "comparisons.csv").open() as comparison_file:
+                self.comparisons = pd.read_csv(comparison_file)
```
glerzing, line 66: When I try it on the real dataset, I have an exception `UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 3576: character maps to <undefined>`.
Explicitly setting `encoding='utf8'` doesn't work either.

amatissart: I can't reproduce the issue. Are you using the versions as specified in "backend/requirements.txt"?

Here with Python 3.9.16, Pandas 1.4.1 :
```bash
$ python manage.py shell
Python 3.9.16 (main, Dec  7 2022, 00:00:00) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from  ml.inputs import MlInputFromPublicDataset

In [2]: ml_input = MlInputFromPublicDataset("https://api.tournesol.app/exports/all/")

In [3]: ml_input.ratings_properties
Out[3]: 
       user_id    entity_id  is_public  trust_score  is_scaling_calibration_user
0         1237  RdAoJhs1yR0       True      0.00000                        False
1         4841  hOWB7o6XUSU       True      0.00000                        False
2          611  TuWrejZE8i8       True      1.00000                         True
3         2196  Isa-qo_vXnw       True      0.00000                        False
4         2892  cOGB3WmylnY       True      1.00000                         True
...        ...          ...        ...          ...                          ...
51124     1459  4_Le9FOsO74       True      0.00000                        False
51125     3083  EFmPGvzfMZI       True      0.00000                        False
51126      150  3AXupc7oE-g       True      0.80000                         True
51127     1888  mL-6oJC2tcg       True      0.97147                         True
51128     5096  GvS2qik3Mr8       True      0.00000                        False

[51129 rows x 5 columns]

In [4]: ml_input.get_comparisons()
Out[4]: 
        user_id     entity_a     entity_b             criteria  score  weight
0          1129  tkiOzcSya18  1jFeYXgarsM  largely_recommended    -10       1
1          1129  tkiOzcSya18  1jFeYXgarsM          reliability    -10       1
2          1129  tkiOzcSya18  1jFeYXgarsM             pedagogy    -10       1
3          1129  tkiOzcSya18  1jFeYXgarsM           importance    -10       1
4          1129  tkiOzcSya18  1jFeYXgarsM      layman_friendly    -10       1
...         ...          ...          ...                  ...    ...     ...
437873     1066  7i1Bq6pr120  AFNAUEKfjpo        better_habits     -1       1
437874     1066  7i1Bq6pr120  AFNAUEKfjpo        backfire_risk     -2       1
437875     5430  Wxp__MZgwAI  d8ntTETk2F0  largely_recommended    -10       1
437876     3416  gKhOJqaD6Pw  2bI7s_d5D8Y  largely_recommended     -8       1
437877     3416  eUB6zkW-e5Y  2bI7s_d5D8Y  largely_recommended     -9       1

[437878 rows x 6 columns]
```

backend/ml/inputs.py
amatissart: The idea would be to use it as a low-level component in the future yes. It's not really usable for now as running Mehestan requires to write the results into the Django database, but I think it would make sense to introduce another abstraction on top. Something like `TournesolRun(input_mode, output_mode, ml_parameters)` which would enable to run the algorithms without any dependence to Django or PostgreSQL. Possibly from notebooks, but the exact use-case has not been specified yet.


But that was not really the purpose of this branch :wink: I'm sorry to have brought this topic on this PR.
Maybe we'd better discuss that separately. I can even delete this class `MlInputFromPublicDataset` completely and move it to a new branch if that can help.

backend/ml/inputs.py
amatissart: I wanted to replicate the definition used in production, but I agree that does not really make sense on the public dataset. I will remove the filter on the number of compared entities.

backend/ml/inputs.py
glerzing: I don't really care which PR or branch we do these corrections, that's fine. And if it's just a lower-level abstraction that's ok too, we'll think later about how to make a high-level practical interface..

backend/ml/inputs.py
glerzing: I have Python 3.9.12 and pandas 1.4.1.
I also tried with Python 3.11 and reproduced the issue. I suspect that it comes from working on Windows, and from working on a Zip file.
It's similar to what's described here (https://stackoverflow.com/questions/68470744/unicodedecodeerror-utf-8-for-pandas-read-csv-from-folder-nested-zip-file), except the solutions don't work for me.

amatissart: Strange. Would it help to open the .csv files in binary mode?
```diff
diff --git a/backend/ml/inputs.py b/backend/ml/inputs.py
index 77c3160d..972abdb6 100644
--- a/backend/ml/inputs.py
+++ b/backend/ml/inputs.py
@@ -62,13 +62,13 @@ class MlInputFromPublicDataset(MlInput):
         with zipfile.ZipFile(dataset_zip) as zip_file:
             zip_root_dir = next(path for path in zipfile.Path(zip_file).iterdir() if path.is_dir())
 
-            with (zip_root_dir / "comparisons.csv").open() as comparison_file:
+            with (zip_root_dir / "comparisons.csv").open(mode="rb") as comparison_file:
                 self.comparisons = pd.read_csv(comparison_file)
                 self.comparisons.rename(
                     {"video_a": "entity_a", "video_b": "entity_b"}, axis=1, inplace=True
                 )
 
-            with (zip_root_dir / "users.csv").open() as users_file:
+            with (zip_root_dir / "users.csv").open(mode="rb") as users_file:
                 self.users = pd.read_csv(users_file)
                 self.users.index.name = "user_id"
```

glerzing: yes it works !

amatissart: Ok, fixed in 75d762c2

Regarding how to use this class to make experimentation easier, I've just opened #1370.

glerzing (approved):

PR #1353, opened 2023-02-11
feat: Add file 'entities.csv' in public dataset export

lfaucon:


backend/tournesol/views/exports.py
```diff
@@ -132,6 +134,47 @@ def write_individual_criteria_scores_file(poll_name: str, write_target) -> None:
...
+        "uploader",
+    ]
+
+    all_metadata = dict(Entity.objects.filter(type=TYPE_VIDEO).values_list("uid", "metadata"))
+
+    criteria_scores = sorted(
+        EntityCriteriaScore.objects
+        .filter(poll__name=poll_name, score_mode=ScoreMode.DEFAULT)
+        .values_list("entity__uid", "criteria", "score")
+    )
```
amatissart, line 158: It doesn't really matter with the current size of the dataset, but is there any specific reason to fetch the videos metadata separately?

It would be possible to fetch individual metadata fields in the EntityCriteriaScore queryset, and to iterate on this queryset without loading the whole list in memory, with something like:

```py
 criteria_scores = (
    EntityCriteriaScore.objects
        .filter(poll__name=poll_name, score_mode=ScoreMode.DEFAULT)
        .values_list(
            "entity__metadata__video_id", 
            "criteria",
            "score",
            "entity__metadata__name",
            "entity__metadata__publication_date",
            "entity__metadata__name",
            "entity__metadata__uploader",
        )
        .order_by("entity__uid", "criteria")
        .iterator()
    )

GresilleSiffle:
Hello,

I created a PR which refactors how the dataset is built and distributed #1359 (it will probably be in conflict with this one).

I suggest to wait a bit before merging this PR (#1353) so that we can integrate the new `entities.csv` file directly in the new building process.

GresilleSiffle:
As discussed on Discord, I'll continue this PR.

PR #1354, opened 2023-02-11
[back] ml: compute scaling uncertainty for calibration contributors

### LINKED ISSUE #1338: Error in uncertainty of scaling-calibration contributors' scores
As pointed out by @glerzing, there is an inconsistency in the algorithms I proposed, when computing the uncertainties on raw scores. In particular, for some unjustified reason, scaling-calibration and non-scaling-calibration contributors are not handled the same way.

The fix is to this line: https://github.com/tournesol-app/tournesol/blob/main/backend/ml/mehestan/global_scores.py#L274
And it should be more like this: https://github.com/tournesol-app/tournesol/blob/main/backend/ml/mehestan/global_scores.py#L298

See Section E.1.1 in the Overleaf: https://www.overleaf.com/project/63d63d642537a23c46fb1775

amatissart:
Fixes #1338 

Based on #1352


lenhoanglnh (approved):
LGTM

glerzing (approved):

PR #1355, opened 2023-02-11
[infra] feat: add the top contributors tweet service

aidanjungo:
Top contributors tweet will be schedule the 1st of every month at 15h08 (because... why not)


GresilleSiffle (approved):
Thank you @aidanjungo 

I pushed few commits to fix some errors occurring during the deployment. I also updated the schedules to make all Twitter Bot timers use the same minutes, it will be easier to remember when the timers are run without having to check the code.

GresilleSiffle:
It could be also nice to tweet at 9 am instead of 15 pm, to not wait all the day for the results ; )

Issue #1232, opened 2022-12-08
QrUnc : use LogSumExp instead of SoftMinMean ?


glerzing:
It may be interesting to use the LogSumExp instead of the SoftMinMean as a smooth minimum for the calculation of QrUnc. One reason is that the SoftMinMean isn't really consistent, you can have x1 < x2 and yet SoftMinMean(x1, y) > SoftMinMean(x2, y). Moreover, (I didn't understand why, but) in the paper we say that both elements of SoftMeanMin "represent upper bounds on uncertainty". So we want SoftMeanMin(x, y) < x and SoftMeanMin(x, y) < y. But this is not true, as shown in the following figure, it's the contrary.

The magenta curve represents the min and the red curves the SoftMinMean.

![image](https://user-images.githubusercontent.com/47578089/206081514-05c7aca9-7a2a-498e-8392-76d9430d24ec.png)

The LogSumExp (https://en.wikipedia.org/wiki/Smooth_maximum) looks like it fits more what we are trying to do :

![image](https://user-images.githubusercontent.com/47578089/206081715-d2bfcd7b-aeda-4950-8e62-175a0af9b104.png)

Here is the code to generate and modify the images, and with the implementation of LogSumExp :
https://colab.research.google.com/drive/1QcNQXFxGcu9PnLPIhyGafCH6QyNx936t?usp=sharing

glerzing:
LogSumExp may not be ideal too, because we may like QrUnc to be higher or equal to the "true" uncertainty, not strictly lower.  And because of the arbitrary choice of alpha.

But I think SoftMinMean is problematic because of its inconsistencies (e.g. SoftMinMean(0, 0) = 0, SoftMinMean(1, 0) = 0.26 and SoftMinMean(10, 0) ≈ 0).

Is using a regularized min necessary ? Should we consider using a simple minimum ? (which would simplify the formula)

lfaucon:
@lenhoanglnh Can you look into @glerzing 's recommendations?

Maybe a session about how we handle uncertainty in our algorithms to teach the whole team would be a big help (same as we did for EigenTrust, Voting rights, Video Suggestions, ...).

glerzing:
I wonder why when the uncertainty increases, QrUnc first increases and then decreases

![image](https://user-images.githubusercontent.com/47578089/216798448-51278448-2bb1-4666-bb48-b02de1833024.png)

lenhoanglnh:
[...]

PR #1360, opened 2023-02-13
Use LogSumExp, fix the biases, add tests (#1232)

glerzing:
Feel free to propose changes, especially if you have a better idea Adrien for preventing divisions by zero in `L_prime`.
I didn't add tests for QrUnc, because the current formula doesn't really have the properties I would expect.

### DIFF
--- a/backend/ml/mehestan/global_scores.py
+++ b/backend/ml/mehestan/global_scores.py
@@ -125,9 +125,9 @@ def compute_scaling(
                 )
             ) - s_nqmab
 
-            s = QrMed(1, 1, s_nqmab, delta_s_nqmab)
-            s_nqm.append(s)
-            delta_s_nqm.append(QrUnc(1, 1, 1, s_nqmab, delta_s_nqmab, qr_med=s))
+            s = QrMed(1, 1, s_nqmab - 1, delta_s_nqmab)
+            s_nqm.append(s + 1)
+            delta_s_nqm.append(QrUnc(1, 1, 1, s_nqmab - 1, delta_s_nqmab, qr_med=s))
             s_weights.append(scaling_weights[user_m])
 
         s_weights = np.array(s_weights)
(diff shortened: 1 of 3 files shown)


backend/ml/mehestan/global_scores.py
```diff
@@ -125,9 +125,9 @@ def compute_scaling(
                 )
             ) - s_nqmab
 
-            s = QrMed(1, 1, s_nqmab, delta_s_nqmab)
-            s_nqm.append(s)
-            delta_s_nqm.append(QrUnc(1, 1, 1, s_nqmab, delta_s_nqmab, qr_med=s))
+            s = QrMed(1, 1, s_nqmab - 1, delta_s_nqmab)
+            s_nqm.append(s + 1)
+            delta_s_nqm.append(QrUnc(1, 1, 1, s_nqmab - 1, delta_s_nqmab, qr_med=s))
```
amatissart, line 130: Good catch, thanks!

backend/ml/mehestan/primitives.py
```diff
@@ -30,7 +31,15 @@ def QrMed(W: float, w: Union[pd.Series, float], x: pd.Series, delta: pd.Series):
...
     def L_prime(m: float):
         x_minus_m = x - m
-        return W * m - np.sum(w * x_minus_m / np.sqrt(delta_2 + x_minus_m ** 2))
+        denominator = np.sqrt(delta_2 + x_minus_m ** 2)
+        # Avoid division by 0 by setting the result of the division
+        # to the sign of x_minus_m (+1 or -1) if the denominator is 0
+        division_result = np.where(
+            denominator > 0,
+            np.divide(x_minus_m, denominator, where=denominator > 0),
+            np.sign(x_minus_m),
```
amatissart, line 40: `denominator` is 0 only when both `delta` and `x_minus_m` are 0. So `np.sign(x_minus_m)` can be replaced with 0.

Note that `L_prime` is on the critical path from a performance point of view. So it would be nice to move the complexity outside of this function as much as possible.  
Another approach could be to define `delta_2`  so that it can no longer be zero: 
```py
delta_2 = np.where(delta > 0, delta ** 2, np.spacing(0))
```

amatissart (approved):

Issue #1289, opened 2023-01-17
[front] Refactor `Donate` page


aidanjungo:
We should add more information on this page, e.g.:
- What we do with the money
- How much money we have
- What we would to with more money

lfaucon:
- [ ] Propose a mock-up of what we target for the Donate Page (could be done on MIRO/Figma)

PR #1362, opened 2023-02-15
[front] feat: rework the donation page layout

OoIwazaruoO:
**related to** #1289

---

This PR reworks the donation page layout.

The page now uses the generic `<TitledPaper>` to display the different ways of contributing. Also the uTip and PayPal links have been made more visible to make it clear what is expected from the users.

The next PRs will expand the page to describe what is done with the received money.

**to-do**
- [x] review the spacing between the components (must be coherent across the page)
- [x] review the wording
- [x] split the translation of the `ContentHeader`

### preview

screencapture-localhost-3000-about-donate-2023-02-21-12_12_41


frontend/src/pages/about/Donate.tsx
```diff
@@ -41,130 +32,130 @@ const useStyles = makeStyles((theme) => ({
...
-                rel="noreferrer"
-                className={classes.link}
-              >
-                Discord
-              </a>
-              .
-            </Trans>
-          </Typography>
+      <ContentHeader title={t('about.donate')} />
+      <FundingSection linkToSupportPage={false} />
```
GresilleSiffle, line 40: We should move `<FundingSection>` inside `<ContentBox maxWidth="lg">` to include it in the page container. It will automatically benefits from the margins and padding defined by `ContentBox`.

GresilleSiffle (approved):

GresilleSiffle:
To simplify the page, we took the time to replace the `<div>` tags with their custom CSS by standard Material UI components.

aidanjungo:
It looks nice, two small comments:

- I would reduce a bit the size of the utip and paypal logo, they look big in comparison of the buttons.
- Should be add a comment to say that the direct transfer should be preferred (because it means less fees) or maybe put the direct transfer section first.

GresilleSiffle:
> I would reduce a bit the size of the utip and paypal logo, they look big in comparison of the buttons.

Sure. We made them big on purpose, but we can of course adjust their size. I can reduce the size from 110 pixels to 90.

We used big logos to fill the available horizontal space, to make the page appear less empty on desktop, and nice on mobile. We also made them click-able, the buttons are just here in case the users don't understand it's possible to click on the logo.

> Should be add a comment to say that the direct transfer should be preferred (because it means less fees) or maybe put the direct transfer section first.

Yes I wanted to talk about this. I personally prefer using direct transfers but I don't know if they are user friendly enough to be displayed first. On one hand I think uTip and PayPal are easy to use, in few clicks it's possible to make a donation (if the users have already an account). On the other hand direct transfers require more steps to set up, but are more efficient.

We can:
- display uTip/PayPal first (current)
- display direct transfers first
- display uTip/PayPal first and add an info MUI Snackbar inviting users to check direct transfer if they want to be more efficient

PR #1366, opened 2023-02-16
[back] refactor: Add `time_ahead` function and harmonize

aidanjungo:
- I created a `time_ahead` function which is the opposite of the existing `time_ago` function.
- I used those two function when it was possible

### DIFF
--- a/backend/core/utils/tests/test_time.py
+++ b/backend/core/utils/tests/test_time.py
@@ -1,4 +1,4 @@
-from core.utils.time import time_ago
+from core.utils.time import time_ago, time_ahead
 
 
 def test_time_ago():
@@ -8,3 +8,12 @@ def test_time_ago():
     assert time_ago(hours=10) > time_ago(days=1)
     assert time_ago(days=1) > time_ago(days=1, hours= 3)
     assert time_ago(days=6, hours=23) > time_ago(weeks=1)
+
+
+def test_time_ahead():
+    """Test method time_ahead."""
+    
+    assert time_ahead(hours=1) < time_ahead(hours=10)
+    assert time_ahead(hours=10) < time_ahead(days=1)
+    assert time_ahead(days=1) < time_ahead(days=1, hours= 3)
+    assert time_ahead(days=6, hours=23) < time_ahead(weeks=1)
--- a/backend/core/utils/time.py
+++ b/backend/core/utils/time.py
@@ -11,7 +11,7 @@ def time_ago(**kwargs):
     return timezone.now() - timedelta(**kwargs)
 
 
-def time_ahead(**kwargs):
+def time_ahead(**kwargs) -> datetime:
     """
     Return a `datetime` in the future relative to now.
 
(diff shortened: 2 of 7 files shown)


backend/core/utils/time.py
```diff
@@ -9,3 +9,9 @@ def time_ago(**kwargs):
     """Return a time in the past as specified as keyword arguments."""
 
     return timezone.now() - timedelta(**kwargs)
+
+
+def time_ahead(**kwargs):
+    """Return a time in the future as specified as keyword arguments."""
```
GresilleSiffle, line 15: Perfect. Let's make the documentation more accurate.

```suggestion
    """
    Return a `datetime` in the future relative to now.

    The keyword arguments are directly passed to `datetime.timedelta`.
    """
```

backend/core/utils/time.py
```diff
@@ -9,3 +9,9 @@ def time_ago(**kwargs):
     """Return a time in the past as specified as keyword arguments."""
 
     return timezone.now() - timedelta(**kwargs)
+
+
+def time_ahead(**kwargs):
```
GresilleSiffle, line 14: The following suggestion requires importing datetime in addition to timedelta.

```suggestion
def time_ahead(**kwargs) -> datetime:
```

backend/core/utils/time.py
```diff
@@ -9,3 +9,9 @@ def time_ago(**kwargs):
     """Return a time in the past as specified as keyword arguments."""
```
GresilleSiffle, line 9:
```suggestion
    """
    Return a `datetime` in the past relative to now.

    The keyword arguments are directly passed to `datetime.timedelta`.
    """
```

GresilleSiffle (commented):
Thank you :ok_hand: 

I made some suggestions to make the functions' documentation more accurate. We will be ready to merge after that.

backend/core/utils/tests/test_time.py
```diff
@@ -8,3 +8,12 @@ def test_time_ago():
     assert time_ago(hours=10) > time_ago(days=1)
     assert time_ago(days=1) > time_ago(days=1, hours= 3)
     assert time_ago(days=6, hours=23) > time_ago(weeks=1)
+
+
+def test_time_ahead():
```
GresilleSiffle, line 13: Out of scope: in the future, if we want to make those tests more robust, we could freeze the time and compare the results of `time_ago` and `time_ahead` to now. Currently those tests don't check that the returned times are relative to now, which is a key part of the implementation of `time_ago` and `time_ahead. 

For instance we could check that `time_ahaed(hours=1)` returns a `datetime` object 3600 seconds in the future.

aidanjungo: Yes I thought the same, but I wasn't sure it it was overkill to test that

GresilleSiffle: The functions are really small indeed, but it will be quite easy to test their relativity to now. In my opinion it's not overkill. Is it worth the invested time? That's also a question : ]

GresilleSiffle (approved):

PR #1371, opened 2023-02-21
[back] refactor: `get_comparisons_data` now takes a datetime as argument

GresilleSiffle:
This PR adds a datetime argument to `get_comparisons_data`. 

The `get_comparisons_data` is not limited anymore to the comparisons made during the previous week, and is now able to retrieve up-to-date data.

I took the time to add this argument because the previous implementation was a bit confusing to me, and might be confusing for new developers too. At a first glance, it was not obvious that calling a function named `get_comparisons_data` will be limited to the comparisons of the previous week. Having an explicit datetime parameter will allow the developers to understand what the function does without the need of reading its implementation.

```python
# all* public comparisons
get_comparisons_data('videos')

# all public comparisons until 2 days ago
get_comparisons_data('videos', time_ago(days=2))

# all public comparisons until the specified datetime
get_comparisons_data('videos', datetime)

# *comparisons made exactly during the function call are excluded
```


backend/tournesol/tests/test_exports.py
```diff
@@ -40,6 +40,8 @@
         {"DATASETS_BUILD_DIR": "ts_api_test_datasets"}, settings.APP_TOURNESOL
     ),
 )
+
+
 class ExportTest(TestCase):
     @MockNow.Context()
```
tbrnd, line 46:
```suggestion
```

tbrnd (approved):
@GresilleSiffle I took the liberty of resolving the conflicts, PR LGTM 👍

GresilleSiffle:
Thank you @tbrnd :+1:

PR #1375, opened 2023-02-21
[back] docs: the install doc invites explicitly the users to configure the settings

bbosson:
Fixed little errors in the readme when setup project

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -50,13 +50,16 @@ Django application; and how to install and configure a PostgreSQL server.
   [Linux](https://www.postgresqltutorial.com/install-postgresql-linux/)).
 
 - Create a config file named `settings-tournesol.yaml`. You can find an
-  [example](backend/documentation/settings-tournesol.yaml) in documentation
+  [example](documentation/settings-tournesol.yaml) in documentation
   folder. You can put this file in different locations:
     - `/etc/tournesol/settings-tournesol.yaml` (linux),
     - Anywhere but add the `path/to/file` in SETTINGS_FILE environment
       variable,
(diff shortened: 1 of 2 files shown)


backend/README.md
```diff
@@ -50,13 +50,16 @@ Django application; and how to install and configure a PostgreSQL server.
...
-  [example](backend/documentation/settings-tournesol.yaml) in documentation
+  [example](documentation/settings-tournesol.yaml) in documentation
   folder. You can put this file in different locations:
     - `/etc/tournesol/settings-tournesol.yaml` (linux),
     - Anywhere but add the `path/to/file` in SETTINGS_FILE environment
       variable,
     - Anywhere but add the `path/to/file` in SETTINGS_FILE variable `.env`
       file.
+  
+- Add in the file `settings-tournesol.yaml`, custom values for the variable `CORS_ALLOWED_ORIGINS`
```
tbrnd, line 61: For my understanding, why was this needed?

bbosson: This is a variable to be customized because in the example provided (documentation/settings-tournesol.yaml) this variable is empty. It is necessary to satisfy the CORS when calling Front to Back

GresilleSiffle: That's right, the back end only accepts cross origin requests if their origins have been explicitly configured.

Thank you @bbosson for taking the time to improve the documentation. As they are other variables that should be configured like `MEDIA_ROOT`, `STATIC_ROOT`, etc. I think we could invite the reader to simply configure the settings as they are described in the `documentation/settings-tournesol.yaml`.

If you don't mind, I'm going to adapt the README now : )

GresilleSiffle (approved):

PR #1377, opened 2023-02-22
[back] refactor: round floating point numbers of the dataset

GresilleSiffle:
As previously discussed, this PR rounds the computed individual and collective scores in the dataset.

I chose a two digits precision after the decimal point.

### DIFF
--- a/backend/tournesol/lib/public_dataset.py
+++ b/backend/tournesol/lib/public_dataset.py
@@ -13,16 +13,16 @@ from django.utils import timezone
 
 from tournesol.entities.base import UID_DELIMITER
 
-# The maximal decimal precision of floating point numbers appearing in the
-# dataset.
+# The standard decimal precision of floating point numbers appearing in the
+# dataset. Very small numbers can use a higher precision.
 FLOAT_PRECISION = 2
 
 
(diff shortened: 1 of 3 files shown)


backend/tournesol/lib/public_dataset.py
```diff
@@ -288,7 +299,7 @@ def write_individual_criteria_scores_file(poll_name: str, write_target) -> None:
             "public_username": criteria_score.username,
             "video": criteria_score.uid.split(UID_DELIMITER)[1],
             "criteria": criteria_score.criteria,
-            "score": criteria_score.score,
+            "score": round(criteria_score.score, FLOAT_PRECISION),
             "voting_right": criteria_score.voting_right,
```
amatissart, line 303: The individual voting right could be rounded too. But the smallest value is currently ~ 0.0184, and even smaller values could be reached in the future, so I think we should keep at least 3 digits here.

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1377/commits/11e601b8d432bde585ace16d69d3cc1e4f415927

amatissart (approved):

PR #1379, opened 2023-02-22
[front] feat: add the search bar on mobile

OoIwazaruoO:
Thanks @aidanjungo and @lfaucon for the review

- [x] Find a better way to display the login/register buttons on very small screen (Demo video down the list)
- [x] new components for LargeScreenTopBar and MobileTopBar
- [x] Multiple condition display replaced by a ternary
- [x] Logo and Search Components moved into single file

https://user-images.githubusercontent.com/50485333/221577971-d980aa4b-5be0-4700-b76b-c4f19901f26c.mp4

### DIFF
--- a/frontend/src/features/frame/components/topbar/TopBar.tsx
+++ b/frontend/src/features/frame/components/topbar/TopBar.tsx
@@ -2,8 +2,8 @@ import React from 'react';
 
 import AppBar from '@mui/material/AppBar';
 import Grid from '@mui/material/Grid';
-import Hidden from '@mui/material/Hidden';
 import Toolbar from '@mui/material/Toolbar';
+import useMediaQuery from '@mui/material/useMediaQuery';
 
 import { useTheme } from '@mui/material';
 import { useCurrentPoll } from 'src/hooks/useCurrentPoll';
@@ -18,6 +18,7 @@ export const topBarHeight = 80;
 const TopBar = () => {
   const theme = useTheme();
   const { options } = useCurrentPoll();
+  const isSmallScreen = useMediaQuery(theme.breakpoints.down('md'));
 
   return (
     <AppBar
@@ -39,12 +40,7 @@ const TopBar = () => {
         }}
       >
         <Grid container sx={{ width: '100%' }} alignItems="center">
-          <Hidden mdDown>
-            <TopBarDesktop />
-          </Hidden>
-          <Hidden mdUp>
-            <TopBarMobile />
-          </Hidden>
+          {isSmallScreen ? <TopBarMobile /> : <TopBarDesktop />}
         </Grid>
       </Toolbar>
     </AppBar>
--- a/frontend/src/features/frame/components/topbar/TopBar.tsx
+++ b/frontend/src/features/frame/components/topbar/TopBar.tsx
@@ -67,6 +67,7 @@ const Logo = () => {
     <Grid
       item
       md={4}
+      sm={2}
       xs={3}
       sx={{
         display: 'flex',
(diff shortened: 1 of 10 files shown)


aidanjungo:
It look very nice, I'm looking forward that we deploy this features. I haven't check the code, just tested it. 
I notice when the width of the screen is very narrow the space on the left feel a bit too small:
image

frontend/src/features/frame/components/topbar/TopBar.tsx
```diff
@@ -140,14 +152,41 @@ const TopBar = () => {
...
+                  <IconButton
+                    aria-label="Close the searchbar"
+                    onClick={() => searchOpeningHandler(false)}
+                  >
+                    <CloseIcon />
+                  </IconButton>
+                </Grid>
+              </Grid>
+            )}
+            {!mobileSearchOpen && <AccountInfo />}
```
lfaucon, line 188: Code Style:

I would find it easier to read with a ternary operator rather than repeating `!mobileSearchOpen && ...`. With the code below, I see more easily what's there when `mobileSearchOpen` is `true`or when it's `false`

```
            {mobileSearchOpen 
             ?
             <Grid container width="100%">
                <Grid item xs={11}>
                  {options?.withSearchBar && <Search />}
                </Grid>
                <Grid item xs={1}>
                  <IconButton
                    aria-label="Close the searchbar"
                    onClick={() => searchOpeningHandler(false)}
                  >
                    <CloseIcon />
                  </IconButton>
                </Grid>
              </Grid>
              : <>
                <Logo />
                <IconButton
                aria-label="Open the searchbar"
                onClick={() => searchOpeningHandler(true)}
              >
                <SearchIcon fontSize="large" />
              </IconButton>
              <AccountInfo />
               </>
              }
```

frontend/src/features/frame/components/topbar/TopBar.tsx
```diff
@@ -140,14 +152,41 @@ const TopBar = () => {
...
-          <Logo />
+        <Grid container sx={{ width: '100%', maxWidth: '100%' }}>
           <Hidden mdDown>
+            <Logo />
             <Grid item md={4}>
               {options?.withSearchBar && <Search />}
             </Grid>
+            <AccountInfo />
+          </Hidden>
+          <Hidden mdUp>
```
lfaucon, line 163: Code Style:

Because the `mdDown` and `mdUp` are now separated, maybe it would be useful to create new components `MobileTopBar` and `LargeScreenTobBar` ?

Then
```
const TopBar = () => {
  [...]

  return (
        [...]
        <Grid container sx={{ width: '100%', maxWidth: '100%' }}>
          <Hidden mdDown>
            <LargeScreenTopBar />
          </Hidden>
          <Hidden mdUp>
            <MobileTopBar />
          </Hidden>
        </Grid>
        [...]
    )
```

lfaucon (changes requested):
I commented some minor code style suggestions, but it's probably only my personal preference so find to be ignored :)


I'm requesting changes because the Loged-out case is not well handled (see screenshot)
Screenshot 2023-02-26 at 20 31 01

GresilleSiffle:
There is not enough horizontal space on very small screens to display everything on the single line.

We talked about it one day, instead of displaying two buttons log in and sign up on very small screens, we could instead display a single button [log in] that opens a menu with the items log in and sign up. Exactly like the current profile button. On larger screens we can keep the current display.

What do you think  @lfaucon @OoIwazaruoO ?

GresilleSiffle (approved):
Good to me.

In a following pull request we may want to add a placeholder text to the search input. This text will be defined per poll.

@lfaucon do you want to have a final look?

lfaucon (approved):

Issue #1001, opened 2022-06-19
[front] feat: multi-page lists navigation 


lfaucon:
The pagination component should allow to quickly access any page. 

Common usecases would be:
- access old comparisons
- watch page 5 of tournesol all time best two days after having watched page 1,2,3,4
- intentionally find videos with lower scores

GresilleSiffle:
We talked also about having the pagination at the end and at the top of video lists to not force the users to scroll down to access the pages.

I could be a good addition, what do you think @lfaucon ?

GresilleSiffle:
**(a)** For instance, the Django admin works as follow:

- you can always access the two first pages
- you can always access the two last pages
- you can always access the next three pages
- you can always access the previous three pages

![capture](https://user-images.githubusercontent.com/39056254/192972100-d2816d85-eef3-4cd4-87ac-acf9d2e4b522.png)

**(b)** There is also a more complex behaviour, when the user is less than 6 pages close to the first or the last page, all the remaining pages are displayed. 

![capture](https://user-images.githubusercontent.com/39056254/192972305-ed661f2a-4dd1-4053-824f-4a35ba89c3b8.png)

What about implementing the behaviour **(a)** to start? Then we could add -100 / -10 / +10 / +100 additional buttons to navigate more quickly, or maybe a text input allowing to type a page number.

If we don't add a text input, the - X / + X buttons can be useful especially when navigation big lists like the all time recommendations, the rate-later list, the comparisons list, etc.

bbosson:
We agreed to implement behavior (a) and add additional -100 / -10 / +10 / +100 buttons to navigate faster, or perhaps a text input to type a page number.

PR #1381, opened 2023-02-22
[front] multi-page lists navigation (#1001)

bbosson:
Refactoring of the Pagination component to present the list of pages.
Allow quick navigation to 3 pages before or 3 pages after.
Ability to go to -100 -10 +10 +100 pages
Added a field to enter the page number.

Pagination on large screens: 
image

Pagination on small screens: 
image

Requested change in issue #1001

### DIFF
--- a/dev-env/run-docker-compose.sh
+++ b/dev-env/run-docker-compose.sh
@@ -65,7 +65,7 @@ function wait_for() {
       return 0
     fi
     echo "Waiting for $service_name to be ready..."
-    sleep 3
+    sleep 4
   done
   echo "$service_name is unreachable."
   exit 1
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -39,12 +39,6 @@
   "score": {
     "totalScoreBasedOnRankingChoice": "Score based on selected ranking parameters"
   },
-  "pagination": {
-    "videos": "videos",
-    "previousButton": "Previous {{limit}}",
-    "nextButton": "Next {{limit}}",
-    "page": "Page :"
-  },
   "personalCriteriaScores": {
     "reasonWhyItCannotBeActivated": {
       "notLoggedIn": "You must be logged in to display personal scores.",
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -42,9 +42,8 @@
   "pagination": {
     "videos": "videos",
     "previousButton": "Previous {{limit}}",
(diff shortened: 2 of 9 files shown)


GresilleSiffle:
Thank you @bbosson for this pull request :+1: 

I'm going to review it properly today.

There is something we can already improve: the spacing between the buttons -100 / -10 and +10 / +100. The buttons are currently stuck to each other and the UI will be more beautiful if we use either:
- a spacing of 8 or 16 pixels between them
- or a Material UI button group, https://mui.com/material-ui/react-button-group/

The spacing 8 / 16 has currently my personal preference but we can also try the button group.

bbosson:
Thank you for this first feedback. I am looking forward to the next ones and we can also discuss this at lunchtime. 

I have updated the screenshot.

amatissart:
At first glance, the most visible button on the screenshot is "10 suiv >", whereas the most common action is to go to the next page. Should we also add buttons "+1" and "-1" to go to the next/previous page?

Let's also keep in mind the component should be usable on mobile. On small screens, I would suggest to hide the input field and show less buttons.
image

bbosson:
Pull request closed for next developpement of pagination responsive support

GresilleSiffle:
On mobile we could have a completely different layout. 

Proposition: display the - 100 / - 10 / prev / next / + 10 / +100 on a first line, and the list of pages just below.

To save space on the second line we can reduce the number of page buttons displayed on mobile. To save space on the first line we could replace the labels `prev` / `suiv` maybe by symbols?

What do you think @amatissart @bbosson ?

bbosson:
Absolutely, in fact Louis made exactly the same proposal to the daily.

And don't make the buttons disabled by default.

I remove the Page field on small screens.

bbosson:
I have processed the various feedbacks in a new version. Could you tell me what should be changed now?

Pagination on large screens: 
image

Pagination on small screens: 
image

aidanjungo:
I haven't check the code, just the behavior on every page, it looks great, I just find a small layout issue on the `Rate later` page, with only 4 video on my list, it look like this:

image

 Is it normal that button are display when there is only one page?

bbosson:
> I haven't check the code, just the behavior on every page, it looks great, I just find a small layout issue on the `Rate later` page, with only 4 video on my list, it look like this:
> 
> image
> 
> Is it normal that button are display when there is only one page?

Oh no, I had looked at this point at the beginning but I forgot to retest, I'm correcting.

GresilleSiffle:
Thank you @bbosson for your work. We are currently reviewing the code with the team and we will merge soon.

GresilleSiffle (approved):

PR #1382, opened 2023-02-23
[front] fix redirection from legacy params on comparison page with autofill

amatissart:
When redirecting from legacy params videoA and videoB, the state `selectorA` and `selectorB` was already initialized with empty uids. That breaks assumptions made by the video selectors when using the autofill feature.

### DIFF
--- a/frontend/src/features/comparisons/Comparison.tsx
+++ b/frontend/src/features/comparisons/Comparison.tsx
@@ -1,6 +1,7 @@
 import React, { useState, useEffect, useCallback, useMemo } from 'react';
-import { Redirect, useHistory, useLocation } from 'react-router-dom';
+import { useHistory, useLocation } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
+import { Location } from 'history';
 
 import { CircularProgress, Grid, Typography, Card } from '@mui/material';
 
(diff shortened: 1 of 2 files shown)


amatissart:
An end-to-end test has been added, but I had to use a fixed delay to trigger the bug. So that's not perfect. But hopefully this legacy implementation and the related tests will be removed soon.

OoIwazaruoO:
I don't know if you have the same issue but when I start a comparison from the browser-extension on the localhost the metadata from the video A do not appear. 
noMetadata

amatissart:
> I don't know if you have the same issue but when I start a comparison from the browser-extension on the localhost the metadata from the video A do not appear. 

That's probably because no YouTube API key was configured in your dev-env. Without this key, videos that are not present in the local database are created without any metadata from Youtube. If you select a video that already exists in the dev-env database, the metadata should be visible as expected.
See instructions in https://github.com/tournesol-app/tournesol/tree/main/backend#set-up-a-google-api-key to setup an API key.

So it's not related to this change.

OoIwazaruoO:
> > I don't know if you have the same issue but when I start a comparison from the browser-extension on the localhost the metadata from the video A do not appear.
> 
> That's probably because no YouTube API key was configured in your dev-env. Without this key, videos that are not present in the local database are created without any metadata from Youtube. If you select a video that already exists in the dev-env database, the metadata should be visible as expected. See instructions in https://github.com/tournesol-app/tournesol/tree/main/backend#set-up-a-google-api-key to setup an API key.
> 
> So it's not related to this change.

Ok thanks I will check that

OoIwazaruoO (approved):

Issue #1070, opened 2022-07-24
feat: Handle deletion of videos on Youtube


lfaucon:
It's not very clear yet all the parts of Tournesol that are impacted and should be impacted in the case of a video deleted by the content creator or removed from Youtube for any reason. This nevertheless gives bad user experience, especially if one such video comes up during the tutorial.

1. Understand all the parts of Tournesol that may be ipacted
2. Decide what behaviour we want Tournesol to have
3. Plan and execute implementation

aidanjungo:
Should we detect that automatically or let user signal that?

lfaucon:
Goal would be automated detection (for example at the same time that we currently refresh a video metadata). If that does not work, then as a last resort letting user signal it, but then we still need trusted moderators to verifies the signals.

lfaucon:
- [ ] First simple solution to investigate and implement: The frontend can find out that the video has been deleted (for example because the thumbnail is missing: https://i.ytimg.com/vi/PN8-fBE50Yg/mqdefault.jpg). Recommendation lists we could collapse the videocard (with a message "deleted video (expand)"). The extension could simply remove the deleted video and show 3 instead of 4 or 11 instead of12 recommendations.

amatissart:
If the title is not displayed, and the thumbnail is not available either, is it worth displaying any other metadata at all?  The user would not be able to recognize the video, even if they compared it in the past.

In this case, I would just keep a minimal collapsed video card without the possibility to expand it.

GresilleSiffle:
**proposition**

1. implement a method allowing the front end to detect when an entity is not available anymore

2. do not display non available entities in the public and individual recommendations

3. display a folded entity card in the rate-later list, in order to inform the user this video is not available anymore. the only action possible is to delete the video

4. display a folded entity card in the My compared items page?

5. rework the video analysis page?

6. rework the comparison page to prevent comparing non available entities?

GresilleSiffle:
I took some screenshots to better visualize the recommendations.

---

![capture2](https://user-images.githubusercontent.com/39056254/220928121-1d49f94b-98cc-4e3e-98cc-773671392244.png)

As a user knowing there was a video here, with an unknown name, that is not available anymore doesn't seem relevant or useful. It can be confusing also because it looks like a title indicating the video above or below is not available.

---

![capture](https://user-images.githubusercontent.com/39056254/220928621-24bfb468-5859-448e-8f33-ee1651ee0e5c.png)

[...]

PR #1384, opened 2023-02-23
[front] feat: hide unavailable videos in list

OoIwazaruoO:
**related to** #1070

---

This PR has for goal to find and apply a better way to display a video in `<EntityList>` when it is not available anymore. For the moment I focused on the display of unavailable videos on the list of recommendations, video ratings and rate later pages.

On the recommendation and video ratings page a  "video unavailable" message is displayed with the possibility to see the unavailable entity by toggling a button. The rate later page display a delete button in addition of the message so we can still delete the video from the rate later list.

- [x] Make AvailableEntity work for more types than just videos (only check availability of video for the moment)
- [x] Add actionsIfUnavailable props to the  EntityList and  AvailableEntity component
- [x] Pass the actionsIfUnavailable from EntityList to AvailableEntity
- [x] Pass the actionsIfUnavailable props from AvailableEntity into  EntityCard actions  props when the entity is unavailable

### preview

In the rate-later list, actions are displayed.

unavailable

unavailable2

On other lists, no action is displayed.

unavailable3

unavailable4

### DIFF
--- a/frontend/src/components/entity/AvailableEntity.tsx
+++ b/frontend/src/components/entity/AvailableEntity.tsx
@@ -1,20 +1,52 @@
 import React, { useState, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 
-import { Box, Grid } from '@mui/material';
+import { Box, Grid, Typography } from '@mui/material';
 import { idFromUid } from '../../utils/video';
 import { entityCardMainSx } from './style';
+import { ActionList } from 'src/utils/types';
+import { TypeEnum } from 'src/services/openapi';
 
-export const EntityNotAvailable = ({ type }: { type: string }) => {
+/*
+ * return a different error element related to the type of the entity and actions
+ */
+export const EntityNotAvailable = ({
+  type,
+  unavailableActions,
+  uid,
+}: {
+  type: string;
+  unavailableActions?: ActionList;
+  uid: string;
+}) => {
   const { t } = useTranslation();
 
-  if (type == 'video') {
+  if (type == TypeEnum.VIDEO) {
     return (
       <Box mx={1} my={2}>
-        <Grid container sx={entityCardMainSx}>
-          <Box mx={1} my={2}>
-            {t('video.notAvailableAnymore')}
-          </Box>
+        <Grid
+          mx={1}
+          my={2}
+          container
+          sx={entityCardMainSx}
+          justifyContent="space-between"
+          alignItems="center"
+        >
+          <Grid ml={1} my={2}>
+            <Typography >
+              {t('video.notAvailableAnymore')}
+            </Typography>
+          </Grid>
+          <Grid justifyContent="center" alignItems="flex-end">
+            {unavailableActions &&
+              unavailableActions.map((Action, index) =>
+                typeof Action === 'function' ? (
+                  <Action key={index} uid={uid} />
+                ) : (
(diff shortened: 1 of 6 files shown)


frontend/src/features/entities/EntityList.tsx
```diff
@@ -14,6 +15,26 @@ interface Props {
...
 }
 
+const AvailableEntity = ({
+  children,
+  uid,
+}: {
+  children: React.ReactNode;
+  uid: string;
+}) => {
+  const [isAvailableOnYoutube, setIsAvailableOnYoutube] = useState(false);
```
amatissart, line 25: The `EntityList` can be used with any `entity.type`. So this component should not assume the entity is a Youtube video.

frontend/src/features/entities/EntityList.tsx
```diff
@@ -14,6 +15,26 @@ interface Props {
...
+
+  useEffect(() => {
+    const img = new Image();
+    img.src = `https://i.ytimg.com/vi/${idFromUid(uid)}/mqdefault.jpg`;
+    img.onload = function () {
+      setIsAvailableOnYoutube(img.width !== 120);
+    };
```
amatissart, line 32: If an additional request is required to check the video availability, then it would be more reliable to use the HTTP status of the response with `fetch()`. The thumbnail of an unavailable video returns a 404 error.

OoIwazaruoO: I've just tried to make it work with fetch but I ran in a cors policy wall.

GresilleSiffle:
Note that in the ratings page, the users lose the possibility to change the visibility public/private for ratings of unavailable entities.

GresilleSiffle (approved):
The implementation is OK for me.

Do we all agree with the new behaviour ? In the lists, a message is displayed instead of unavailable entities. No metadata, no action (except remove from the rate-later list).

@amatissart @lfaucon @aidanjungo

aidanjungo:
> Do we all agree with the new behavior ? In the lists, a message is displayed instead of unavailable entities. No metadata, no action (except remove from the rate-later list).

Personally I would have prefer to have the possibility to access info (at least channel and title). I hate when a video is removed from YouTube and I have no way to know which video it was.

lfaucon (approved):
> Do we all agree with the new behaviour ? In the lists, a message is displayed instead of unavailable entities. No metadata, no action (except remove from the rate-later list).

I'm fine with the new behaviour. I also agree with @aidanjungo. Possibly a button to extend that would show all the information that was previously available and saved in our backend is useful. Any reason to not give access to it at all?

frontend/src/components/entity/AvailableEntity.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useState, useEffect } from 'react';
+
+import LoaderWrapper from 'src/components/LoaderWrapper';
+import { TypeEnum } from 'src/services/openapi';
+import { ActionList } from 'src/utils/types';
+import { idFromUid } from 'src/utils/video';
+
+/**
+ * Return an <EntityNotAvailable> if the entity doesn't seem to be available
```
GresilleSiffle: Let's rework this comment. The mentionned `<EntityNotAvailable>` component doesn't exist anymore.

GresilleSiffle (approved):

PR #1386, opened 2023-02-25
[infra] Add timers to watch account creation

aidanjungo:
- [x] Add timers to run it every 1 hour and every day
- [x] Create a private Discord channel for these alert and get the webhook
- [x] Add the webhook name in the `settings` file
- [ ]  Add the webhook to the secrets
- [ ] Test on staging

I add two alerts, which will pop in a private Discord channel (to avoid leaking some mail domain):

- hourly: if more than 5 account are created with a trusted email domain
- daily: if more than 10 account are created with a trusted email domain

I think it will not raise alert too often, but those values could be change in the future.


infra/ansible/roles/django/tasks/main.yml
```diff
@@ -272,6 +272,44 @@
     enabled: yes
     daemon_reload: yes
 
+# scheduled task: watch account creation hourly
+
+- name: Copy Tournesol API watch-account-creation-1h service
+  template:
+    dest: /etc/systemd/system/tournesol-api-watch-account-creation-1h.service
+    src: tournesol-api-watch-account-creation-1h.service.j2
```
GresilleSiffle, line 280: I need to check, but I think we can use the same template for the 1h and 24h service.

Instead of having two .service.j2 file in Ansible, we could have only one using a template variable for `{{ since_n_hours }}`.

infra/ansible/roles/django/templates/tournesol-api-watch-account-creation-1h.j2
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Tournesol API inactive users clean-up
```
GresilleSiffle, line 2:
```suggestion
Description=Tournesol API watch accounts created during the last 1 hour 
```

infra/ansible/roles/django/templates/tournesol-api-watch-account-creation-1h.timer.j2
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Tournesol API inactive users clean-up
```
GresilleSiffle, line 2:
```suggestion
Description=Tournesol API watch accounts created during the last 1 hour 
```

infra/ansible/roles/django/templates/tournesol-api-watch-account-creation-24h.j2
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Tournesol API inactive users clean-up
```
GresilleSiffle, line 2:
```suggestion
Description=Tournesol API watch accounts created during the last 24 hours 
```

GresilleSiffle (approved):

PR #1389, opened 2023-02-27
[back] add machine name for email alerting

aidanjungo:


aidanjungo:
I don't know i should use the name of host in the test? If it is also imported from settings, is it useful?

GresilleSiffle:
> I don't know i should use the name of host in the test? If it is also imported from settings, is it useful?

I think we can. This should not cause any problem.

See: https://github.com/tournesol-app/tournesol/pull/1389/commits/e67ddc3edf0a586659ce13e8844f5af1c708c956

GresilleSiffle (approved):

Issue #1319, opened 2023-01-28
[extension] fix: extension buttons break responsive layout on medium screens on Youtube. 


raphaelp-github:
The like button is superimposed on the subscribe button.

![image](https://user-images.githubusercontent.com/34160362/215267810-ebcf9a91-3883-4a19-8733-743b27c4860a.png)

Screen size : 14"

Thank's

lfaucon:
Thanks for the reporting!

abhiram11:
We can remove the text `Rate Later` and `Rate Now` and just keep the icons there, but that would require manipulation of text nodes which is a bit tricky.

I can try to implement it, I'll wrap the text node with a `span` or something and then try to hide the text only for screen sizes between ~ sm:md (`656px to 900px` would do)

abhiram11:
Alternatively, adding this snippet below in addRateButtons.css gives the below output, it still needs more specificity.

```
@media (min-width: 656px) and (max-width: 900px) {
  button.tournesol-rate-button {
    font-size: 0;
  }
}
```
![image](https://user-images.githubusercontent.com/20269286/221615388-29ea01ba-cb16-4692-b42a-94d9ba3e80c9.png)

lfaucon:
Removing the text and keeping the icons looks good!

The layout issue occurs also on screens of width 1100px:

![Screenshot 2023-02-27 at 21 30 06](https://user-images.githubusercontent.com/8818081/221678534-aa37da6b-71aa-4a66-a2a0-767ef58f08e8.png)

It seems to start occuring from about 1250px until 1000px , then from 999px the layout changes and it looks ok again (see screenshot)

![Screenshot 2023-02-27 at 21 33 14](https://user-images.githubusercontent.com/8818081/221679348-646b5d79-a818-435a-999b-d11984e383da.png)

Then it starts to look bad around 850px until 639px, then the layout changes again at 638px:

![Screenshot 2023-02-27 at 21 36 51](https://user-images.githubusercontent.com/8818081/221680307-ee64fd2f-b6c2-49f9-b122-c790d00a5188.png)


As a conclusion I woulc propose hiding the text for the intervals 1250px until 1000px and 850px until 639px, but I'm not completely sure that this generalises to everyone, and it may also break with future versions of Youtube.

Because it is only a small piece of code, I'm happy with it.

If you want to open a PR @abhiram11 that will be very welcomed 👍

abhiram11:
[...]

PR #1390, opened 2023-02-28
[extension] fix: removes the text of rate buttons on some screen sizes

abhiram11:
This PR tries to be a temporary fix to #1319. 

There are too many factors that decide the width at which we should remove the text from the `Rate Later` and `Rate Now` Buttons. They are:
1. If you are subscribed or not. If you are, their button takes more size and has dropdown for Notification settings (All, Personalized, None)
2. If they have a Membership feature set up, this adds another `Join` button to the UI
3. If their channel name is too long.

tournesol Rate

We can add UI customization based on the three factors mentioned above or just go with the pixel numbers decided in this commit that may work well for the general use case. We can at least keep `1017px` value since it is a break point of Youtube UI change as well.

**Please Note:** I found out that during inspect in browser and toggling to Mobile view/Responsive view, the buttons do not show up at all for sizes ~`1420px` and console shows this error:
```
Refused to frame 'https://tournesol.app/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors https://www.youtube.com moz-extension: chrome-extension:".
```

### DIFF
--- a/browser-extension/src/addRateButtons.css
+++ b/browser-extension/src/addRateButtons.css
@@ -27,18 +27,6 @@ button.tournesol-rate-button:disabled {
   background: #ffc800;
 }
 
-@media (min-width: 656px) and (max-width: 990px) {
-  button.tournesol-rate-button {
-    font-size: 0;
-  }
-}
-
-@media (min-width: 1017px) and (max-width: 1366px) {
-  button.tournesol-rate-button {
-    font-size: 0;
-  }
-}
-
 img.tournesol-button-image {
   margin-left: -6px;
   margin-right: 6px;
@@ -48,4 +36,32 @@ img.tournesol-button-image.tournesol-rate-later {
   /* Compensate the extra margins inside the Material UI "Add" icon */
   margin-left: -10px;
   margin-right: 2px;
+}
+
+@media (min-width: 656px) and (max-width: 990px) {
+  button.tournesol-rate-button {
+    font-size: 0;
+  }
+
+  img.tournesol-button-image {
+    margin: 0;
+  }
+
+  img.tournesol-button-image.tournesol-rate-later {
+    margin: 0;
+  }
+}
+
+@media (min-width: 1017px) and (max-width: 1366px) {
+  button.tournesol-rate-button {
+    font-size: 0;
+  }
+
+  img.tournesol-button-image {
+    margin: 0;
+  }
+
+  img.tournesol-button-image.tournesol-rate-later {
+    margin: 0;
+  }
 }
--- a/browser-extension/src/addRateButtons.css
+++ b/browser-extension/src/addRateButtons.css
@@ -27,6 +27,18 @@ button.tournesol-rate-button:disabled {
   background: #ffc800;
 }
 
+@media (min-width: 656px) and (max-width: 990px) {
+  button.tournesol-rate-button {
+    font-size: 0;
+  }
+}
+
+@media (min-width: 1017px) and (max-width: 1366px) {
+  button.tournesol-rate-button {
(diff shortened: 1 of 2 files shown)


lfaucon:
> We can at least keep 1017px value since it is a break point of Youtube UI change as well.

When testing the breakpoints, I found that 1000px was one but not 1017px. Could it be a difference between chrome/firefox ? or something else?

lfaucon:
Please also increment the patch version of the extension

abhiram11:
> > We can at least keep 1017px value since it is a break point of Youtube UI change as well.
> 
> When testing the breakpoints, I found that 1000px was one but not 1017px. Could it be a difference between chrome/firefox ? or something else?

I'll look into this, might be possible that I was on 90% zoom on browser or some other case.

abhiram11:
I'll look into I'll look into this, might be possible that I was on 90% zoom on browser or some other case.

> Please also increment the patch version of the extension

Do you want me to change `version` in `manifest.json`? It is `2.5.1` right now. If patch version means something else please let me know 😅

lfaucon:
> Do you want me to change `version` in `manifest.json`? It is `2.5.1` right now. If patch version means something else please let me know 😅

Correct.

"Patch version" means incrementing it to 2.5.2
"Minor version" would mean incrementing to 2.6.0
But your PR is a small fix, so "patch" is the right thing to increment.

You can look up the concept of semantic versioning if you are not familiar with it. It's an important thing to know about is sodftware engineering.

abhiram11:
Thank you for taking the time for explaining it 😊 It's been a while since I did updates in practice

GresilleSiffle:
Hello @abhiram11, thanks for the pull request. 

I realized we can also display the icons in the center of the buttons easily by applying the media queries to `img.tournesol-button-image` and by setting `margin: 0`.
 
capture

abhiram11:
> Hello @abhiram11, thanks for the pull request.
> 
> I realized we can also display the icons in the center of the buttons easily by applying the media queries to `img.tournesol-button-image` and by setting `margin: 0`.
> 
> capture

Hi @GresilleSiffle , I have updated the CSS as per the requirement. I was a bit confused earlier because I didn't find `tournesol-rate-now-button` being styled anywhere. Now it works as below though: 

tsnl margin

As per the syntax, all the media queries are shifted to the bottom as well.

GresilleSiffle:
Thank you, the new changes look good to me :+1: 

I have one last question about the media queries: with the current CSS between 990px and 1017px the buttons will still appear with their full text. Do we really want/need this interval?

What about this instead?

```css
@media (min-width: 656px) and (max-width: 1000px) {
/* ... */
}

@media (min-width: 1001px) and (max-width: 1366px) {
/* ... */
}
```

Over more simple, what about removing the media feature `min-width`?

```css
@media (max-width: 1366px) {
/* ... */
}
```

abhiram11:
I added the textnode for `990px and 1017px` because without the text in that range, it seemed there was empty space that we can make use of and could be filled.
I'll be happy to implement the CSS without the `min-width` if required.

GresilleSiffle:
I see, it's true that the UI becomes wide again when the resolution goes under ~ 1000 pixels.

I don't have a very strong opinion. The current PR is fine for me, even if I'd personally use only `max-width: 1366px`. It simplifies the code and avoid duplication, at the cost of a less explicit UI between `990px and 1017px`.

Let's keep it as is for now ; )

GresilleSiffle (approved):

abhiram11:
Let me know if things change in the future or if new features are added that affect this styling, would be happy to change it! 🚀

Issue #818, opened 2022-04-18
[back] Youtube Metadata should save full datetime


lfaucon:
We currently only save the date of video publication. Although it is not a game changer, it would be best to save the exact datetime (including hours, minutes and seconds). This is especially useful for example in the filter that fetches the videos from the previous 24 hours.

lfaucon:
@GresilleSiffle Let's tackle this one. Maybe starting from @benjaminlebigot work-in-progress. It should be easy and it would improve filters based on "last 24h" and allow us to do #1214

PR #1391, opened 2023-02-28
[back] feat: added video datetime in metadata

tbrnd:
PR addressing #818

### DIFF
--- a/backend/tournesol/tests/test_api_video.py
+++ b/backend/tournesol/tests/test_api_video.py
@@ -26,25 +26,25 @@ class VideoApi(TestCase):
         User.objects.create(username=self._user, email="user@test")
 
         self.video_1 = VideoFactory(
-            metadata__publication_date="2021-01-01",
+            metadata__publication_date="2021-01-01T19:44:24.686532",
             metadata__uploader="uploader1",
             tournesol_score=1.1,
             rating_n_contributors=2,
         )
         self.video_2 = VideoFactory(
-            metadata__publication_date="2021-01-02",
+            metadata__publication_date="2021-01-02T19:44:24.686532",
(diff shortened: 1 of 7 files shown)


GresilleSiffle:
Thank you @tbrnd 

It looks good to me. I'll just update a little bit the front end before merging, so that the publication date are displayed in a more human friendly way.

capture2

amatissart:
What about existing data? Do we need to add a migration or the new serializer is compatible with the date-only format?

GresilleSiffle:
> What about existing data? Do we need to add a migration or the new serializer is compatible with the date-only format?

The new serializer is compatible with the date-only format, but will consider all videos as published at midnight UTC. I think adding a migration could be good to avoid having different kind of data in the database.

GresilleSiffle:
It looks like I broke the e2e tests with the new date display. I'll fix them.

backend/tournesol/tests/test_api_video.py
```diff
@@ -311,9 +311,9 @@ def test_get_video_date_filters(self):
         self.assertEqual(resp.status_code, 200, resp.data)
         self.assertEqual(resp.data["count"], 1)
 
-        resp = client.get("/video/?date_lte=2021-01-03T00:00:00")
+        resp = client.get("/video/?date_lte=2021-01-03T19:45:00")
         self.assertEqual(resp.status_code, 200)
-        self.assertEqual(resp.data["count"], 3)
+        self.assertEqual(resp.data["count"], 2)
```
tbrnd, line 316: realized this this morning but this test should fail, 3 is expected (19:44:24 < 19:45:00), please do not merge before this is fixed

GresilleSiffle (approved):

PR #1400, opened 2023-03-03
[front][ext] fix: disable Plausible in login iframe, add url tags in pageviews from extension

amatissart:
The browser extension uses an invisible iframe "/login" to fetch or refresh the access token used for authentication.

This PR adds a parameter "?dnt=1" in the URL of this iframe, to disable tracking of this invisible page.

["UTM tags" ](https://plausible.io/docs/manual-link-tagging )are also used on the comparison iframe and links used in the extension menu.

### DIFF
--- a/browser-extension/src/addRateButtons.js
+++ b/browser-extension/src/addRateButtons.js
@@ -102,7 +102,7 @@ function addRateButtons() {
         chrome.runtime.sendMessage({
           message: 'displayModal',
           modalOptions: {
-            src: `https://tournesol.app/comparison?embed=1&uidA=yt%3A${videoId}`,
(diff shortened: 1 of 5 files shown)


GresilleSiffle (approved):
Thank you @amatissart 

Should we also add `?utm_source=extension&utm_medium=menu` to the link used by the function `openTournesolHome`?

I tried the new extension but I was not able to make Plausible filter by `utm_source` and `utm_medium`, even if I saw that the events sent to Plausible contain these parameters.

GresilleSiffle:
I supposed I was not able to see the UTM sources because of this:

> Note that all referral sources are counted only when they start a new session on your site. This is why you don't see all your own referral sources if you click to test several different UTM tagged links at the same time. Only the first one would be included in the "Top Sources" report. You would need to start a new session by for instance waiting more than 30 minutes or using a different device, browser or IP address.

See: https://plausible.io/docs/top-referrers

PR #1404, opened 2023-03-04
add contributors metrics

adrienpacifico:
Ça fait des nouveaux graphs:

### preview

screencapture-localhost-8501-Community-evolution-2023-03-04-18_10_48

### DIFF
--- a/data-visualization/pages/03_Community_evolution.py
+++ b/data-visualization/pages/03_Community_evolution.py
@@ -4,11 +4,10 @@ A page dedicated to the evolution of the Tournesol community.
 List of concepts:
 
     New contributor:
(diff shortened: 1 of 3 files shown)


GresilleSiffle:
As discussed, a more robust way to count the contributions is to count only the ratings of the main criterion `largely_recommended`.

GresilleSiffle (commented):
It looks good to me, what do you think @aidanjungo ?

aidanjungo:
Nice to have these trends, maybe labels could be more descriptive and name of the series could be remove if there is only one

GresilleSiffle (approved):

PR #1413, opened 2023-03-04
[tests] chore: upgrade to cypress 12, optimize e2e tests workflow

amatissart:
Based on #1408 

* Cypress 8 :arrow_right: 12
* The e2e workflow now builds a production bundle to access the frontend, instead of starting the webpack dev server. This should make the job less resource-intensive when running on GitHub CI, and hopefully more robust.
* Tests regarding the frontend and the browser extension runs in parallel in 2 distinct sub-jobs


amatissart:
End-to-end tests runs in about 6 to 8 minutes on this branch, whereas they last between 8 and 12 minutes on "main".

lfaucon (approved):
Works for me. Thanks for the improvements!

Issue #1396, opened 2023-03-02
[back] Include a metadata export including the date and remove root folder


aidanjungo:
Small remaining task from #1143 

- [ ] Add a text file to store metadata such as the date
- [ ] Remove the root folder included in the zip (the name of the folder was written with the date)

@amatissart @lfaucon Should we add the hyper parameters used to run the algo in this file?

GresilleSiffle:
We could use the json format to make the metadata file easily readable with the std Python.

An example (not a spec).

```json
{
  "creation_date": "2023-03-02",
  "generated_by": "https://api.tournesol.app",
  "algorithms": {
    "byztrust": {
      "version": "2023.02.09",
      "parameters": {}
    },
    "mehestan": {
      "version": "2023.02.09",
      "parameters": {}
    }
  }
}
```
- creation_date: creation date of the archive, iso format
- the origin of the dataset, useful if several instances of Tournesol exists (currently staging, production)
- the version key is just an idea for the future, if the algorithms are reworked

lfaucon:
> Should we add the hyper parameters used to run the algo in this file?

Oh yes, that would be great !

PR #1418, opened 2023-03-05
[back] Include a metadata and remove root folder in the public dataset

aidanjungo:
#1396

### DIFF
--- a/backend/ml/inputs.py
+++ b/backend/ml/inputs.py
@@ -60,15 +60,13 @@ class MlInputFromPublicDataset(MlInput):
             dataset_zip, _headers = urlretrieve(dataset_zip)  # nosec B310
 
         with zipfile.ZipFile(dataset_zip) as zip_file:
-            zip_root_dir = next(path for path in zipfile.Path(zip_file).iterdir() if path.is_dir())
-
-            with (zip_root_dir / "comparisons.csv").open(mode="rb") as comparison_file:
+            with (zipfile.Path(zip_file) / "comparisons.csv").open(mode="rb") as comparison_file:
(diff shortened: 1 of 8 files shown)


aidanjungo:
Pending questions:

- What is the best way to include the commit hash in the metadata?
- Which algo parameter should be included in the metadata?
- Should be gather all the these parameters in a single file? (maybe not relevant if we decide to create a `tournesol-algo` package)

GresilleSiffle:
> Should be gather all the these parameters in a single file? (maybe not relevant if we decide to create a tournesol-algo package)

If the question is: should we write all those parameters in a single metadata file? my answer is: yes I think it's useful for now. Even if the algorithms are part of an external python package, they might be configurable, and in this case providing the configuration in the dataset output is useful.

GresilleSiffle:
> What is the best way to include the commit hash in the metadata?

We can write the hash in the `/etc/tournesol/settings.yml` in a Django setting `TOURNESOL_VERSION`.

aidanjungo:
@amatissart if you have idea of other algo parameters that we should include in the metadata, let me know

amatissart (approved):
LGTM

@aidanjungo  I added a few more parameters defined in "mehetan/run.py" and "trust_algo.py".  I also moved `OVER_TRUST_BIAS` and `OVER_TRUST_SCALE`  to "mehestan" as they are used when computing global scores and are more related to Mehestan than Byztrust.

In the longer term, we may want to persist such metadata in the database, for a better traceability about the past runs. As is, there is no guarantee that the values exported in "metadata.json" were the same when scores were computed.

PR #1423, opened 2023-03-05
[tests] fix: close youtube promotional banner when testing browser extension

amatissart:
Should fix broken e2e tests, observed in #1419


amatissart:
Arf now, the other test, about the "rate later" buttons fails :sweat:

PR #1426, opened 2023-03-06
Front reload unavailable video in comparisons page

OoIwazaruoO:
**replaced by** https://github.com/tournesol-app/tournesol/pull/1431 (conflicts solved)

---

This PR has for goal to find and apply a good behavior to implement on the comparison page when a video is unavailable.
For the moment when a video is unavailable it displays the tournesol loader till a new video is loaded and displayed.

However, it might be disconcerting not to inform the user the video is not available. So we are looking for the better way to tell the user.

Some options could be : 

- Display a message informing the video is not available with a counter instead of the tournesol loader and reload a new video at the end of the counter.
- Display that message with no counter and wait for the user to reload a new video manually.
- Display the message in place of the video player
- Display the message on a light overlay on top of the video display

To do
- [ ] Fix : the comparison slider flash when a video is not available

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -26,6 +26,7 @@
     "criteriaRatedLow": "Rated low:",
     "nbViews": "{{nbViews}} views",
     "seeRecommendedVideosSameUploader": "See recommended videos of the same uploader",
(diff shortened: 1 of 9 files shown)


GresilleSiffle:
I'll clean this branch on Wednesday.

PR #1431, opened 2023-03-08
[front] display unavailable video message on comparison page for unavaible video

GresilleSiffle:
This PR attempts to fix the conflicts between https://github.com/tournesol-app/tournesol/pull/1426 and the branch `main`.

@OoIwazaruoO I've imported all your code as-is here, you can continue on this branch.

**to-do**

- [x] simplify the introduced complexity if possible
- [x] hunt bugs


GresilleSiffle:
Hello @tournesol-app/maintainers 

I've reviewed those changes and I'd like your input before going further.

**(1) about the usage `setTimeout` to modify the states**

The function `handleChange` and the `useEffect` hook, responsible of asking for new videos when an unavailable one is returned by the Auto button, use the `setTimeout` to set a state after a certain delay. I don't know if it's a common practice or something that should be avoided. What do you think @amatissart and @lfaucon ?

Regarding the `useEffect`, the timeout is used to enable the access again to the function `getVideoForComparison`, previously blocked by the `setLoading(true)`.

See:

https://github.com/tournesol-app/tournesol/blob/ab7845d8b9adb56ee9ce9baadc0d258a42e93a13/frontend/src/features/entity_selector/EntitySelector.tsx#L246

**(2) about the auto-load feature**

When opening a comparison link, the `EntitySelector`s will automatically replace the unavailable videos by available ones. As a result, opening or sharing a previously made comparison involving unavailable entities redirects to a different comparison. It's not possible anymore to see or edit those comparisons.

In my opinion this is a small problem, but a problem still: redirecting users to an unexpected place, and preventing them to edit their comparisons could be a source of frustration. This is something I'd like to avoid as much as possible in the UX: unexpected behaviours.

Propositions:
- allow the users to disable the auto-load feature in the settings (out of scope)
- allow the users to see and update their own comparisons even if the auto-load feature is enabled

**(3) about the `getVideoForComparison`**

We currently call a non-generic function in a supposedly generic effect.

As we don't have a generic `getEntityForComparison`, we could bypass this feature for non-video entities.

The feedback are more than welcome on this : )

frontend/src/features/entity_selector/EntitySelector.tsx
```diff
@@ -184,21 +202,82 @@ const EntitySelectorInnerAuth = ({
...
+  ]);
+
+  /**
+   * Ask a new entity if an auto-reload has been asked.
+   */
+  useEffect(() => {
+    if (entityAvailability === ENTITY_AVAILABILITY.UNAVAILABLE) {
+      if (autoReload && !loading) {
+        setLoading(true);
+        getVideoForComparison(
```
amatissart, line 247: As pointed by @GresilleSiffle, this component is not specific to "videos", so we should check the current poll before calling `getVideoForComparison`.

frontend/src/features/entity_selector/EntitySelector.tsx
```diff
@@ -184,21 +202,82 @@ const EntitySelectorInnerAuth = ({
...
+          idFromUid(otherUid || ''),
+          idFromUid(uid || '')
+        ).then((newUid) => {
+          onChange({ uid: `${UID_YT_NAMESPACE}${newUid}`, rating: null });
+          setTimeout(() => {
+            setLoading(false);
+          }, 800);
```
amatissart, line 254: why this delay?

amatissart (commented):
The behavior on this branch has a few problems:
 * it's not possible to access a past comparison related to an unavailable video. The video is automatically replaced with another video without any explanation, which is very confusing for the user.
 * when selecting an unavailable video manually in one of the selectors, I can see the  message "Only available videos can be compared". However it's still possible to submit the comparison.

But more importantly, I'm not confident with how the `EntitySelector` uses `setTimeout` in multiple places . It's not clear what is the problem they are supposed to solve, and if they are actually fix, especially when some requests would be slower than expected.

This problem seems to arise because the video availability is achieved independently from the existing requests to the API. This creates a lot of potential race conditions.  
I think it could make sense to move the state related to the selected entities in a new dedicated state/context, in order to reduce the complexity of  `EntitySelector` and define a single validation flow, to check the video availability, fetch the contributor rating, etc.

amatissart (changes requested):
After discussion, let's remove the "autoReload" feature for now. 

We can allow unavailable videos to be compared (or their past comparisons to be modified). The message used on the entity selector should be modified to clarify that.

cc @OoIwazaruoO

GresilleSiffle (commented):
I can't approve a pull request I opened myself, but the code looks good to me :+1: 

It's far more simple now : ) @amatissart do you want to have a final look before approving?

### Next step

We now have a small display bug making the duration not easy to read when the entities are not available (see the bottom right corner of the capture):
 
capture

frontend/src/features/entity_selector/EntitySelector.tsx
```diff
@@ -296,16 +312,39 @@ const EntitySelectorInnerAuth = ({
...
-        <EmptyEntityCard compact loading={loading} />
-      )}
+      <Box position="relative">
+        {(rating && (
+          <EntityCard
+            compact
+            entity={rating.entity}
+            settings={showRatingControl ? toggleAction : undefined}
+          ></EntityCard>
+        )) || <EmptyEntityCard compact loading={loading} />}
```
amatissart, line 322: (nitpicking) 
Is it not equivalent to the previous version, that I find more readable?

```suggestion
        {rating ? (
          <EntityCard
            compact
            entity={rating.entity}
            settings={showRatingControl ? toggleAction : undefined}
          ></EntityCard>
        ) : (
          <EmptyEntityCard compact loading={loading} />
        )}
```

amatissart (approved):
LGTM. Thanks for the simplification @OoIwazaruoO @GresilleSiffle !

frontend/src/features/entity_selector/EntitySelector.tsx
GresilleSiffle: Good point!

Issue #431, opened 2022-01-09
[extension] feat: Search on youtube displays recommendations from Tournesol's search


lfaucon:
Display 1-2 results of Tournesol's search on Youtube search page similarly to recommendations added to the home page.

Youtube's search:
![Screenshot 2022-01-09 at 13 33 04](https://user-images.githubusercontent.com/8818081/148682407-78e561d5-bf52-48bd-98e3-73dcf80d0226.png)

Notes: 
- Use language preference of the user
- Example search: https://api.tournesol.app/video/?search=climate
- If no video are returned from Tournesol nothing should be displayed
- Elon 5 years ago is really not great on Climate Change
- Let's try to implement it without too much code duplication

lfaucon:
Initial work was done in PR: #440 
But branch is stale and we wil need to extract code from it to recover the work done.

GresilleSiffle:
After having discussed with Adrien and Aidan:
- we diplay 3 videos in a single column, the top 3 of the back end answer (all time recommendations with the search parameter)
- the title is click-able and redirects to the YT video
- the thumbnail is click-able and redirects to the YT video

Following task:
- display Recommended by Tournesol inside the yellow background on both the YT homapage and the search results
- display a button `View more` after the search results (aligned to the right) that redirects the users to the Tournesol recommendations with the same search parameter

lfaucon:
> use English only for now

What does it mean exactly? Why is it difficult to get languages the same way as for the home page?

OoIwazaruoO:
> 

No no, I mixed English and French in the display by mistake it was just a fix to do. I can see how it can be misinterpreted I correct it.

PR #1436, opened 2023-03-13
[ext] feat: add search recommendations

OoIwazaruoO:
**related to** #431

---

This PR has for goal the display of tournesol video recommandations based on the users search on youtube. ( issue #431 )
At the moment it displays 3 videos based on the users search, this will be editable later once the extension has configurations.

To do:
- [x] display 3 videos in a single column, the top 3 (the first three) of the back end answer (all time recommendations with the search parameter)
- [x] the title is click-able and redirects to the YT video
- [x] the thumbnail is click-able and redirects to the YT video
- [x] the channel name is click-able and redirect to the YT channel
- [x] use English only for now
- [x] handle the dark mode, as on the Tournesol recommendations on the YT home page
- [x] translate in French (the mechanism introduced by https://github.com/tournesol-app/tournesol/pull/1220 could be used)
- [x] display Recommended by Tournesol inside the yellow background on both the YT home page and the search results
- [x] display a button View more after the search results (aligned to the right) that redirects the users to the Tournesol recommendations with the same search parameter

Menu with search on and off

onoff

Home page light and dark

youtubeHomeLight

youtubeHomeDark

Search page light and dark

image

image


GresilleSiffle:
The current way the search is implemented is not deterministic.

(1) If I search several times the term "animaux" I'll get a different restult for each request.

(2) The results are not order by "relevance" score (Tournesol score + search term).

It means the order I get from the extension is not the same I get from the website. A search with the term "animaux" gave me "6 MENSONGES sur l'HYGIÈNE" in first position, while in the all time recommendations from Tournesol are about ethics, animals, citizenship, slaughter houses, etc. If users have recommened videos about ethics (scores ~ 50), I don't expect "6 MENSONGES sur l'HYGIÈNE" in first position (score ~ 13).

(3) The search mechanism use the same code of the YT home page recommendation. It means that once the back end has returned the search results, those results are filtered by dates and mixed according to some proportion. That's why the results feel "not Tournesol" and a bit random.

Let's discuss together in order to create a good search system.

If think the search recommendations should not be reprocessed by the extension. At least, the distinction between all-time recommendations and recent recommendations should be clear, and videos should not be mixed.

OoIwazaruoO:
The view more is a little too big, make it smaller by imitating a smaller button of tournesol.
Sometimes when you click on the learn more button from the search page and come back to the previous page the search does not reload,we can fix this by making the learn more link open in a new tab (target=_blank rel=noopenener).
There is a utm_src param on the learn more link, add it on the view more button.
The manifest version is not up to date it should be 2.6.0.
On the tournesol app the max score is displayed  only if it is higher than zero and the min score only if it is lower than zero, apply the same behaviour on the youtube search.
It can be counter-intuitive that the title and channel name are not clickable on the home page, make it behave like on the search page (after this PR)
Use the preference language for the view more link
Get the search recommandation query by a more robust method than substring
Check if the position relative is still needed on the menu search enable button
Replace the else by an else if in the background file when request message is tournesolRecommandations or tournesolSearchRecommandation

GresilleSiffle:
The recent merge of https://github.com/tournesol-app/tournesol/pull/1220 made this branch in conflict with `main`.

**to-do**
- [ ] solve the conflicts and align the branch with `main` (me and @OoIwazaruoO )
- [ ] invite all the team to test this new feature
- [ ] double check if we can replace the `setInterval` by something else

GresilleSiffle (commented):
Thanks a lot @OoIwazaruoO 

The layout the home page recommendations is better now, and the layout of the search results is great :+1: 

I really really like this feature. For now it doesn't replace the search on the Tournesol website (mainly because the website has a pagination, filters, etc.), but I think it should not be seen as a replacement, but rather as a way to make the users discover new channels and the top recommendations of Tournesol about specific topics.

**future considerations (could be done in another PR)**

**(a)**

I think it could be better to split the home page recommendations and the search results in two content scripts. The more the file `addTournesolRecommendations.js` grows the harder it will be to maintain it.

To avoid code duplications between the home page recommendations and the search results, we could make the content scripts dynamically load a JavaScript module. Contrary to the content scripts, this module will be able to use standard JavaScript import statements to use factorized code.

```
# example: the folder features contains factorized code
features/banner.js
features/recommendationsContainer.js

# example: the folder contentScripts contains only content scripts and their imported module 
contentScripts/homePage.js # dynamically loads `homePageModule.js`
contentScripts/homePageModule.js # can import `lib/banner.js` and more
contentScripts/searchResults.js # dynamically loads `searchResultsModule.js`
contentScripts/searchResultsModule.js # can import `lib/banner.js` and more
``` 

[An example here.](https://github.com/otiai10/chrome-extension-es6-import/blob/master/src/js/content_script.js)

What do you think @tournesol-app/maintainers ?

**(b)**

I didn't have the time to investigate how to replace the `setTimeout`. It seems to work correctly but I was wondering if it was robust enough.

browser-extension/.eslintrc.json
```diff
@@ -5,7 +5,7 @@
     "webextensions": true
   },
   "parserOptions": {
-    "ecmaVersion": 8,
+    "ecmaVersion": "latest",
```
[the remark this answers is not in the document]
OoIwazaruoO: solved by [42476dd](https://github.com/tournesol-app/tournesol/pull/1436/commits/42476dda44c0927ed35e693e67fb45aa771434eb)

browser-extension/src/models/banner/BannerOptions.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+}
+
+// TODO: these values are placeholder values that should be updated.
+export const defaultBannerOptions = new BannerOptions(
+  '2020-01-01T00:00:00Z',
+  '2024-01-01T00:00:00Z',
+  'https://docs.google.com/forms/d/e/1FAIpQLScQzlEKBSA3MqxI0kaPazbyIUnZ4PjFcrR8EFiikG1quyAoiw/viewform?usp=pp_url&entry.939413650=',
+  'https://docs.google.com/forms/d/e/1FAIpQLSf9PXr-f8o9QqDR-Pi63xRZx4y4nOumNDdwi_jvUWc6LxZRAw/viewform?usp=pp_url&entry.1924714025=',
+  'browser_extension_study_2023'
```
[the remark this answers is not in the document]
OoIwazaruoO: solved by [5605c4b](https://github.com/tournesol-app/tournesol/pull/1436/commits/5605c4b81f12804bb9ed98359ecaa92b0ccf6476)

aidanjungo:
I tried to use it and it seems really good 😀🌻

GresilleSiffle:
I'm currently reviewing the code.

The new classes make the extension code a lot easier to read and maintain, thanks a lot @OoIwazaruoO :+1: I may suggest new names for some classes and files to better match their behaviour.

**to-do**

- [ ] the YT button Filters appears on top of the Tournesol search results instead of below

This is confusing because it looks like this button affects the Tournesol search results but in fact it only affects the YouTube search results. If possible, I suggest to display the Tournesol search results before this button.

See:

capture2

browser-extension/src/models/banner/BannerOptions.js
```diff
@@ -0,0 +1,24 @@
...
+    this.TS_BANNER_ACTION_FR_URL = TS_BANNER_ACTION_FR_URL;
+    this.TS_BANNER_ACTION_EN_URL = TS_BANNER_ACTION_EN_URL;
+    this.TS_BANNER_PROOF_KW = TS_BANNER_PROOF_KW;
+  }
+}
+
+// TODO: these values are placeholder values that should be updated.
+export const defaultBannerOptions = new BannerOptions(
+  '2020-01-01T00:00:00Z',
+  '2022-01-01T00:00:00Z',
```
GresilleSiffle, line 20: The study is now in progress, let's not forget to update those values.

See: https://github.com/tournesol-app/tournesol/commit/daa91a0175777158012a559e9777c245dfc60d8e

browser-extension/src/models/banner/Banner.js
```diff
@@ -0,0 +1,156 @@
+import { isNavigatorLang } from '../../utils.js';
+import { defaultBannerOptions } from './BannerOptions.js';
+
+export class Banner {
```
GresilleSiffle, line 4: This `Banner` is not generic yet, mainly because of the specific translation keys.

Let's add comment so that we can make this class generic later.

```suggestion
/**
 * A banner displaying a message and an action button.
 * 
 * At the moment this class is specific to the extension research study, but
 * could be refactored to be generic.
 * 
 * - the texts could be parameters of the constructor
 * - the action button could be optional
 * - etc.
 */
export class Banner {
```

browser-extension/src/models/banner/Banner.js
```diff
@@ -0,0 +1,156 @@
...
+    this.TS_BANNER_DATE_END = options.TS_BANNER_DATE_END;
+    this.TS_BANNER_ACTION_FR_URL = options.TS_BANNER_ACTION_FR_URL;
+    this.TS_BANNER_ACTION_EN_URL = options.TS_BANNER_ACTION_EN_URL;
+    this.TS_BANNER_PROOF_KW = options.TS_BANNER_PROOF_KW;
+
+    // Create the banner once at the initialisation
+    this.banner = this.createBanner();
+  }
+
+  displayBanner() {
```
GresilleSiffle, line 16: I think there is no need to use the `Banner` suffix here, as we are already in a banner.

This change implies the update of all calls to  `displayBanner`.

```suggestion
  display() {
```

browser-extension/src/models/banner/Banner.js
```diff
@@ -0,0 +1,156 @@
...
+    this.banner.classList.add('displayed');
+  }
+
+  hideBanner() {
```
GresilleSiffle, line 20: See my previous comment about `displayBanner`.

browser-extension/src/models/tournesolVideosBox/TournesolVideosBox.js
```diff
@@ -0,0 +1,219 @@
+import { convertDurationToClockDuration } from '../../utils.js';
+
+export class TournesolVideosBox {
```
GresilleSiffle, line 3: Singular if this class represents the container of a single video.

```suggestion
export class TournesolVideoBox {
```

(minor: we can even use the same vocabulary as in the front end `TournesolVideoCard`)

GresilleSiffle (commented):
Here are the first batch of changes we can already apply.

The new architecture is really good! You'll see in my comments that there are several suggestions related to the class names. My intention is to prevent as much as possible any potential excessive increase of the code complexity by making the roles of the classes easy to understand.

browser-extension/src/models/banner/Banner.js
OoIwazaruoO: solved by [723eba7](https://github.com/tournesol-app/tournesol/pull/1436/commits/723eba728af057c42584a07c4e7d91a4c002787b)

browser-extension/src/models/tournesolVideosBox/TournesolVideosBox.js
OoIwazaruoO: solved by [04c4b77](https://github.com/tournesol-app/tournesol/pull/1436/commits/04c4b77d4bdad156925140fbed04ba0f6097365a)

browser-extension/src/models/banner/BannerOptions.js
OoIwazaruoO: solved by [7426da6](https://github.com/tournesol-app/tournesol/pull/1436/commits/7426da6081f4634f2c78b87540d3daa3216ad861)

browser-extension/src/models/tournesolComponent/TournesolComponent.js
```diff
@@ -0,0 +1,167 @@
+import { TournesolVideosBox } from '../tournesolVideosBox/TournesolVideosBox.js';
+
+export class TournesolComponent {
+  constructor(parent, banner) {
+    this.isExpanded = false;
+    this.parent = parent;
+    this.banner = banner;
+  }
+
+  getComponent() {
```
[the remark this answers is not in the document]
OoIwazaruoO: solved by [26c23ad](https://github.com/tournesol-app/tournesol/pull/1436/commits/26c23ad4db7992c9627c5757f695f97568aedb64) , [d337ad2](https://github.com/tournesol-app/tournesol/pull/1436/commits/d337ad2fe5c2d76920d749489ee9aa317397ad16) and [1366ee0](https://github.com/tournesol-app/tournesol/pull/1436/commits/1366ee07ed9fcb01bb2d510eeb88c85789b6a23c)

GresilleSiffle:
**to-do**
- [ ] backport in this branch the fix added by https://github.com/tournesol-app/tournesol/pull/1523

GresilleSiffle:
I'm progressing in the review, only 10 files remaining.

I currently have a display bug when the search is on. Did I break something in my last commits?

capture

browser-extension/src/displayHomeRecommendations.js
```diff
@@ -0,0 +1,32 @@
...
+      './models/tournesolRecommendations/TournesolRecommendations.js'
+    )
+  );
+
+  const tournesolHomeRecommendations = new TournesolRecommendations();
+
+  const process = () => {
+    if (location.search.includes('tournesolSearch')) {
+      chrome.storage.local.set({ searchEnabled: true });
+    }
```
GresilleSiffle, line 13: Is this still required?

browser-extension/src/displaySearchRecommendations.js
```diff
@@ -0,0 +1,46 @@
...
+      true
+    );
+  const tournesolSearchRecommendations = new TournesolSearchRecommendations(
+    tournesolSearchRecommendationsOptions
+  );
+
+  const process = () => {
+    if (location.search.includes('tournesolSearch')) {
+      chrome.storage.local.set({ searchEnabled: true });
+    }
```
GresilleSiffle, line 35: I don't remember the use case of this parameter.

browser-extension/src/models/tournesolContainer/TournesolContainer.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  createHTMLElement() {
+    // Create container
+    let tournesol_container = document.createElement('div');
+    tournesol_container.id = 'tournesol_container';
+
+    // Add inline-block div
+    const inline_div = document.createElement('div');
+    inline_div.setAttribute('class', 'inline_div');
```
GresilleSiffle, line 17: This div is not inline anymore, let's change the class name and the comment.

browser-extension/src/models/tournesolContainer/TournesolContainer.js
```diff
@@ -0,0 +1,168 @@
...
+      .then((svg) => (refresh_button.innerHTML = svg));
+
+    refresh_button.className = 'tournesol_simple_button';
+    refresh_button.onclick = () => {
+      refresh_button.disabled = true;
+      this.parent.loadRecommandations();
+    };
+    inline_div.append(refresh_button);
+
+    // Add title
```
GresilleSiffle, line 73:
```suggestion
    // Learn more 
```

browser-extension/src/models/tournesolContainer/TournesolContainer.js
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1436/commits/32084d1d13b49c909f378edfe57e5069137fc141

browser-extension/src/displaySearchRecommendations.js
OoIwazaruoO: This parameter is to check if the search works well in the tests

GresilleSiffle:
I'm slowly reaching the end of the review : ) I need to check if all bug fixes that have been applied on main since the beginning of this branch are present in this branch. I also need to check the updated tests.

browser-extension/src/displaySearchRecommendations.js
GresilleSiffle: I was able to directly set during the tests the key `searchEnabled` to `true` in the local storage. It allowed me to delete the condition `location.search.includes('tournesolSearch')` but since I merged `main` into this branch it doesn't work anymore.

I'm checking what's happening.

GresilleSiffle (approved):

PR #1451, opened 2023-03-20
[front] add test of pagination component

bbosson:
Add test on newly version of pagination component

checked:

- The default navigation
- The calculation of the total number of pages
- The offset of elements when choosing a page 
- The click on the different buttons -1 / -10 / -100 / +1 / +10 / +10
- That the -100 / +100 buttons are not displayed by default
- That the buttons are well deactivated according to the selected page

### DIFF
--- a/frontend/src/components/Pagination.spec.tsx
+++ b/frontend/src/components/Pagination.spec.tsx
@@ -21,7 +21,7 @@ const defineMatchMedia = (width: number) => {
     value: width,
   });
   window.matchMedia = jest.fn().mockImplementation((query) => {
-    console.log(width == theme.breakpoints.values.sm);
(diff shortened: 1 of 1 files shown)


GresilleSiffle:
Thank you @bbosson 

I'm currently reviewing the code.

GresilleSiffle (approved):
Good job @bbosson 

The tests fail correctly when we introduce bugs in the component `<Pagination>` :ok_hand: 

I made few changes to make the tests:
- more easy to read (shorter title, grouped by screen size, etc.)
- more robust (usage of `getBy` instead of `queryBy` when appropriate)
- less coupled (avoid repeating the same checks in several tests)

Issue #1051, opened 2022-07-15
URLs trailing slash


glerzing:
In most or our backend API URLs, there is a trailing "/". Some URLs (around 7 among 50) don't have it. From what I read, it's not really important whether there is a trailing slash or not. But it may be better to use a consistent rule, for homogeneity.

Moreover, the Tournesol URLs can be used both with or without the trailing slash. While it may be better for the users, that don't have to care about it, some websites speculate that it may be interpreted in some situations as duplicate content, and have a negative effect on SEO. I don't know if this is an issue, probably not.

GresilleSiffle:
Regarding the back end, we could indeed starts by adding at trailing slash at the end of each URL for consistency within the API.

The `APPEND_SLASH` settings should ensure the compatibility with old URLs wandering in the front end, but it doesn't seem to work on my local setup (I didn't test a lot).

https://docs.djangoproject.com/en/4.0/ref/settings/#append-slash

PR #1460, opened 2023-03-22
[back] fix: comparison preview url without trailing slash

aidanjungo:
The comparison preview was not display for url like:
<https://tournesol.app/comparison?uidA=yt%3AWHUJL-EjsGY&uidB=yt%3ADjwxOH3N-gc>

This is a quick fix, but a more general one should be done with #1051 

Questions:
- Is it OK to use the same name for the path?
- Should we copy/paste all the `DynamicWebsitePreviewComparisonTestCase` to test without the trailling slash?

### DIFF
--- a/backend/tournesol/tests/test_api_preview.py
+++ b/backend/tournesol/tests/test_api_preview.py
@@ -255,6 +255,15 @@ class DynamicWebsitePreviewComparisonTestCase(TestCase):
         # check is not very robust.
         self.assertNotIn("Content-Disposition", response.headers)
 
+        # Test without the trailing slash
+        response = self.client.get(
+            self.preview_url[:-1], {"uidA": self.valid_uid, "uidB": self.valid_uid2}
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.headers["Content-Type"], "image/png")
+        self.assertNotIn("Content-Disposition", response.headers)
+
+
     @patch("requests.get", mock_yt_thumbnail_response)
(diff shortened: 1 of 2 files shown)


GresilleSiffle:
> Is it OK to use the same name for the path?

No, different URLs should use different names, even if in our case both URLs are bound to the same view.

The `name` attribute is used to reverse the URL in case of need i.e. to get the path from the name. It's better to use different names to avoid any confusion.

See: https://docs.djangoproject.com/en/4.1/topics/http/urls/#naming-url-patterns

GresilleSiffle:
> Should we copy/paste all the DynamicWebsitePreviewComparisonTestCase to test without the trailling slash?

I think we just need to ensure the new URL is bound the expected view. If the view is already tested there is no need to re-test everything.

I think you can simply update the test `test_anon_200_get_existing_entities`.

GresilleSiffle:
I wonder, the Django setting `APPEND_SLASH` doesn't work in our case?

aidanjungo:
> I wonder, the Django setting `APPEND_SLASH` doesn't work in our case?

It is set to True, but it does not work

GresilleSiffle:
Let's investigate with the DRF parameter `trailing_slash`.

https://www.django-rest-framework.org/api-guide/routers/

GresilleSiffle:
Should be fixed by https://github.com/tournesol-app/tournesol/pull/1460/commits/c7bb432e11c9b0303bc3bb8b8a1f91b40d0e254b

GresilleSiffle (approved):

PR #1478, opened 2023-03-28
[back] Use EntityPollRating model to display the Tournesol score in the admin

aidanjungo:


backend/tournesol/admin.py
```diff
@@ -84,6 +84,11 @@ class EntityAdmin(admin.ModelAdmin):
     )
     actions = ["update_metadata"]
 
+    def get_queryset(self, request):
+        qst = super().get_queryset(request)
+        qst = qst.with_prefetched_poll_ratings(poll_name="videos")
```
GresilleSiffle, line 89: We saw with @aidanjungo that this method may not be appropriate here.

It makes the related `EntityPollRating` object available in a list, which prevent us to easily order the entities by Tournesol score or number of comparisons.

```python
    @admin.display(description="comparisons", ordering="???")
    def get_n_comparisons(obj):
        return obj.single_poll_ratings[0].n_comparisons
```

We may need to retrieve the Tournesol score, n comparisons and n contributors in a different way (don't forget to check the generated SQL query and the performance of the ordering).

aidanjungo:
@GresilleSiffle after discussion with Adrien and Louis, we decide to use the EntityPollRating as the "main" page to display all the information, it simplify a lot of things. We let just the minimum info in the Entity admin.

Issue #974, opened 2022-06-09
[front] feat: display a weekly collective objective in the comparison page


GresilleSiffle:
### context

In order to engage contributors to add comparisons, I propose to create and display a collective goal on the comparison page. A collective weekly goal could be 1000 comparisons per week. This way, contributors can collectively estimate how close / far they are from the weekly objective, can be motivated to add more comparisons, and have a direct feedback of their individual implication.

Here is how Common Voice display its daily objective after 5 contributions: 

![capture](https://user-images.githubusercontent.com/39056254/172807516-9b2c9e5f-a6dc-4311-914b-439580cb8198.png)

This feature can be part of the UX effort to make the Tournesol application more enjoyable.

(see: https://commonvoice.mozilla.org/fr)

### UX proposition

I imagine a progress bar at the top of the comparison page:

```
Thanks to everyone, Tournesol has reached 60% of its weekly objective of 1000 comparisons.
[----------------------->                    ]               4,2% comes directly from you.
```

In the progress bar, we could use the buffer feature of the MUI progress component to display in a different color the personal contribution of the logged user.

The component must display:
- a sentence explaining what the progress bar represents
- a progress bar representing the collective progress, with a buffer representing the logged user progress
- a little message displaying the percentage "added" by the logged user 

### technical specification

The developers must be able to display this component anywhere in the application. 

When seen by anonymous users, the UI can be:

```
Thanks to everyone, Tournesol has reached 60% of its weekly objective of 1000 comparisons.
[----------------------->                    ]                    Help us reach this goal!
```

Where `Help us reach this goal!` is a light grey link to the sign-up page.

### technical specifications

**back end**
- [ ] **(a)** identify and develop a way to retrieve the comparisons made since the last Monday and today (could be part of the current stats API)
- [ ] **(b)** identify and develop a way to retrieve the comparisons made by the logged user since the last Monday and today (could be part of the current `/users/me/comparisons/` API, an additional GET parameter could be enough)
- [ ] add new API tests

Regarding the point **(a)**, it's important to choose a good name for the new metric, to avoid confusion with the existing one. We already have `added_last_month`, we could use `added_this_week` to create a difference between sliding and non-sliding metrics (last VS. this).

Example:
- added **last** month: number of comparisons made during the last 30 days
- added **this** month: number of comparisons made since the first day of the current month

In our case, we need the weekly metric. We can update the `ComparisonsStatisticsSerializer` to add the new `added_this_week` field.

```python
[...]

PR #1480, opened 2023-03-29
UI added for weekly, refactor such as translate text soon with feedback

abhiram11:
This PR is opened to work on #974 while considering the following criteria:
1) The Progress bar uses Buffer from material UI, but I have removed the continuous animation of the buffer from it.
2) We can add the user's contribution parameter later, until then we can go with the `Help us reach this goal` text.
3) I will add the Typography text to the "translation.json" in upcoming commits.
4) I will also refactor this whole "Component" into another file based on your preference on the filename, file location, etc.
5) I have also added the `stats` and `comparison count` related code as comments, if my former PR gets accepted then this can be added and tested.
6) Please let me know the color choices on the progress bar as well. Styling and other finishing touches are also to be done.

Here's how it looks right now:
trnsl

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -72,7 +72,10 @@
     "submitAComparison": "Submit a comparison",
     "inactivePoll": "This poll is closed.",
     "inactivePollComparisonCannotBeSubmittedOrEdited": "No comparison can be submitted or modified.",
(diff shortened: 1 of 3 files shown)


abhiram11:
Hi, since this PR was pushing Frontend code but had backend code that was depricated, I am closing this PR and creating a new PR which had code of latest `main`.

PR #1482, opened 2023-03-30
[back] fix: server error when concurrent requests attempt to init the same video

amatissart:
When 2 requests attempt to create the same video, the second creation may fail after retrieving the video metadata.

In this case, the method `create_from_video_id()` will retry to fetch the object from the database.


aidanjungo (approved):
LGTM

PR #1484, opened 2023-03-31
[dataviz] Correcting dependency for data-visualization

pierremancini:
With numpy==1.22.4, the command "streamlit run Tournesol_Data_Visualization.py" produce an the error "ImportError: numpy.core.multiarray failed to import" in User comparisons's page.

Installing numpy==1.24.2 in virtual env corrected the problem.

Cf. https://stackoverflow.com/questions/20518632/importerror-numpy-core-multiarray-failed-to-import?page=1&tab=scoredesc#tab-top

I am new to contributing on an open source project. I hope I'm not missing something @tournesol-app/maintainers .

### DIFF
--- a/data-visualization/requirements.txt
+++ b/data-visualization/requirements.txt
@@ -1,5 +1,5 @@
 streamlit==1.18.1
-numpy==1.22.4
+numpy==1.24.2
 seaborn==0.11.1
 matplotlib==3.3.4
 pillow==9.3.0


lfaucon:
Thanks for the fix @pierremancini 

@aidanjungo Would it be helpful to have very basic tests for the streamlit app as part of the CI that we run on Github Actions?

lfaucon:
Thanks for the fix @pierremancini 

@aidanjungo Would it be helpful to have very basic tests for the streamlit app as part of the CI that we run on Github Actions?

aidanjungo:
> Thanks for the fix @pierremancini
> 
> @aidanjungo Would it be helpful to have very basic tests for the streamlit app as part of the CI that we run on Github Actions?

Yes we plan to do something like that with @GresilleSiffle

aidanjungo (approved):

aidanjungo:
Thank you for the modification, we recently update numpy in the backend, so it is a good thing that we stay with the same version.

Issue #1259, opened 2022-12-22
Sorting of rated entities


lfaucon:
TODO:

- [ ] Rename "Personnal recommendation" -> "Public personnal recommendation
- [ ] Enable sorting by personnal score and global score in the "My compared items" page

Linked to: https://github.com/tournesol-app/tournesol/issues/606

GresilleSiffle:
Note that even if it's possible to sort the rated entities by Tournesol score, the UI doesn't display display this score in the `<EntityCard>` yet.

PR #1488, opened 2023-04-04
[back][front] feat: rated entities can be sorted by tournesol score

bbosson:
**related to** #1259 

---

Enable sorting by personnal score and global score in the "My compared items" page

### DIFF
--- a/backend/tournesol/views/ratings.py
+++ b/backend/tournesol/views/ratings.py
@@ -31,8 +31,8 @@ ALLOWED_GENERIC_ORDER_BY_VALUES = [
     "-n_comparisons",
     "collective_score",
     "-collective_score",
-    "contributor_rating_criteria_score",
-    "-contributor_rating_criteria_score",
+    "individual_score",
+    "-individual_score",
 ]
 
 # Appended to the positional arguments of all calls to QuerySet.order_by()
@@ -46,6 +46,7 @@ def get_annotated_ratings(poll: Poll):
     Return a `ContributorRating` queryset with additional annotations like:
         - the number of comparisons made by the user for the entity
         - the date of the last comparison made for this entity
+        - etc.
 
     This queryset expects to be evaluated with a specific poll, user and
     entity.
@@ -69,7 +70,7 @@ def get_annotated_ratings(poll: Poll):
         .values("tournesol_score")
     )
 
-    contributor_rating_criteria_score = (
+    individual_score = (
         ContributorRatingCriteriaScore.objects
         .filter(contributor_rating=OuterRef("pk"), criteria=poll.main_criteria)
         .values("score")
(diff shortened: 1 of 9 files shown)


backend/tournesol/views/ratings.py
```diff
@@ -54,9 +58,21 @@ def get_annotated_ratings():
...
+    contributor_rating_criteria_score = (
+        ContributorRatingCriteriaScore.objects.filter(contributor_rating=OuterRef("pk"), criteria="largely_recommended")
+        .values("score")
+    )
+
     return ContributorRating.objects.annotate(
         n_comparisons=Subquery(n_comparisons),
         last_compared_at=Subquery(last_compared_at),
+        collective_score_largely_recommended=Subquery(collective_score),
+        contributor_rating_criteria_score=Subquery(contributor_rating_criteria_score)
```
lfaucon, line 75: Is it a worry for performance that this is now making 5 queries instead of 3? Could we refactor the code a bit such that we actually do only a single annotation for the ordering that is requested?

backend/tournesol/views/ratings.py
```diff
@@ -54,9 +58,21 @@ def get_annotated_ratings():
         .order_by("-datetime_lastedit")
     )[:1]
 
+    collective_score = (
+        EntityPollRating.objects.filter(poll=OuterRef("poll"), entity=OuterRef("entity"))
+        .values("tournesol_score")
+    )
+
+    contributor_rating_criteria_score = (
+        ContributorRatingCriteriaScore.objects.filter(contributor_rating=OuterRef("pk"), criteria="largely_recommended")
```
lfaucon, line 67: This should be replaced `poll.main_criteria` such that it works for other polls too.

backend/tournesol/views/ratings.py
GresilleSiffle: We tested the new query on staging and production and it doesn't look costly. Moreover the execution time of this query set will be further reduced by the presence of the filters on the poll, on the user, and by the pagination (offset / limit).

I think we can keep it as is for now.

backend/tournesol/views/ratings.py
bbosson: Thank you for this clarification, it is changed.

bbosson:
> Hello @bbosson
> 
> For my information, are you waiting for a review or is this PR still a draft?

I opened the Pull Request for a first reading, but I wanted to finalize the automatic tests. It's done. You can reread, Thanks

frontend/src/utils/constants.tsx
```diff
@@ -207,7 +213,12 @@ export const polls: Array<SelectablePoll> = [
     withSearchBar: true,
     topBarBackground: null,
     comparisonsCanBePublic: true,
-    extraMetadataOrderBy: ['duration', 'publication_date'],
+    extraMetadataOrderBy: [
+      'duration',
+      'publication_date',
+      'collective_score',
+      'individual_score',
```
GresilleSiffle, line 220: I'm currently reworking this part.

It works perfectly as is, but we can make those two metadata available for all polls.

The `extraMetadataOrderBy` key is expected to contain only non-generic metadata, i.e. metadata specific to an entity type. For instance the metadata `duration` is specific to the `video` entity type, and may be irrelevant for other types, like the candidates. 

`collective_score` and `individual_score` are generic, it means we use them directly in the file `RatingOrderByInput.tsx`.

GresilleSiffle (approved):

PR #1492, opened 2023-04-05
[front] feat: add a weekly collective goal widget in the comparison UI

abhiram11:
This PR is opened to work on #974 and is an update of #1480 and works on the following criteria:

1. The Progress bar uses Buffer from material UI, but I have removed the continuous animation of the buffer from it. However, since we do not have the user's weekly comparison count yet, we can just use the default Progress bar (i.e., not Buffer). and go with the "Help us reach this goal" text.
2. The Typography text is added to BOTH `translation.json`. Since I am not familiar with French, I have used google translate to convert `en` to `fr`, I request you to provide me with the preferred sentences for that part.
3. I can refactor this whole "Component" into another file based on your preference on the filename, file location, etc.
4. Please let me know the color choices on the progress bar as well. Styling and other finishing touches are also to be done.

Here's how it looks right now:
trnsol

https://user-images.githubusercontent.com/20269286/229989271-b5228d00-6190-40bc-95fb-c236b436616a.mp4

### NOTE:
Please update the code in `ComparisonsStatistics.ts` as below:

```
/* istanbul ignore file */
/* tslint:disable */
/* eslint-disable */

export type ComparisonsStatistics = {
    total: number;
    added_last_month: number; // remove after deprecation/backward compatibility handling
    added_current_week: number;
    added_last_30_days: number;
};

```

**P.S.: I will run the yarn schema and related commands soon.**


frontend/src/pages/comparisons/Comparison.tsx
```diff
@@ -1,13 +1,16 @@
...
-import { Box, Typography } from '@mui/material';
+import { Box, LinearProgress, Typography } from '@mui/material';
 
 import { ContentBox, ContentHeader } from 'src/components';
 import { useCurrentPoll } from 'src/hooks/useCurrentPoll';
 import Comparison from 'src/features/comparisons/Comparison';
 import ComparisonSeries from 'src/features/comparisonSeries/ComparisonSeries';
 
+import { useStatsRefresh } from 'src/hooks/useStatsRefresh';
+import { useLoginState } from 'src/hooks';
```
abhiram11, line 12: I'll remove the login state and other unnecessary code as well

GresilleSiffle:
Hello @abhiram11 

Thank you very much for all your work! It's very good to see the community goals "for real". I'll do a first review tomorrow afternoon.

As the current community goal widget is rather big in the screenshot, we may want:
- to give the possibility to the users to fold it and unfold it (a back-end setting would be ideal to save this state)
- to not display it during the tutorial to keep the UI simple
- to move it in an hypothetical user dashboard page (which doesn't exist yet and is out of scope of this PR : )

Please, don't rework your PR for now, I'll do a proper review tomorrow with more details and I'll ask feedback from the rest of the team.

abhiram11:
Thank you for the positive words, @GresilleSiffle : ) ! I also had the same thought of waiting for a review before coding more into this feature! 👍🏻

frontend/src/features/comparisons/statsSlice.ts
```diff
@@ -0,0 +1,30 @@
+import { RootState } from '../../app/store';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+import { PollStats } from 'src/utils/types';
+import { DEFAULT_POLL_STATS } from 'src/utils/api/stats';
+
+export const initialState: PollStats = DEFAULT_POLL_STATS;
+
+export const statsSlice = createSlice({
```
GresilleSiffle, line 9: This slice will contain only the stats of one poll at a time for now, but the name `statsSlice` could be confusing. 

In the back end API, the statistics are represented with a JSON containing several polls in addition to the number of users. Ex:

```json
{
  "active_users": {
    "total": 0,
    "joined_last_month": 0,
    "joined_last_30_days": 0
  },
  "polls": [
    {
      "name": "string",
      "compared_entities": {
        "total": 0,
        "added_last_month": 0,
        "added_last_30_days": 0
      },
      "comparisons": {
        "total": 0,
        "added_last_month": 0,
        "added_last_30_days": 0,
        "added_current_week": 0
      }
    }
  ]
}
```

We have two options here:
- **(a)** saving the complete JSON object returned by the API in the store, which include all polls stats
- **(b)** saving only one poll at a time, and renaming the slice `pollStatsSlice` (singular)

The options **(a)** has my preference for few reasons:
- when we query the back end, it already answers with the statistics of all polls without any filter (so it doesn't cost anything to save everything)
- it opens the possibility to display the stats of different polls on the same page
- it makes the statsSlice, a real statsSlice, in sync with the stats API (and not a subpart of a response)
- if a developer wonder if some stats are available in the store, the answer will always be yes : )

I don't have a strong preference, it will be fine to have only one poll for now in this PR. You can choose @abhiram11  : )

GresilleSiffle (commented):
Hello @abhiram11 

I didn't have the time to completely review everything today unfortunately, but the feature looks very promising :+1: :+1: :+1: Here are some comments we can already address.  

Note that I had to merge the `main` branch in this one on my laptop to fix the outdated API schema. For instance the new field `added_last_30_days` was not recognized as a valid field. You don't have this error on your side? Let's merge the up-to-date `main` branch in this one and push the merge commit.

### P.S. : I have added the `added_last_30_days` and similar object types manually, as shown in the top/initial comment of this PR. I am also attaching a screenshot in the next commit as another reminder for that.

> The Progress bar uses Buffer from material UI, but I have removed the continuous animation of the buffer from it. However, since we do not have the user's weekly comparison count yet, we can just use the default Progress bar (i.e., not Buffer). and go with the "Help us reach this goal" text.

You did well, I think we can go with the default progress bar for now.

> The Typography text is added to BOTH translation.json. Since I am not familiar with French, I have used google translate to convert en to fr, I request you to provide me with the preferred sentences for that part.

That's perfect, your efforts are really appreciated. We will review the translations just before the merge.

> I can refactor this whole "Component" into another file based on your preference on the filename, file location, etc.

Yes please : ) I added a comment about it.

> Please let me know the color choices on the progress bar as well. Styling and other finishing touches are also to be done.

Sure, for now I'm focus on the code, I propose to review the colours before merging.

GresilleSiffle:
Here is a proposition to simplify the layout, with:

- a small title
- the exact number of comparisons made and the target
- the % representing the progression
- only on two lines
 
capture

GresilleSiffle:
Few ideas for the future. I put them here so I don't forget : )

The goal will be to represent the logged user participation, without overloading the UI.

capture

frontend/src/features/comparisons/statsSlice.ts
abhiram11: Hi @GresilleSiffle , I'm going forward with the **(a)** approach, should we rename the entire `PollStats` to something like `Stats` or `AllStats` ? I have:

1. Separated the entire component into an `<CollectiveGoalWeeklyProgress>` and placed it in `src/features/goals`.
2. The "PollStats" now contains the entire data received from "/stats/" API in a spreaded format.
3. Working on removing states in `HomeVideos, etc.` that were the alternative from redux slice to store the `/stats/` api data. This would also affect files such as `UsageStatsSection` as well.
4. The UI has also been minimalized as per your requirement, I would prefer with going with the Tooltip approach afterwards since it encourages client/user's curiosity to hover on their "Green coloured +60" contribution : )

abhiram11:
Hi, would you also like me to refactor this function into a fetch from the statsSlice redux? As the API being called in the same `/stats/` API as at `HomeVideos`, `UsageStatsSection`, and the `CollectiveWeeklyProgress` that I am adding in this PR.

https://github.com/tournesol-app/tournesol/blob/2a308f3b29ed745082f7e9b146e7f4002956ee7b/frontend/src/pages/about/PublicDownloadSection.tsx#L19

frontend/src/features/comparisons/statsSlice.ts
GresilleSiffle: If you have chosen the approach **(a)** then we can go with:
- `statisticsSlice` for the Redux slice
-  `Statistics` for the type of the initial state (it's a type already available, you can import it with `import { Statistics } from 'src/services/openapi';`)

> Separated the entire component into an <CollectiveGoalWeeklyProgress> and placed it in src/features/goals.

Perfect.

> The "PollStats" now contains the entire data received from "/stats/" API in a spreaded format. Otherwise, keeping the "active_users", "compared_entities", etc. Type Definitions in the redux would introduce complexity, but let me know if you would want it that way.

Good. It matches the behaviour of the slice `userSettingsSlice` :+1: 

> Working on removing states in HomeVideos, etc. that were the alternative from redux slice to store the /stats/ api data. This would also affect files such as UsageStatsSection as well.

Exactly. You are not forced to do it in the same pull request, but it will be useful. The components previously using the state will need to use Redux selectors. You can find an example here:
- usage: https://github.com/tournesol-app/tournesol/blob/0dcf9fa5dd1674a35b9f757e9917c6d828180479/frontend/src/features/settings/preferences/VideosPollUserSettingsForm.tsx#L36
- definition: https://github.com/tournesol-app/tournesol/blob/0dcf9fa5dd1674a35b9f757e9917c6d828180479/frontend/src/features/settings/userSettingsSlice.ts#L33

> The UI has also been minimalized as per your requirement, I would prefer with going with the Tooltip approach afterwards since it encourages client/user's curiosity to hover on their "Green coloured +60" contribution : )

Great, let's see how the first implementation looks like.

Do not hesitate to ping me again when you are ready for a review ; )

abhiram11: Sure, I can move forward with these ideas since most of it is already done. I'm currently dealing with memory leak issues in the useEffects where I'm calling `/stats/` api... along with some changes. I'll push a commit in an hour for your reference!

abhiram11: We can later use cache and stale times effectively to call the `stats` API from components like `UsageStatsSection`, `PublicDownloadsSection`, and `CollectiveWeeklyProgress`, etc.

abhiram11:
This new commit consists of the following changes:

- The redux store `statsSlice` FETCHES only at `HomeVideos` component or when we click on `Submit Comparison` when in `/comparison`.
- The rest of the components such as (the ones that received props from HomeVideos) `UsageStatsSection`, `RecommendationsSection`, `ComparisonSection`, and `CollectiveGoalWeeklyProgress`, they receive the values of `/stats/` API via the redux store.
- As of now, `<CollectiveGoalWeeklyProgress />` is not conditioned based on `tutorial or series` values, I am having some issues testing that : Instead of showing the tutorial "modal" screen, I am getting a memory leak warning of useEffect instead. However, the UI is modified as per @GresilleSiffle 's advice.
- I have also updated `PublicDownloadSection` where it was using `/stats` API in a similar way as of getPollStats function.
- Some other minute changes, please see the attached screenshot and video to replicate this commit. In case there are any issues, please let me know! : )

### Note: Please change these values manually to replicate the commit:
typeChanges

GresilleSiffle:
Hello,

I'm currently reading the code : ) I'm going to merge `main` in this branch so that we don't have to manually edit the OpenAPI files.

GresilleSiffle:
The OpenAPI schema is now up-to-date, and running the following command should allow us to work more comfortably without editing the generated files.

```
./scripts/generate-services-from-openapi.sh
```

GresilleSiffle:
Hello @abhiram11 

I tried to simplify the code, can you tell me if you see have the memory leak issue?

abhiram11:
Hi @GresilleSiffle , the memory leak warning still exists. It is shown only the first time when I start the dev environment -> Click on `Compare the Videos` Button that's in `Contribute` component of the Home Page. 

The warning appears even if I am already logged in, or if I am logged out then : Click on Compare the Videos -> Login Page where I do Login -> `/Comparison` component (warning appears when this is mounted)

GresilleSiffle:
> I would also like to thank you for simplifying using redux-thunk in statsSlice.ts, that helped me learn a lot as well: : ) +1

Wonderful!

> The warning appears even if I am already logged in, or if I am logged out then : Click on Compare the Videos -> Login Page where I do Login -> /Comparison component (warning appears when this is mounted)

I'm going to check this.

GresilleSiffle:
It looks like this error exists in `main` at the time of writing. It's not something we introduced with this pull request so we are fine.

The error seems to be caused by the unmount of the component `<ComparisonSeries>`. It is probably unmounted during the execution of the following effect:

```js
      const alts = await getAlts();
      if (alts.length > 0) {
        setAlternatives(alts);
        return alts;
      }
```

I'm going to create a separate issue, and we can continue working in this branch : )

GresilleSiffle:
### Preview

capture

### When the group reach 25 % of the goal

capture2

### When the group reach 75 % of the goal

capture3

GresilleSiffle (approved):
It works!

I'm ready to merge : ) Thank you @abhiram11 for all your work. Soon you'll see it in production 🥳

GresilleSiffle:
Here is a preview of a setting I'm currently working on:

capture

abhiram11:
Super excited !! 🚀🚀 The preference of weekly progress UI looks interesting as well!

frontend/src/features/comparisons/statsSlice.ts
```diff
@@ -0,0 +1,37 @@
...
+    builder.addCase(fetchStats.fulfilled, (state, action) => {
+      state.active_users = action.payload.active_users;
+      state.polls = action.payload.polls;
+    });
+  },
+});
+
+export const selectStats = (state: RootState) => state.stats;
+
+export default statsSlice.reducer;
```
amatissart, line 37: It's not clear when exactly the "stats" object will be initialized with the values from the server. The current implementation seems to assume that `<HomeVideos>`  is responsible for calling `fetchStats()`. But that does not seem to be a correct assumption. For example, a user may directly access the comparison page without loading the home page first.  In this case, the "stats" will not be fetched, and the `<CollectiveGoalWeeklyProgress>` will show 0 comparison, without any request to the server.

I must say that I am not convinced of the benefit of using a Redux Slice in this case. I think we expect the stats values to be refreshed on every session anyway. So creating a custom hook to memorize the stats based on the current poll during the current session could be simpler to avoid this kind of bugs.

amatissart (changes requested):
Thanks for working on this. It looks very good :+1: 

I think there is still a issue though, regarding how and when the stats are retrieved from the server in the first place (see my comment).

frontend/src/features/comparisons/statsSlice.ts
abhiram11: We can use React Query and its cache and stale time features to optimize our backend calls. Would like to know the team's advice and feasibility on this as well, and feel free to let me know if some changes are to be made.

GresilleSiffle: Thanks for the feedback, I'll refactor this tomorrow.

> The current implementation seems to assume that <HomeVideos> is responsible for calling fetchStats(). But that does not seem to be a correct assumption. For example, a user may directly access the comparison page without loading the home page first. 

That's true. Instead we can fetch the statistics in the `<App>` component by using a hook, the same way we fetch the user's preferences.

> I must say that I am not convinced of the benefit of using a Redux Slice in this case. I think we expect the stats values to be refreshed on every session anyway. So creating a custom hook to memorize the stats based on the current poll during the current session could be simpler to avoid this kind of bugs.

Yes it's not useful indeed to save the data in the local storage as we want them to be refreshed for every session. It's still easily possible to manage the statistics global state with Redux without the need of saving the data in the local storage. The package `redux-persist` we use offers the possibility to declare [which states should be persisted](https://github.com/rt2zz/redux-persist#blacklist--whitelist) between sessions.

```javascript
const persistConfig = {
  key: 'root',
  storage,
  // The following line does the trick.
  blacklist: ['stats']
};
```

We could persist in the local storage user specific settings, and keep the rest outside. What do you think about it @amatissart ? 

I still see some benefits of using Redux to manage those two kind of global states (session-only and persisted), such as:
- having a unified interface that avoids to "re-invent the wheel"
- it "forces" the contributors to follow a documented path when they work on global states

amatissart: I have no strong opinion about this. Your approach looks good to me, if you think this pattern could indeed be easier to replicate for other features in the future.

GresilleSiffle: Alright, let's go.

> if you think this pattern could indeed be easier to replicate for other features in the future.

To be honest I'm not strongly convinced too. What motivates me is the fact that we (I?) have a good opportunity to discover what kind of limitations we will face with Redux, by actually using it.

GresilleSiffle:
The tests were failing because the store used by the test `App.test.tsx` was not fully and properly mocked.

The commit https://github.com/tournesol-app/tournesol/pull/1492/commits/05a2d162bd2dedc0a1110f477a252a20cb448699 should fix it (the code is based on `Login.spec.tsx`).

amatissart (approved):

frontend/src/features/comparisons/statsSlice.ts
abhiram11: Hi @GresilleSiffle , I was trying to find something more robust to move forward with this approach.
I found out that `React Query` can be a good (future?) substitute for this one. Here's a clip that I believe touches on the exact obstacle that we might foresee: https://youtube.com/clip/Ugkxlyz_ErUUjudZYtywHwmyDVBub7fg6gfI 

Adding the link here for future references : )

Issue #1409, opened 2023-03-04
[front] Add `preferences` in the settings menu


aidanjungo:
It could look like that, we could start with just the implementation of the existing setting to change the number of comparison before a video is removed from the rate later list, then we could have other category when we add other settings.

<img width="1156" alt="image" src="https://user-images.githubusercontent.com/53896963/222921464-92f84b14-29e6-4f20-b4c4-5fc0ce0077c4.png">

GresilleSiffle:
Before implementing this task I think we have a decision to take.

Some settings are poll specific like the rate-later auto remove, and others are global. What is the target UX?

I see several options:

**(a)** Display all settings of all poll in the same preferences page. For a more user friendly experience, the page could display several forms (one per poll for now).

**(b)** Display only the global settings and the settings related to the current poll in the preferences page. If the users want to change the settings of several poll they will have to use the poll selector to access the different forms.

I have a preference for the option **(b)** as it makes the UI more simple:
- no duplicated field in the same page (auto-remove will be displayed for each poll)
- less forms in the same page

GresilleSiffle:
Here is a draft of the UI:

Global settings, like the the configuration of the UI or the tutorial could be on the same page, just before the poll specific form.

![capture](https://user-images.githubusercontent.com/39056254/231127819-0c33d6c9-b959-468b-9b26-f61d93e00d67.png)

PR #1501, opened 2023-04-11
[front] feat: add user's preferences page & form

GresilleSiffle:
**related to** #1409 

---

This PR adds a new Preferences page in the settings.

This page displays a form allowing the users to update their settings related to the poll videos. The page currently displays only the videos form, but is designed to be able to also display global settings and forms for other polls.

The added form is currently generic, it can be re-used by any poll to display the generic settings. The day we will have a poll with specific settings, let's say videos, we will have to refactor it a bit.

The form is divided in sections, each section contains settings related to the same category. There is only one section for now, the rate-later section, but we may have more sections in the future.

**to-do**

_create the page preferences_
- [x] create the new page
- [x] display it in the settings menu
- [x] correctly display the human name of the `videos` poll

_create the new preferences form_
- [x] load the settings from the API
  - [x] update the store afterward 
- [x] add the rate-later auto-remove setting
- [x] display errors field by field
- [x] add tests

### Preview

Two fields are displayed to have a preview of how the form will look like with more fields. 

capture

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -312,8 +312,7 @@
     "autoRemove": "Auto-remove",
     "autoRemoveHelpText_one": "Entities will be removed from your rate-later list after {{rateLaterAutoRemove}} comparison.",
     "autoRemoveHelpText_other": "Entities will be removed from your rate-later list after {{rateLaterAutoRemove}} comparisons.",
-    "updatePreferences": "Update preferences",
-    "errorOccurredWhileRetrievingPreferences": "Sorry, an error has occurred, cannot retrieve your preferences."
+    "updatePreferences": "Update preferences"
   },
   "username": "Username",
   "metrics": {
@@ -629,8 +628,15 @@
   },
   "noVideoCorrespondsToSearchCriterias": "No video corresponds to your search criteria.",
   "preferences": {
+    "errorOccurredWhileRetrievingPreferences": "Sorry, an error has occurred, cannot retrieve your preferences.",
     "preferences": "Preferences",
     "preferencesOfThePoll": "Preferences of the poll:"
+  },
+  "poll": {
+    "videos": "videos",
+    "presidential2022": "election FR 2022",
+    "entityCandidate": "candidate",
+    "entityVideo": "video"
   },
   "profile": "Profile",
   "signup": {
@@ -677,12 +683,6 @@
     "better_habits": "Does it make people adopt habits that benefit themselves and beyond?",
     "backfire_risk": "Is it adapted to viewers with opposing beliefs? Does it prevent misconceptions or undesirable reactions?",
     "reliability": "Is the presented information trustworthy, robustly backed and properly nuanced?"
-  },
-  "poll": {
-    "presidential2022": "election FR 2022",
-    "videos": "videos",
-    "entityCandidate": "candidate",
-    "entityVideo": "video"
   },
   "videoMetadata": {
     "duration": "Duration",
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -306,14 +306,14 @@
     "deleteAccount": "Delete account"
   },
   "pollUserSettingsForm": {
-    "errorOccurredWhileRetrievingPreferences": "Sorry, an error has occurred, cannot retrieve your preferences.",
     "errorOccurredDuringPreferencesUpdate": "Sorry, an error has occurred, cannot update your preferences.",
     "preferencesUpdatedSuccessfully": "Preferences updated successfully.",
     "rateLater": "Rate-later list",
     "autoRemove": "Auto-remove",
     "autoRemoveHelpText_one": "Entities will be removed from your rate-later list after {{rateLaterAutoRemove}} comparison.",
     "autoRemoveHelpText_other": "Entities will be removed from your rate-later list after {{rateLaterAutoRemove}} comparisons.",
-    "updatePreferences": "Update preferences"
+    "updatePreferences": "Update preferences",
+    "errorOccurredWhileRetrievingPreferences": "Sorry, an error has occurred, cannot retrieve your preferences."
   },
   "username": "Username",
   "metrics": {
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -309,7 +309,7 @@
     "errorOccurredDuringPreferencesUpdate": "Sorry, an error has occurred, cannot update your preferences.",
     "preferencesUpdatedSuccessfully": "Preferences updated successfully.",
     "rateLater": "Rate-later list",
-    "autoRemove": "Auto-remove",
+    "autoRemove": "Automatic removal",
     "autoRemoveHelpText_one": "Elements will be removed from your rate-later list after {{rateLaterAutoRemove}} comparison.",
     "autoRemoveHelpText_other": "Elements will be removed from your rate-later list after {{rateLaterAutoRemove}} comparisons.",
     "updatePreferences": "Update preferences"
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -305,6 +305,16 @@
     "exportAllData": "Export all data",
     "deleteAccount": "Delete account"
   },
+  "pollUserSettingsForm": {
+    "errorOccurredWhileRetrievingPreferences": "Sorry, an error has occurred, cannot retrieve your preferences.",
+    "errorOccurredDuringPreferencesUpdate": "Sorry, an error has occurred, cannot update your preferences.",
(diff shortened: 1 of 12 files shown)


frontend/src/features/settings/preferences/GenericPollUserSettingsForm.tsx
```diff
@@ -0,0 +1,140 @@
...
+import {
+  ApiError,
+  TournesolUserSettings,
+  UsersService,
+} from 'src/services/openapi';
+
+interface Props {
+  // Allowed polls. Using literal allows us to use the form settings[pollName]
+  // with TypeScript.
+  pollName: 'videos';
```
GresilleSiffle, line 21: Let's see if we can avoid to hardcode the polls' names here.

GresilleSiffle: We could use `keyof` but the resulting type would include not only the polls but also all other additional keys that the object `TournesolUserSettings` would contain, such as the future global settings. 

```javascript
pollName: keyof TournesolUserSettings;
```

GresilleSiffle: We could use the directly the `polls` array defined in the file `constants.tsx` to list all polls, but it doesn't work.

This array contains the poll `presidentielles2022` which is not a key of `TournesolUserSettings`. In fact, the prop `pollName` should not allow all polls, but only those defined by `TournesolUserSettings` to avoid triggering TypeScript errors.

```javascript
// Let's pretend 'videos' and 'presidentielle' come from polls.map(poll => poll.name);
const pollNames = ['videos', 'presidentielle2022'] as const;

interface Props {
  pollName: typeof pollNames[number];
}

// ...

// The following line produce an error as `pollName` can be 'presidentielle2022', which is not
// a key of `userSettings`.
const [rateLaterAutoRemove, setRateLaterAutoRemove] = useState(
  userSettings ? userSettings[pollName]?.rate_later__auto_remove ?? 4 : 4
);
```
```

Comment (GresilleSiffle) [reply frontend/src/features/settings/preferences/GenericPollUserSettingsForm.tsx:21]: I finally opted for the form:

```javascript
interface Props {
  pollName: typeof YOUTUBE_POLL_NAME;
}
```

It could be better if it was possible to extract the allowed polls from the constant `polls`, but I don't know if it's possible with TypeScript. My idea was:

```javascript
// This line produces the error:
//   A 'const' assertions can only be applied to references to enum members,
//   or string, number, boolean, array, or // object literals.
const pollNames = polls.filter(poll => poll.hasSettings === true).map(poll => poll.name) as const;

interface Props {
  pollName: typeof pollNames[number];
}
```

frontend/src/features/settings/preferences/GenericPollUserSettingsForm.spec.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+  afterEach(() => {
+    storeDispatchSpy.mockClear();
+  });
+
+  describe('Success', () => {
+    it('displays the defined values after a submit', async () => {
+      const { rateLaterAutoRemove, submit } = setup();
+
+      fireEvent.change(rateLaterAutoRemove, { target: { value: 16 } });
```
GresilleSiffle, line 131: Let's check that the value of the field `rateLaterAutoRemove` is 8 before triggering the event.

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1501/commits/1990c4ff34adac427337450cd8ec88e2a32fdefa

frontend/public/locales/fr/translation.json
```diff
@@ -625,6 +635,17 @@
     }
   },
   "noVideoCorrespondsToSearchCriterias": "Aucune vidéo ne correspond à vos critères de recherche.",
+  "preferences": {
+    "errorOccurredWhileRetrievingPreferences": "Désolé, une erreur est survenue, impossible de récupérer vos préférences.",
+    "preferences": "Préférences",
+    "preferencesOfThePoll": "Préférences du scrutin :"
```
amatissart, line 641: It seems to be the first time that we use the word "scrutin" in the context of video comparisons. 

Maybe we could stick to a more generic wording like "Préférences concernant {pollName}" ?

frontend/src/features/settings/preferences/GenericPollUserSettingsForm.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const { showSuccessAlert, showErrorAlert } = useNotifications();
+
+  const [disabled, setDisabled] = useState(false);
+  const [apiErrors, setApiErrors] = useState<ApiError | null>(null);
+
+  // List of user's settings.
+  const userSettings = useSelector(selectSettings).settings;
+  const [rateLaterAutoRemove, setRateLaterAutoRemove] = useState(
+    userSettings ? userSettings[pollName]?.rate_later__auto_remove ?? 4 : 4
+  );
```
amatissart, line 40: can be simplified
```suggestion
  const [rateLaterAutoRemove, setRateLaterAutoRemove] = useState(
    userSettings?.[pollName]?.rate_later__auto_remove ?? 4
  );
```

It could be relevant to define the default value 4 in constant.tsx, and explain that the default behavior is actually controlled by the backend.

frontend/public/locales/en/translation.json
```diff
@@ -305,6 +305,15 @@
     "exportAllData": "Export all data",
     "deleteAccount": "Delete account"
   },
+  "pollUserSettingsForm": {
+    "errorOccurredDuringPreferencesUpdate": "Sorry, an error has occurred, cannot update your preferences.",
+    "preferencesUpdatedSuccessfully": "Preferences updated successfully.",
+    "rateLater": "Rate-later list",
+    "autoRemove": "Auto-remove",
```
amatissart, line 312:
```suggestion
    "autoRemove": "Automatic removal",
```

frontend/public/locales/fr/translation.json
```diff
@@ -311,6 +311,16 @@
...
     "deleteAccount": "Supprimer votre compte"
   },
+  "pollUserSettingsForm": {
+    "errorOccurredDuringPreferencesUpdate": "Désolé, une erreur est survenue, impossible de mettre à jour vos préférences.",
+    "preferencesUpdatedSuccessfully": "Préférences mises à jour avec succès.",
+    "rateLater": "Liste à comparer plus tard",
+    "autoRemove": "Suppression automatique",
+    "autoRemoveHelpText_one": "Les entités seront supprimées de votre liste à comparer plus tard après {{rateLaterAutoRemove}} comparison.",
+    "autoRemoveHelpText_many": "Les entités seront supprimées de votre liste à comparer plus tard après {{rateLaterAutoRemove}} comparisons.",
+    "autoRemoveHelpText_other": "Les entités seront supprimées de votre liste à comparer plus tard après {{rateLaterAutoRemove}} comparisons.",
```
amatissart, line 321: maybe "élements" instead of "entités"?

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1501/commits/ab1b0ead398c986fab8854d74271706eab1df673

frontend/public/locales/fr/translation.json
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1501/commits/ab1b0ead398c986fab8854d74271706eab1df673

frontend/src/features/settings/preferences/GenericPollUserSettingsForm.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1501/commits/8d5769a2c005d11cddfeecd806fc3614328f74b7

frontend/src/features/settings/userSettingsSlice.ts
```diff
@@ -15,15 +16,22 @@ export const fetchUserSettings = createAsyncThunk(
...
+    // Replace all user's settings of all polls by new ones.
+    replaceSettings: (state, action) => {
+      state.settings = action.payload;
+    },
+  },
   extraReducers: (builder) => {
-    /**
-     * Replace all user's settings by new ones.
-     */
+    // Replace all user's settings of all polls by new ones.
```
amatissart, line 26: Side question, not directly related to this PR:
The settings are currently persisted when the user logs out. Maybe they would be expected to be removed from local storage on logout?

amatissart (approved):

frontend/src/features/settings/userSettingsSlice.ts
GresilleSiffle: Good idea, let's do it.

Issue #1467, opened 2023-03-25
[back] feat: Website preview for recommendation page


lfaucon:
For links to the recommendation page such as: https://tournesol.app/recommendations?date=Month&language=en display only the default preview when shared on social media. It would make them a lot more attractive to have one. I propose to have one similar to the screenshot below:

<img width="915" alt="Screenshot 2023-03-25 at 17 47 32" src="https://user-images.githubusercontent.com/8818081/227730747-1cf87d07-2933-451f-b4a1-70625434c506.png">

At bottom is a shadow to make it noticeable that it is a list and that it can be scrolled on the website.

PR #1504, opened 2023-04-13
[back] feat: Recommendation page website preview

OoIwazaruoO:
**related to** #1467

---

This PR has for goal the display of tournesol recommendations on the recommendations preview based on the url search parameters. Currently the preview is the tournesol website default preview.

- [x] retrieve recommendations
- [x] edit the preview
    - [x] display the headline
    - [x] display the overlay
    - [x]  display recommendations boxes
        - [x] display videos metadata
            - [x] display yt thumbnail
            - [x] display videos title
            - [x] display videos views
            - [x] display videos date
            - [x] display videos channel
            - [x] display timer
        - [x] display tournesol data
            - [x] display tournesol logo 
            - [x] display tournesol score
            - [x] display rating number
            - [x] display contributors number
- [x]  update request data to get a more accurate recommendations result
    - [x] update request language data
    - [x] update request date data 
- [x] update font
- [x] remove the grey background from tournesol logo
- [x] display no recommendations message when no recommendations are found
- [x] Refactor the code to pass the lint test

Current Result for recommendations?language=en%2Cfr&date=&search=guerre

image

Current Result for recommendations?language=en&date=&search=guerre (no recommendations for search guerre for English language) and for recommendations?date=Month&language=en (there is no recommendation for the last month on localhost)

image

Current Result for recommendations?date=Year&language=en (there is recommendation for the last year on localhost)

image


lfaucon (changes requested):
Thanks a lot for this awesome feature!

One thing I am worried about the implementation is that it may often and easily start failing without us being aware of it. This is because it depends a lot on both how the frontend urls are constructed and also the backend recommendation route.

To avoid this problem, I think we should try to have some sort of e-2-e test here: open the frontend and navigate to "last week" & "english" (or some set of filtering options), then fetch the preview based on the frontend URL. I don't think we will be able to test that the expected videos are on the preview, but it would be great if we could.

Instead or additionally, 5-6 unit tests with hardcoded urls is definitely a must.

amatissart:
> One thing I am worried about the implementation is that it may often and easily start failing without us being aware of it. This is because it depends a lot on both how the frontend urls are constructed and also the backend recommendation route.
> 
> To avoid this problem, I think we should try to have some sort of e-2-e test here: open the frontend and navigate to "last week" & "english" (or some set of filtering options), then fetch the preview based on the frontend URL. I don't think we will be able to test that the expected videos are on the preview, but it would be great if we could.

I added 2 small unit tests, and one e2e test to check that frontend parameters are correctly transformed.

lfaucon (approved):

PR #1512, opened 2023-04-17
[front] chore: clear user's settings in the store after a logout

### LINKED ISSUE #1510: [front] chore: clear user's settings in the store after a logout

ShinseiNova:
Close #1510

The goal is to prevent the user to leave tracks of its preferences in the browser after having finished to use Tournesol.

### DIFF
--- a/frontend/src/hooks/useLoginState.ts
+++ b/frontend/src/hooks/useLoginState.ts
@@ -15,7 +15,12 @@ export const useLoginState = () => {
 
   const logout = useCallback(() => {
     dispatch(logoutAction());
-    dispatch(clearSettings());
+    /**
+     * Ensure that the user's settings retrieved after the login are cleared 
+     * after a logout, to not leave trace of the user's preferences in the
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useLoginState.ts
```diff
@@ -14,6 +15,7 @@ export const useLoginState = () => {
 
   const logout = useCallback(() => {
     dispatch(logoutAction());
+    dispatch(clearSettings());
```
GresilleSiffle, line 18: It could be good to explain to the other developers the reason behind this clear.

```suggestion
    /**
     * Ensure that the user's settings retrieved after the login are cleared 
     * after a logout, to not leave trace of the user's preferences in the
     * browser. 
     */
     dispatch(clearSettings());
```

GresilleSiffle (commented):
It works :+1: 

We can add a small comment before merging.

GresilleSiffle (approved):

PR #1522, opened 2023-04-23
Unavailable videos comparisons list

OoIwazaruoO:
***Related to #1070#issuecomment-1517741134***

---------------------------

This PR has for goal the display of unavailable entity on the comparisons list page.
The unavailable video must be highlighted with a message overlay on top of the image life for the comparison page

- [ ] Fix entity card uneven on the comparisons list

### DIFF
--- a/frontend/src/components/entity/EntityCard.tsx
+++ b/frontend/src/components/entity/EntityCard.tsx
@@ -53,7 +53,7 @@ const EntityCard = ({
   const [settingsVisible, setSettingsVisible] = useState(!isSmallScreen);
 
   useEffect(() => {
-    setContentDisplayed(isAvailable);
+    setContentDisplayed(isAvailable || compact);
   }, [isAvailable]);
 
   const displayEntityCardScores = () => {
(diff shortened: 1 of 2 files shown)


frontend/src/components/entity/EntityCard.tsx
```diff
@@ -114,6 +115,30 @@ const EntityCard = ({
...
+                    fontSize: '0.6rem',
+                  },
+                }}
+              >
+                <Typography textAlign="center" fontSize="inherit">
+                  {entity.type == TypeEnum.VIDEO
+                    ? t('video.notAvailableAnymore')
+                    : t('entityCard.thisElementIsNotAvailable')}
+                </Typography>
+              </Box>
```
GresilleSiffle, line 140: The same components are also used by `<EntitySelectorInnerAuth>`.

Let's factorize them in their own component.

GresilleSiffle (commented):
It looks good to me.

I'm going to refactor a duplicate piece of code before merging.

GresilleSiffle (changes requested):
I just realized that those changes break the visual of the My comparisons page on the `presidentielle2022` poll.

We should fix it.

Issue #1023, opened 2022-06-29
[back] feat: Add option to exclude compared videos from recommendations


lfaucon:
**Acceptance criteria:**

Similar to the unsafe option, the request to recommendations (for example 
https://api.tournesol.app/polls/videos/recommendations/?limit=20&date_gte=2023-03-17T19%3A00%3A00.000Z&unsafe=true&metadata%5Blanguage%5D=en&score_mode=default, see class `PollsRecommendationsView`) accepts an option `exclude_compared_entities` to only include not compared videos. Note that the route currently does not use authentication, if the user is not authenticated the option should be ignored.

PR #1526, opened 2023-04-25
[back] feat: Exclude already compared in recommendations

sintaz1992:
#1023

### DIFF
--- a/backend/tournesol/views/polls.py
+++ b/backend/tournesol/views/polls.py
@@ -14,7 +14,7 @@ from drf_spectacular.utils import (
 from rest_framework import serializers
 from rest_framework.generics import ListAPIView, RetrieveAPIView
 
-from tournesol.models import Entity, Poll
+from tournesol.models import ContributorRating, Entity, Poll
 from tournesol.models.entity_score import ScoreMode
 from tournesol.models.poll import ALGORITHM_MEHESTAN
 from tournesol.serializers.entity import EntityCriteriaDistributionSerializer
@@ -148,6 +148,13 @@ class PollRecommendationsBaseAPIView(PollScopedViewMixin, ListAPIView):
             rating_n_contributors__gte=settings.RECOMMENDATIONS_MIN_CONTRIBUTORS,
             tournesol_score__gt=0,
         )
+
+    def exclude_rated_entities(self, queryset, filters, user):
+        if filters["exclude_rated_entities"] and user.is_authenticated:
+            user_rating_qs = ContributorRating.objects.filter(user=user)
+            rated_entities = [rating.entity_id for rating in user_rating_qs]
+            return queryset.exclude(id__in=rated_entities)
+        return queryset
 
     def sort_results(self, queryset, filters):
         """
@@ -297,10 +304,12 @@ class PollsRecommendationsView(PollRecommendationsBaseAPIView):
 
     def get_queryset(self):
         poll = self.poll_from_url
+        user = self.request.user
         queryset = Entity.objects.all()
         queryset, filters = self.filter_by_parameters(self.request, queryset, poll)
         queryset = self.annotate_and_prefetch_scores(queryset, self.request, poll)
         queryset = self.filter_unsafe(queryset, filters)
+        queryset = self.exclude_rated_entities(queryset, filters, user)
         queryset = self.sort_results(queryset, filters)
         return queryset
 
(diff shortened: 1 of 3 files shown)


GresilleSiffle:
Thank you @sintaz1992 , it's nice to see you around : )

I haven't reviewed the code yet, but the tests look good and exhaustive :+1: 

Note for the reviewers: as a precaution let's check the performance of the new SQL query for a user with a big number of compared entities (~ 500 or more). The `/recommendations/` endpoint is currently one our endpoints able to answer in more than 500 ms.

backend/tournesol/views/polls.py
```diff
@@ -149,6 +149,13 @@ def filter_unsafe(self, queryset, filters):
             tournesol_score__gt=0,
         )
 
+    def exclude_rated_entities(self, queryset, filters, user):
+        if filters["exclude_rated_entities"] and user.is_authenticated:
+            user_rating_qs = ContributorRating.objects.filter(user=user)
```
GresilleSiffle, line 154: I think I found a possible bug here.

`ContributorRating.objects.filter(user=user)` is an approximation of the entities rated by the user, it doesn't represent exactly the rated entities.

In fact it's possible for a user to create one or more `ContributorRating` objects without comparing the related entities. This is what happens when we paste YouTube URLs in the comparison page. When receiving the URL,  and even before the comparison is submitted, the front end component creates a  `ContributorRating` for the couple user / entities.

**option (a)**

A better approximation could be (SQL pseudo code):
- SELECT the `ContributorRating`
- INNER JOIN `ContributorRatingCriteriaScore`
- ON `ContributorRatingCriteriaScore.contributor_rating_id` = `ContributorRating.id`
- WHERE `ContributorRating.user_id` = authenticated user id
- AND `ContributorRating.poll_id` = poll.id
- AND `ContributorRatingCriteriaScore.criteria` = poll.main_criteria

Here we are doing a SQL INNER JOIN to keep only the compared entities. The compared entities are the entities having at least a computed score for the main criteria (largely recommended for the `videos` poll).

This is still an approximation as the score are computed only after a run of the algorithm, meaning it's possible to compare an entity and to have the `ContributorRatingCriteriaScore` table empty (until the next run).

**option (b)**

If we want to exclude the compared entities in the real time, we need to filter on the rating given by the user at the time of the comparison instead of the computed score which exists only after a certain amount of time in the database.

SQL pseudo code:

- SELECT the Comparison
- JOIN  ComparisonCriteriaScore
- ON ComparisonCriteriaScore.comparison_id = Comparison.id
- WHERE Comparison.user_id = authenticated user id
- AND Comparison.poll_id = poll.id
- AND ComparisonCriteriaScore.criteria = poll.main_criteria

If an entity has been compared multiple times by the user, it will appear several times in the results of the previous SQL query. I don't know how it will impact the performance of global recommendations queryset, and if it's relevant to filter the duplicates. If it is we either need to filter the duplicates in the SQL query (might be tricky), or to filter them directly in Python with `set` (may be slower than a filter in the SQL query).

I propose to investigate the feasibility of the **option (b)** first, as it represents what we want to have minus the duplicates. We can check the performance of option b:
- with the duplicates
- without the duplicates

GresilleSiffle (commented):
Hello @sintaz1992 @lfaucon 

Thanks a lot for the pull request. This feature will greatly improve the recommendations displayed by the extension.

I found a little problem with the current SQL, it doesn't seem to do exactly what we want. In its current state, the query will exclude more entities than expected.

I proposed two options. One "easy" to implement which is still an approximation but instead of filtering more entities than expected, it filters less entities than expected. And another one that may be a bit more tricky but does exactly what we want.

@sintaz1992 tell me what do you think. Do you have the `dev-env` with the Docker containers up and running? It will help to test the performances of the SQL queries as it's possible to download the public database of the production server when starting the containers.

backend/tournesol/views/polls.py
lfaucon: > SQL pseudo code:
> 
> * SELECT the Comparison
> * JOIN  ComparisonCriteriaScore
> * ON ComparisonCriteriaScore.comparison_id = Comparison.id
> * WHERE Comparison.user_id = authenticated user id
> * AND Comparison.poll_id = poll.id
> * AND ComparisonCriteriaScore.criteria = poll.main_criteria

I don't understand why do you recommend to use the ComparisonCriteriaScore too? Can there be comparison object without the associated criteria scores?

GresilleSiffle: It looks unnecessary indeed, it's possibly a leftover of the option (a) pseudo query.

Currently our API doesn't allow users to create comparison without the main criterion.

(Nevertheless, I remember correctly we have comparisons in the database that don't have a related `ComparisonCriteriaScore` for the main criterion, but the suggested SQL join won't fix the issue).

lfaucon (approved):

GresilleSiffle (approved):
Good!

PR #1531, opened 2023-04-29
[solidago] Move algorithms for individual scores in Solidago, Introduce Continuous Bradley-Terry model (CBT)

amatissart:
The legacy algorithm (based on matrix inversion) and the new CBT are part of Solidago.
The algorithm used in Tournesol is unchanged.

Reference paper for CBT:
https://www.overleaf.com/project/5f44dd8e84c8540001bf1552 (section 3)


backend/ml/tests/test_mehestan_individual.py
```diff
@@ -0,0 +1,121 @@
...
+                for idx in range(20)
+            ]
+        )
+        scores = compute_method(comparisons)
+        scores.sort_values("raw_score", inplace=True)
+
+        # Video are ranked according to the chain of comparisons
+        assert list(scores.index) == sorted(scores.index)
+
+        # video_10 is in the middle with score 0.0
```
lfaucon, line 49:
```suggestion
        # Compared videos are indexed 0 to 20 included; video_10 is in the middle with score 0.0
```

backend/ml/tests/test_mehestan_individual.py
```diff
@@ -0,0 +1,121 @@
...
+        assert scores.loc["video_10"].raw_score == pytest.approx(0.0, abs=1e-4)
```
lfaucon, line 50:
```suggestion
        assert scores.loc["video_10"].raw_score == pytest.approx(0.0, abs=1e-4)
        assert scores.loc["video_00"].raw_score == pytest.approx(scores.loc["video_20"].raw_score, abs=1e-4)
```

lfaucon (commented):
Tests look great. Only small doubts about `test_comparisons_monotony` but we discussed it. OK

backend/ml/tests/test_mehestan_individual.py
```diff
@@ -0,0 +1,138 @@
...
+            ("A", "B", random.randint(-10, 10)),
+            ("A", "C", random.randint(-10, 10)),
+            ("A", "D", random.randint(-10, 10)),
+            ("B", "C", random.randint(-10, 10)),
+            ("B", "D", random.randint(-10, 10)),
+            ("C", "D", random.randint(-10, 10)),
+        ]
+        comparisons_df = pd.DataFrame(comparisons, columns=["entity_a", "entity_b", "score"])
+        scores = compute_method(comparisons_df)
+        assert scores.raw_score.mean() == pytest.approx(1.0, abs=1e-4)
```
glerzing, line 67: I don't understand that test. Why do we check that it is "1.0" instead of 0 ? And why do we pick random integers, instead of just the same arbitrary number each time ?

backend/ml/tests/test_mehestan_individual.py
```diff
@@ -0,0 +1,138 @@
...
+            ("C", "D", 0),
+            # strong preferences
+            ("A", "E", 9),
+            ("B", "F", 10),
+        ]
+        comparisons_df = pd.DataFrame(comparisons, columns=["entity_a", "entity_b", "score"])
+        scores = compute_method(comparisons_df)
+
+        # F is preferred to E
+        assert scores.loc["F"].raw_score > scores.loc["E"].raw_score
```
glerzing, line 88: It's a bit confusing because "traditionally", in Tournesol when we say that the comparison score between B and F is 10, we mean that B is better than F. But here it seems to mean the contrary.

backend/ml/tests/test_mehestan_individual.py
```diff
@@ -0,0 +1,138 @@
...
+            ("E", "H", 1),
+        ]
+        comparisons_df = pd.DataFrame(comparisons, columns=["entity_a", "entity_b", "score"])
+        scores = compute_method(comparisons_df)
+
+        assert scores.loc["A"].raw_score == pytest.approx(0.0, abs=1e-4)
+        assert scores.loc["A"].raw_score == pytest.approx(scores.loc["E"].raw_score, abs=1e-4)
+        assert scores.loc["B"].raw_score == pytest.approx(scores.loc["F"].raw_score, abs=1e-4)
+        assert scores.loc["C"].raw_score == pytest.approx(scores.loc["G"].raw_score, abs=1e-4)
+        assert scores.loc["D"].raw_score == pytest.approx(scores.loc["H"].raw_score, abs=1e-4)
```
glerzing, line 138: I've read this sentence in the paper that with CBT : 

> Essentially, we consider that when a
> contributor reports rab, it is as though answered they preferred a a (non-integer) number 1 + rab
> of times, and they reported preferring b a number 1 − rab of times

Shouldn't we have a test for this ?
Like, have 3 videos A, B, C and D with the same rating and set the comparisons :
("A", "B", 0)
("A", "C", 0)
("A", "D", 0)
("B", "C", 0)
("B", "D", 0)
("C", "D", 0)
("A", "E", 2)
("B", "E", 8)
("C", "F", 5)
("D", "F", 5)
assert score.loc["E"].raw_score == pytest.approx(scores.loc["F"].raw_score, abs=1e-4)

backend/tournesol/tests/test_ml_run.py
```diff
@@ -235,6 +236,7 @@ def test_tournesol_scores_different_trust(self):
             places=3,
         )
 
+    @pytest.mark.skip("Uncertainty behaves differently with BBT")
```
glerzing, line 239: Another potential simple test is:
  user1 rates +10
  user2 rates -10
  assert(score == 0)
  user1 rates +10
  assert(score > 0)

backend/ml/mehestan/individual_bbt.py
```diff
@@ -0,0 +1,142 @@
...
+    result.index.name = "entity_id"
+    return result
+
+
+# if __name__ == "__main__":
+#     from ml.inputs import MlInputFromPublicDataset
+#     ml_input = MlInputFromPublicDataset("https://api.tournesol.app/exports/all/")
+#     comp = ml_input.get_comparisons(user_id=116, criteria="largely_recommended")
+#     res = compute_individual_score(comp)
+#     print(res.sort_values("raw_score", ascending=False))
```
glerzing, line 142: Debug code to remove before pushing.

backend/ml/tests/test_mehestan_individual.py
amatissart: `1.0` instead of `0.0` was a typo.  
The random values are used to show that the property is verified for any comparison set.

backend/ml/tests/test_mehestan_individual.py
lenhoanglnh: It almost works but does not quite. It took us a bit of time to figure this one out! 
It would work perfectly if the scores of A, B, C and D were equal at optimum.
However, they will themselves be moved by different amounts towards negative scores.

backend/ml/tests/test_mehestan_individual.py
amatissart: fixed in 8ee16ea581017333f0

backend/ml/tests/test_mehestan_individual.py
amatissart: fixed in 8ee16ea581017333f0

backend/ml/tests/test_mehestan_individual.py
```diff
@@ -0,0 +1,138 @@
...
+            ("A", "B", random.randint(-10, 10)),
+            ("A", "C", random.randint(-10, 10)),
+            ("A", "D", random.randint(-10, 10)),
+            ("B", "C", random.randint(-10, 10)),
+            ("B", "D", random.randint(-10, 10)),
+            ("C", "D", random.randint(-10, 10)),
+        ]
+        comparisons_df = pd.DataFrame(comparisons, columns=["entity_a", "entity_b", "score"])
+        scores = compute_method(comparisons_df)
+        assert scores.raw_score.mean() == pytest.approx(0.0, abs=1e-4)
```
glerzing, line 67: I still don't understand why we use random.randint. Unless the seed is the same each time and it randomly happens to work, I would expect that this test can fail. I would have done something like :
``` Python
        n = 8
        comparisons = [
            ("A", "B", n),
            ("A", "C", n),
            ("A", "D", n),
            ("B", "C", n),
            ("B", "D", n),
            ("C", "D", n),
        ]
```
You could also set n = random.uniform(-10, 10) if you want, but setting a different value for each comparison seems buggy.

backend/ml/mehestan/individual_bbt.py
```diff
@@ -0,0 +1,147 @@
...
+            pd.DataFrame(
+                {
+                    "entity_a": scores.entity_b,
+                    "entity_b": scores.entity_a,
+                    "score": -1 * scores.score,
+                }
+            ),
+        ]
+    )
+    r = scores_sym.pivot(index="entity_a", columns="entity_b", values="score") / R_MAX
```
glerzing, line 110: Shouldn't it be (1.0 + R_MAX) rather than R_MAX ?

glerzing: BTW this doesn't seem explicit in overleaf.

backend/ml/mehestan/individual_bbt.py
```diff
@@ -0,0 +1,147 @@
...
+        indices, r_ab = coord_to_subset[coord]
+        old_theta_a = theta[coord]
+        theta_b = theta[indices]
+        new_theta_a = coordinate_optimize(r_ab, theta_b, precision=EPSILON / 10)
+        theta[coord] = new_theta_a
+        if abs(new_theta_a - old_theta_a) < EPSILON:
+            unchanged.add(coord)
+        else:
+            unchanged.clear()
+    return theta
```
glerzing, line 87: Can't this function be adapted for @njit ?

backend/ml/tests/test_mehestan_individual.py
glerzing: Yes, the random values actually make sense.

lenhoanglnh (commented):
I think this can wait for another version, but it would be nice to better isolate the different components of the algorithms in different task-specific folders.

Typically, continuous bradley terry should be in a folder outside mehestan callled ```comparisons_to_scores```.
Similarly, optimize.py and indiv_matrix_inversion.py should be in a folder outside mehestan called ```solvers```.
Finally, the primitives like QrMed, BrMean and others could be in a folder called ```resilient_primitives```.

solidago/src/solidago/mehestan/individual/indiv_matrix_inversion.py
```diff
@@ -0,0 +1,86 @@
...
+                        "entity_a": scores.entity_b,
+                        "entity_b": scores.entity_a,
+                        "score": -1 * scores.score,
+                    }
+                ),
+            ]
+        )
+
+        # "Comparison tensor": matrix with all comparisons, values in [-R_MAX, R_MAX]
+        r = scores_sym.pivot(index="entity_a", columns="entity_b", values="score")
```
lfaucon, line 40: out-of-scope:
Would this pivot become a memory bottleneck? 
In my understanding that makes us use O(videos^2) (400M with our current 20k videos), while we could use only O(comparisons) with a sparse matrix. It seems that the below operations would work exactly in the same way.

UPDATE: As it is for individual comparisons, it may be fine. However it opens us to an attack from a user that would make 50k comparisons between 100k videos and crash our server when the ML run allocates >10Gb for them. This is easily feasible with current rate limits

solidago/src/solidago/mehestan/individual/continuous_bradly_terry.py
```diff
@@ -0,0 +1,156 @@
...
+                pd.DataFrame(
+                    {
+                        "entity_a": scores.entity_b,
+                        "entity_b": scores.entity_a,
+                        "score": -1 * scores.score,
+                    }
+                ),
+            ]
+        )
+        r = scores_sym.pivot(index="entity_a", columns="entity_b", values="score") / self.r_max
```
lfaucon, line 118: Same comment

solidago/src/solidago/mehestan/individual/continuous_bradly_terry.py
```diff
@@ -0,0 +1,156 @@
...
+            initial_scores = np.zeros(n_entities)
+        else:
+            initial_scores = pd.Series(initial_entity_scores, index=r.index)
+            initial_scores.fillna(0.0, inplace=True)
+            initial_scores = initial_scores.to_numpy()
+        theta_star_numpy = self.coordinate_descent(coord_to_subset, initial_scores=initial_scores)
+        delta_star_numpy = np.zeros(len(theta_star_numpy))
+        for idx in range(len(theta_star_numpy)):
+            indices, _r_ab = coord_to_subset[idx]
+            delta_star_numpy[idx] = Delta_theta(theta_star_numpy[indices])
```
lfaucon, line 137: Is it expected that the uncertainty for video_A depends only on the individual scores of the compared videos to video_A.

The formula seems to use `theta_ab`, maybe you should do the difference? 
```suggestion
            delta_star_numpy[idx] = Delta_theta(theta_star_numpy[indices] - theta_star_numpy[idx])
```
?

backend/tournesol/lib/public_dataset.py
```diff
@@ -264,9 +264,8 @@ def write_metadata_file(write_target, data_until: Optional[datetime] = None) ->
                 "TRUSTED_EMAIL_PRETRUST": TRUSTED_EMAIL_PRETRUST,
                 "VOUCH_DECAY": VOUCH_DECAY,
             },
+            "individual_scores": individual_scores_algo.get_metadata(),
```
lfaucon, line 267: Small inconsistency: We don't guarantee that the scores computed are indeed computed with this algo. The worse edgecase would be if we deploy a new algorithm on Sunday evening right before the re-computation of the dataset

solidago/src/solidago/mehestan/individual/continuous_bradly_terry.py
```diff
@@ -0,0 +1,156 @@
...
+            {
+                "raw_score": theta_star_numpy,
+                "raw_uncertainty": delta_star_numpy,
+            },
+            index=r.index,
+        )
+        result.index.name = "entity_id"
+        return result
+
+    def get_metadata(self) -> dict:
```
lfaucon, line 149: I'm very glad we created the Solidago library :)

We will regularly encounter the question: "Is it solidago's or tournesol's responsibility?"

For the case of this method `get_metadata` I would slightly (not a strong opinion) be on the side of tournesol's responsibility. Typically we want to only make changes to Solidago when we implement a new algorithm, not when we add a feature to Tournesol.

solidago/src/solidago/mehestan/individual/indiv_matrix_inversion.py
```diff
@@ -0,0 +1,86 @@
+import numpy as np
+import pandas as pd
+
+from .base import IndividualScoresAlgorithm
+
+DEFAULT_ALPHA = 1.0  # Signal-to-noise hyperparameter
+
+
+class MatrixInversionInvididualAlgo(IndividualScoresAlgorithm):
```
lfaucon, line 9: Can we quickly brainstorm another model name for this one? A name that would describe the model, not the technique. Otherwise CBT should be called `CoordinateDescentIndividualAlgorithm`

lfaucon (approved):

solidago/src/solidago/mehestan/individual/continuous_bradly_terry.py
amatissart: This `.pivot()` call has been removed from CBT in  3f48497

solidago/src/solidago/mehestan/individual/continuous_bradly_terry.py
amatissart: Yes indeed. Fixed in 99ba5e5

backend/tournesol/lib/public_dataset.py
amatissart: Indeed. We could consider saving the parameters in the database. But the situation could become more complex with online updates where scores for different users / entities will be computed at different times with potentially different configurations of the algorithm.

This was briefly discussed when implementing https://github.com/tournesol-app/tournesol/pull/1418

solidago/src/solidago/mehestan/individual/indiv_matrix_inversion.py
amatissart: I agree this name is not great. I'm open to suggestions.

solidago/src/solidago/mehestan/individual/continuous_bradly_terry.py
amatissart: Even if Tournesol is responsable for building the metadata attached to the public dataset, I think that it makes sense for each algorithm to expose its entire parameter set. Otherwise Tournesol's backend would need to be know about solidago implementation details to write the public dataset, and it would make switching between algorithms more cumbersome.

Or do you have another alternative in mind?

solidago/src/solidago/mehestan/individual/indiv_matrix_inversion.py
lfaucon: Following up yesterday's discussion, I'd propose: `HookeIndividualScoresAlgorithm` (my prefered option, as in https://fr.wikipedia.org/wiki/Loi_de_Hooke) or `QuadraticIndividualScoresAlgorithm`

I also like better **not** to shorten Algorithm -> Algo

Issue #1527, opened 2023-04-26
[front] fix: do not call `/stats/` when the home page is loaded by the extension hidden iframe


GresilleSiffle:
When the extension loads the Tournesol home page in the hidden iframe, it triggers an HTTP request to the back end `/stats/` endpoint. This request makes several SQL queries that are not required, they should be avoided in this context.

Moreover, a request to `/stats/` can take more than 500ms and can trigger the latency monitoring alerts. 

**to-do (quick fix)**
- [x] the front end should detect when it is loaded in the hidden iframe (see the `dnt` query parameter)
- [x] the HTTP request to `/stats/` should be skipped

### Technical specifications

Now that the quick fix has been deployed we can take the time to develop a more robust method that will lazy load the stats only when at least one component needs them. The statistics will not be retrieved at the beginning of the user's navigation session as it was before.

Here are the steps to follow:

1 - Create a React context `StatsContext`

This context will contain the statistics and will provide the methods to fetch them from the API an updated them locally. It can be good to take inspiration from Redux:
- the context never returns the saved statistics themselves
- but provides accessors that return a deep copy of the saved statistics 

2 - The context should provide two methods 

`refreshStats`

This function call the API `/stats` and replace the value of the `Statistics` object in the context. It replaces the current `fetchStats` provided by our Redux slice.

`getStats`

If there is no statistics yet in the context, call `refreshStats`, and return a deep copy of the retrieved stats. It replaces the current `selectStats` provided by our Redux slice.

If there are statistics in the context, and they have been refreshed more than 4 seconds ago, call `refreshStats`, and return a deep copy of the retrieved stats.

Else, return the statistics of the context.

3 - Unplug all components using stats from Redux, and use the context's methods instead.

4 - Delete the `statsSlice`.

GresilleSiffle:
**quick fix** We can use a new query parameter set by the iframe to allow the front end to skip the call to `/stats`. 

**improvement** We can fetch the statistics lazily only when a component requires them. We may need to remove the stats from Redux and use a React context to have more control on the process.

PR #1532, opened 2023-05-01
[ext][front] fix: do not call the API /stats/ when it's not needed

GresilleSiffle:
**related to** #1527 

---

This temporary fix should reduce the number of useless call to /stats/ generated by the browser extension hidden iframe, and appease our monitoring system.

As discussed with Adrien, we agreed to replace this temporary fix by a more robust system that will lazily fetch the statistics from the back end when the front end needs to display them (see #1527)


lfaucon (approved):
Lgtm

PR #1536, opened 2023-05-03
[front] feat: FAQEntries can be folded / unfolded

ShinseiNova:


GresilleSiffle (dismissed):

GresilleSiffle:
**to-do**
- [ ] sharing a link with an anchor should unfold the related question
- [ ] clicking on an item in the table of content should unfold it
- [ ] clicking on the arrow button of a folded item should unfold it
- [ ] clicking on the arrow button of an unfolded item should fold it
- [ ] several questions can be unfolded at the same time

GresilleSiffle (approved):

PR #1542, opened 2023-05-05
[front] refactor: add StatsContext with its lazy provider

ShinseiNova:
**related to** #1527 

---

**to-do**
- [x] use the new functions to retrieve the stats on the home page
- [x] double check if we correctly designed the lazy provider (no infinite loop, no memory leak, no react error, etc.)

**if the design is appropriate**
- [x] handle the dependency arrays warning (if the dependencies are good, make the warnings silent)
- [x] unplug the stats from Redux
- [x] update all components using the stats
- [x] `getStats` should not returned the cached stats if the last refresh is older than 4 seconds
- [x] clean the prototype code in the FAQ


GresilleSiffle:
We can also remove the usage of the URL parameter `page_not_displayed` from the browser extension.

GresilleSiffle (approved):
Hello @amatissart 

We have replaced the `statsSlice` by a lazy React context provider. We made several tests and everything looks good on our side.

Do you also want to have a look at the `StatsContext` file to see if it matched the initial idea?

frontend/src/features/statistics/StatsContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const loading = useRef(false);
+  const lastRefreshAt = useRef(0);
+
+  const [stats, setStats] = useState(initialState);
+
+  const refreshStats = useCallback(async () => {
+    setStats(await StatsService.statsRetrieve());
+    loading.current = false;
+    lastRefreshAt.current = Date.now();
+  }, []);
```
amatissart, line 48: As `setStats()` will trigger a re-render, and possibly reevaluate `getStats()`, the references should be updated beforehand:

```suggestion
  const refreshStats = useCallback(async () => {
    const newStats = await StatsService.statsRetrieve();
    loading.current = false;
    lastRefreshAt.current = Date.now();
    setStats(newStats)
  }, []);
```

frontend/src/hooks/useStats.ts
```diff
@@ -0,0 +1,15 @@
+import { useState, useEffect, useContext } from 'react';
+import StatsContext from 'src/features/statistics/StatsContext';
+import { Statistics } from 'src/services/openapi';
+
+export const useStats = () => {
+  const [stats, setStats] = useState<Statistics>();
+
+  const { getStats } = useContext(StatsContext);
```
amatissart, line 8: It's possible to use `getStats()` to initialize the state, so that `stats` is never undefined:
```suggestion
  const { getStats } = useContext(StatsContext);
  const [stats, setStats] = useState<Statistics>(getStats());
```

frontend/src/features/statistics/StatsContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+    if (
+      stats.polls.length !== 0 &&
+      currentTime - lastRefreshAt.current >= EXPIRATION_TIME
+    ) {
+      refreshStats();
+      return stats;
+    }
+
+    return stats;
+  }, [stats, refreshStats]);
```
amatissart, line 73: `loading` is not used when stats are outdated, so concurrent requests can be triggered in this case.
The refresh should never be called when another request is pending:

```js
  const getStats = useCallback(() => {
    const currentTime = Date.now();

    if (loading.current) {
      return stats;
    }

    // Initialize the stats if they are empty, or if they are outdated.
    if (
      stats.polls.length === 0 ||
      currentTime - lastRefreshAt.current >= EXPIRATION_TIME
    ) {
      loading.current = true;
      refreshStats();
    }
    return stats;
  }, [stats, refreshStats]);

amatissart (commented):
Looks good to me overall, with a few suggestions in comments to fix some edge cases.

frontend/src/features/statistics/StatsContext.tsx
GresilleSiffle: Should be fixed by https://github.com/tournesol-app/tournesol/pull/1542/commits/f1e37f3fcf7aad563122847088fd7d9064a444b6

frontend/src/hooks/useStats.ts
GresilleSiffle: Should be fixed by https://github.com/tournesol-app/tournesol/pull/1542/commits/19e07570caca58f070b8e0eb4e39e716bf4627ed

PR #1543, opened 2023-05-05
[back] optim: use Numba to compile critical path in Mehestan primitives and scalings computation

amatissart:
While working on the BBT implementation (#1531), I introduced some optimizations that can be merged first to simplify the review process. This PR doesn't have any impact on the algorithm output.

1. Use `@numba.njit` on some critical functions. Numba provides just-in-time compilation to run Python functions (under some conditions) with near-native performance. In particular, Numba supports most Numpy operations. 

2. Optimize scalings computation by keeping data about reference_users (i.e "scaling calibration users") in memory, and avoiding expensive indexation

### DIFF
--- a/backend/ml/inputs.py
+++ b/backend/ml/inputs.py
@@ -180,7 +180,7 @@ class MlInputFromDb(MlInput):
     @cached_property
     def ratings_properties(self):
         # This makes sure that `get_scaling_calibration_users()` is evaluated separately, as the
-        # table names mentionned in its RawSQL quey could conflict with the current queryset.
+        # table names mentionned in its RawSQL query could conflict with the current queryset.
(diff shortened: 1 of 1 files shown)


backend/ml/inputs.py
```diff
@@ -179,13 +179,16 @@ def get_comparisons(self, criteria=None, user_id=None) -> pd.DataFrame:
 
     @cached_property
     def ratings_properties(self):
+        # This makes sure that `get_scaling_calibration_users()` is evaluated separately, as the
+        # table names mentionned in its RawSQL quey could conflict with the current queryset.
```
glerzing, line 183: quey => query

glerzing (approved):

Issue #1508, opened 2023-04-17
[front] feat: allow the users to update their recommendation settings


GresilleSiffle:
### Context

The back end currently contains 3 settings related to the front end recommendations page:
- `recommendations__default_date`
- `recommendations__default_language`
- `recommendations__default_unsafe`

**(a)**

The front end preferences page should allow the users to visualize those settings and to update them. The form `<VideosPollUserSettingsForm>` should contain a new section "Recommendations page" below the current Rate-later list section.
 
Example (the setting `recommendations__default_unsafe` is missing)

![capture](https://user-images.githubusercontent.com/39056254/232485499-6645e3a6-e34e-4397-a432-9adc415e43fc.png)

**(b)**

For logged users, the recommendations page should use the users' preferences as default value for the filters. 

### Acceptance criteria

**(a)** and **(b)**

### Future improvements

It could be nice to also allow the users to update their settings directly from the recommendations page. A small round button could define the current filters' values as the user's preferences.

This new button could be displayed in the top right corner of the `<SearchFilter>` component.

If this feature is accepted, a new issue should be created.

GresilleSiffle:
The default languages field will be added in the front end after https://github.com/tournesol-app/tournesol/issues/1560

GresilleSiffle:
The default languages field will be added in the front end after the following task has been developed https://github.com/tournesol-app/tournesol/issues/1560

PR #1553, opened 2023-05-11
[front] feat: add default unsafe preference for recommendations page

ShinseiNova:
**related to #1508**


GresilleSiffle:
Here is the expected behaviour:

If a user (anonymous or authenticated) clicks on a specific recommendation URL containing recommendation filters, his/her personal preferences should not override those filters. This allow the users to share links between each other. This is the current behaviour.

If an anonymous user clicks on the recommendation link of the front end's side bar, the recommendation should be filtered using the poll's default recommendation filters. This is the current behaviour.

If an authenticated user clicks on the recommendation link of the front end's side bar:
- the poll's default filters should be used to initialize the filters
- the poll's default filters should be overridden by configured user settings (configured means != than undefined/null)

Example:

```bash
# considering those poll default filters
date=month&unsafe=false

# and the following user setting
unsafe=true

# the expected recommendation link in the sidebar is
https://tournesol.app/recommendations/?date=month&unsafe=true

# and not
https://tournesol.app/recommendations/?unsafe=true
```

GresilleSiffle (changes requested):
Thank you @ShinseiNova 

Here is a first batch of improvements we can already work on. Let me know if you have any question.

GresilleSiffle (commented):
The code is almost ready!

It could be good to have some functions that do the matching between the user's settings and the values expected by the front end. Currently it's the responsibility of the side bar, and in my opinion it should be moved in external functions for a greater readability / maintainability.

For instance the value `ALL_TIME` of the setting `recommendations__default_date` is translated in an empty string by the side bar. The boolean value `false` or the setting `recommendations__default_unsafe` is also translated in an empty string.

GresilleSiffle (commented):
I removed the default languages setting for now.

As discussed with the rest of the team we don't want to maintain at the same time a preference system based on the local storage only (updating the lang from the recommendations page) and  a system based on the API (updating the lang from the preferences page).

We will first update the extension to allow the users to select their preferred languages, then we will re-introduced the default languages setting.

frontend/public/locales/en/translation.json
```diff
@@ -328,8 +328,15 @@
     "preferencesUpdatedSuccessfully": "Preferences updated successfully.",
     "comparisonPage": "Comparison page",
     "rateLater": "Rate-later list",
+    "recommendationsPage": "Recommendations page",
     "updatePreferences": "Update preferences"
   },
+  "videosUserSettingsForm": {
+    "recommendations": {
+      "defaultUploadDate": "Default value of the Uploaded filter",
```
lfaucon, line 336:
```suggestion
      "defaultUploadDate": "Default value of the uploaded date filter",
```

frontend/public/locales/fr/translation.json
```diff
@@ -336,8 +336,15 @@
     "preferencesUpdatedSuccessfully": "Préférences mises à jour avec succès.",
     "comparisonPage": "Comparaison (page)",
     "rateLater": "Liste à comparer plus tard",
+    "recommendationsPage": "Recommandations (page)",
     "updatePreferences": "Mettre à jour les préférences"
   },
+  "videosUserSettingsForm": {
+    "recommendations": {
+      "defaultUploadDate": "Valeur par défaut du filtre Mis en ligne",
```
lfaucon, line 344:
```suggestion
      "defaultUploadDate": "Valeur par défaut du filtre date de mise en ligne",
```

frontend/src/features/settings/preferences/VideosPollUserSettingsForm.tsx
```diff
@@ -43,6 +48,16 @@ const VideosPollUserSettingsForm = () => {
     DEFAULT_RATE_LATER_AUTO_REMOVAL
   );
 
+  // Recommendations (page)
+  const [recoDefaultUnsafe, setRecoDefaultUnsafe] = useState(false);
```
lfaucon, line 52: When reloading the page, after having set my configuration to True, I see it showing false for a second, then showing my selected value. I'd recommend not displaying until the settings are fetched.

lfaucon (approved):

frontend/src/features/settings/preferences/VideosPollUserSettingsForm.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1553/commits/dab7cc70653f759fe23a1ad83c727faf74efbb68

PR #1561, opened 2023-05-20
[front] feat: add id to all HomeVideos' section titles

GresilleSiffle:
Ex:

https://tournesol.app#research now leads to the research section


lfaucon (approved):
Thanks!

Issue #557, opened 2022-02-09
Define priorities and strategies for Tournesol Wiki


amatissart:
This issue arise during the last retrospective. The role of [wiki.tournesol.app](https://wiki.tournesol.app) is unclear.

- Some of the content present on the Wiki would be more useful in the frontend (e.g to guide first-time users).
- The Wiki stayed mostly unchanged in the last few months.
- Is there a specific need to provide "advanced" documentation in the Wiki (e.g for researchers)?
- Need for internationalization? Or english only ?
- How should the Wiki be integrated with the Tournesol platform?

--- 
List of all pages on the current Wiki:
https://wiki.tournesol.app/wiki/Special:AllPages

amatissart:
@lenhoanglnh During the discussion, it appeared that you are probably the person with the best view of how useful the wiki could be. It would be nice to discuss about this at some point.

amatissart:
The decision has been made to remove the wiki, currently unmaintained.

- [ ] Export wiki dump in a more flexible format.
**Wiki pages in Markdown format:** [wiki_pages_markdown.zip](https://github.com/tournesol-app/tournesol/files/10834913/wiki_pages_markdown.zip)

- [ ] Move relevant content to the frontend:  
#1388 
- [ ] Remove references to the wiki
- [ ] Remove and disable wiki deployment

GresilleSiffle:
We can also remove the corresponding oauth applications created in the back end admin.

lfaucon:
Last step to close this ticket:

- [ ] Clean-up infrastructure code to remove everything needed for the wiki

PR #1565, opened 2023-05-21
infra: disable Tournesol wiki

amatissart:
Once the wiki is disabled and no issue is observed, the infra configuration will be removed

Related to #557

### DIFF
--- a/infra/ansible/roles/mediawiki/tasks/main.yml
+++ b/infra/ansible/roles/mediawiki/tasks/main.yml
@@ -94,8 +94,7 @@
 
 - name: Enable Nginx Mediawiki configuration
   file:
-    src: /etc/nginx/sites-available/mediawiki
-    dest: /etc/nginx/sites-enabled/mediawiki
+    path: /etc/nginx/site-enabled/mediawiki
     state: absent  # 2023-05-21: wiki is disabled
   notify:
     - Reload Nginx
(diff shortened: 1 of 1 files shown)


infra/ansible/roles/mediawiki/tasks/main.yml
```diff
@@ -96,7 +96,7 @@
   file:
     src: /etc/nginx/sites-available/mediawiki
     dest: /etc/nginx/sites-enabled/mediawiki
-    state: link
+    state: absent  # 2023-05-21: wiki is disabled
```
GresilleSiffle, line 99: I try a deployment on my local VM but this command returns an error (sorry I didn't copy it).

I was able to make the deployment works with the following code instead:

```suggestion
    path: /etc/nginx/site-enabled/mediawiki
    state: absent  # 2023-05-21: wiki is disabled
```

GresilleSiffle (commented):
Hello @amatissart 

I had to slightly modify the code to make it work. Have you got the same problem?

infra/ansible/roles/mediawiki/tasks/main.yml
amatissart: Good catch , thanks.

GresilleSiffle (approved):

PR #1569, opened 2023-05-22
[back] feat: Add sum of trust scores on EntityPollRating

lfaucon:
Changes:
- The unsafe option uses the ratings (we really need this change, because otherwise adding comparison on a video adds it to the recommendations before the score is recomputed)
- Adding the pre-computed field `sum_trust_scores` on the EntityPollRating model. This field is computed right at the end of the ML (I think it is the right place for this)


lfaucon:
@amatissart @GresilleSiffle Do you have suggestions for fixing all the tests failing because of the change in this PR?

backend/tournesol/models/entity_poll_rating.py
```diff
@@ -67,4 +74,18 @@ def update_n_ratings(self):
             .distinct("user")
             .count()
         )
-        self.save(update_fields=["n_comparisons", "n_contributors"])
+
+        self.save(update_fields=["n_comparisons", "n_contributors", "sum_trust_scores"])
+
+    @staticmethod
+    def bulk_update_sum_trust_scores(poll):
+        EntityPollRating.objects.filter(poll=poll).update(
```
GresilleSiffle, line 82: Do we want to use the Django's method `bulk_update` here to benefit from the `batch_size` attribute?

We had some problems before with SQL queries locking tables for too long while updating all rows with `update`.

GresilleSiffle: More details on the differences between `update` and `bulk_update`:
- https://docs.djangoproject.com/en/4.1/ref/models/querysets/#bulk-update

backend/tournesol/migrations/0053_entitypollrating_sum_trust_scores.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.0.9 on 2023-05-22 20:25
```
GresilleSiffle, line 1: Just in case this file has been generated after our last batch upgrade: let's not forget to run `pip install -r requirements.txt` to sync our local packages.

backend/tournesol/tests/test_api_polls.py
```diff
@@ -116,15 +117,21 @@ def setUp(self):
...
 
         results = response.data["results"]
         self.assertEqual(len(results), 3)
-        self.assertEqual(results[0]["tournesol_score"], 4.4)
-        self.assertEqual(results[1]["tournesol_score"], 3.3)
-        self.assertEqual(results[2]["tournesol_score"], 2.2)
+        # TODO why does this fail?
+        # self.assertEqual(results[0]["tournesol_score"], 4.4)
+        # self.assertEqual(results[1]["tournesol_score"], 3.3)
+        # self.assertEqual(results[2]["tournesol_score"], 2.2)
```
amatissart, line 134: It seems to work now, on my dev-env at least :)

backend/tournesol/migrations/0053_entitypollrating_sum_trust_scores.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('tournesol', '0052_alter_entity_last_metadata_request_at'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='entitypollrating',
+            name='sum_trust_scores',
+            field=models.FloatField(default=0., help_text='Sum of trust scores of the contributors who rated the entity'),
```
amatissart, line 16: All recommendations will be reset as unsafe when this branch is deployed.  

Should we add a custom operation in this migration to compute the values (e.g with `EntityPollRating.bulk_update_sum_trust_scores(poll)`),  without waiting for the next ml run?

amatissart (commented):
The strategy for deploying this change needs to be clarified. As is, all videos will be considered unsafe after applying the migration, until the next ml run. Computing the values for `sum_trust_scores` in the database migration would help to deploy the change without any disruption.

As shown by the failures in the e2e tests, this also affects the dev-env. All user trust scores are currently null in the dev-env. So in addition to computing `sum_trust_scores` in the migration we may want to update the dev-env database beforehand.

backend/tournesol/models/entity_poll_rating.py
lfaucon: I made the two options available and tested

backend/tournesol/migrations/0053_entitypollrating_sum_trust_scores.py
lfaucon: Done

backend/tournesol/tests/factories/entity.py
```diff
@@ -15,6 +16,25 @@ class Meta:
 
     uid = factory.Sequence(lambda n: "uid_%s" % n)
 
+    @classmethod
+    def create(cls, *args, make_safe_for_poll: Poll = True, **kwargs):
```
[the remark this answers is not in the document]
lfaucon: I reduced the scope of this PR. I will make this change in the second half

backend/tournesol/tests/factories/entity.py
```diff
@@ -15,6 +16,25 @@ class Meta:
...
+        entity = super().create(*args, **kwargs)
+        """
+        If `make_safe_for_poll` is True, the entity will be made safe for the default Poll by
+        creating an EntityPollRating with a high `sum_trust_score`. If a poll is provided it will
+        be made safe for the specified Poll. If something else is provided, nothing will be done.
```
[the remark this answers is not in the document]
lfaucon: I reduced the scope of this PR. Good suggestion :)

backend/tournesol/migrations/0053_entitypollrating_sum_trust_scores.py
lfaucon: As we quickly discussed. I propose to cut the PR in two, first adding the sum_trust_scores and their computations, then usingthe sum_trust_scores once they have been computed

lfaucon:
@amatissart @GresilleSiffle This PR is ready to be merged from my point of view if you approve it

backend/tournesol/models/entity_poll_rating.py
GresilleSiffle: Looks good!

I was wondering, do we really want to maintain the two options in our code base? Or are they temporarily here so we can profile the two options?

It could be good to test (if it has not been done already) the the `update` and the `bulk_update` on staging or a local machine with a dump of the production database. If we are confident that we can update thousands and thousands of  `EntityPollRating` in one query, I still think a priori that `bulk_update` is a better alternative. Using batch size of X is like saying "we don't know how our database server will react if we try to update more than X rows at the same time". And currently I don't know how our database will react if we update our 36 000 ratings in a single query.

backend/tournesol/models/entity_poll_rating.py
```diff
@@ -67,4 +75,38 @@ def update_n_ratings(self):
...
+            sum_trust_scores=Coalesce(Subquery(
+                Entity.objects.filter(
+                    id=OuterRef('entity_id'),
+                    contributorvideoratings__poll=OuterRef('poll'),
+                ).annotate(
+                    s_t_s=Sum("contributorvideoratings__user__trust_score")
+                ).values("s_t_s")[:1]
+            ), 0.)
+        )
```
amatissart, line 99: This approach does not seem quite correct: for a given pair (entity, user) a `ContributorRating` may exist in the database even if the user has not evaluated the entity (for example when the user marked their rating as private or public, without submitting the comparison). 

So only ratings where at least one  `ContributorRatingCriteriaScore` exists should be considered (or almost equivalently, where a  `ContributorRatingCriteriaScore` on the main criterion exists);

This implementation should work:
```py
       EntityPollRating.objects.filter(poll=poll).update(
            sum_trust_scores=Coalesce(
                Subquery(
                    ContributorRating.objects.filter(
                        entity_id=OuterRef("entity_id"),
                        poll=OuterRef("poll"),
                    )
                    .filter(
                        Exists(
                            ContributorRatingCriteriaScore.objects.filter(
                                contributor_rating=OuterRef("id")
                            )
                        )
                    )
                    .values("entity")
                    .annotate(s_t_s=Sum("user__trust_score"))
                    .values("s_t_s")[:1]
                ),
                0.0,
            )
        )
 ```

amatissart (changes requested):
The query computing `sum_trust_scores` needs to be adapted, as explained in the comment.

@lfaucon Depending on your availability, would you like me to apply the change and adapt the tests to cover this edge-case?

backend/tournesol/models/entity_poll_rating.py
lfaucon: Addressed in e005d9a

GresilleSiffle (approved):

backend/tournesol/models/entity_poll_rating.py
lfaucon: Thanks for the review and catching this issue @amatissart !
Merging now, and opening the next steps which will change the recommendations.

PR #1571, opened 2023-05-23
[front] fix: remove extra trailing slash in comparison URL

GresilleSiffle:
This should fix the duplicates "comparison" and "comparison/" entries in our web analytics tool.


amatissart (approved):
Thanks :+1: 

The URLs could also be updated in the e2e tests:
https://github.com/tournesol-app/tournesol/blob/ec8a7296d61ec1a19fe7f71b8a17eff3fd1eaacf/tests/cypress/e2e/frontend/comparisonPage.cy.ts#L256
https://github.com/tournesol-app/tournesol/blob/ec8a7296d61ec1a19fe7f71b8a17eff3fd1eaacf/tests/cypress/e2e/frontend/comparisonPage.cy.ts#L273

PR #1572, opened 2023-05-23
[front] fix: remove extra HTTP request made by VideosPollUserSettingsForm

GresilleSiffle:
This pull request fixes the behaviour of the Preferences page.

Before, both the page and its form were fetching the user's settings from the API. Now only the page is fetching the settings and the form use them instead of making a second HTTP call.

The bug was introduced by https://github.com/tournesol-app/tournesol/commit/4222f6226dbb1992f539de02a2e2e2ad2338fc25

The tests of the form have been updated to reflect this behaviour, and a new test file `Preferences.spec.tsx` has been added to test the logic specific to the page.

This PR should also simplify the development of this other PR https://github.com/tournesol-app/tournesol/pull/1598

**to-do**
- [x] update the tests
- [x] create a new test for the Preferences page
  - ensure the settings are retrieved from the API
  - ensure the dispatch function is called one time with the correct parameter
  - ensure an error message appears in case of HTTP error


frontend/src/pages/settings/preferences/Preferences.tsx
```diff
@@ -39,10 +46,13 @@ const PreferencePage = () => {
...
         dispatch(replaceSettings(response));
       }
+
+      setLoading(false);
     }
 
     retrieveProfile();
-  }, [t, dispatch, contactAdministrator]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
GresilleSiffle, line 55: I'm not satisfied with this `eslint-disable` here, but I haven't found an elegant solution to remove it yet.

I added it because when I run the `Preferences.spec.tsx` tests with the complete array of dependencies `[t, dispatch, contactAdministrator]`, this effect is triggered two times instead of one, which  makes  the tests fail. When using the app, the effect is only triggered one time as expected.

I first thought it was caused by the fact that the function `dispatch` is "spied", but even with a incomplete dependency array like `[t]` the effect was still triggered two times.

amatissart (approved):

PR #1575, opened 2023-05-23
[front] feat: saving search filters from recommendations page

ShinseiNova:
image
image


GresilleSiffle:
We closed this PR as we estimated the amount of code required to implement this feature properly was not worth it for the moment:
- we expect the users to not update their preferences every day
- the number of preferences is small enough for them to be updated directly on the Preferences page
- with the proposed solution the UI will invite the users to always update their preferences, even when they are opening a shared link or when they are just browsing the recommendations

We might want to implement it when the number of settings for the recommendations page increase, or when we expect some settings to be updated very often.

PR #1581, opened 2023-05-24
[front] feat: add an info msg explaning the behaviour of the default search filter.

GresilleSiffle:
This PR adds a paragraph in the Preferences page explaining how to use the default search filters.

I think it may not be obvious for all users to:
1. discover the Preferences page
2. realize that some are related to the default search filters of the recommendations page
3. understand when those default filters are used

To address the point 1 (and a bit of the point 2) we added a link to the preferences in the recommendations page :heavy_check_mark: 

To address the point 3 (and one more bit of the point 2) I added a small paragraph inviting the users to customize their preferences, and explain how to access the newly configured recommendations page.

This small addition should make the settings more friendly, and I hope avoid us to receive too much questions about "why the recommendations page doesn't use my preferences".

### preview
  
capture

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -329,7 +329,7 @@
     "comparisonPage": "Comparison page",
     "rateLater": "Rate-later list",
     "recommendationsPage": "Recommendations page",
-    "customizeYourDefaultSearchFilter": "Customize <1>the default search filters</1> according to your own preferences. Those filters are applied <4>only</4> when you access the recommendations from the left <6>side bar</6>.",
+    "customizeYourDefaultSearchFilter": "Customize <1>the default search filters</1> according to your own preferences. Those filters are applied <4>only</4> when you access the recommendations from the <6>main menu</6>.",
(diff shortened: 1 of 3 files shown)


frontend/public/locales/en/translation.json
```diff
@@ -329,6 +329,7 @@
     "comparisonPage": "Comparison page",
     "rateLater": "Rate-later list",
     "recommendationsPage": "Recommendations page",
+    "customizeYourDefaultSearchFilter": "Customize <1>the default search filters</1> according to your own preferences. Those filters are applied <4>only</4> when you access the recommendations from the left <6>side bar</6>.",
```
amatissart, line 332: Maybe "main menu" would be clearer for mobile users? As you prefer.

amatissart (approved):

frontend/public/locales/en/translation.json
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1581/commits/96048af3d463996ada0073864c2d7a5c72cb00a9

PR #1593, opened 2023-05-29
[front] style: add missing content padding in the Rate Later page

lfaucon:
I found that some padding on the rate later page would make it look better on mobile.

Without padding (current state):

Screenshot 2023-05-29 at 21 28 19

With some padding (this PR):

Screenshot 2023-05-29 at 21 29 00

### DIFF
--- a/frontend/src/pages/rateLater/RateLater.tsx
+++ b/frontend/src/pages/rateLater/RateLater.tsx
@@ -193,6 +193,7 @@ const RateLaterPage = () => {
           spacing={2}
           justifyContent="space-between"
           alignItems="stretch"
+          sx={{ p: 1 }}
         >
           <Grid item sm={6} display="flex" width="100%">
             <Paper sx={{ p: 2, display: 'flex', width: '100%' }}>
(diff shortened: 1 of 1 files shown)


GresilleSiffle:
Thanks @lfaucon 

The Rate Later page was indeed lacking padding, and we haven't taken the time to fix it yet.

Instead of defining a custom padding I think we can use the same mechanism that we use on all other pages. Instead of

```typescript
      <ContentBox noMinPaddingX maxWidth="lg">
```

We can use:

```typescript
      <ContentBox maxWidth="lg">
```

See:
https://github.com/tournesol-app/tournesol/blob/63e063531c2bf512715ea9293ce16d5db0fe2e12/frontend/src/pages/rateLater/RateLater.tsx#L186

GresilleSiffle (approved):

lfaucon:
@GresilleSiffle Thanks for taking over and merging the PR

I like that we handle the padding in a uniform manner on the top component, however I also tried to fix that the padding was not aligned between the first components and the list of video cards, which your fix did not address. See screeenshot:

Screenshot 2023-05-30 at 18 19 00

Issue #1519, opened 2023-04-19
[front] the component `<ComparisonSeries>` is unmounted during a state update


GresilleSiffle:
- see on `main` at https://github.com/tournesol-app/tournesol/commit/ba3d8f03cfc141218233cb57926c5d30179bd082
- this doesn't seem to impact the user experience.

---

### How to reproduce

- start the `dev-env`
- open the front end home page
- open the browser JavaScript console
- click on the button Compare the videos in the Contribute section
- see the following message in the console

```
Warning: Can't perform a React state update on an unmounted component.
This is a no-op, but it indicates a memory leak in your application.
To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

**expected**

No error in the console.

### Details

This may be caused by the component `<ComparisonSeries>`, which is instantly unmounted for users having more than 4 comparisons or having skipped the tutorial.

```javascript
      const alts = await getAlts();
      if (alts.length > 0) {
        setAlternatives(alts);
        return alts;
      }
```

abhiram11:
Looking into this

PR #1595, opened 2023-05-30
solves unmounting and memory leak issues of ComparisonSeries

abhiram11:
Solves #1519 by declaring a variable `isMounted` in the useEffect of `<ComparisonSeries>`. This variable checks if the component is mounted whenever we call any setState within the useEffect.

Working behavior can be seen below:

https://github.com/tournesol-app/tournesol/assets/20269286/6531876c-8c0f-4bfb-94d0-dd5202588575
 
I request you to test it as well and let me know if there are any changes to be made.

### DIFF
--- a/frontend/src/features/comparisonSeries/ComparisonSeries.tsx
+++ b/frontend/src/features/comparisonSeries/ComparisonSeries.tsx
@@ -128,12 +128,14 @@ const ComparisonSeries = ({
    * series.
    */
   useEffect(() => {
+    let isMounted = true;
+
     async function getAlternativesAsync(
       getAlts: () => Promise<Array<Entity | Recommendation>>
     ) {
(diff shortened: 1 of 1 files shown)


GresilleSiffle (changes requested):
The proposed fix works.

I think we can simplify even more the code and make it more maintainable. See my comment here: https://github.com/tournesol-app/tournesol/pull/1595#issuecomment-1569797819

abhiram11:
Thanks for the detailed explanation! I also encountered a bit of confusion when working with `skipped` during the weekly comparison count PR. I'll add the inital check in the useEffect and commit the changes. It will surely make the solution cleaner now that I understand the root cause of it!

GresilleSiffle (approved):
It works!

Issue #1412, opened 2023-03-04
[front] Add possibility to modify criteria order and display status in `Preferences`


aidanjungo:
Should be done after #1410 and #1411 

Could use something like [that](https://codesandbox.io/s/material-ui-sortable-list-with-react-smooth-dnd-swrqx) but with the possibility to select or not (check box?) each criteria.

Issue #1411, opened 2023-03-04
[front] modify setting order in comparison page according to the settings


aidanjungo:
Should be done after #1410 

Should respect the order of criteria in the setting list. Nothing in the list correspond to the current implementation.

GresilleSiffle:
I don't understand exactly what is the "default setting selection for the comparison page" here.

Are we talking about the back end setting `comparison__criteria_order`?

If no, what is the difference between this ticket and https://github.com/tournesol-app/tournesol/issues/1412?

PR #1598, opened 2023-05-31
[front] feat: configurable criteria order and default display

ShinseiNova:
**related to #1411 & #1412**

---

This PR allows the users to select the criteria they want to see always displayed on the comparison UI.

**added**
- new field "optional criteria displayed" in the preferences form 

**changed**
- the submit button is now sticky 
- the layout of all settings page has been reworked to give more space to the content 

### Preview

We plan to remove the question mark from each criterion label here to simplify the interface. They are more useful on the comparison page.

#### Layout column

image

#### Layout row

This layout will fallback to the column layout on mobile

image

capture


frontend/src/features/settings/preferences/fields/ComparisonCriteriaOrder.tsx
```diff
@@ -0,0 +1,191 @@
...
+        ))}
+        {checkedCriteria.length === 0 && (
+          <Alert severity="info">
+            <Typography>
+              {t('pollUserSettingsForm.displayedCriteriaEmpty')}
+            </Typography>
+          </Alert>
+        )}
+      </Grid>
+    </Grid>
```
amatissart, line 187: 2 small comments on this component. (I know this is still a draft, so feel free to ignore my comments :wink: ) 

1. When checking criteria, the height of this block may change and make other sections appear at the bottom of the screen. I find that disturbing for the user. Setting a `min-height` would help to prevent the layout from changing, ideally to fit all criteria just in 1 column, no more no less.

2. An additional breakpoint may be required on medium screens. Some space is currently lost on the right of the screen, and force the buttons to wrap on a second line:
image

GresilleSiffle (commented):
The code is good for me : )

This first implementation allows the users to select and order the optional criteria that are always displayed.

As suggested, the two columns containing the criteria have a minimum height to avoid a resize of the whole page each time a criterion is selected. This creates an empty gap on mobile between the two columns, but the absence of screen shake is in my opinion worth the trade.

The global layout of all the settings page has been updated to increase the horizontal space available for the page content.

As the configuration of the drag and drop navigation can represent a challenge on mobile, this first PR only provides the move up and move down buttons. It also reduce the amount of code to review.

**Questions**

**(a)** Do we want to wait for the drag and drop prototype before making this new setting available to all users? 

**Future considerations**

We need to add the related end-to-end tests, and to update the component tests.

As the height of the page is increasing, we might want in the future to add a Floating Action Button to allow the users to save their changes  without having to scroll down. 

We may want to add a direct link from the comparison UI to the related preferences.

aidanjungo:
I tested the behavior of this new setting, is seems really nice to use to me. The only point (not really linked to this PR) is the "Update Preference" button which is a the bottom of the page, it could be missed by the user, maybe we should move it or automatically save the changes.

GresilleSiffle (approved):
I made several adjustments to the style and the wording to make the interface more simple.

From my point of view we are ready to merge :ok_hand:

amatissart (approved):

PR #1610, opened 2023-06-12
Fix: Not needed translations in `fr` language detected

NilsJacobsen:
## Description
I was running `@inlang/cli` on the project. It aims to lint and machine translate translation messages. The lint report detected missing reference messages. I removed the unnecessary messages in the `fr` resource. The messages I removed turned out to be plural keys that where already covered by `_other`. 

No the lint report is fine.

## Further information
I added the cli to your project so everybody can check the status of the messages by executing `npx inlang lint`. Also `machine translate` and `open editor` are available see [docs](https://inlang.com/documentation/apps/inlang-cli)


NilsJacobsen:
I also added a translation status badge:

[translation badge](https://inlang.com/editor/github.com/NilsJacobsen/tournesol?ref=badge)

GresilleSiffle:
Hello @NilsJacobsen 

Thanks for opening a pull request : ) I'm going to divide my comment in different sections so we can talk easily refer to each point independently.

Here are my current thoughts, I haven't run the code read, I just read the changes. 

**(a)**

I saw in your changes that you added some files at the root of the repository, but the newly added tool targets only the front end right? The root of the repository is a collection of different software with different translation mechanisms, I wonder if `inlang` shouldn't be added in the front end folder instead.

**(b)**

Moreover, even if we might not use it correctly, the suffix `_many` is useful in French. See the bullet point "PluralRules category for compact notation numbers in French":
- https://icu.unicode.org/download/68
 
If we delete the keys ending with `_many` in the translations, a run of `yarn i18n:parse` will add them again because they are considered required by the tool and the language.

Change request: don't delete those keys :upside_down_face: 

**(c)**

I saw you removed the code `\u00a0` used to create non-breakable space. I'm gonna see what happens without them, but the goal was probably to keep words such as "score personnel :" always on the same line, preventing the `:` to be displayed alone on the next line.

What was the motivation of these changes? It was warning triggered by `inlang`? If yes, we may want to configure `inlang` to disable those warnings.

NilsJacobsen:
Hey @GresilleSiffle, 

> a)

The problem is that `inlang.config.js` needs to be on the root, so i needed to add the cli on the root too. I see that this is a pain for your project. So I want to push that the cli can access the inlang.config.js from inside a frontend folder. SO we can remove `package.json`, `.gitignore` and `yarn.lock`. -> https://github.com/inlang/inlang/issues/919 

> b)

Thanks for the reply. Ah u use `fr` as the default language right? So I'm gonna change that. Then it makes sense to keep them as a template for other languages.

> c) 

The lint report was showing where the project has missing references. So while having `en` as default it was showing me missing references for the _many keys. That one should be resolved by switching the default to `fr`. The `\u00a0` was cleaned up by the i18next plugin. But I also see that this is important for you. Let me open an issue there at inlang.  -> https://github.com/inlang/inlang/issues/921

I could also this to your lint command in the `package.json` so you never ship a new feature with only the reference language. In that case you could simply open the editor and enter it or use the CLI to machine translate.

I try to push the issues. So you could seamlessly use inlang in your project. Keep u updated.

GresilleSiffle:
Hello @NilsJacobsen 

I read the documentation of Inlang CLI, thank you again for sharing I didn't know this tool.

Unfortunately, adding Inlang CLI to our project seems to add an extra layer that solves problems we current don't have, or appear to be superficial enough to me (thus we might need Inlang in the future). It's always a bit delicate to refuse a PR because you took the time to open it and to work on it, and be sure we are very grateful for this.

Here is my reasoning:

Inlang CLI can lint the translations following some basic rules by default: detecting obsolete translation keys, missing keys, or identifying identical patterns. The configured script `yarn i18n:parse` already automatically remove obsolete keys and add the missing ones. Identical patterns are not checked, but they are usually detected during code reviews. In other word, the tool solves a problem we currently don't have.

Inlang CLI can automate the translation process by calling a translation service. This is a great feature. Currently all the maintainers are more or less bilingual fr/en, and because we add small portion of text feature by feature, we usually don't spend extra time translating. Nevertheless machine translation is an alternative we might consider in the future, as well as online collaborative translation platforms.

So my proposition is the following: don't merge this pull request today, and wait for the day we encounter a use case that would benefit from Inlang (such as automatically translating the platform in several other languages, other something else). It will help the team to adopt the tool, to configure it properly and to use it.

It's just my personal opinion, and I may be missing something so I'm going to ping the others maintainers in case they have different opinions @amatissart @lfaucon

GresilleSiffle:
> Makes it easy to extract strings and shows the live text next to the keys in code. Is the community already using such a tooling?

As fas as I know we haven't used it yet. I'll git it a try : )

> I agree since you only have two languages, it is fine. But jap when you scale globally it probably makes sense in the future.

Yes probably, we will keep Inlang and other alternatives in mind when we will add new languages.

I'm going to close this pull request for now.  Thanks for your work and your time @NilsJacobsen

PR #1623, opened 2023-06-19
[tests] e2e: add test for criteria order setting

ShinseiNova:


GresilleSiffle:
It works :+1: 

I suggest to replace the labels used in `aria-label` by strings that are more human friendly. They are intended to be read by assistive technologies, so better to not used the symbols `_` or `-` between words.

PR #1624, opened 2023-06-19
[front] refactor: remove anchors in faq url for previews

ShinseiNova:


GresilleSiffle:
Good, I'll fully test tomorrow.

We still need to handle one last thing: the presence of both `#` and the parameter `scrollTo`. Even if it should not happen, it's still possible to provide those two inputs in the URL. We can decide how we want the application to behaves when this happens.

GresilleSiffle (approved):

PR #1628, opened 2023-06-20
[front] feat: add previewLang parameter to FAQ url

ShinseiNova:
The goal is to use it in the preview so we can have the translated question on discord for example.


GresilleSiffle:
Currently, when we ask for the preview of a question, it will be returned and translated according to the header accept-language of the request.

If we paste a question on a social network, we do not control this header, and the preview will probably be always in English.

We opened this PR to see how complex it is to to control this behaviour with a query parameter, and to see if it adds hard to maintain code in the front end.

Let's talk about the following questions during a meeting:
- do we want to allow the users to control the language of the preview?
- do we want the update the parameter `previewLang` when the user changes the lang of its UI?

PR #1631, opened 2023-06-22
[dev-env] improvement: define separate volume for node_modules

amatissart:
`node_modules` contains a _lot_ of small files, and this contributes to make the dev-env initialisation really slow on some systems when mounted volumes introduce a significant overhead. This was observed when using Docker Desktop on Windows or MacOS.

Using a separate volume (non shared with the host filesystem) should improve performance.


aidanjungo (approved):
Works fine on macOS and takes less than 5 min to create the dev-env 😃

GresilleSiffle (approved):

Issue #1651, opened 2023-06-27
[back] feat: create the banner API in the backoffice


GresilleSiffle:
to do: copy the spec i wrote on my laptop here

PR #1652, opened 2023-06-28
[back] feat: create the banner API in the backoffice

ShinseiNova:
**related to #1651**


GresilleSiffle (commented):
It works!

We can add one more test to ensure that the banners are retrieved in the expected language. If the target language is not available, we expect the banner to be translated in English. If the English is not available, we expect empty strings.

backend/backoffice/models/banner.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    class Meta:
+        unique_together = ["banner", "language"]
+
+    def __str__(self) -> str:
+        return self.action_label
+
+
+class BannerActionLink(models.Model):
```
amatissart, line 135: Is there a specific to define `title`, `text`, `action_label` etc in separate models?  
I think I would find it slightly easier to maintain to define all localized content in a single model (e.g `BannerLocale`)

GresilleSiffle (approved):

Issue #1655, opened 2023-06-28
[front] feat: allow the users to define some notification preferences in the sign up form


GresilleSiffle:
In the sign up form, below check box "I have read and agree...", add:
- a checkbox allowing the users to opt in the search studies email notifications
- a checkbox allowing the users to opt in the new features email notifications

Those two checkboxes are optional, and it must be clear that they are optional. One option could be to use an horizontal `<Delimiter>` between the Terms of Service consent and the two checkboxes.

Plug those new fields to the API.

PR #1657, opened 2023-06-28
[front] feat: notification settings at sign up

GresilleSiffle:
**related to** https://github.com/tournesol-app/tournesol/issues/1655

**waiting for** #1671

---

This PR allows the users to define their notification preferences at sign up.

The two new fields use a MUI Switch instead of a check box to accentuate their difference with the mandatory field "I'm at least 15 years old...". Let me know if we want only check boxes or switches in the form.

As suggested by @aidanjungo I changed the labels of the notification fields by adding the prefix "I want to" to make them more explicit. 

~~The default value used to initialize the preferred language is the user's `navigator.language` (with a fallback to English if the language is not supported).~~ We don't want to create an impressive form, the `notifications__lang` will be implicitly initialized with the language of the UI, no field will be displayed.

#### full screen

(**edit:** the preferred language field doesn't exist anymore.)

capture

| before | after | alternative |
|---|---|---|
| capture | capture | capture2 |


frontend/src/pages/signup/Signup.tsx
```diff
@@ -1,54 +1,87 @@
...
       const createdUser = await AccountsService.accountsRegisterCreate({
-        requestBody: formObject as RegisterUserRequest,
+        requestBody: {
+          email: (formData.get('email') as string) ?? '',
+          username: (formData.get('username') as string) ?? '',
+          password: (formData.get('password') as string) ?? '',
+          password_confirm: (formData.get('password_confirm') as string) ?? '',
+          settings: {
+            general: {
+              notifications__lang: notififLang,
```
amatissart, line 79: As discussed I would use the current UI language to initialize this settings (i.e use `i18n.resolvedLanguage`).

It would also make unnecessary to add one more language detection mechanism in `isNavigatorLang()`.

amatissart (approved):

PR #1658, opened 2023-06-29
[front] feat: live chip on talk entry for ongoing talks

ShinseiNova:
These are usually the colors used to define a live event.
Yet, does it catch too much the eye ? Or is it ok since it is its purpose ?
image


GresilleSiffle (approved):
Here is the current look:

capture

PR #1665, opened 2023-07-03
[front] feat: connect the videos home page to the /banners/ API

ShinseiNova:
In this first PR, we only display the banner that has the highest priority.
We don't deal with the dates or the security advisories.


GresilleSiffle:
### Preview

capture

GresilleSiffle (approved):

PR #1666, opened 2023-07-04
[front] fix: remove the extra trailing slash in `recommendations`

GresilleSiffle:
This will fix the duplicate entries we have in Plausible.

In the following capture, we can see both `recommendation` and `recommendations/`

capture

### DIFF
--- a/browser-extension/src/manifest.json
+++ b/browser-extension/src/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "Tournesol Extension",
-  "version": "3.1.2",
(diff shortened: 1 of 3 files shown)


amatissart (approved):
:+1: 

A similar change needs to be applied in the browser extension:
https://github.com/tournesol-app/tournesol/blob/e112af6ee58e4ca38f5477b796c94f7ba186bbe6/browser-extension/src/models/tournesolContainer/TournesolContainer.js#L77

PR #1546, opened 2023-05-08
[ext]: Adjust amount of video based on screen size

mbourand:
I did some changes to the tournesol recommendations on the home page of youtube.

Here are the changes that were made:
- The number of videos displayed now varies based on the size of the screen (previously it was always 4).
- The video size and spacing between them now matches that of YouTube.
- The displayed video duration was moved slightly higher and to the left to better align with the YouTube design.
- The method for displaying how much time has passed since a video's release was simplified.

Please let me know if you have any feedback on these changes or if there are any further improvements that you think could be made.

PR #1667, opened 2023-07-05
[ext] style: make videos wrap onto multiple lines

GresilleSiffle:
Replace the PR https://github.com/tournesol-app/tournesol/pull/1546 by @mbourand

---

This PR makes the video cards "wrap" onto multiple lines when the screen size is reduced. Each line displays as many videos as YouTube does, 4 being the maximum for now.

A more advanced refactor should allow the extension to display as many videos as YouTube (not limited by the current maximum of 4). See the spec here: https://github.com/tournesol-app/tournesol/pull/1667#issuecomment-1632437999


GresilleSiffle:
With only CSS we can make the video cards wrap when the sum of their width exceeds the width of the flex container:

cap2
 
It looks good in with all screen sizes, with one exception. When there is space for only three videos, the fourth video is displayed alone on the second line:

capture

This is caused by the variable `videosPerRow` saying that one row always contains 4 videos (even if this conceptual row is displayed on two lines).

We can decide the behaviour we want:
1. Keeping the code as-is, considering this edge-case is acceptable.
2. Or when the number of video that can be displayed on a single row is odd (and more than 1), change the variable `videosPerRow` to this value. This can reduce the total number of recommendations when the Tournesol container is expanded when there is only the space for 3 videos in a visual row (total = 3 videos * 3 visual rows instead of 4 videos * 3 visual rows). This will match the YT visual.
3. Define a specific total number of recommendations for each `videosPerRow` value (it's like point 2, but we decide how many videos will be displayed when the container is expended). This allows us to control the length of the expanded Tournesol container.

Examples with the option 2:

- available space: it's the number of videos that can fit in one row without being wrapped
- total nbr reco.: total number of displayed video when the Tournesol container is expanded
- total nbr line: number of line displayed by the Tournesol container when expanded

| available space | total nbr reco. | total nbr line |
|-----------------|-----------------|----------------|
| 4               | 12 (4 * 3)      | 3              |
| 3               | 9 (3 * 3)       | 3              |
| 2               | 12 (4 * 3)      | 6              |
| 1               | 12 (4 * 3)      | 12             |

Example with the option 3:

If the number of rows displayed when the Tournesol container is expanded is more important than the total number of recommendations (or if we just want to control this number of rows), then the option 3 allows us to fully customize the results.

| available space | total nbr reco. | total nbr line |
|-----------------|-----------------|----------------|
| 4               | 12              | 3              |
| 3               | 9               | 3              |
| 2               | 6               | 3              |
| 1               | 4               | 4              |

GresilleSiffle:
The option 1 requires no JavaScript, it works out of the box.

The option 2 should be slightly faster to implement than the option 3, but looks a bit like hack (if visual space = 3 then videosPerRow = 3).

The option 3 makes the behaviour nearly fully customizable, and requires to rework how the `videosPerRow` and `rowsWhenExpanded` are initialized. It should be relatively easy to implement.

New option 4: we could merge option 1 and create an issue to implement option 2 or 3.

What do we prefer @aidanjungo @amatissart @lfaucon ?

lfaucon:
It's a bit complicated, I am not sure I fully understand all the options you suggest.

I propose the below format to discuss each option, and I would propose the below target behaviour.


| Youtube's video per row | Number of reco (expanded) | Number of reco (default) | Number of rows (expanded) | Number of rows (default) |
|-----------------|-----------------|----------------|-----------------|----------------|
| 4 | 12 | 4 | 3 | 1 |
| 3 | 12 | 3 | 4 | 1 |
| 2 | 12 | 4 | 6 | 2 |
| 1 | 12 | 2 | 12 | 2 |


I think this can be implemented by putting all 12 videos in the same wrap block with the right width, but such that videos that overflow are not visible. Then adapting only the height of the recommendation box.

browser-extension/src/addTournesolRecommendations.css
```diff
@@ -301,16 +306,18 @@ html[darker-dark-theme][dark] .tournesol_simple_button:disabled img {
 
 .time_span {
   position: absolute;
-  bottom: 0;
-  right: 0;
-  color: white;
-  background-color: var(--yt-spec-static-overlay-background-heavy);
+  bottom: 6px;
+  right: 6px;
```
amatissart, line 310: (very minor) 3px seems closer to the Youtube design to my eyes
```suggestion
  bottom: 3px;
  right: 3px;
```

amatissart (approved):

PR #1688, opened 2023-07-20
feat: Add `unsafe` status in API responses, adapt the entity card implementation

### LINKED ISSUE #1686: [back] refactor: Unsafe should be handled in the backend
### To-do
- [ ] The backend should have a unique centralized mechanism to determine if an entity is safe or unsafe
- [ ] The backend serializer for entities should include in API response metadata concerning the safe status of entities (see proposition below). 
- [ ] The entity card in the frontend should display the grey sunflowers and explaination tooltips based on the backend responses.

### Proposition
The frontend and backend should maintain a list of keywords/reasons that make entity unsafe.

The serializer output could roughly look like:
```
{
    "uid": "yt:ePU32cKdtPs",
    "tournesol_score": 42,
    "unsafe": {
        status: bool
        reasons: List[Enum]
    }
}
```

amatissart:
Closes #1686

This required a bit more refactoring than expected, but also gave a good opportunity to delete some obsolete functions and components in the frontend.


frontend/public/locales/fr/translation.json
```diff
@@ -19,8 +19,8 @@
   "video": {
     "notAvailableAnymore": "Cette vidéo n'est plus disponible sur YouTube.",
     "labelShowSettings": "Voir les paramètres de cette video",
+    "insufficientScore": "Le score de cette vidéo est inférieur au seuil de recommendabilité défini par Tournesol.",
```
amatissart, line 22: @GresilleSiffle  I opted for "seuil de recommendabilité" instead of "seuil de fiabilité", as I feared it could be confused with the criterion "Fiable et non-trompeur".  Suggestions are welcome.

GresilleSiffle (approved):
It looks very good to me. Thank you for cleaning obsolete functions in the front end.

I saw there was a little inconsistency in the preview API, that we can fix. The score of an entity is displayed with a grey color when it is below the hardcoded value 0, instead of `RECOMMENDATIONS_MIN_TOURNESOL_SCORE`.

If I'm not mistaken the line:

https://github.com/tournesol-app/tournesol/blob/2ca180cb5a7e38371c006ab93fd8b1ea4655e93f/backend/tournesol/views/previews/default.py#L258

... should be replaced by `if poll_rating.is_recommendation_unsafe`, right ?

Issue #1491, opened 2023-04-04
Upgrade the algorithm for individual scores


lenhoanglnh:
Currently, and perhaps surprisingly, the algorithm that transform a contributor's comparisons into its scores fails to guarantee a desirable monotonicity property. Namely if a contributor further pushes his comparison between A and B towards A, then the score of A should increase and that of B should decrease.
https://arxiv.org/pdf/2211.01179.pdf

Essentially, the reason why this property fails is because, when the contributor says A is vastly better than B, the corresponding spring has a very small stiffness, which can make the comparison nearly meaningless for the algorithm. 

The reason why this annoying property holds is because the algorithm demands a quadratic loss, so that it can be minimized by a straightforward matrix inversion. Thereby, we avoid doing optimization, with related issues such as nondeterminism and lack of guarantee of small distance to the minimum (in the parameter space).

I have previously proposed an alternative, based on "Binomial Bradley Terry", which I think satisfies monotonicity (to be verified). Moving towards this would provide better properties, but would require greater computation care to perform the optimization. Additionally, doing optimization will make it significantly easier to include other kinds of data, e.g. likes/dislikes or direct scoring.

Next step is for Lê to clarify the new algorithms.

amatissart:
A draft for the implementation of individual scores based on BBT is in #1531 

Remaining tasks
- [ ] Review (and potentially add more) unit tests in https://github.com/tournesol-app/tournesol/blob/indiv-scores-bbt/backend/ml/tests/test_mehestan_individual.py
- [ ] Any other check before we can opt to replace the current approach with BBT?
- [ ] Validate the value of alpha
- [ ] Observe the impact on scores distribution, and adapt the value of `W`

Issue #1490, opened 2023-04-04
Shift the zero


lenhoanglnh:
Right now, a non-scored video is set to be a median video (as its score is 0, and 0 is roughly the median score).
A safer approach would be to consider that a non-scored video is, say, in the bottom 5%.
This would correspond to saying that the score 0 corresponds to the score of the 5 percentile.

It seems that this can be approximately fixed by a few lines, when computing the global scores:
https://github.com/tournesol-app/tournesol/blob/main/backend/ml/mehestan/global_scores.py#L311

Namely, we first estimate the bottom 5-percentile of scaled individual scores. Denote ```tau_hat``` this  estimate.
Then before aggregation by QrMed of scaled individual scores, we subtract ```tau_hat``` to all scaled individual scores.

I think that this does not conflict (too much) with the rest of the algorithm (but please test it!).

This corresponds to fixing Equation (14), page 14: https://arxiv.org/pdf/2211.01179.pdf

amatissart:
I run a few experiments on the branch [`shift-zero-experiments`](https://github.com/tournesol-app/tournesol/tree/shift_zero_experiment), with variations on 4 parameters:

![runs](https://github.com/tournesol-app/tournesol/assets/4726554/469f1cf3-f680-4ff9-8d10-d684bcd6e2db)

amatissart:
For reference, with the current algorithm as existing on `master`:

* individual scores:
![indiv_scores_](https://github.com/tournesol-app/tournesol/assets/4726554/7d9b2284-6786-4e23-8952-789048806199)

* global scores:
![global_scores](https://github.com/tournesol-app/tournesol/assets/4726554/7c0d6318-1294-4ef1-8d9f-98cd7a0ecac3)

PR #1689, opened 2023-07-20
New implementation for Tournesol scores: CBT + shifting zero

amatissart:
Depends on #1688

* Use CBT to compute individual scores (see #1491)

* Shift individual scores before aggregation (see #1490)
This PR supersedes #1629


backend/tournesol/models/poll.py
```diff
@@ -87,10 +86,8 @@ def entity_cls(self):
 
     @property
     def scale_function(self):
-        if self.algorithm == ALGORITHM_MEHESTAN and self.sigmoid_scale is not None:
-            return lambda x: (4 * MEHESTAN_MAX_SCALED_SCORE / TAU) * \
-                             np.arctan(self.sigmoid_scale * x)
-        return lambda x: x
+        # TODO: deprecate sigmoid scale stuff
```
lfaucon, line 89: Is there more to do than the current change? If not, let's not leave the comment here. If yes let's do it or make the comment more helpful

backend/twitterbot/settings.py
```diff
@@ -1,8 +1,8 @@
 # Quality criteria
 MIN_NB_CONTRIBUTORS = 5
 MIN_NB_RATINGS = 15
-MIN_RELIABILITY_SCORE = 10
-MIN_TOURNESOL_SCORE = 10
+MIN_RELIABILITY_SCORE = 25
```
lfaucon, line 4: Makes me think the 95% / 5% prior on "should be largely recommended" could be adapted for other criteria. 50/50 would be more appropriate for "entertaining". 

(not a serious suggestion)

solidago/pyproject.toml
```diff
@@ -18,9 +18,9 @@ classifiers = [
 ]
 keywords = ["tournesol", "collaborative recommendations", "judgement aggregation", "comparison based", "mehestan"]
 dependencies = [
-    "pandas>=1.5.3,<2.0",
+    "pandas>=1.5.3,<3.0",
```
lfaucon, line 21: Not upgrading to pandas 2 like Tournesol?

lfaucon (approved):

solidago/pyproject.toml
amatissart: Pandas has been upgraded in Tournesol to take advantage of performance improvements. I don’t want to block the usage of Solidago in external projects that would use Pandas v1.

backend/tournesol/models/poll.py
amatissart: Not much except that the field `sigmoid_scale` should be removed from the model `Poll` and from the database. And I preferred not to include db migrations that would conflict with other ongoing work.

Issue #1696, opened 2023-07-26
[front] feat: each tab of the video selector should display a short description


GresilleSiffle:
### Context

During our last brainstorm about the video selector, we decided to display a short description below each tab.

**rate-later**

This list is probably empty for new users. The message should invite the users to add videos to their rate-later list thanks to the browser extension and thanks to the + button displayed next to each video.

**compared**

The message could be "Compared elements will appear here".

**Top of the month**

...

**To connect**

Those descriptions should not be displayed in the tabs of the video analysis page. We could display them only when a props `displayDescriptions` is set to `true`. 

### Open questions

Do we want to allow the users to display / hide those description on demand? The `Alert` component of Material UI can display a close icon in the top right corner of its container, we could use it to close the description. An information icon (i) in the top right corner of the tabs' content could display the description.

PR #1697, opened 2023-07-26
[front] feat: display a description of the video selector tabs

ShinseiNova:
**related to #1696**

---

We have multiple ways of handling the display of the description.

Different approaches :
- (1) Always displayed by default, if the user close it, it's stored locally
- (2) Never displayed, there is an icon if we want to show it manually
- (3) Displayed and cant close if NbComp < 2*TutoLength, then it has (2) behavior

(1) Everyone will see it at least once but we need to implement storage
(2) Users might not pay attention and never read the descriptions
(3) New users will see it since they are the ones who should need it the most but we don't give them the choice

image
image

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -112,6 +112,10 @@
   },
   "tabsBox": {
     "errorOnLoading": "An error occurred while loading the results.",
+    "rateLater": "Your rate-later videos will appear here. You can add some to your list by clicking on the '+' sign on the video cards. You can also add them directly from the extension.",
+    "compared": "Compared videos will appear here.",
+    "recommendations": "Recommended videos will appear here.",
+    "toConnect": "Videos isolated from your other comparisons appear here. Compare them to improve the recommendations",
     "emptyList": "This list is empty."
   },
   "stackedCandidatesPaper": {
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -115,7 +115,7 @@
     "rateLater": "Your rate-later videos will appear here. You can add some to your list by clicking on the '+' sign on the video cards. You can also add them directly from the extension.",
     "compared": "Compared videos will appear here.",
     "recommendations": "Recommended videos will appear here.",
-    "toConnect": "Videos isolated from your other comparisons appear here. Compare them to improve the recommendations",
+    "toConnect": "Videos isolated from your other comparisons appear here. Compare them to improve the recommendations.",
     "emptyList": "This list is empty."
   },
   "stackedCandidatesPaper": {
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -112,9 +112,9 @@
   },
   "tabsBox": {
     "errorOnLoading": "An error occurred while loading the results.",
-    "rateLater": "Your rate-later videos will appear here. You can add some to your list by clicking on the '+' sign on the video cards. You can also add them directly from the extension.",
-    "compared": "Compared videos will appear here.",
-    "recommendations": "Recommended videos will appear here.",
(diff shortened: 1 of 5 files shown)


frontend/public/locales/en/translation.json
```diff
@@ -112,6 +112,10 @@
   },
   "tabsBox": {
     "errorOnLoading": "An error occurred while loading the results.",
+    "rateLater": "Your rate-later videos will appear here. You can add some to your list by clicking on the '+' sign on the video cards. You can also add them directly from the extension.",
+    "compared": "Compared videos will appear here.",
+    "recommendations": "Recommended videos will appear here.",
```
GresilleSiffle, line 117: The "recommended videos" might be understood as the videos Tournesol recommend you to compare, instead of the videos the community is recommending.

As we may create a suggestions tab, which will contain videos Tournesol recommend you to compare, we can add few words to make the description more explicit.

```suggestion
    "recommendations": "Videos recommended by the community appear here.",
```

Moreover, I think we can use the indicative present. Instead of "will appear here", we can write simply "appear here".

frontend/public/locales/en/translation.json
```diff
@@ -112,6 +112,10 @@
   },
   "tabsBox": {
     "errorOnLoading": "An error occurred while loading the results.",
+    "rateLater": "Your rate-later videos will appear here. You can add some to your list by clicking on the '+' sign on the video cards. You can also add them directly from the extension.",
```
GresilleSiffle, line 115: Instead of displaying a button "Install the extension", we can make the description container shorter by using a simple HTML link directly in the paragraph. This will be less intrusive with users having already installed the extension.

Example of a HTML link used inside a translation: 

https://github.com/tournesol-app/tournesol/blob/e925beb7408eafb684717c2d6fb335f6c49695d6/frontend/src/pages/about/TermsOfService/TermsOfService.tsx#L47

GresilleSiffle (commented):
Hello !

I didn't review the code but I made few comments about the translations.

frontend/public/locales/fr/translation.json
```diff
@@ -116,6 +116,10 @@
   },
   "tabsBox": {
     "errorOnLoading": "Une erreur est survenue lors du chargement des résultats.",
+    "rateLater": "Vos vidéos à comparer plus tard apparaissent ici. Vous pouvez en ajouter sur le site web via le bouton '+'. Vous pouvez aussi les ajouter directement depuis YouTube grâce à <2>l'extension</2>.",
+    "compared": "Les vidéos comparées apparaissent ici.",
+    "recommendations": "Les vidéos recommandées apparaissent ici.",
```
amatissart, line 121: Similarly to the english version, these messages could be made more explicit:

```suggestion
    "compared": "Les vidéos que vous avez comparées apparaissent ici.",
    "recommendations": "Les vidéos recommandées par la communauté apparaissent ici.",
```

frontend/src/features/entity_selector/EntityTabsBox.tsx
```diff
@@ -43,11 +108,21 @@ const EntityTabsBox = ({
...
   entityTextInput,
+  displayDescription = false,
 }: Props) => {
   const { t } = useTranslation();
+
   const [tabValue, setTabValue] = useState(tabs[0]?.name);
   const [status, setStatus] = useState<TabStatus>(TabStatus.Ok);
   const [options, setOptions] = useState<RelatedEntityObject[]>([]);
+  const [toggleDescription, setToggleDescription] =
+    useState(displayDescription);
```
amatissart, line 119: (minor)
Using "toggle" to name the state sounds confusing to me. What about something like `isDescriptionVisible` ?

```suggestion
  const [isDescriptionVisible, setIsDescriptionVisible] =
    useState(displayDescription);
```

amatissart (approved):

PR #1700, opened 2023-07-27
[front] feat: add the fill entity selector setting to the preferences

ShinseiNova:
**related to #1699**

Default value is set to true for now so the behavior is the same as the current one until we change it.
Should we set the default to false ?

We might not want to merge it for now since the use of the setting isn't implemented so that wouldn't make much sense.


GresilleSiffle:
> Should we set the default to false ?

`true` seems good here to keep the default behaviour aligned with the current behaviour.

> We might not want to merge it for now since the use of the setting isn't implemented so that wouldn't make much sense

Yes. The strategy we have followed for this kind of feature was:
1. merge the back end first
2. implement the behaviour in the front end
3. add a new field in the user preferences

It allows us to merge one, two or three pull requests depending on the complexity. As the point (1) and (3) are already done, I'll implement the point (2) in this PR.

GresilleSiffle:
**to-do**

*back end*
- [ ] add the new setting to the tests

*front end*
- [ ] add missing e2e tests:
  - test the default behaviour
  - test the behaviour of the comparison UI when the setting is true

GresilleSiffle (changes requested):
Let's check how the tutorial is affected before merging.

GresilleSiffle (dismissed):

PR #1701, opened 2023-07-31
[front] feat: implement the fill entity selector feature

ShinseiNova:
**following #1699 and #1700**

---
The buttons have the same behavior as the current ones.
No video (entity) is loaded if the setting is set to false, default value is true.
image

### DIFF
--- a/frontend/public/locales/en/translation.json
+++ b/frontend/public/locales/en/translation.json
@@ -100,12 +100,14 @@
   "entitySelector": {
     "newVideo": "Select a new video automatically",
     "autoEntityButton": "Auto",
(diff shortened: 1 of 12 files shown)


aidanjungo:
The preference option and the entity selector works well, two comments that could improve it a bit.

1) Better placement of the video selector when the big button is clicked

image

It is rather small and a bit weird to have it above or bellow the button.


2) When the "auto" button is clicked there is nothing to indicate the next video is loading.

GresilleSiffle:
**to-do**
- [ ] fix the tests
- [ ] restore the loading effect after a click on Auto

PR #1703, opened 2023-08-01
feat: Add filtering option for already compared videos

lfaucon:
Changes:
* chore: update api yaml schema
* Add filtering option for already compared videos available for logged-in users
* Change frontend URLs to use a parameter `advanced` that support the safe option and the exclude option instead of `unsafe`
* Adapt the preview link redirect to support the `advanced=unsafe` option

### DIFF
--- a/backend/tournesol/views/previews/recommendations.py
+++ b/backend/tournesol/views/previews/recommendations.py
@@ -79,6 +79,8 @@ def get_preview_recommendations_redirect_params(request):
                 now = timezone.now()
                 date_gte = now - datetime.timedelta(seconds=seconds_max_age)
                 query["date_gte"] = date_gte.isoformat(timespec="seconds")
+        elif key == "advanced":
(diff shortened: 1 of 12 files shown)


amatissart (approved):

aidanjungo:
Two comment which could improve this PR (or be in following one):

- Because of the cache, when we compare a "not compare before" video and go back to the recommendation page, the video did not disappear as it should
- As a new filter option is should be added in the user preferences as the other filters

lfaucon:
Thanks for your suggestions @aidanjungo 

> * Because of the cache, when we compare a "not compare before" video and go back to the recommendation page, the video did not disappear as it should

=> Let's discuss this in the next team meeting

> * As a new filter option is should be added in the user preferences as the other filters

=> I create the issue https://github.com/tournesol-app/tournesol/issues/1712 for this

PR #1721, opened 2023-08-21
[back] chore: upgrade several dependencies

GresilleSiffle:
This PR upgrades several Python packages used by the back end.

In the near future we may want upgrade `django-oauth-toolkit` to the latest major release 2.x to benefit from the security improvements. This might involve updating the set-up of the dev-env.

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -30,11 +30,11 @@ drf-spectacular==0.26.4
 # Pillow is used to generate the Previews for links shared on social media
 Pillow==9.5.0
 # Needed for postgres database
-psycopg2-binary==2.9.5
+psycopg2-binary==2.9.7
 # PyYAML is used for reading the settings 
 PyYAML==6.0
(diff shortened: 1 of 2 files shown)


backend/requirements.txt
```diff
@@ -18,23 +18,23 @@ django-oauth-toolkit==1.7.1
...
 # Pillow is used to generate the Previews for links shared on social media
-Pillow==9.4.0
+Pillow==9.5.0
 # Needed for postgres database
-psycopg2-binary==2.9.5
+psycopg2-binary==2.9.7
 # PyYAML is used for reading the settings 
 PyYAML==6.0
 # tqdm is used to display a progressbar in some slow ML algorithms
-tqdm==4.65.0
```
amatissart, line 37: Note that this dependency is used only in `Entity.recompute_quantiles()` which is obsolete, and could probably be removed completely.

amatissart (approved):

Issue #1727, opened 2023-08-24
refactor: reduce duplicated code by creating a generic component for the settings displayed as switch field


GresilleSiffle:
**better implementing it once** https://github.com/tournesol-app/tournesol/pull/1730 **is merged**

---

The following components share the same code:
- `AutoSelectoEntities`
- `RecommendationsDefaultUnsafe`
- `RecommendationsDefaultExcludeCompared`
- etc

Issue #1725, opened 2023-08-22
[back] refactor: add the Search on/off setting of the extension to the API (and to the extension storage)


GresilleSiffle:
- [x] add a new setting in the API, `extension__recommendations_search` or similar
  - add it to the API tests  
- [x] add a new setting in the front end's Preferences page, in the section Extension
- [x] make the extension use this new setting

PR #1730, opened 2023-08-24
[ext][front] refactor: move the action "search on/off" to the preferences page

GresilleSiffle:
**related to** #1725 

---

This PR moves the extension setting Search on/off to the extension local preferences page, and to the website preferences page. 

It introduces a new API `getSingleSetting` to query any setting from the `background.js` script. This function queries the Tournesol API and is able to fallback to the extension storage in case of need (it also supports the "legacy" settings). 

**to-do**

*back end*
- [x] add the new setting
- [x] update the API tests

*front end*
- [x] display the new setting in the Preferences page

*browser extension*
- [x] add the setting in the local preferences
- [x] use this setting instead of the legacy local storage key

### Next steps

- it's the good moment to refactor the duplicated code in the front end (see https://github.com/tournesol-app/tournesol/issues/1727)

### Preview

capture


backend/core/serializers/user_settings.py
```diff
@@ -86,6 +86,8 @@ class VideosPollUserSettingsSerializer(GenericPollUserSettingsSerializer):
         ("ALL_TIME", "all_time"),
     ]
 
+    extension__search_reco = serializers.BooleanField(required=False)
```
amatissart, line 89: Maybe we could add more context about this setting in a help text:

```suggestion
    extension__search_reco = serializers.BooleanField(
        required=False,
        help_text=(
            "Whether Tournesol recommendations should be integrated in Youtube.com search results."
        ),
    )
```

amatissart (approved):

Issue #1439, opened 2023-03-14
Improve HTTPS configuration


amatissart:
- [x] Disable deprecated TLS ciphers (CBC based)
Probably in this nginx configuration:
https://github.com/tournesol-app/tournesol/blob/845634e76e176448b81793cd956286b27e649ec5/infra/ansible/roles/nginx/templates/nginx.conf.j2#L35-L35

- [x] Higher HSTS max-age: 180 days instead of `max-age=300`

GresilleSiffle:
**to-do**
- [ ] apply the HSTS max-age on Grafana and Plausible

PR #1741, opened 2023-08-30
sec(infra) set HSTS max-age to 180 days (instead of 300 seconds)

GresilleSiffle:
**related to** #1439

---

(a)

I read that Mozilla recommend to set max-age to [one year](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#examples), and by doing so our website become eligible for [HTST preload](https://hstspreload.org/?domain=tournesol.app).  Curiously even with our current value of 300 seconds, it looks like we are already considered preloaded by hstspreload, but to confirm this we need to be sure we are part of the Chrome preload list (for instance, we are not part of the [Firefox preload list](https://hg.mozilla.org/mozilla-central/raw-file/tip/security/manager/ssl/nsSTSPreloadList.inc)).

There [several warnings](https://hstspreload.org/?domain=tournesol.app#opt-in) about being preloaded, so it may be a good move to first set max-age to 180 days or lower before going further.

(b)

I didn't update grafana and plausible max-age, because I'd like to see if they are automatically affected by the `includeSubDomains` options. 

See:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
- https://hstspreload.org/?domain=tournesol.app

### DIFF
--- a/infra/ansible/roles/django/templates/tournesol.j2
+++ b/infra/ansible/roles/django/templates/tournesol.j2
@@ -57,7 +57,15 @@ server {
     add_header X-Frame-Options "SAMEORIGIN";
     add_header X-XSS-Protection "1; mode=block";
     add_header X-Content-Type-Options nosniff;
-    add_header Strict-Transport-Security "max-age=300; includeSubDomains; preload; always;";
+
+    # HSTS set to 180 days (15552000 seconds)
+    # For information, according to Mozilla 1 year is acceptable, and 2 year
+    # is recommended.
+    #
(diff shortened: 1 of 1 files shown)


amatissart (approved):
A similar change could be applied in
`infra/ansible/roles/monitoring/templates/grafana.j2`
and
`infra/ansible/roles/plausible_analytics/templates/website_analytics.j2`

GresilleSiffle:
> A similar change could be applied in
infra/ansible/roles/monitoring/templates/grafana.j2
and
infra/ansible/roles/plausible_analytics/templates/website_analytics.j2

Yes, I would like to see the behaviour of the `includeSubDomains` option before.

Issue #1746, opened 2023-09-04
back: creating two contributor rating


GresilleSiffle:
### Context

In the logs we found a HTTP 500 error returned by the following route: 

```
POST /users/me/contributor_ratings/videos/

django.db.utils.IntegrityError: duplicate key value violates unique constraint "tournesol_contributorrat_user_id_entity_id_poll_i_749f0366_uniq
```

### Spec

The code should not fail, and the route should not an error 500. Instead, like the rate-later API, code should gracefully fail within a try except IntegrityError and the route return a HTTP 409 Conflict error.

GresilleSiffle:
After some tests on `main` before and after the contributor ratings API refactor (897f9006bb4a8e25eb5c6ae0683741eb411f08c7 and f578c0f264a31dec4b64dd4dec4ae47a32539fb2), I confirm that the API currently returns HTTP 400 Bad Request (not as explicit as 409 Conflict but ok still).

The errors 500 we can see in the logs (2023-05-09) don't seem to be related to the regular behaviour of the route, but may happen in specific cases.

GresilleSiffle:
At the time of writing, the `ContributorRatingCreateSerializer.validate` method returns an exception if the rating already exists.

https://github.com/tournesol-app/tournesol/blob/897f9006bb4a8e25eb5c6ae0683741eb411f08c7/backend/tournesol/serializers/rating.py#L76

It's possible that two concurrent requests successfully pass the `validate` without raising any validation error, making the method `ContributorRatingList.perform_create` able to fail.

https://github.com/tournesol-app/tournesol/blob/897f9006bb4a8e25eb5c6ae0683741eb411f08c7/backend/tournesol/views/ratings.py#L187

GresilleSiffle:
According to Grafana, during the day 2023-09-05, the errors 500 caused by a POST to `/users/me/contributor_ratings/videos/` always follow two 201 Created from the same remote ip address.

I haven't checked the other days yet, but it may be a pattern followed by the front end. Maybe when the auto suggestion is enabled.

PR #1748, opened 2023-09-06
[back] fix: HTTP 500 is now 400 when unique contrainst of rating is violated

GresilleSiffle:
**related issues** #1746 

---

### Description

Usually the API `POST /users/me/contributor_ratings/{poll_name}/` correctly returns HTTP 400 when an already existing rating is validated.

There is an exception to this. According to the logs, we received concurrent requests (possibly bearing the same JSON payload), passing the validation, and causing the API to fail with an error 500. This PR makes the back end fail gracefully by returning a 400 Bad Request, instead of a 500 Interval Server Error.

Nevertheless, there might still be an issue in the front end causing two requests to be sent instead of one.

---

Here is the query I used in Grafana to investigate the problem, with different time ranges:

- time range from 2023-09-05 17:12:04 to 2023-09-05 17:12:05
- time range from 2023-09-05 10:25:23 to 2023-09-05 10:25:26

Between each time range there is an ip address sending several requests very close to each other. I suppose that at least two of those requests have the same JSON payload.

```
{filename="/var/log/nginx/json_access.log"} | json | request_uri = "/users/me/contributor_ratings/videos/"
```

### Checklist

### DIFF
--- a/backend/tournesol/views/ratings.py
+++ b/backend/tournesol/views/ratings.py
@@ -1,6 +1,7 @@
 """
 API endpoint to interact with the contributor's ratings.
 """
+from django.db import IntegrityError
 from django.db.models import Prefetch, prefetch_related_objects
 from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiParameter, extend_schema, extend_schema_view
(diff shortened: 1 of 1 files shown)


backend/tournesol/views/ratings.py
```diff
@@ -184,7 +185,20 @@ def get_queryset(self):
         return ratings
 
     def perform_create(self, serializer):
-        contributor_rating = serializer.save()
+        # In some cases, several concurrent requests might pass the
+        # serializer's validation and cause the view to fail here.
+        try:
+            contributor_rating = serializer.save()
+        except IntegrityError as err:
+            raise ValidationError(
```
amatissart, line 193: Maybe we could also add a warning here, just in case this `except` would catch other kinds of `IntegrityError` we don't expect.

Something like:
```python
        try:
            contributor_rating = serializer.save()
        except IntegrityError as err:
            logging.warning("Got IntegrityError when creating ContributorRating", exc_info=True)
            raise ValidationError(
            ...
```

amatissart (approved):

Issue #1749, opened 2023-09-06
[back] feat: add the subsamples API


GresilleSiffle:
Define the specification of the /subsample/ API.

### Proposition

(a) This new API provides a single route:
- `GET /users/me/subsamples/{poll_name}/`
- `GET /users/me/subsamples/videos/`

(b) It divides all user's `ContributorRatingCriteriaScore` in 20 buckets. 

- they are filtered on the poll's main criterion (`largely_recommended` for the poll videos) 
- public and private ratings are included, it's a personal route
-both safe and unsafe entities are included

(c) If a user has less than 20 `ContributorRatingCriteriaScore` on the criterion `largely_recommended`, the number of buckets should be equal to this number of `ContributorRatingCriteriaScore`.

(c) It returns a list of "sub-sample" JSON object.

```json
{
  "entity": {
    "uid": "",
    "type": "video",
    "metadata": {}
  },
  "collective_rating": {
    "n_comparisons": 14,
    "n_contributors": 6,
    "tournesol_score": 22.516694294829456,
    "unsafe": {
      "status": false,
      "reasons": []
    }
  },
  "individual_rating": {
    "is_public": true,
    "n_comparisons": 1
  },
  "subsample_metadata": {
    "bucket": "1"
  }
}

```

### More

The following SQL query divided all `ContributorRatingCriteriaScore` of the user with `id` 1 into 20 buckets. The next step is to select only one entity per bucket. 

```sql
SELECT
  tcrcs.id,
  tcrcs.score,
  NTILE(20) OVER(
    ORDER BY tcrcs.score DESC
  ),
  tournesol_entity.metadata->'name'

FROM tournesol_contributorratingcriteriascore AS tcrcs

JOIN tournesol_contributorrating
  ON tournesol_contributorrating.id = tcrcs.contributor_rating_id

JOIN tournesol_entity
  ON tournesol_entity.id = tournesol_contributorrating.entity_id

WHERE criteria = 'largely_recommended'
  AND tournesol_contributorrating.user_id = 1;
```

PR #1765, opened 2023-09-13
[back] feat: add the sub-samples API

GresilleSiffle:
**related issues** #1749 

---

### Description

This PR adds the `/users/me/subsamples/` API. (Should it be renamed subsample singular instead?)

This new endpoint is private, and return a sub-sample of the user's rated entities as described by #1749  

To create a sub-sample, the view uses a combination of SQL query and Python filtering. The performance of the Python code seems good, ~ 0.009 second to filter a list of ~2000 compared entities on my development environment with Docker. 

The SQL query is more greedy. In production it takes ~ 0.1 second (with the `EXPLAIN ANALYZE` overhead) to process ~3000 compared entities with all ORM annotations except `.annotate_n_comparisons()`, which looks very good to me. Nevertheless with `annotate_n_comparisons()` it takes ~ 1 second. If we plan to display the sub-samples in a compact entity card, we may remove this annotation for now.

**to-do**
- [x] return the entity without making extra SQL requests
- [x] dynamically determine the main criterion of the poll
- [x] add the `collective_rating` key to the serializer
- [x] add the `individual_rating` key to the serializer
- [x] add the `subsample_metadata` to the serializer
- [x] document the API
- [x] add the API tests
- [x] test the performance of the view with a user having a lot compared videos

### Checklist


GresilleSiffle:
I just realized that the OpenAPI schema is not well made.

backend/tournesol/views/subsamples.py
```diff
@@ -0,0 +1,85 @@
...
+                order_by="-criteria_scores__score"
+            ))
+        )
+
+        sub_sample = []
+        buckets = len(qst)
+
+        for i in range(min(buckets, sub_sample_size)):
+            sub_sample.append(
+                random.choice([rating for rating in qst if rating.bucket == i + 1])  # nosec
```
lfaucon, line 54: All ratings are fetched with all annotations from prefetched global and individual scores. That's suboptimal, but maybe it works perfectly fine and does not need to be optimized?

Proposal 1:
Don't do the bucketing on the database side, fetch everything from a sorted qs, use np.array_split for the bucketing. I expect it would be faster than the for loop you are using that then goes through the whole list to group each bucket together.
```
sorted_ratings = list(qst)  # assuming the qst has order_by
sub_sample_size = min(sub_sample_size, len(sorted_ratings)
sub_sample = [random.choice(bucket) for bucket in np.array_split(sorted_ratings, sub_sample_size)]
```

Proposal 2:
Make a first query for only the `uid`, fetch everything, sample from buckets, then do a second query with the selected list of uids that includes all the metadatas.

backend/tournesol/serializers/subsample.py
```diff
@@ -0,0 +1,22 @@
+from rest_framework import serializers
+from rest_framework.fields import IntegerField
+
+from tournesol.serializers.entity import RelatedEntitySerializer
+from tournesol.serializers.poll import CollectiveRatingSerializer
+from tournesol.serializers.rating import ExtendedInvididualRatingSerializer
+
+
+class SubSampleMetadataSerializer(serializers.Serializer):
+    bucket = IntegerField(read_only=True)
```
lfaucon, line 10: This is not very important, would it just be in order the numbers 0, 1, 2, 3, 4, etc.. ?

If yes, we don't need to include this in the result

backend/tournesol/tests/test_api_subsamples.py
```diff
@@ -0,0 +1,184 @@
...
+        self.client.force_authenticate(self.user2)
+        response = self.client.get(self.base_subsamples_url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        results = response.data["results"]
+        self.assertEqual(len(results), 4)
+        self.assertEqual(response.data["count"], 4)
+
+        # The results:
+        #  - are ordered by score descending
```
lfaucon, line 112: I don't think you assert this, but it would be good to check it.

backend/tournesol/serializers/subsample.py
GresilleSiffle: In our case, when the `offset` parameter is provided, the first bucket returned will not necessarily be 1.

backend/tournesol/views/subsamples.py
```diff
@@ -0,0 +1,85 @@
...
+from drf_spectacular.utils import OpenApiParameter, extend_schema
+from rest_framework import generics
+from rest_framework.permissions import IsAuthenticated
+
+from tournesol.models import ContributorRating
+from tournesol.serializers.subsample import SubSampleSerializer
+from tournesol.views.ratings import ContributorRatingQuerysetMixin
+
+# The number of ranked group used to divide the rated entities.
+DEFAULT_BUCKET_SIZE = 20
```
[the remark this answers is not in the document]
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1765/commits/ced3295e13acd3f7dc71ea1c27ce0bebba094da2

backend/tournesol/views/subsamples.py
```diff
@@ -0,0 +1,85 @@
...
+                criteria_scores__criteria=poll.main_criteria
+            )
+            .annotate(bucket=Window(
+                expression=Ntile(sub_sample_size),
+                order_by="-criteria_scores__score"
+            ))
+        )
+
+        sub_sample = []
+        buckets = len(qst)
```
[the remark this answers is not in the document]
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1765/commits/ced3295e13acd3f7dc71ea1c27ce0bebba094da2

backend/tournesol/views/subsamples.py
GresilleSiffle: Good idea!

The for loop seems to be the most efficient part of the code. It's the SQL query, especially the annotation `annotate_n_comparisons`, that severely hits the performances. The SQL query can take up to 1 second for a user having compared ~3000 entities.

I'll try the proposal 2, to see how it improves the performance. And maybe the proposal 1, even if I think it will have a positive impact only if the SQL window function is not optimized.

GresilleSiffle: I implemented the proposal 2. It greatly improves the performance (see: https://github.com/tournesol-app/tournesol/pull/1765/commits/cdee9f41450c6c6ee8045c0feb85efd64425bfd0).

Currently the buckets are made by the database and the ratings are selected with Python. It currently takes ~ 5 milliseconds to perform the SQL query with the window function, and according to the Python `time` module ~ 5 more milliseconds to iterate over 2300 ratings.

If we need to improve this we have two strategies:
- create the buckets and select the ratings in one SQL query / no Python
- select all ratings by SQL and create the buckets and select the entities with Python only

I haven't tried any of those strategies yet as 10 ms seems to be good enough. Let me know what do you think. We can still try by curiosity : )

backend/tournesol/tests/test_api_subsamples.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1765/commits/9d23dcb343083fe663062326d8b801ec61e98659

backend/tournesol/views/subsamples.py
```diff
@@ -0,0 +1,94 @@
...
+            .filter(
+                poll=poll,
+                user=self.request.user,
+                entity_id__in=[rating[0] for rating in selected],
+                criteria_scores__criteria=poll.main_criteria
+            ).order_by("-criteria_scores__score")
+        )
+
+        for idx, item in enumerate(sub_sample):
+            item.bucket = idx + 1
```
amatissart, line 64: Any specific reason to start indexing by 1 and not 0 ?

backend/tournesol/views/subsamples.py
```diff
@@ -0,0 +1,94 @@
...
+        sub_sample_size = min(sub_sample_size, len(all_ratings))
+        selected = [random.choice(bucket)  # nosec
+                    for bucket in np.array_split(all_ratings, sub_sample_size)]
+
+        # Only the previously selected ratings are retrieved and annotated.
+        sub_sample = (
+            ContributorRating.objects.annotate_n_comparisons()
+            .annotate_collective_score()
+            .annotate_individual_score(poll=poll)
+            .prefetch_related(self.get_prefetch_entity_config())
```
amatissart, line 54: The annotations can be removed here. They are typically useful to order the queryset by their values, but that's not relevant for the subsample API.

On the other hand, you want to prefetch the `criteria_scores`  to collect all of them at once in a single query when serializing individual criteria scores.

```suggestion
            ContributorRating.objects
            .prefetch_related(self.get_prefetch_entity_config())
            .prefetch_related("criteria_scores")
```

backend/tournesol/views/subsamples.py
```diff
@@ -0,0 +1,94 @@
...
+        # not to affect performance.
+        all_ratings = (
+            ContributorRating.objects
+            .filter(
+                poll=poll,
+                user=self.request.user,
+                criteria_scores__criteria=poll.main_criteria
+            )
+            .order_by("-criteria_scores__score")
+            .values_list("entity_id")
```
amatissart, line 42: (very minor)
When selecting a single field, [`values_list()` ](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#values-list) accepts an argument `flat=True` to get a list of values, instead of a list of tuples. This would avoid using `rating[0]` on line 58, and may improve readability a bit.

amatissart (commented):
Fetching all `entity_id` before building the buckets looks elegant indeed :+1:

backend/tournesol/views/subsamples.py
GresilleSiffle: Not really. At the beginning I thought we would display the bucket number in the front end, and 1 appeared more user friendly than 0 to identify the first bucket. Now I think using 0 makes everything easier on the back end side.

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1765/commits/310e9fadff85473790ed5ba5f84496e1d655be96

backend/tournesol/views/subsamples.py
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1765/commits/385baec531bb09f9416fe398f00f5c4f13cf9f13 and https://github.com/tournesol-app/tournesol/pull/1765/commits/1e689640a8f398234b99428f53dbb50a62daab5f

amatissart (approved):

lfaucon (approved):

PR #1774, opened 2023-09-21
[back] feat: add `entity` and `collective_rating` fields in recommendations API

amatissart:
Related to https://github.com/tournesol-app/tournesol/issues/606

---

#### New fields 
* Entity metadata are moved to `entity`
* Collective criteria scores, number of comparisons, unsafe status etc are moved to `collective_rating`
* 'total_score" computed based on the current criteria weights is moved to `recommendation_metadata`

#### Legacy fields
All legacy fields are preserved for backwards compatibility with the browser extension that relies on the API `/polls/<name>/recommendations ` 

### Checklist

### DIFF
--- a/backend/.pylintrc
+++ b/backend/.pylintrc
@@ -165,7 +165,8 @@ disable=raw-checker-failed,
         too-many-ancestors,
         too-few-public-methods,
         use-maxsplit-arg,
-        cyclic-import
+        cyclic-import,
+        duplicate-code
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/backend/tournesol/models/entity.py
+++ b/backend/tournesol/models/entity.py
@@ -52,16 +52,21 @@ class EntityQueryset(models.QuerySet):
             )
         )
 
-    def with_prefetched_contributor_ratings(self, poll, user):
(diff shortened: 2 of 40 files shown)


GresilleSiffle:
Thank you for the refactor! As I read the code, I'll take notes here.

(a) Let's not forget to create an issue to refactor the extension.
- [ ] issue created for the extension 
- [ ] issue created to remove the legacy fields from back end

> All legacy fields are preserved for backwards compatibility with the browser extension that relies on the API `/polls/<name>/recommendations`

(b) I saw that the model serializer `RecommendationSerializer` (and other serializers based on it) now requires the annotation `@extend_schema_serializer` to prevent the display of all fields from the `Entity` model.

It should be possible, either in this PR or in a second one, to make `RecommendationSerializer` inherit from `Serializer` instead of `ModelSerializer`, to avoid having to declare and exclude all entity fields. If we don't need the `.create` and `.update` methods from `ModelSerializer` maybe it's better to use a simple `Serializer`. We may have less code to write, and it should be easier to maintain.

What do you think?

(An example in the sub-sample PR: https://github.com/tournesol-app/tournesol/pull/1765/files#diff-72f6fb3c530aa32cb9000635d41ceaced58f1e924485f6b2907b1f1d52f7c25a )

GresilleSiffle (approved):

Issue #1779, opened 2023-09-23
[ext] bug: Unnecessary alert and iframe stuck on Home Page


lfaucon:
Regularly (when auth is expired?) an alert to log in again appears on Youtube to request loging in again. Then it opens an iframe that is stuck on the Home page of Tournesol.

GresilleSiffle:
There are several issues with the browser extension:

(a) There is an alert displayed when the YouTube page is loading. This alert is not required. Instead, we can try to refresh the user's access token in the background or display a login iframe when needed.

(b) Some extension requests to the API can fail when the YT page starts loading. This is because the script responsible of updating the extension token is run after the YT page has finished loading, causing all previous requests to fail if the token of the extension is expired or null.

Nevertheless this shouldn't display any login iframe and all following page reload should work.

PR #1784, opened 2023-09-26
[ext] fix: remove unecessary browser alerts

GresilleSiffle:
**related issues** #1779

---

### Description

This PR aims to improve the UX of the browser extension, and addresses the issue **(a)** described in this comment https://github.com/tournesol-app/tournesol/issues/1779#issuecomment-1742648000. The other points require more investigation.

The browser alert telling the users to refresh their connection to Tournesol has been deleted. We consider that an outdated access token should be refreshed automatically in the background, without noticing the users. Some cases still trigger the display of the log in iframe: when the refresh token present in the front end is "expired" (automatically deleted in the API database by a scheduled task when its life time exceed 1 week), the users have to log in again.

### Checklist

### DIFF
--- a/browser-extension/src/utils.js
+++ b/browser-extension/src/utils.js
@@ -36,7 +36,6 @@ export const fetchTournesolApi = async (url, method, data) => {
     body['body'] = JSON.stringify(data);
   }
   return fetch(`https://api.tournesol.app/${url}`, body)
-    .then((r) => r)
     .catch(console.error);
 };
 
--- a/browser-extension/src/utils.js
+++ b/browser-extension/src/utils.js
@@ -16,13 +16,6 @@ export const alertUseOnLinkToYoutube = () => {
   alertOnCurrentTab('This must be used on a link to a youtube video');
 };
 
-export const alertInvalidAccessToken = () => {
-  alertOnCurrentTab(
-    'Your connection to Tournesol needs to be refreshed.\\n\\n' +
-      'Please log in using the form below.'
-  );
-};
-
 export const fetchTournesolApi = async (url, method, data) => {
   const headers = {
     Accept: 'application/json',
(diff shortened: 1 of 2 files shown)


GresilleSiffle:
At the moment, I have not been able to reproduce the bug concerning the blocked iframe on the home page, as described in the issue. I can reproduce it by following an unnatural workflow and modifying the database manually, but I don't understand yet how a user can end up in this situation.

Here is a workflow that leads to this soft lock:
- connect to the front end, and receive fresh access token + refresh token
- delete the access token + refresh token from the API (this is what happens when a refresh token is expired)
- go to YouTube and try to add a video to the rate-later list
- the iframe appears and the Tournesol home page is displayed

This situation should not happen. It shouldn't be possible for a non-expired token to be deleted by the API. When this happens, the front end will not try to refresh the access token (because it is still valid according to its expire date), and will try to call the API with an access token that doesn't exist anymore. The user is then forced to log out and log in again.

There are possibly other workflows that could lead to a soft lock that I need to investigate:
- the expire date stored in the front end's local storage is incorrect, or doesn't have the correct time zone
- or anything similar that could make the front end believe the user is correctly logged (being redirected to the home page when the extension opens the login page means the front end considers the user as logged)
- the user is correctly logged, but the iframe doesn't close automatically
- something else?

browser-extension/src/utils.js
```diff
@@ -43,17 +36,7 @@ export const fetchTournesolApi = async (url, method, data) => {
...
-        // 401 Unauthorized with an access token means either
-        // - the token has expired
-        // - the token has been crafted
-        if (r.status === 401 && access_token) {
-          alertInvalidAccessToken();
-        }
-      }
-      return r;
-    })
+    .then((r) => r)
```
amatissart, line 39: Then `then()` clause is no longer necessary:
```suggestion
```

amatissart (approved):

PR #1817, opened 2023-10-27
Fix dev-env api setup

sigmike:
The installation of the python packages in the dev-env api container failed with this error:

```
ERROR: File "setup.py" or "setup.cfg" not found. Directory cannot be installed in editable mode: /solidago
(A "pyproject.toml" file was found, but editable mode currently requires a setuptools-based build.)
WARNING: You are using pip version 21.2.4; however, version 23.3.1 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
```

It looks like using `pyproject.toml` requires a newer version of `setuptools`. Upgrading to a newer Debian worked.

I also made the container stop immediately when the setup fails.

### DIFF
--- a/backend/dev-env/Dockerfile
+++ b/backend/dev-env/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1
-FROM python:3.9-slim-bullseye
+FROM python:3.9-slim-bookworm
 ENV PYTHONUNBUFFERED=1
 WORKDIR /backend
 
--- a/dev-env/docker-compose.yml
+++ b/dev-env/docker-compose.yml
@@ -31,6 +31,7 @@ services:
     entrypoint: ["/bin/bash", "-c"]
     command:
       - |
+        set -eux
         pip install -r tests/requirements.txt -r requirements.txt
         python manage.py migrate
         python manage.py createcachetable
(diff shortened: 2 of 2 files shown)


backend/dev-env/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1
-FROM python:3.9-slim-bullseye
+FROM python:3.9-slim-bookworm
```
amatissart, line 2: I would prefer to avoid that change, to stay consistent with the version used on production. There must be something else going on to explain why it works in the CI on `main` and not on your system :thinking: 

Is it possible that your system use an older version of the image `python:3.9-slim-bullseye` for some reason? Have you tried `docker pull python:3.9-slim-bullseye` ?

sigmike: > I would prefer to avoid that change, to stay consistent with the version used on production.

That's a good reason indeed.

> Have you tried `docker pull python:3.9-slim-bullseye` ?

It indeed fixed the problem.

amatissart: Ok nice. It's not clear why this is needed though. On my system it seems that `docker compose up --build` will try to update the base image. I suppose it may depend on whether the specific tag was explicitly pulled on the system before, or on the `docker compose` version itself.

Issue #1822, opened 2023-11-01
[infra] feat: mention @core-team in Discord msgs when URLs monitoring fails


GresilleSiffle:
The role id should be securely copied to the server secrets, and made available to the script `/usr/local/bin/external-urls-monitoring.sh`

PR #1824, opened 2023-11-01
[infra] feat: notify Discord role when URLs checks fail

GresilleSiffle:
**related issues** #1822

---

### Description

This PR makes the external URLs monitoring script able to highlight a specific Discord role when it fails.

The goal is to explicitly and quickly notify the core team when the staging and production server are unreachable.

### Checklist


infra/ansible/scripts/get-vm-secrets.sh
```diff
@@ -88,6 +88,9 @@ export TWBOT_ACCESS_TOKEN_EN
 TWBOT_ACCESS_TOKEN_SECRET_EN=$(get_settings_value .TWITTERBOT_CREDENTIALS.\"@TournesolBot\".ACCESS_TOKEN_SECRET)
 export TWBOT_ACCESS_TOKEN_SECRET_EN
 
+DISCORD_ROLE_ID_URLS_CHECKS="$(ssh "$VM_USER@$VM_ADDR" -- sudo cat /root/discord_role_id_urls_checks)"
```
GresilleSiffle, line 91: In the current state, if `/root/discord_role_id_urls_checks` doesn't exist the command will fail and will stop the deployment.

If we want to keep this feature optional, we could use:

```bash
# won't fail, the error will be displayed in the stdout
sudo cat /root/discord_role_id_urls_checks || true

# won't fail, the error will be silent
sudo cat /root/discord_role_id_urls_checks 2> /dev/null || true
```

In theory I think that `DISCORD_ROLE_ID_URLS_CHECKS` should be optional, to let people deploy a Tournesol instance without being forced to use Discord alerts. Thus using `2> /dev/null` could be appropriate if we don't want to display an error when an optional setting is missing

amatissart: `2> /dev/null` is fine.  

Alternatively, I think it would be acceptable to define the variable `discord_role_id_urls_checks` publicly in "inventory.yml" in the `tournesol-prod` section.

GresilleSiffle: Sure, addressed in:
- https://github.com/tournesol-app/tournesol/pull/1824/commits/ed9bac3065607e5f57b518797ee5f22c54b9584c
- https://github.com/tournesol-app/tournesol/pull/1824/commits/55cfb179a5631fe0e5c555ddcdb8321c2b44cb47

amatissart (approved):

Issue #1829, opened 2023-11-07
[back] perf: the /stats/ API sometimes answers in more than 500 ms 


GresilleSiffle:
### Context

The `/stats/` API has recently triggered several monitoring alerts because of it its response time.

Here are some ideas that could improve the route:

(a) Make it poll specific. Instead of computing the number of comparisons and entities available in each poll, we could compute only the stats of the given poll. This should save up to 4 SQL queries by poll that is discarded this way.

- we could add the endpoint `/stats/{poll}/` in addition to `/stats/`
- we could add a query parameter filter the results `/stats/?poll=videos`

(b) Optimize some SQL queries if relevant. We may be able to speed up the queries computing the stats of the last 30 days by adding an index.

(c) If the performance are still bad after (a) and (b), investigate the cause and find solutions : )

PR #1830, opened 2023-11-07
[back][front] perf: stats API can be filtered by poll

GresilleSiffle:
**related issues** #1829

---

### Description

This PR adds a query parameter to the `/stats/` API that allows to retrieve the stats of a single poll, instead of all at once.

### Checklist


frontend/src/features/statistics/StatsContext.tsx
```diff
@@ -52,23 +52,26 @@ export const StatsLazyProvider = ({
...
-    if (
-      stats.polls.length === 0 ||
-      currentTime - lastRefreshAt.current >= EXPIRATION_TIME
-    ) {
-      loading.current = true;
-      refreshStats();
-    }
+      if (
+        stats.polls.length === 0 ||
+        currentTime - lastRefreshAt.current >= EXPIRATION_TIME
```
amatissart, line 65: add a condition to force refresh when current `poll` is not present in `stats.polls` ?

GresilleSiffle: Addressed by https://github.com/tournesol-app/tournesol/pull/1830/commits/9df3994a527b3c68cfcd4fc738d79b3f5108ed54

I finally opted for a ref, instead of checking if the poll was present in `stats.polls`. This allows to handle the specific case where `getStats` is called two times in less than 4 seconds, a first time with a poll, a second time with no poll.

frontend/src/features/statistics/StatsContext.tsx
```diff
@@ -52,23 +53,28 @@ export const StatsLazyProvider = ({
...
+  const getStats = useCallback(
+    (poll: string | undefined) => {
+      const currentTime = Date.now();
 
-    if (loading.current) {
-      return stats;
-    }
+      if (loading.current) {
+        return stats;
+      }
```
amatissart, line 62: In some cases, this means `getStats()` could return stats about the wrong poll, because a concurrent request is running. This is probably not a problem for now, as there is a single one active poll, but it would be worth adding a FIXME.

```suggestion
      // FIXME: this function may return stats about the wrong poll here.
      // Maybe `useStats()` should always depend on `useCurrentPoll()` to simplify this logic.
      if (loading.current) {
        return stats;
      }
```

amatissart (approved):

frontend/src/features/statistics/StatsContext.tsx
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1830/commits/094f0306861a316dcc50357ac463bc810dca94af

amatissart (approved):

Issue #1820, opened 2023-10-30
[front] refactor: connect the Auto button to the new API


GresilleSiffle:
This issue is the front end part of #1785

PR #1835, opened 2023-11-10
[front] refactor: connect Auto btn to suggestions/tocompare API

GresilleSiffle:
**related issues** #1820

---

The Auto button is now connected to the `suggestions/tocompare` API.

The first click sends a request to the API. The results will be stored locally.  Subsequent clicks will yield UIDs from this local list without triggering further requests until all local results have been returned. When the list is empty, a new request is made to the API.

### Checklist


frontend/src/features/rateLater/autoSuggestions.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const isAutoSuggestionsEmpty = (poll: string) =>
+  AUTO_SUGGESTIONS[poll] == null || AUTO_SUGGESTIONS[poll].length === 0;
+
+export const fillAutoSuggestions = (
+  poll: string,
+  uids: string[],
+  exclude?: string[]
+) => {
+  AUTO_SUGGESTIONS[poll] = uids.filter((uid) => !exclude?.includes(uid));
```
lfaucon, line 15: This operation and a few others in these methods are unnecessarily of quadratic complexity.
Casting exclude to a `Set` and calling `excludeSet.has` should be faster. Especially if we crank it up in the future to more than 100 uids.

(see https://www.tech-hour.com/javascript-performance-and-optimization)

frontend/src/utils/video.ts
```diff
@@ -60,140 +64,40 @@ export function idFromUid(uid: string): string {
...
+          (item: { entity: RelatedEntity }) => item.entity.uid
+        ),
+        exclude
+      );
+    }
   }
 
-  newUid = await retryRandomPick(5, uid, uidOther, newSuggestions);
+  if (isAutoSuggestionsEmpty(poll)) {
+    return null;
```
lfaucon, line 97: Will this case ever happen? What would/could be the cause?
How is `null` handled by the video comparison card component?

lfaucon (approved):
Nice to see a lot of code deleted & simplified

frontend/src/features/rateLater/autoSuggestions.ts
lfaucon: I see after this comment that it is only used with the exclude list of length 2, so no impact for the current usage. But I'd still recommend the very easy optimization to protect us from future dev using these arguments differently

frontend/src/utils/video.ts
GresilleSiffle: It's very unlikely but it can happen after a new installation without any data, like the first deployment ever of a new Tournesol instance, or a local set up without the `dev-env`.

I think I'm gonna remove this line because the following call  to `autoSuggestionsRandom` already checks if the suggestion array is null.

> How is `null` handled by the video comparison card component?

When the value null is returned, like before, a click on the Auto button does nothing visually. There is no feedback and it's a bit surprising, users may think the Auto button is not working.

Even if this is very situational, it still can happen if the API is down. A task could be to add a visual hint, like a `notistack` message "Sorry, no element found.".

frontend/src/features/rateLater/autoSuggestions.ts
GresilleSiffle: Good, fixed by https://github.com/tournesol-app/tournesol/pull/1835/commits/2e5afe38c70a901d4e45712a20be2402148d4f6e

Issue #1268, opened 2022-12-29
feat: support pasting URL and uid in Search bar


glerzing:
Modify the frontend so that, if we enter a uid in the search bar (or something that looks like a uid), if the uid corresponds to an existing video, it directly redirects to the video page (e.g. tournesol.app/entities/yt:dGJzpQwA090), even if the video doesn't have a positive score.

(the problem is that currently, if the video doesn't have a positive score, it will not be displayed unless you check "include all videos" in the advanced filter options)

Feel free to reject this issue if it's not relevant.

lfaucon:
I like the idea! We have wip for something similar: Showing the search in the video selector https://github.com/tournesol-app/tournesol/pull/1706

yuyujijin:
Hi, just pushed to the linked branch (commit https://github.com/tournesol-app/tournesol/commit/70b644c861dd3f5faa23b43c81049cac3f1385b0) :)) Hope it work as intended !

PR #1841, opened 2023-11-15
[front] feat: support pasting URL and UID in search bar

yuyujijin:
**related issues** #1268

---

### Description

Added support for search by directly typing video URL or video UID. 
Added test to verify it works as intended.

### Checklist

- [X] I added the related issue(s) id in the related issues section (if any)
- [X] I read the development guidelines of the [CONTRIBUTING.md][development-guidelines]
- [X] The tests pass and have been updated if relevant
- [X] The code quality check pass  (there's actually none for front-end)


frontend/src/features/frame/components/topbar/Search.tsx
```diff
@@ -55,7 +55,13 @@ const Search = () => {
     searchParams.delete('search');
     searchParams.append('search', search);
     searchParams.delete('offset');
-    history.push('/recommendations?' + searchParams.toString());
+
+    // Parse the searchParams to see if it matches a youtube video
+    // And if so stripes it UID and automatically push to the specific Tournesol page
+    const videoId = extractVideoId(search);
```
amatissart, line 61: One issue when using `extractVideoId()` on user input is that any 11-letters word will be considered as a YouTube video id. For example searching for "algorithmes" will redirect to `/entities/yt:algorithms` which is unexpected :worried:.

Maybe we should add an option to `extractVideoId()` to require the `yt:` prefix in this case. As video ids (without prefix `yt:`) are already present in the search index, the video would still be present in the search results, so I think that would be an acceptable trade-off. 

Note that the implementation of `extractVideoId()` has been slightly modified in #1847 and should be easier to maintain. Make sure to rebase on the latest "main" if you want to modify it. And rebasing would also fix the flaky test that caused the CI pipeline to fail (fixed in #1840)

tests/cypress/e2e/frontend/search.cy.ts
```diff
@@ -0,0 +1,39 @@
...
+            cy.intercept('GET', `/polls/videos/entities/yt:${videoUID}/**`).as('videopoll')
+
+            // Visit main page (should work on every page since search bar is the same)
+            cy.visit('/');
+
+            // Input the video URL
+            cy.get('input[id="searchInput"]').click().type(`${videoURL}{enter}`);
+
+            // Check if AJAX returns 200
+            cy.wait(`@videopoll`).its('response.statusCode').should('eq', 200)
```
amatissart, line 17: that's very nice to check the correct request has been triggered, but I would also expect to check that the user has been redirected to the expected page. That's more directly related to the feature we want to test: 

```suggestion
            // Input the video URL
            cy.get('input[id="searchInput"]').click().type(`${videoURL}{enter}`);
            
            // Redirection to video page
            cy.url().should('include', `/entities/yt:${videoUID}`)

            // Check if AJAX returns 200
            cy.wait(`@videopoll`).its('response.statusCode').should('eq', 200)
```

amatissart (commented):
Thanks for this PR!  
The implementation behaves as expected, except in one edge case. See the comments.

yuyujijin:
Hi @amatissart,
the idea of adding an option to require the `yt:` prefix seems to be a good compromise !

If I understood correctly, does the option should be **a function args** (that could be false by default) that forces the user to put `yt:<id>` when searching for videos by id in order to fix the 11 character problem ?

```
export function extractVideoId(idOrUrl: string, needYtPrefix = false)
```

amatissart:
> If I understood correctly, the option should be **a function args** (that could be false by default) that forces the user to put `yt:<id>` when searching for videos by id in order to fix the 11 character problem ?

Yes that sounds good to me :+1:

yuyujijin:
> > If I understood correctly, the option should be **a function args** (that could be false by default) that forces the user to put `yt:<id>` when searching for videos by id in order to fix the 11 character problem ?
> 
> Yes that sounds good to me 👍

So just to be sure : you have to put `yt:` when searching by UID or when wanting to be redirected (the feature I'm implementing) ? I think I need to be sure what is the desired behaviour in order not to break anything haha

yuyujijin:
> Here is what I would expect, if that's looks good to you.
> 
> Query type	Example	Expected behavior of the search bar
> Video URL	`https://www.youtube.com/watch?v=xbBsSEProGY`	Redirected to entity page on Tournesol.
> Tournesol UID	`yt:xbBsSEProGY`	Redirected to entity page on Tournesol.
> Youtube video ID	`xbBsSEProGY`	Unchanged.
> Redirected to search results which contain the video.
> 11-letter word	`algorithmes`	Unchanged.
> Redirected to search results which contain videos related to the query.

perfectly clear ! working on it :))

GresilleSiffle:
Thank you @yuyujijin for the last comments.

We will probably review and merge your work tomorrow :+1: 

If there is no objection from you, we will probably remove the comments that describe self explanatory code.

PR #1846, opened 2023-11-16
 [front] attempt to reduce memory usage during build 

amatissart:
### Description

Node `max_old_space_size` (also called "heap size limit") is set at 2GB by default, and may be too high when deploying to the staging environment.

### Checklist


amatissart:
It does not seem to make a significant difference. I will try to find another solution, maybe by adding a swap file to handle the peak during deploy.

Issue #1852, opened 2023-11-28
[front] feat: add `EntityContext` support


GresilleSiffle:
A notice about the presence of an unsafe entity context should be displayed in the following page:
- [x] entity analysis
  - https://github.com/tournesol-app/tournesol/pull/1844
- [x] recommendations
  - https://github.com/tournesol-app/tournesol/pull/1853
- [x] comparison
  - https://github.com/tournesol-app/tournesol/pull/1858 
- [x] comparisons
  - https://github.com/tournesol-app/tournesol/pull/1865
- [x] ratings
  - https://github.com/tournesol-app/tournesol/pull/1855 
- [x] rate-later
  - https://github.com/tournesol-app/tournesol/pull/1853 
- [x] entity selector?
  - https://github.com/tournesol-app/tournesol/pull/1870
  - [x] update the subsamples API
    - https://github.com/tournesol-app/tournesol/pull/1872
  - [x] update the unconnected_entities API
    - https://github.com/tournesol-app/tournesol/pull/1873

The back end serializers may need to be updated to provide the entity contexts.

PR #1865, opened 2023-12-13
[back][front] feat: display context alerts in /comparisons/

GresilleSiffle:
**related issues** #1852

---

### Description

This PR makes the comparisons API able to return entity contexts, and updates the front end to display them.

Contrary to other views, like rate-later and recommendations, the comparisons API doesn't prefetch the contexts in the `get_queryset` method. This allows to not unnecessarily traverse the relationships `entity_1.single_poll_ratings.poll.all_entity_contexts.texts`  (and `entity_2...`) to retrieve the localized contexts, and saves several SQL queries. 

In our case, the `poll` instance is already fetched at the beginning of each HTTP request, making the contexts directly available with `poll_from_url.all_entity_contexts`.

**to do**

*back end - /comparisons/ API*
- [x] review the SQL query
- [x] review the prefetch configuration
- [x] review all endpoints
- [x] update the tests

*front end*
- [x] display the notice in the My comparisons page
- [x] review the CSS

### Preview

capture

I was not very satisfied by the default look of the context alert when two entity cards have different height. I set the `align-content` CSS property of the `EntityCard` to `space-between` to improve the style. Here are the side effects. 

| before | after | before | after |
|---|---|---|---|
| capture | capture | capture2 |  capture |

### Checklist


frontend/src/components/entity/style.ts
```diff
@@ -7,7 +7,7 @@ export const entityCardMainSx: SxProps = {
   border: '1px solid #DCD8CB',
   boxShadow: '0px 0px 8px rgba(0, 0, 0, 0.02), 0px 2px 4px rgba(0, 0, 0, 0.05)',
   borderRadius: '4px',
-  alignContent: 'flex-start',
+  alignContent: 'space-between',
```
amatissart, line 10: I am not really convinced by this change, but for now I can't find a better one :smile:

GresilleSiffle: Me too :upside_down_face: 

Do you know what you don't like precisely? On my side I was scared of unexpected visual side effects in the entity lists.

There is another way, maybe more "natural", to display the entity context alert at the bottom of entity cards: do not stretch the entity cards height.

 If we replace this layout...

capture

...by this layout...

capture2

...the entity context alert will be automatically displayed at the bottom. If I remember correctly we stretched the entity card on purpose to make them all visually equal. Reverting this could be seen as a "step backward".

GresilleSiffle: Another option could be to use `space-between` only when there is a context. I initially discarded this option because I didn't want to introduce more complexity in the way we define the CSS of entity cards. Nevertheless this can be a good "in between" solution.

amatissart: I submitted another approach in https://github.com/tournesol-app/tournesol/pull/1871

amatissart (approved):

Issue #1850, opened 2023-11-28
[ext] Add the ability to connect to the development environment


sigmike:
It would be useful to make the extension connect to our local development environment instead of the production. It makes testing much easier for features that require specific responses from the backend (e.g. the banners).

Here are the static changes I had to make to make the extension connect to the dev-env:

<details>
  <summary>
    Diff
  </summary>

```diff
diff --git a/browser-extension/src/addModal.js b/browser-extension/src/addModal.js
index 0d85bd7d..4527be9a 100644
--- a/browser-extension/src/addModal.js
+++ b/browser-extension/src/addModal.js
@@ -13,11 +13,11 @@ const EXT_MODAL_VISIBLE_STATE = 'flex';
 const EXT_MODAL_INVISIBLE_STATE = 'none';

 // unique HTML id of the Tournesol iframe
 const IFRAME_TOURNESOL_ID = 'x-tournesol-iframe';
 // URL of the Tournesol login page
-const IFRAME_TOURNESOL_LOGIN_URL = 'https://tournesol.app/login?embed=1&dnt=1';
+const IFRAME_TOURNESOL_LOGIN_URL = 'http://localhost:3000/login?embed=1&dnt=1';

 /**
  * YouTube doesnt completely load a page, so content script doesn't
  * launch correctly without these events.
  *
diff --git a/browser-extension/src/addRateButtons.js b/browser-extension/src/addRateButtons.js
index c45ed998..ac0f27d6 100644
--- a/browser-extension/src/addRateButtons.js
+++ b/browser-extension/src/addRateButtons.js
@@ -112,11 +112,11 @@ function addRateButtons() {
       iconSrc: chrome.runtime.getURL('images/compare.svg'),
       onClick: () => {
         chrome.runtime.sendMessage({
           message: 'displayModal',
           modalOptions: {
-            src: `https://tournesol.app/comparison?embed=1&utm_source=extension&utm_medium=frame&uidA=yt%3A${videoId}`,
+            src: `http://localhost:3000/comparison?embed=1&utm_source=extension&utm_medium=frame&uidA=yt%3A${videoId}`,
             height: '90vh',
           },
         });
       },
     });
diff --git a/browser-extension/src/addVideoStatistics.js b/browser-extension/src/addVideoStatistics.js
index ffe0e115..388b7dee 100644
--- a/browser-extension/src/addVideoStatistics.js
+++ b/browser-extension/src/addVideoStatistics.js
[...]

PR #1868, opened 2023-12-19
[ext] Allow the extension to connect to dev-env

sigmike:
**related issues** #1850 

---

### Description

These changes add a new script `prepareExtension.js` that generates the `manifest.json` and a new `config.js`. This allows us to generate an extension that connects to the dev-env (and other environments can be added).

I ended up making a script instead of using a vite plugin because the plugins actually implied some unnecessary changes to the extension code, we don't need all the features the plugin provides, and I didn't feel comfortable adding so many new dependencies. The only dependency added here is node. But if we ever want to use TypeScript or React in the extension I think the best solution is to use the rollup version of [this plugin](https://github.com/crxjs/chrome-extension-tools): it's quite minimal and is very similar to what I did in the script.

To get all the content scripts to connect to the right servers, we need to import the config. Since content scripts can't use static imports, this would require wrapping all of them inside an anonymous async function to use the dynamic import. To avoid this, and to make development easier by enabling static imports in content scripts I copied the above plugin's method of replacing the content scripts in the manifest with small import wrappers that only import the actual content script.

The only references to the Tournesol production servers that I didn't change are inside HTML files:

- [the Tournesol logo](https://github.com/tournesol-app/tournesol/blob/a5dc09f/browser-extension/src/browserAction/menu.html#L57)
- [the link to Tournesol in the options](https://github.com/tournesol-app/tournesol/blob/a5dc09f/browser-extension/src/options/options.html#L27)
- [the iframe in the options](https://github.com/tournesol-app/tournesol/blob/a5dc09f/browser-extension/src/options/options.html#L331)

These references are not very important to change when we are working on the extension, but it may be confusing to have links to the production servers when we are setting up the environment to not use production. I'm not sure what the best way is to change them. The easiest and safest way is probably to write these references in javascript at runtime.

### Checklist


browser-extension/src/browserAction/menu.js
```diff
@@ -24,7 +25,7 @@ function get_current_tab_video_id() {
  */
 function openTournesolHome() {
   chrome.tabs.create({
-    url: 'https://tournesol.app?utm_source=extension&utm_medium=menu',
+    url: `${frontendUrl}?utm_source=extension&utm_medium=menu`,
```
GresilleSiffle, line 28: **minor**

Here we replace `https://tournesol.app` by `https://tournesol.app/`.

I still need to check but I think our web analytics software will correctly consider both URL as the same home page. Unfortunately this is not the case of other pages. For instance `https://tournesol.app/recommendations` and `https://tournesol.app/recommendations/` will be considered as two different pages.

Even if this is not a problem for the Tournesol home page, I wonder if using a trailing slash here will not invite future contributors to write front end URL in `prepareExtention.js` with a trailing slash, which will create duplicates in our web analytics dashboard.

Maybe a comment in the `prepareExtention.js` would be enough? Maybe a comment and also removing the trailing slash on the home page URL is better?

```javascript

// please DO NOT add a trailing slash to front end URL, this prevents
// creating duplicates in our web analytics tool
const config = getForEnv(
  {
    production: {
      frontendUrl: 'https://tournesol.app',
      frontendHostEquals: 'tournesol.app',
      apiUrl: 'https://api.tournesol.app/',
    },
    'dev-env': {
      frontendUrl: 'http://localhost:3000',
      frontendHostEquals: 'localhost:3000',
      apiUrl: 'http://localhost:8000/',
    },
  },
  env
);
```

GresilleSiffle (commented):
It looks very good to me :ok_hand: it was very easy to use my local environment.

Regarding the remaining hardcoded links in the options, as you suggested we can dynamically write them with JavaScript. There is already a script `options.js` responsible of configuring the behaviour of the page `options.html`.

browser-extension/src/browserAction/menu.js
sigmike: Done in 928ad8d82deaba547b188184128c6a584b46298e.

browser-extension/prepareExtension.js
```diff
@@ -0,0 +1,120 @@
...
+    'models/*',
+    'config.js',
+  ],
+};
+
+const config = getForEnv(
+  {
+    production: {
+      frontendUrl: 'https://tournesol.app/',
+      frontendHostEquals: 'tournesol.app',
```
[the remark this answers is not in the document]
sigmike: Done in 7746b4bf4c4ad1917709c9f11e335b406b217dba.

sigmike:
> Regarding the remaining hardcoded links in the options, as you suggested we can dynamically write them with JavaScript. There is already a script `options.js` responsible of configuring the behaviour of the page `options.html`.

I did that in 21414cabda8d3eeb565ed173954fc2023e1fe61b.

As discussed elsewhere I also:

- moved the version into package.json: 0c5910ec64aba159d6108b94579ffc8766d18d55
- added a `yarn prepare` to prepare the script (and also added a `yarn prepare:dev`): 55eda5377d93cd442dc05ee8c7bc80eb284d80bb

GresilleSiffle:
Thank you.

@amatissart do you also want to have a look before merging?

I read the documentation about .eslintrc file cascading, and if I understood correctly it could be a good idea to define the key `root: true` in the root .eslintrc file. This will prevent unexpected configuration merging if ESLint checks all parent folders until the root directory.

> By default, ESLint looks for configuration files in all parent folders up to the root directory. This can be useful if you want all of your projects to follow a certain convention, but can sometimes lead to unexpected results. To limit ESLint to a specific project, place "root": true inside the .eslintrc.* file or eslintConfig field of the package.json file or in the .eslintrc.* file at your project’s root level. ESLint stops looking in parent folders once it finds a configuration with "root": true.

https://eslint.org/docs/latest/use/configure/configuration-files#cascading-and-hierarchy

sigmike:
> it could be a good idea to define the key `root: true` in the root .eslintrc file.

It's indeed good to not merge the files, but as I understand it, to do that the `root: true` should be in the src's eslintrc, not the root one. It means eslint will not merge it with the root eslintrc when parsing src files. I did that in aa2ca2a181d68d02e9220a2a2ebd8eccd85dccaf.

sigmike:
I discovered that the `prepare` script is automatically started when you run `yarn install`. I think it doesn't matter too much here, but it may still be quite surprising to have the extension switched back to production environment because you added a package or something.

amatissart (approved):

PR #1873, opened 2024-01-04
[back] feat: add entity contexts to unconnected entities API

GresilleSiffle:
**related issues** #1852

---

### Description

This PR adds the support of entity contexts in the unconnected entities API.

Moreover the logic used by the comparisons API and the unconnected entities API has been factorized in `PollScopedViewMixin`. Now all poll related views can access the entity contexts directly from `self.entity_contexts`.

**to-do**
- [x] update the tests

### Checklist


backend/tournesol/serializers/comparison.py
```diff
@@ -89,7 +89,7 @@ def to_representation(self, instance):
         poll = self.context.get("poll")
         ent_contexts = self.context.get("entity_contexts")
 
-        if poll is not None:
+        if poll is not None and ent_contexts is not None:
```
amatissart, line 92: What are the cases where `poll` or `ent_contexts` would be None ?

amatissart (approved):

backend/tournesol/serializers/comparison.py
GresilleSiffle: The `ExportComparisonsView` class uses the `ComparisonSerializer` to write the comparisons of the logged user in a .csv file. In this case the serializer context doesn't contain any entity context.

If we consider that a comparison should not be represented without its related entity contexts, then we may want to remove this condition `and ent_contexts is not None`. In this case we should also consider adding the entity contexts in the public dataset.

Because they are here to "bias" the comparisons I think it could be a good idea to include those entity contexts in the public dataset.

GresilleSiffle: I created an issue to address this question.

PR #1876, opened 2024-01-08
[back][ext] feat: recommendations API can now return random entities

GresilleSiffle:
### Description

This PR adds two features:

**(1)** A new API `polls/{name}/recommendations/random/` that returns random recommendations, designed to be more efficient than the regular recommendations route.

**(2)** The browser extension now uses this new API to improve the loading time of recommendations and their diversity.

**context**

This PR aims to reduce the number of latency alerts triggered by the API recommendations, by providing API clients, such as the browser extension, an alternative and efficient way to request random recommended entities.

Those latency alerts are currently triggered by the browser extension, when fetching the top 250 of all time recommendations to perform a random selection. Thanks to the new API, the extension doesn't need to fetch a big amount of entities to display a diverse bundle of entities.

Note that the new API is more efficient because it uses a simple SQL query that provides less features than the regular recommendations route:
- no search by text
- no search by weighted criteria score
- not possible to return unsafe recommendations

ℹ️ **consequences and important changes**

This PR significantly changes the strategy used by browser extension to display diverse entities.

(a) Before, only the top 250 of all time recommendations could be displayed. Now all recommended videos have a chance to be displayed, even those with a Tournesol score of 20.1, like on the web site.

(b) The number of cached recommendation results requested by the extension will increase from 2 every 10 minutes (for a given set of languages) to ~ 40, up to a lot more depending on how many time the users refresh their recommendations with the extension refresh button. This may impact our infrastructure.

**to-do**

*back end - /recommendations/*
- [x] remove the random logic

*back end - /recommendations/random/*
- [x] add a /recommendations/random/ view
- [x] create a new serializer
  - and the `random` parameter in the metadata?
- [x] document the view
- [x] test the view

*extension*
- [x] request random recommendations
  - should the recommendation be in the top X ?
- [x] randomly initialize the query parameter `random` for each instance of `TournesolRecommendations`
- [x] make the query parameter `random` vary for each refresh
- [x] increase the `limit` parameters for more randomness
  - actually 5 for old videos
  - actually 7 for new videos
- [x] connect the extension to the new route

### Checklist

### DIFF
--- a/backend/tournesol/serializers/poll.py
+++ b/backend/tournesol/serializers/poll.py
@@ -123,3 +123,9 @@ class RecommendationsFilterSerializer(serializers.Serializer):
         help_text="If true and a user is authenticated, then entities compared by the"
         " user will be removed from the response",
     )
+
+
+class RecommendationsRandomFilterSerializer(serializers.Serializer):
+    random = serializers.IntegerField(default=None)
+    date_lte = serializers.DateTimeField(default=None)
+    date_gte = serializers.DateTimeField(default=None)
--- a/backend/tournesol/serializers/poll.py
+++ b/backend/tournesol/serializers/poll.py
@@ -139,7 +139,7 @@ class RecommendationsFilterSerializer(serializers.Serializer):
 
 
 class RecommendationsRandomFilterSerializer(serializers.Serializer):
-    random = serializers.IntegerField(
+    bundle = serializers.IntegerField(
         default=None,
         help_text="Successive calls can return the same cached results. Vary"
         " this parameter to request new results."
--- a/backend/tournesol/serializers/poll.py
+++ b/backend/tournesol/serializers/poll.py
@@ -145,7 +145,7 @@ class RecommendationsFilterSerializer(serializers.Serializer):
 class RecommendationsRandomFilterSerializer(serializers.Serializer):
     random = serializers.IntegerField(
         default=None,
-        help_text="Successive calls will produce the same cached results. Vary"
+        help_text="Successive calls can return the same cached results. Vary"
         " this parameter to request new results."
     )
     date_lte = serializers.DateTimeField(default=None)
--- a/backend/tournesol/serializers/poll.py
+++ b/backend/tournesol/serializers/poll.py
@@ -104,6 +104,12 @@ class RecommendationSerializer(ModelSerializer):
 
 
 class RecommendationsFilterSerializer(serializers.Serializer):
+    random = serializers.IntegerField(
+        default=None,
+        help_text="If defined, the entities will be randomized instead of sorted by score."
+        " Because the API results are cached, using the same digit will return the same set of"
+        " entities. To get new entities, use a different digit."
+    )
     date_lte = serializers.DateTimeField(default=None)
     date_gte = serializers.DateTimeField(default=None)
     search = serializers.CharField(default=None, help_text="A search query to filter entities")
--- a/backend/tournesol/serializers/poll.py
+++ b/backend/tournesol/serializers/poll.py
@@ -103,6 +103,29 @@ class RecommendationSerializer(ModelSerializer):
         read_only_fields = fields
 
 
+class RecommendationRandomSerializer(serializers.Serializer):
+    entity = RelatedEntitySerializer(source="*", read_only=True)
+    collective_rating = ExtendedCollectiveRatingSerializer(
+        source="single_poll_rating",
+        read_only=True,
+        allow_null=True,
+    )
+    entity_contexts = EntityContextSerializer(
+        source="single_poll_entity_contexts",
+        read_only=True,
+        many=True
+    )
+
+    class Meta:
+        model = Entity
+        fields = [
+            "entity",
+            "collective_rating",
+            "entity_contexts",
+        ]
+        read_only_fields = fields
+
+
 class RecommendationsFilterSerializer(serializers.Serializer):
     random = serializers.IntegerField(
         default=None,
--- a/backend/tournesol/serializers/poll.py
+++ b/backend/tournesol/serializers/poll.py
@@ -105,23 +105,23 @@ class RecommendationSerializer(ModelSerializer):
 
 class RecommendationRandomSerializer(serializers.Serializer):
     entity = RelatedEntitySerializer(source="*", read_only=True)
-    collective_rating = ExtendedCollectiveRatingSerializer(
-        source="single_poll_rating",
-        read_only=True,
-        allow_null=True,
-    )
     entity_contexts = EntityContextSerializer(
         source="single_poll_entity_contexts",
         read_only=True,
         many=True
+    )
+    collective_rating = ExtendedCollectiveRatingSerializer(
+        source="single_poll_rating",
+        read_only=True,
+        allow_null=True,
     )
 
     class Meta:
         model = Entity
         fields = [
             "entity",
-            "collective_rating",
             "entity_contexts",
+            "collective_rating",
         ]
         read_only_fields = fields
 
--- a/backend/tournesol/serializers/poll.py
+++ b/backend/tournesol/serializers/poll.py
@@ -127,12 +127,6 @@ class RecommendationRandomSerializer(serializers.Serializer):
 
 
 class RecommendationsFilterSerializer(serializers.Serializer):
-    random = serializers.IntegerField(
-        default=None,
-        help_text="If defined, the entities will be randomized instead of sorted by score."
-        " Because the API results are cached, using the same digit will return the same set of"
-        " entities. To get new entities, use a different digit."
-    )
     date_lte = serializers.DateTimeField(default=None)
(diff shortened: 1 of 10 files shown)


GresilleSiffle:
The check `run end-to-end tests / e2e-browser-extension (pull_request)` is failing because the browser extension uses the route that comes with this PR (not deployed in production yet).

backend/tournesol/serializers/poll.py
```diff
@@ -103,6 +103,29 @@ class Meta:
         read_only_fields = fields
 
 
+class RecommendationRandomSerializer(serializers.Serializer):
```
lfaucon, line 106: I'm slightly worried about the code duplication. This Serializer is nearly exactly the same thing as RecommendationSerializer defined above.

Also the name is not very well chosen. "Random" relates to how the entities are selected not to what data we want to include in the response.

backend/tournesol/views/polls.py
```diff
@@ -295,6 +295,13 @@ def annotate_and_prefetch_scores(self, queryset, request, poll: Poll):
         criteria_weight = self._build_criteria_weight_condition(
             request, poll, when="all_criteria_scores__criteria"
         )
+
+        # FIXME: we can significantly improve the performance of the queryset
+        # by filtering the criteria on their names, to remove those that are
+        # not present in the request.
+        #
+        #   ex:
+        #       all_criteria_scores__criteria__in=[...]
```
lfaucon, line 304: Can we move this suggestion in a ticket instead of a comment?

An other good optimization would be that the route used 95% of the time simply reads the tournesol_score on EntityPollRating (i.e. the main criteria) without doing the join with the criteria scores.

Only when users specify other criteria or want to weight each criteria manually, we'd use this route.

backend/tournesol/utils/constants.py
```diff
@@ -18,4 +18,5 @@
 COMPARISON_MAX = 10.0
 
 # Default weight for a criteria in the recommendations
+# FIXME: the default weight used by the front end is 50, not 10
```
lfaucon, line 21: They don't necessarily need to be the same. The backend can assume 10 when nothing is specified & the frontend can initialize a user's search by default with a value of 50.

I vaguely remember discussing with @amatissart to make it 0 by default. I don't remember well what makes a criteria included or not in the scores used for ranking.

backend/tournesol/tests/test_api_polls_reco_random.py
```diff
@@ -0,0 +1,141 @@
...
+
+        resp = self.client.get(f"{self.url_path}?date_gte=2021-01-01")
+        results = resp.data["results"]
+        uids = [res["entity"]["uid"] for res in results]
+
+        self.assertEqual(resp.status_code, status.HTTP_200_OK)
+        self.assertEqual(len(results), 3)
+        self.assertIn(self.video_2.uid, uids)
+        self.assertIn(self.video_3.uid, uids)
+        self.assertIn(self.video_4.uid, uids)
```
lfaucon, line 141: Missing a test that actually checks that videos are ordered & selected randomly.

Can add 100 videos, then make two requests varying the ?random param and assert they return different lists. Assert they don't return the list 1,2,3,4,5 (that would be returned if ordered by scores).

browser-extension/src/background.js
```diff
@@ -12,13 +12,9 @@ import {
...
-const oversamplingRatioForRecentVideos = 3;
-const oversamplingRatioForOldVideos = 50;
-// Higher means videos recommended can come from further down the recommandation list
-// and returns videos that are more diverse on reload
-
-const recentVideoProportion = 0.75;
-const recentVideoProportionForAdditionalVideos = 0.5;
+const RECENT_VIDEOS_RATIO = 0.75;
+const RECENT_VIDEOS_EXTRA_RATIO = 0.5;
+const BUNDLE_SIZE_F = 3;
```
lfaucon, line 17: What does `BUNDLE_SIZE_F` stand for?

browser-extension/src/background.js
```diff
@@ -192,42 +186,36 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
...
         const threeWeeksAgo = getDateThreeWeeksAgo();
         const recommendationsLangs =
           await getRecommendationsLanguagesAuthenticated();
 
-        // Only one request for both videos and additional videos
         const recentParams = new URLSearchParams([
           ['date_gte', threeWeeksAgo],
-          ['limit', recentVideoToLoad + recentAdditionalVideoToLoad],
+          ['limit', (recentToLoadRow1 + recentToLoadExtra) * BUNDLE_SIZE_F],
+          ['random', request.queryParamRandom],
```
lfaucon, line 212: To my understanding this parameter could have any name and is just helpful to invalidate the cache?

maybe we could name it `cache_avoidance_param` ?


It looks a bit like a `random_seed` but that's not how it actually works either.

backend/tournesol/serializers/poll.py
GresilleSiffle: That's right. I first thought about adding specific fields in this serializer, but I finally decided to keep it simple.

I can make a "base" serializer out of it, and `RecommendationSerializer` could simply extend it with its own metadata.

browser-extension/src/background.js
GresilleSiffle: Ha this name caused me trouble : ) It was hard to find a "good" one.

It's a factor used to define the size the returned video bundle: 3 means we want the API to return three times as many videos as we need.

browser-extension/src/background.js
GresilleSiffle: We had this conversation with @amatissart .

`seed` / `random_seed` could have worked, but it's not really a seed, and using the same parameter won't necessarily return the same results after the cache has expired.

`cache_avoidance_param` could work, it's still a bit confusing to me. For instance calling several times in a row `cache_avoidance_param=true` will return the same cached results, even if the name suggests that it won't.

What do you think about `bundle`? `bundle=1` means we want the first cached video bundle identified by the integer 1.

backend/tournesol/serializers/poll.py
GresilleSiffle: Addressed by https://github.com/tournesol-app/tournesol/pull/1876/commits/cd27180532c252b68ffd80b3757691172a4ab68a

`RecommendationBaseSerializer` is a base class containing all fields a recommendation serializer should always provide: the data related to the serialized entity, its context, and the collective ratings.

`RecommendationSerializer` inherits from `RecommendationBaseSerializer` and adds ts own metadata.

The inner class `Meta` of `RecommendationSerializer` is explicitly defined as recommended by the DRF documentation, but we could make it inherit from `RecommendationBaseSerializer` if we prefer.

`RecommendationSerializer` is not a model serializer anymore, it's now a regular serializer. This serializer is a representation built from several different models, and is not designed to be able to create or update a specific Django model.

browser-extension/src/background.js
lfaucon: BUNDLE_OVERFETCHING_FACTOR ?

browser-extension/src/background.js
lfaucon: I like bundle the best :)

backend/tournesol/views/polls_reco_random.py
```diff
@@ -0,0 +1,88 @@
...
+
+    permission_classes = []
+    serializer_class = RecommendationBaseSerializer
+    poll_parameter = "name"
+
+    @method_decorator(cache_page_no_i18n(60 * 10))
+    def get(self, request, *args, **kwargs):
+        response = self.list(self, request, *args, **kwargs)
+        if request.query_params.get("exclude_compared_entities") == "true":
+            patch_vary_headers(response, ["Authorization"])
```
lfaucon, line 87: Doesn't this line means that the cache is separate between the users. I.e. if 1000 users request that route once. The cache will contain 1000 entries and never successfully hit?

GresilleSiffle: Yes, this is something I still need to fix.

GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1876/commits/cf9f06202b38b6f2af4468f11a8ae8ad9422b039

browser-extension/src/background.js
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1876/commits/4c1cb676aeca5578c0066e8cac502d78058f761a

browser-extension/src/background.js
GresilleSiffle: Fixed by https://github.com/tournesol-app/tournesol/pull/1876/commits/4c1cb676aeca5578c0066e8cac502d78058f761a

backend/tournesol/views/polls.py
GresilleSiffle: See: https://github.com/tournesol-app/tournesol/issues/1881

Issue #861, opened 2022-04-26
[front] Investigate possibilities for Tournesol to be a PWA


lfaucon:
- [x] Learn about PWA: https://web.dev/progressive-web-apps/
- [x] Make a small presentationation to the team with the most important things to understand about PWA
- [x] If it is decided, make the necessary changes such that Tournesol can be a PWA

adrausch:
Presentation in english?

lfaucon:
For anytime when ready & in French should be good :)

lfaucon:
To-do short term:
- Check what can be done easily and do it.
- Discuss with the team about what cannot be done easily and would be valuable

Issue #1867, opened 2023-12-19
[back] clean-up: delete the /video/ API


GresilleSiffle:
### Context

The `/video/` API has been replaced by `/entities/` and `/poll/{poll_name}/`. Only the POST `/video/` route has not been refactored yet.

If this legacy API is not required anymore, we could delete it. Moreover, deleting the legacy serializers will simplify the front end services generation, and type definition.

We may want to refactor the POST `/video/` route before. Note that this view uses a custom request throttle rate that should also be used by the new view.

⚠️ We think it is not used, but we should double check that it is indeed not used and change its usages where needed

PR #1877, opened 2024-01-11
[front] feat: complete PWA manifest

amatissart:
Related to https://github.com/tournesol-app/tournesol/issues/861

---

### Description

This PR adds several features related to the PWA manifest.

Here is what it looks like on Android with the app installed from Chrome (it's also possible to install the app via Firefox on Android or Safari on iOS, but in this case some of those features will be missing).

#### Tournesol is listed as a "share target" from other apps

When a video URL is shared to Tournesol, the Tournesol app will open automatically the appropriate entity page.
This PR makes sure to add the video in the database on behalf of the logged-in user. 

### Checklist


frontend/public/images/screenshot_comparison_mobile.png
lfaucon, line 1: Could we simplify these two screenshots removing SELECT, AUTO, A, B, v, ADD OPTIONAL CRITERIA, (i) ..., and the footer ?

frontend/src/pages/entities/EntityAnalysisPage.tsx
```diff
@@ -75,7 +82,33 @@ const EntityAnalysisPage = () => {
...
+    }
+    if (!isLoggedIn) {
+      return false;
+    }
+    const videoId = extractVideoId(uid);
+    if (!videoId) {
+      return false;
+    }
+    try {
+      await VideoService.videoCreate({
```
lfaucon, line 97: Related to #1867. It could be easily forgotten. Commenting only so that it gets mentioned on the ticket.

frontend/src/app/SharedContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+  if (!videoId) {
+    return <Redirect to="/entities/invalid" />;
+  }
+
+  const destination = `/entities/yt:${videoId}`;
+  if (!isLoggedIn) {
+    return (
+      <Redirect
+        to={{
+          pathname: '/login',
```
lfaucon, line 30: It's quite unlikely that someone would use the PWA and not be logged in, but browsing the pages /entities or adding a new entity does not require login does it?

lfaucon (approved):
Works great for me :) Thank you

GresilleSiffle (approved):
It also works for me when using Chrome : )

I have a lot of performance issue with the PWA installed with Firefox on mobile, I didn't investigate yet, but it's probably not related to our code.

frontend/public/images/screenshot_comparison_mobile.png
amatissart: Do you prefer these?

GresilleSiffle: If the screenshots are here to let the users know how the application looks like, I have a preference for keeping the real captures. This way we are fully transparent with the users. Moreover I don't know if it's a good practice to display illustrations instead of real captures at the installation.

amatissart: Ok that's also a good incentive to simplify the real UI :+1:

PR #1899, opened 2024-01-24
Browser extension.v3.5.0

Andrea-Bardinet:
Extension update : 

On a youtube video page, add a "Watch on Tournesol" button, next to "Rate Later" and "Rate Now".

I've renamed the files and the addRateButton function to addVideoButton for a more consistent name.
Pressing the button opens a new tab on the tournesol.app/video/<id> page.

Tested on Chromium and Firefox.

Capture d’écran du 2024-01-24 05-41-06

### DIFF
--- a/browser-extension/src/addVideoButtons.js
+++ b/browser-extension/src/addVideoButtons.js
@@ -15,12 +15,12 @@ const TS_ACTIONS_ROW_BEFORE_REF = 'bottom-row';
  *
  * This part is called on connection for the first time on youtube.com/*
  */
-document.addEventListener('yt-navigate-finish', addRateButtons);
+document.addEventListener('yt-navigate-finish', addVideoButtons);
 
 if (document.body) {
-  addRateButtons();
+  addVideoButtons();
 } else {
-  document.addEventListener('DOMContentLoaded', addRateButtons);
+  document.addEventListener('DOMContentLoaded', addVideoButtons);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


browser-extension/src/addVideoButtons.js
```diff
@@ -159,5 +159,14 @@ function addRateButtons() {
         });
       },
     });
+
+    addVideoButton({
+      id: 'tournseol-watch-button',
+      label: 'Watch on Tournesol',
+      iconSrc: chrome.runtime.getURL('images/watch.svg'),
+      onClick: () => {
+        window.open(`${frontendUrl}/video/${videoId}`, '_blank');
```
GresilleSiffle, line 168: The URL `/video/{video_id}` is a legacy URL we don't want to use anymore. We can use the new URL instead:

```suggestion
        window.open(`${frontendUrl}/entities/yt:${videoId}?utm_source=extension&utm_medium=video_button`, '_blank');
```

I also added a couple query parameters to make the team able to see how users interact with the browser extension (those stats are private and anonymous).

GresilleSiffle (commented):
Thank you @Andrea-Bardinet for your spontaneous contributions :clap: 

This looks like a nice addition. We already have a link to the Tournesol website "hidden" in the browser actions menu, but displaying the button under the video is far more visible.

I'm personally ok with this new feature. What do you think @tournesol-app/maintainers and @aidanjungo ?

If we decide to merge, I'll update the .svg icon before: I'll use a free Font Awesome icon to keep the design coherent between all icons, and to ease the licence management.

GresilleSiffle:
Let's not forget to update the file `package.json`.

GresilleSiffle (approved):

Issue #1883, opened 2024-01-17
[ext] Manifest V3 support


sigmike:
Chrome will eventually stop supporting manifest V2, so we should prepare support for manifest V3.

Some references:

- https://developer.chrome.com/blog/resuming-the-transition-to-mv3?hl=en ("we highly recommend completing migration to Manifest V3 before June 2024")
- https://developer.chrome.com/docs/extensions/develop/migrate/what-is-mv3
- https://developer.chrome.com/docs/extensions/develop/migrate/checklist?hl=en
- https://extensionworkshop.com/documentation/develop/manifest-v3-migration-guide/

sigmike:
@GresilleSiffle @amatissart @lfaucon Firefox has supported manifest V3 since version 109, released in January 2023. Do we want to keep a manifest V2 version to maintain compatibility with Firefox versions < 109? Or should we fully migrate to V3? It's not very difficult to support both with the new system that generates the manifest, but it does add a bit of complexity to handle both cases.

Do you know if we have any users using an old Firefox?

GresilleSiffle:
Hello @sigmike 

Thank for all those references. We are going to check in our analytics what % of users use Firefox < 109.

amatissart:
@sigmike If the support is good enough on Firefox too, I think it would reasonable to support manifest v3 only.
There is a small residue of Firefox 108 in the logs, but that seems to be specific to Android where the extension is not currently available anyway.

sigmike:
@GresilleSiffle @amatissart Ok thank you.

So I've done some more testing and we can't really have a manifest V3 that works well on all browsers (yet?). To sum up, it doesn't work well on Firefox, but works well on Chrome.

So I suggest we start building 2 versions of the extension: one for Firefox (with manifest V2 for now, and maybe V3 later) and one for Chrome-based browsers (with manifest V3). And maybe one day both extensions will be identical, but we might still want or need browser specific features.

Or we can just ignore manifest V3 for now, and try again when it's about to become mandatory.

The details of the problems:

1. Background: Chrome requires a `service_worker` for the `background` key, and Firefox doesn't support it (yet, probably). We can use both `script` and `service_worker` at the same time for a manifest that works on both, but it's only supported from Chrome 121 which was released in January 2024. [Reference](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background).

[...]

PR #1912, opened 2024-02-06
[ext] Manifest V3

sigmike:
**related issues** #1883

---

### Description

This PR implements the changes to support manifest V3. The configure and build commands still produce a manifest V2.

You can produce a manifest V3 version for Chrome with:

    EXTENSION_BROWSER=chrome node prepareExtension.js

And for Firefox:

    EXTENSION_BROWSER=firefox MANIFEST_VERSION=3 node prepareExtension.js

For Firefox it defaults to manifest V2 because of UX and CORS issues, see https://github.com/tournesol-app/tournesol/issues/1883#issuecomment-1916692065.

If we want to release a manifest V3 version for Chrome the remaining task is to update the build scripts, and we may also want to change the `yarn configure` commands.

### Checklist

### DIFF
--- a/browser-extension/prepareExtension.js
+++ b/browser-extension/prepareExtension.js
@@ -1,5 +1,5 @@
 import {
-  getForEnv,
+  selectValue,
   generateImportWrappers,
   writeManifest,
   writeConfig,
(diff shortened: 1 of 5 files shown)


GresilleSiffle:
Thank you again @sigmike for all your contributions :clap: 

The current way Firefox handles the host permissions is indeed quite complex and I'm worried about the UX. I hope it will change in the future. To use the manifest v3 version on Firefox, I had to manually enable the permissions for each host from the extension menu, 3 clicks + 1 click per host :face_with_spiral_eyes: (see the image below).

For our extension to work those host permissions are required, [but it looks like they are optional for Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/host_permissions#requested_permissions_and_user_prompts). If for specific reasons we want to publish a manifest v3 version for Firefox in the future, we may need to add an onboarding UI inviting the users to accept the permissions.

On Chromium, the host permissions were accepted by default. Let's not forget to update our GitHub workflow [before the end of manifest v2](https://developer.chrome.com/docs/extensions/develop/migrate/mv2-deprecation-timeline?hl=fr#june_2024_chrome_mv2_deprecation_pre-stable_rollout) to build both the Firefox and the Chromium version.

capture

GresilleSiffle (approved):
Good to me :+1: 

Maybe we can add a section in the `README.md`, just after the section `Prepare the extension`, or inside, to explain how to create a manifest v3 version with the commands listed in this PR description.

amatissart (approved):

Issue #1931, opened 2024-03-06
[front] feat: in comparison, swipe up / down on entity card to trigger the Auto button


GresilleSiffle:
**sub-task of** https://github.com/tournesol-app/tournesol/issues/1274

The swipe up has been implemented.

The swipe down behaviour needs to be described and implemented. Also mind the behaviour in the tutorial.

**expected behaviour**

A swipe down should behave like a swipe up: return a new entity to compare, or if the history has been browsed "up", return the previous entry in the history.

PR #1954, opened 2024-04-16
[front] feat: swipe down to go back in the history

GresilleSiffle:
**related issues** #1931

---

### Description

**to-do**

- [x] add the values of the query parameters `uidA` and `uidB` to the history
- [x] put things in the correct files
- [x] transform `UidHistory` in `SuggestionHistory` and remove the hook `useSuggestions`

*tests*

- [x] test the class `UidHistory`
- [x] test the class `SuggestionPool`

**questions**

(a) Should we have one suggestion history per entity selector?

*Yes.* It's rather surprising to be able to go back in the history with one selector, after new videos have been suggested in the other selector. Moreover, having a single history position for two selectors create incoherences during the history navigation.

(b) Should we include the video selected with the Select button in the history?

*Yes.*

### Checklist


frontend/src/features/comparisons/Comparison.tsx
```diff
@@ -143,12 +147,20 @@ const Comparison = ({
       let autoUidA = null;
       let autoUidB = null;
 
-      if (!uidA && autoFillSelectorA) {
-        autoUidA = await nextSuggestion(uidA, uidB, pollName);
+      if (uidA) {
+        selectorAHistory.push(pollName, uidA);
+      } else if (autoFillSelectorA) {
+        autoUidA = await nextSuggestion(pollName, [uidA, uidB], 'A');
       }
```
amatissart, line 154: I wonder if it could be simplified by moving the suggestion logic into `UidHistory` to get something like:
```ts
      if (uidA) {
        selectorAHistory.push(pollName, uidA);
      } else if (autoFillSelectorA) {
        autoUidA = await selectorAHistory.getSuggestion(pollName, {exclude: [uidB]})
      }
```

amatissart (commented):
Overall it looks good to me, even if I haven't looked at #1956 yet.

As described in my comment, I wonder the `UidHistory` could be responsible for fetching suggestions and pass history objects instead of using the hook `useSuggestions` + historyId.

frontend/src/features/entity_selector/EntitySelector.tsx
```diff
@@ -318,6 +358,8 @@ const EntitySelectorInnerAuth = ({
                 value={inputValue || uid || ''}
                 onChange={handleChange}
                 otherUid={otherUid}
+                historyId={historyId}
+                historyInsertion={lastSlide === 'up' ? 'right' : 'left'}
```
amatissart, line 362: After some manual testing, I am not convinced this behaviour is that intuitive.
I think I would prefer to make the previous video always accessible with a "down" swipe. That's also easier to explain: after selecting a video, swipe "up" to go the next suggestion, "down" to go back to the previous one.

GresilleSiffle: Good idea :+1:

frontend/src/features/comparisons/Comparison.tsx
```diff
@@ -143,12 +146,31 @@ const Comparison = ({
...
-      if (!uidA && autoFillSelectorA) {
-        autoUidA = await nextSuggestion(uidA, uidB, pollName);
+      // At this moment, the entities still present in selectorAHistory and
+      // selectorBHistory come from a previous comparison session. During this
+      // last session they may have been compared enough times to not be
+      // considered as suggestion by the API anymore. To avoid suggesting
+      // entities that don't match the suggestion strategies used by the API,
+      // we clear the history of all selectors.
+      selectorAHistory.clear();
+      selectorBHistory.clear();
```
amatissart, line 156: Does that mean the history is always reset when the comparison page is reloaded?
If that's the case, it could be more simple and explicit to initialize the history as `useRef()` in this component?

frontend/src/features/entity_selector/EntitySelector.tsx
```diff
@@ -268,13 +270,31 @@ const EntitySelectorInnerAuth = ({
...
   const onSlideExited = async () => {
-    setSlideDirection('up');
-    const newUid = await nextSuggestion(uid, otherUid, pollName);
-    if (newUid) {
-      onChange({ uid: newUid, rating: null });
+    if (slideDirection === 'up') {
+      const newUid = await history?.nextLeftOrSuggestion(pollName, [
+        uid,
+        otherUid,
+      ]);
```
amatissart, line 277: From a UX perspective, I am not sure it makes sense to load new suggestions on both sides of the history. I think it would be more understandable/discoverable to require a "swipe up" to get new suggestions, and to block the "swipe down" (ideally with a bounce animation) when there is no older item in the history.

What do you think?

amatissart (commented):
I have 1 UX question about the behaviour of the "swipe down", and a few minor comments on the implementation.

frontend/src/features/comparisons/Comparison.tsx
GresilleSiffle: > Does that mean the history is always reset when the comparison page is reloaded?

Exactly.

> If that's the case, it could be more simple and explicit to initialize the history as useRef() in this component?

Do you mean initializing it inside the component `Comparison` instead of importing it as a global variable?

GresilleSiffle: Addressed by https://github.com/tournesol-app/tournesol/pull/1954/commits/3117722a8e20280c1912b5bb32280886f7125540

frontend/src/features/entity_selector/EntitySelector.tsx
amatissart: As discussed on Discord, we can keep this version and get feedbacks.

amatissart (approved):

PR #1960, opened 2024-04-22
[front] feat: always exclude compared videos when loading recommendations feed (PWA)

amatissart:
### Description

Following #1920, this PR suggests to exclude already compared videos when loading the initial recommendations page from the PWA. Regular users should be able to find new content at a glance. This would also encourage users to compare videos they have already watched, to make them disappear from the feed.

### Checklist

### DIFF
--- a/frontend/src/pages/feed/FeedCollectiveRecommendations.tsx
+++ b/frontend/src/pages/feed/FeedCollectiveRecommendations.tsx
@@ -13,10 +13,11 @@ const FeedCollectiveRecommendations = () => {
   const searchParams = getDefaultRecommendationsSearchParams(
     pollName,
     options,
-    userSettings
+    userSettings,
+    { forceExcludeCompared: true }
   );
 
-  return <Redirect to={`/recommendations${searchParams.toString()}`} />;
+  return <Redirect to={`/recommendations${searchParams}`} />;
 };
 
 export default FeedCollectiveRecommendations;
--- a/frontend/src/utils/userSettings.ts
+++ b/frontend/src/utils/userSettings.ts
@@ -35,15 +35,17 @@ export const buildVideosDefaultRecoSearchParams = (
   searchParams: URLSearchParams,
   userSettings: VideosPollUserSettings | undefined
 ) => {
-  const advancedFilters: string[] = [];
+  const advancedFilters = new Set(
+    searchParams.get('advanced')?.split(',') ?? []
+  );
   if (userSettings?.recommendations__default_unsafe) {
-    advancedFilters.push('unsafe');
+    advancedFilters.add('unsafe');
   }
   if (userSettings?.recommendations__default_exclude_compared_entities) {
-    advancedFilters.push('exclude_compared');
+    advancedFilters.add('exclude_compared');
   }
-  if (advancedFilters.length > 0) {
-    searchParams.set('advanced', advancedFilters.join(','));
+  if (advancedFilters.size > 0) {
+    searchParams.set('advanced', [...advancedFilters].join(','));
   }
 
   if (userSettings?.recommendations__default_date != undefined) {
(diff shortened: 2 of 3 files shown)


GresilleSiffle (commented):
The logic is good to me :ok_hand: 

Regarding the UX, I suggest to add an indication about this new behaviour in the page Preferences, to avoid any confusion (but before read my comment below :)

capture

As a contributor, this new behaviour is a little bit frustrating to me. When I go to the page Recommendations, it's not always to see what I can compare, but also to see what's going on in the Tournesol community, and what Tournesol is currently recommending. Sometimes I go to see if a video I compared is now recommended.

By forcing the setting `exclude_compared` to true, the platform makes an assumption on my intentions which can go against the preferences I configured. If I want to see all recommendations I'll have to display the filters and uncheck the checkbox each time I open the PWA. I've been checking more and more the recommendations from the PWA recently, but this new behaviour makes me want to use my laptop instead : /

Here is an alternative: consider `exclude_compared` exactly as the other settings. As long as the user hasn't configured the setting `exclude_compared` yet, consider `exclude_compared` as true (true will be the default value, like "month" is the default value of `date`). When the setting has been configured, use the configured value instead.

What do you think?

amatissart:
> As a contributor, this new behaviour is a little bit frustrating to me. When I go to the page Recommendations, it's not always to see what I can compare, but also to see what's going on in the Tournesol community, and what Tournesol is currently recommending. Sometimes I go to see if a video I compared is now recommended.

There are different use-cases for the page "Recommendations". The intention of this PR is to make the PWA behave more like a consumer app to find new content directly from the entry page. But we may need to make the distinction clearer between the recommendations "feed" and the public recommendations page with all videos ordered by Tournesol ratings.

> By forcing the setting `exclude_compared` to true, the platform makes an assumption on my intentions which can go against the preferences I configured. If I want to see all recommendations I'll have to display the filters and uncheck the checkbox each time I open the PWA. I've been checking more and more the recommendations from the PWA recently, but this new behaviour makes me want to use my laptop instead : /

I agree this change can bring confusion about the behaviour the existing setting "exclude_compared". Note that technically you don't need to uncheck the filter though: it's possible to reset the filters by accessing to "Recommendations" from the main menu, but that's still confusing :confused: .

> Here is an alternative: consider `exclude_compared` exactly as the other settings. As long as the user hasn't configured the setting `exclude_compared` yet, consider `exclude_compared` as true (true will be the default value, like "month" is the default value of `date`). When the setting has been configured, use the configured value instead. In other words: are there reasons to consider the behaviour of the setting `exclude_compared` different than the other settings?

That would be more simple indeed. But I fear it may create other problems to apply "exclude_compared" by default in all contexts. (what about the results when clicking on a channel name, etc.).

Initially I thought this PR could be a simple change, but it obviously deserves more discussion :sweat_smile:

lfaucon:
Let's have a discussion

I think we would dispel the difficulty if we had separate pages for separate purposes.

- Top videos
    - title: "Top videos according to the community"
    - description: "...tbd"
    - filters: date & language, default "All time" + configured language
    - user settings: only languages
    - default: include compared
    - motivation: This should be what we show when we talk about how Tournesol identifies high quality important videos
- Search 
    - title: "Best videos"
    - description: ""
    - filters: all filters
    - user settings: same as now to configure what parameters the search takes by default.
    - default: include compared
    - motivation: This is somehow what our current page does best, and should fulfill the same role, whenever a user looks up a channel / searches key words / wants to see unsafe videos / ...
- For you
    - title: "Recommended for you"
    - description: "Videos recommended for you. Use the settings to control what gets shown"
    - filters: no filter on the page, but easy way to navigate to Search, maybe even no pagination.
    - user settings: similar as for search but separate settings parameters to configure what the for you does by default.
    - default: exclude compared
    - motivation: This should be the page that opens by default for logged in recurrent users. Maybe in the future it uses the reco bundles.

GresilleSiffle:
The new pages Top Videos, For You and Search will be tracked by this issue: https://github.com/tournesol-app/tournesol/issues/1983

Issue #1834, opened 2023-11-10
[back] feat: ClassicEntitySuggestionStrategy should use the user's preferred languages when returning recommended entities


GresilleSiffle:
The suggestion strategy `ClassicEntitySuggestionStrategy` can include in its results entities from the recent and all-time recommendations.

The recommendations should be returned according to the user's preferred languages configured in its setting `recommendations__default_languages` (if the poll has such a notion, not all polls have a language associated with their entities).

**pending questions**

(a) What should we return when the user hasn't defined its preferred languages?
- we could use the English language in addition to the `accept-language` request header as ultimate fallback 

(b) What should the strategy do when no recommendations are available in the user's preferred languages?
- we could use the same fallback as (a)

GresilleSiffle:
I moved this issue in the selected column because it's the continuation of a pull request that is about to be merged.

Feel free to move back this issue in the Idea column in case of need.

GresilleSiffle:
The way it has been implemented in #1961 has a small issue.

The recommended entities, which represent 4 / 20 of entities recommended by the `ClassicEntitySuggestionStrategy`, are randomly selected. If a user has `en` and `it` defined as his/her preferred languages, the suggestions will contain mostly in English recommendations, because the API database has far more English entities than Italian. In fact, it's possible that the user will get zero entities matching the Italian language.

This is even more problematic for a user having only `it` defined as his/her preferred language. The view always adds `en` to the preferred languages if they don't contain `en` or `fr`, and it's possible that this user will get only English recommendations even if the database contain enough recommendations in Italian.

 I think the issue is small because:
- the current user base is mainly composed by English and French speaking users, they should not be impacted
- only 4 / 20 suggested entities are affected by the issue, the rest come from the rate-later list + already compared

This could easily be fixed by adding a "fallback" feature in the `ClassicEntitySuggestionStrategy` class. Instead of returning recommended entities based on a single list of languages, the class could:
- return entities based on the user's preferences (without the `en` lang added)
- if there is no recommended entities matching those languages, use `en` instead

PR #1961, opened 2024-04-24
[back] feat: /tocompare/ API uses the user's preferred languages

GresilleSiffle:
**related issues** #1834 

---

### Description

This PR makes the endpoint `/users/me/suggestions/{poll_name}/tocompare/` able to return recommended entities matching the logged user's preferred languages. This should close the milestone _Video selector redesign_.

The preferred languages are determined using the following procedure:
- check the user's settings for the poll
- if the setting hasn't been configured, check the HTTP header `Accept-Language`
- if the languages `en` or `fr` are not present in the preferred languages, add `en`

The fallback language `en` is explicitly hardcoded in the back end. This might not be satisfactory if we want to make the back end generic. This trick ensures new users without entities in their rate-later list will always be suggested entities by the Auto button, even if there is no recommendations matching their preferred languages.

See also the behaviour of the tutorial: https://github.com/tournesol-app/tournesol/blob/4f67252a9a702b0e376d961577ad439b81a3d69d/frontend/src/utils/polls/videos.tsx#L54

**to-do**

*model - User*

- [x] add a method `get_recommendations_default_langs` take a poll name as argument

*view - SuggestionsToCompare*

- [x] call `get_recommendations_default_langs`
- [x] refactor the fallback on lang `en`

*tests*

- [x] test the new language filter added to `ClassicEntitySuggestionStrategy`
- [x] update the API tests

### Checklist

### DIFF
--- a/backend/tournesol/views/suggestions/to_compare.py
+++ b/backend/tournesol/views/suggestions/to_compare.py
@@ -7,6 +7,7 @@ from rest_framework.permissions import IsAuthenticated
 
 from tournesol.lib.suggestions.strategies import ClassicEntitySuggestionStrategy
 from tournesol.serializers.suggestion import EntityToCompare
+from tournesol.utils.http import langs_from_header_accept_language
 from tournesol.views import PollScopedViewMixin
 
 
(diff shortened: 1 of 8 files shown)


backend/tournesol/views/suggestions/to_compare.py
```diff
@@ -38,16 +40,54 @@ class SuggestionsToCompare(PollScopedViewMixin, generics.ListAPIView):
...
+    min_lang_set = ["en", "fr"]
+
+    def _langs_from_accept_language(self, header: str) -> list[str]:
+        """
+        Return a list of language tags from a given Accept-Language HTTP
+        header.
+
+        See https://www.rfc-editor.org/rfc/rfc9110#field.accept-language
+        """
+        return [lang.split(';')[0].strip() for lang in header.split(",")]
```
amatissart, line 56: [`django.utils.translation.trans_real.parse_accept_lang_header`](https://github.com/django/django/blob/4.2.11/django/utils/translation/trans_real.py#L618) does something similar, possibly in a more robust way. But not sure this is a stable API.

amatissart (approved):

backend/tournesol/views/suggestions/to_compare.py
GresilleSiffle: Addressed in https://github.com/tournesol-app/tournesol/pull/1961/commits/77102964668d47942c92cf8274e75d0820af1eb2

PR #1970, opened 2024-05-15
Neurips24

lenhoanglnh:
---

### Description

Initially the goal was mostly data analysis and experiments for NeurIPS 24 submissions.
However, this has spurred two changes to Solidago source files:
- Added import of other csv files of Tournesol public dataset.
- Modified qr_quantile, which was previously flawed. More precisely, qr_quantile had the undesirable property that the addition of a new user with (near) infinite uncertainty decreased the qr_quantile estimate. With the new version of qr_quantile, infinite uncertainty is equivalent to zero voting right.

### Checklist


solidago/src/solidago/primitives.py
```diff
@@ -85,18 +85,23 @@ def _qr_quantile_loss_derivative(
...
+    right_strength = min(1.0, (1-quantile) / quantile)
+    
+    forces = np.where(
+        forces < 0,
+        forces * left_strength,
+        forces * right_strength,
+    )
 
-    return regularization + quantile_term + forces.sum()
+    return regularization + forces.sum()
```
amatissart, line 104: @lenhoanglnh This change seems to change significantly the behaviour of the "zero shift" on current Tournesol data. Is it expected? Should we adjust the quantile parameter?

On "main", after applying the shift with `score_shift_quantile = 0.15`, about 13% of the individual scores are negative. On this branch "neurips24", that would be 37%. 

As a consequence the distribution of Tournesol would be modified, with fewer videos reaching the recommendability threshold (1238 instead of 3013).

(I used the "legacy2023" pipeline, currently deployed on production. But I expect it would similar with the new pipeline).

image

lenhoanglnh: This is unsatisfactory indeed. 
I'm a bit disturbed. It feels like the quantile is now poorly estimated.
Maybe this is because videos with lower scores have higher uncertainty? Or less trust?

lenhoanglnh: OK I looked at the data and indeed, the uncertainties for bad videos are smaller than for good videos, which explains why the quantile increased with the new quantile definition. I see two simple fixes:
- Reduce ```score_shift_quantile = 0.15``` to ```score_shift_quantile = 0.05```.
- Remove uncertainties in quantile estimation.
The former is much more satisfactory.

amatissart (approved):

PR #1973, opened 2024-05-18
[solidago] gbt: estimate asymmetrical uncertainties based on increase of loss by 1

amatissart:
Based on #1970 

I struggled with the sign conventions, but I think I got something that works as expected.

TODO:

- [x] review the definition
@lenhoanglnh the paper suggests to only consider the negative log-likelihood term to estimate the uncertainties. Don't we need to consider the regularization term too? That what is done on this branch, because I observed very high values when it was not present.

- [x] the test data need to be updated with new uncertainties, after some sanity checks on the actual values

- [x] adapt the L-BFGS implementation to use the new uncertainties too (or split the tests)

### Checklist


lenhoanglnh:
>  @lenhoanglnh the paper suggests to only consider the negative log-likelihood term to estimate the uncertainties. Don't we need to consider the regularization term too? That what is done on this branch, because I observed very high values when it was not present.

This is intended. One interesting implications of this is that if a user says A is maximally better then B, then the comparison will yield an infinite right uncertainty on A, and an infinite uncertainty on B. Does this break something?

If the uncertainty is too large (perhaps a feature rather than a bug in principle), the value + 1 in the equation may be changed to a smaller value.

amatissart:
> This is intended. One interesting implications of this is that if a user says A is maximally better then B, then the comparison will yield an infinite right uncertainty on A, and an infinite uncertainty on B. Does this break something?
> 
> If the uncertainty is too large (perhaps a feature rather than a bug in principle), the value + 1 in the equation may be changed to a smaller value.

Ok :+1:  I pushed the modification in bf95cbb.

It seems to work. I was just a bit surprised to see such a big difference compared to the current expected values in the tests files. For example in "data_3.py":

```
Index | Obtained          | Expected                     
1     | 6.156474787981032 | 0.03096002542681708 ± 1.0e-01
2     | 700.7408689591132 | 0.03096002542681708 ± 1.0e-01
```

amatissart:
@lenhoanglnh The uncertainty values close to 700 were actually due to a numerical issue. In practice, there are cases where the log-likelihood term never reaches the threshold $\mathcal L + 1$. So I adapted the implementation to use $10^3$ as the default uncertainty. (I fear storing in the databse $+\infty$ would create other problems).

solidago/src/solidago/solvers/dichotomy.py
```diff
@@ -29,7 +30,7 @@ def solve(
     -------
     out: float
     """
-    ymin, ymax = f(xmin) - value, f(xmax) - value
+    ymin, ymax = f(xmin, *args) - value, f(xmax, *args) - value
```
lfaucon, line 33: minor: Another way to do something similar would be to not change solve, but use it with a `partial` https://docs.python.org/3/library/functools.html#functools.partial

solidago/src/solidago/preference_learning/generalized_bradley_terry.py
```diff
@@ -46,23 +47,25 @@ def cumulant_generating_function_derivative(self) -> Callable[[npt.NDArray], npt
...
-        Returns
-        -------
-        out: float
+    def log_likelihood_function(self) -> Callable[[npt.NDArray, npt.NDArray], float]:
+        """The loss function definition is used only to compute uncertainties.
         """
-        pass
+
+    @cached_property
+    def loss_increase_to_solve(self):
```
lfaucon, line 58: minor: Naming `translated_negative_log_likelihood` (what it is, not what it is meant to be used for) + it's a log likelihood, not a loss

solidago/src/solidago/preference_learning/generalized_bradley_terry.py
```diff
@@ -46,23 +47,25 @@ def cumulant_generating_function_derivative(self) -> Callable[[npt.NDArray], npt
...
+        """The root of this function is used to compute asymetric uncertainties
+        by looking for the delta for which the loss is increased by 1.
+        """
```
lfaucon, line 61:
```suggestion
        """
        This function is a convex negative log likelihood, translated such 
        that its minimum has a constant negative value at `delta=0`. The 
        roots of this function are used to compute the uncertainties 
        intervals. If it has only a single root, then uncertainty on the 
        other side is considered infinite.
        """
```

lfaucon (approved):

PR #1996, opened 2024-07-22
[back] fix: validate score and score_max in ComparisonCriteriaScore form

GresilleSiffle:
### Description

The back end doesn't crash anymore when submitting/editing a `ComparisonCriteriaScore` from the admin interface, with a `score_max` less than `score`.

We should not use the admin interface to edit `ComparisonCriteriaScore` but it is useful for testing and debugging.

### Checklist

### DIFF
--- a/backend/tournesol/models/comparisons.py
+++ b/backend/tournesol/models/comparisons.py
@@ -5,6 +5,7 @@ Models for Tournesol's main functions related to contributor's comparisons
 import uuid
 
 import computed_property
+from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator
 from django.db import models
 from django.db.models import F, ObjectDoesNotExist, Q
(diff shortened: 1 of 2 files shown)


backend/tournesol/models/comparisons.py
```diff
@@ -159,4 +160,13 @@ def save(self, *args, **kwargs):
...
+
+    def clean(self):
+        try:
+            self._validate_score_max()
+        except (TypeError, ValueError) as err:
+            raise ValidationError({"score_max": err.args[0]}) from err
+
+    def save(self, *args, **kwargs):
+        self._validate_score_max()
         return super().save(*args, **kwargs)
```
amatissart, line 172: Maybe something we missed in https://github.com/tournesol-app/tournesol/pull/1947, it does not seem like the case when an invalid score/score_max is tested in "backend/tournesol/tests/test_api_comparison.py". As it is now, it may return a 500 response, I think.

amatissart (approved):

backend/tournesol/models/comparisons.py
GresilleSiffle: Indeed, I created an issue to fix this: https://github.com/tournesol-app/tournesol/issues/2001

PR #2059, opened 2025-02-11
Import YouTube history into the rate later list

sigmike:
### Description

Initial draft to import your watch history into your rate later list.

The bulk import is not done yet, so the import is very slow, but it works.

Remaining to do:

- [x] bulk import
- [x] improve the style of the status box while importing
- [x] translate texts
- [x] improve behavior when the user is not connected (currently it marks all imports as failed)

### DIFF
--- a/backend/tournesol/urls.py
+++ b/backend/tournesol/urls.py
@@ -114,7 +114,7 @@ urlpatterns = [
         name="usersme_ratelater_list",
     ),
     path(
-        "users/me/rate_later/<str:poll_name>/bulk_create",
+        "users/me/rate_later/<str:poll_name>/_bulk_create",
         RateLaterBulkCreate.as_view(),
         name="usersme_ratelater_bulk_create",
     ),
--- a/backend/tournesol/urls.py
+++ b/backend/tournesol/urls.py
@@ -38,7 +38,7 @@ from .views.previews import (
 )
 from .views.previews.recommendations import get_preview_recommendations_redirect_params
 from .views.proof import ProofView
-from .views.rate_later import RateLaterDetail, RateLaterList
+from .views.rate_later import RateLaterBulkCreate, RateLaterDetail, RateLaterList
 from .views.ratings import (
     ContributorRatingDetail,
     ContributorRatingList,
@@ -112,6 +112,11 @@ urlpatterns = [
         "users/me/rate_later/<str:poll_name>/",
         RateLaterList.as_view(),
         name="usersme_ratelater_list",
+    ),
+    path(
+        "users/me/rate_later/<str:poll_name>/bulk_create",
+        RateLaterBulkCreate.as_view(),
+        name="usersme_ratelater_bulk_create",
     ),
     path(
         "users/me/rate_later/<str:poll_name>/<str:uid>/",
--- a/backend/tournesol/views/rate_later.py
+++ b/backend/tournesol/views/rate_later.py
@@ -83,6 +83,7 @@ class RateLaterBulkCreate(RateLaterQuerysetMixin, generics.CreateAPIView):
 
     pagination_class = None
     permission_classes = [IsAuthenticated]
+    throttle_scope = "api_rate_later_bulk_create"
     serializer_class = RateLaterSerializer
 
     def get_serializer(self, *args, **kwargs):
--- a/backend/tournesol/views/rate_later.py
+++ b/backend/tournesol/views/rate_later.py
@@ -71,7 +71,7 @@ class RateLaterList(RateLaterQuerysetMixin, generics.ListCreateAPIView):
         " for a given poll.",
         request=RateLaterSerializer(many=True),
         responses={
-            201: RateLaterSerializer,
+            201: RateLaterSerializer(many=True),
         },
     ),
 )
@@ -81,6 +81,7 @@ class RateLaterBulkCreate(RateLaterQuerysetMixin, generics.CreateAPIView):
     Accepts an array of entities to be added to the rate-later list.
     """
 
+    pagination_class = None
     permission_classes = [IsAuthenticated]
     serializer_class = RateLaterSerializer
 
--- a/backend/tournesol/views/rate_later.py
+++ b/backend/tournesol/views/rate_later.py
@@ -9,6 +9,7 @@ from rest_framework.permissions import IsAuthenticated
 
 from tournesol.models import Entity, RateLater
 from tournesol.serializers.rate_later import RateLaterSerializer
+from tournesol.utils.constants import RATE_LATER_BULK_MAX_SIZE
 from tournesol.views.mixins.poll import PollScopedViewMixin
 
 
@@ -85,6 +86,7 @@ class RateLaterBulkCreate(RateLaterQuerysetMixin, generics.CreateAPIView):
 
     def get_serializer(self, *args, **kwargs):
         kwargs["many"] = True
+        kwargs["max_length"] = RATE_LATER_BULK_MAX_SIZE
         return super().get_serializer(*args, **kwargs)
 
     def perform_create(self, serializer):
--- a/backend/tournesol/views/rate_later.py
+++ b/backend/tournesol/views/rate_later.py
@@ -65,6 +65,35 @@ class RateLaterList(RateLaterQuerysetMixin, generics.ListCreateAPIView):
 
 
 @extend_schema_view(
+    post=extend_schema(
+        description="Add a multiple new entities to the logged user's rate-later list,"
+        " for a given poll.",
+        request=RateLaterSerializer(many=True),
+        responses={
+            201: RateLaterSerializer,
+        },
+    ),
+)
+class RateLaterBulkCreate(RateLaterQuerysetMixin, generics.CreateAPIView):
+    """
+    Create multiple rate-later entries at once.
+    Accepts an array of entities to be added to the rate-later list.
+    """
+
+    permission_classes = [IsAuthenticated]
+    serializer_class = RateLaterSerializer
+
+    def get_serializer(self, *args, **kwargs):
+        kwargs["many"] = True
+        return super().get_serializer(*args, **kwargs)
+
+    def perform_create(self, serializer):
+        rate_later_instances = serializer.save()
+        for rate_later in rate_later_instances:
+            self.prefetch_entity(rate_later)
+
+
+@extend_schema_view(
     get=extend_schema(description="Get an entity from the logged user's rate-later list."),
     delete=extend_schema(description="Delete an entity from the logged user's rate-later list."),
 )
--- a/browser-extension/src/background.js
+++ b/browser-extension/src/background.js
@@ -1,5 +1,6 @@
 import {
   addRateLater,
+  addRateLaterBulk,
   alertOnCurrentTab,
   alertUseOnLinkToYoutube,
   fetchTournesolApi,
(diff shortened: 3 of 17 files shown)


sigmike:
Also, I added a button to the extension dropdown but it only opens a tab to the watch history, so I'm not sure it's useful.

GresilleSiffle:
I'm testing the current version. It works well thank you :+1::+1:

Here are few things that could improve the user experience. 

**(1)** Do not allow the users to click on the import button if they are not connected to Tournesol. For non-authenticated users, all import attempts will fail. It looks like the process is simply not working : /

Suggestions 1:
- rename the button Import history to Tournesol (maybe it's more clear than the current title?)
- for non-authenticated users this button could open the login modal

**(2)** Add an indication that the process in ongoing

Suggestions:
- a simple title could be enough for now
- later we could add an animation

Exemple with a random title (could be "Processing, please wait..." or similar)
capture

**(3)** Make the overlay a bit bigger, to clearly show that something is going on. I hope the users will understand it's not a dialogue box expecting an action like confirm/cancel, but more a control panel.

> Also, I added a button to the extension dropdown but it only opens a tab to the watch history, so I'm not sure it's useful.

I think it can be good to display the button, it shows that the feature exists. Maybe we could rename it "Import history?".

GresilleSiffle:
The videos that are already in the rate-later list appear as successfully imported, sounds good to me.

GresilleSiffle:
> The bulk import is not done yet, so the import is very slow, but it works.

Should we add the bulk import before releasing a first version of this feature? What do you think @amatissart ?

Without the bulk import, the API allows authenticated users to make 120 requests per minute. This is very tight for an import process. If we decide to release a first version without the bulk import, we may want to manually control the pace of the import.

sigmike:
> rename the button "Import history to Tournesol"

OK

> for non-authenticated users this button could open the login modal

OK I'll try that.

> (2) Add an indication showing that the process in ongoing.
> (3) Make the overlay a bit bigger, to clearly show that something is going on.

The title is indeed needed. A loading animation would really convey that something is happening. Maybe one of the CC0 pure CSS animations at https://loading.io/css/?

Or maybe even a description of the process and a "start" button?

> Should we add the bulk import before releasing a first version of this feature?

It doesn't seem very difficult to add the bulk import, so it's probably better to wait for it.

I think I'll make the current API also accept an array of IDs as input. But if you'd rather have a new entry point, tell me.

I don't know how heavy this process is to the backend. If it is, we may want a custom rate limit for this specific call (and that may warrant a specific entry point). The current version of the feature will do a call 1 second after the last request completed.

sigmike:
I made all the changes. It should work well, with a bulk import.

I ended up using a specific API endpoint, but it can probably be merged into the existing create endpoint.

GresilleSiffle:
Thank you very much, I'll have a look tomorrow.

backend/tournesol/urls.py
```diff
@@ -113,6 +113,11 @@
         RateLaterList.as_view(),
         name="usersme_ratelater_list",
     ),
+    path(
+        "users/me/rate_later/<str:poll_name>/bulk_create",
```
amatissart, line 117: To disambiguate more clearly from URLs that contain an entity uid, I would suggest to prefix actions with `_`. This is also more consistent with the existing `"users/me/contributor_ratings/<str:poll_name>/_all/"`

```suggestion
        "users/me/rate_later/<str:poll_name>/_bulk_create",
```

backend/tournesol/views/rate_later.py
```diff
@@ -64,6 +64,35 @@ def perform_create(self, serializer):
...
+    Create multiple rate-later entries at once.
+    Accepts an array of entities to be added to the rate-later list.
+    """
+
+    permission_classes = [IsAuthenticated]
+    serializer_class = RateLaterSerializer
+
+    def get_serializer(self, *args, **kwargs):
+        kwargs["many"] = True
+        return super().get_serializer(*args, **kwargs)
```
amatissart, line 88: A limit on the size of the bulk needs to be defined. This is to avoid timeouts, and abuse of the Youtube API (as the bulk requests could be exploited to get around the existing rate limits). 
Not sure about what a reasonable value would be? 20 ?

I suppose the extension may also need to be adapted to send batches with appropriate size.

```suggestion
    def get_serializer(self, *args, **kwargs):
        kwargs["many"] = True
        kwargs["max_length"] = RATE_LATER_BULK_MAX_SIZE
        return super().get_serializer(*args, **kwargs)
```

GresilleSiffle:
Hello @sigmike 

I created a new PR #2063 that targets this branch to add the following features:
- stop the regular calls to `captureStep` when the end of the history is reached (your feedback is welcome)
- display a "Import complete!" message to the users when the import is complete

> The title is indeed needed. A loading animation would really convey that something is happening. Maybe one of the CC0 pure CSS animations at https://loading.io/css/?

The animation you chose is great.

> Or maybe even a description of the process and a "start" button?

I think it's not necessary because the users already need to click on the Import button to trigger the import. One click seems enough to me.

> I think I'll make the current API also accept an array of IDs as input. But if you'd rather have a new entry point, tell me.

Adding a new entry point is good for now. 

As @amatissart said here https://github.com/tournesol-app/tournesol/pull/2059#discussion_r1973063506 and there https://github.com/tournesol-app/tournesol/pull/2059#discussion_r1973071121 we should limit the number of videos that can be added at the same time by the API. This will make the process more robust by avoiding timeout, and also will give us more control on how we manage the calls to the YouTube API.

20 videos per batch sounds good to me.

sigmike:
Ok, thanks for the suggestions. I'll implement that.

backend/tournesol/views/rate_later.py
sigmike: Done in 2bf25b55a159ce158490d7f08912aa84d044dc0e.

backend/tournesol/urls.py
sigmike: Done in b320048b1053c59713e73de460bd7059dd87b520.

backend/tournesol/views/rate_later.py
```diff
@@ -64,6 +65,36 @@ def perform_create(self, serializer):
         self.prefetch_entity(rate_later)
 
 
+@extend_schema_view(
+    post=extend_schema(
+        description="Add a multiple new entities to the logged user's rate-later list,"
+        " for a given poll.",
+        request=RateLaterSerializer(many=True),
+        responses={
+            201: RateLaterSerializer,
```
GresilleSiffle, line 74:
```suggestion
            201: RateLaterSerializer(many=True),
```

---

The API returns a list of `RateLater` serialized instances, but with this schema definition the front end will expect a single  `RateLater` element. This is not a problem for the web browser extension, but the generated service for the front end will be wrong.

Here is the currently generated service in the front end (file `src/services/openapi/services/UserService.ts`):

```typescript
    public static usersMeRateLaterBulkCreateCreate({
        pollName,
        requestBody,
    }: {
        pollName: string,
        requestBody: Array<RateLaterRequest>,
    }): CancelablePromise<RateLater> {
        return __request(OpenAPI, {
            method: 'POST',
            url: '/users/me/rate_later/{poll_name}/_bulk_create',
            path: {
                'poll_name': pollName,
            },
            body: requestBody,
            mediaType: 'application/json',
        });
    }
``` 

We expect the return type to be `CancelablePromise<Array<RateLater>>` or eventually `CancelablePromise<PaginatedRateLaterList>`.

backend/tournesol/views/rate_later.py
```diff
@@ -64,6 +65,36 @@ def perform_create(self, serializer):
...
+        },
+    ),
+)
+class RateLaterBulkCreate(RateLaterQuerysetMixin, generics.CreateAPIView):
+    """
+    Create multiple rate-later entries at once.
+    Accepts an array of entities to be added to the rate-later list.
+    """
+
+    permission_classes = [IsAuthenticated]
```
GresilleSiffle, line 84: This following suggestion will disable the `limit` and `offset` query parameters automatically added by DRF.

```suggestion
    pagination_class = None
    permission_classes = [IsAuthenticated]
```

GresilleSiffle:
Minus the two previous comments, the PR looks good to me, thanks a lot @sigmike 

**(1) about the YouTube API calls**

@amatissart before merging should we consider defining a new strategy to limit the number of YouTube API generated by the history imports? 

Random ideas:
- completely prevent YouTube API calls when using "bulk" routes of our back end (for now); the metadata of an entity will be refreshed after its first comparison
- allow a few calls per user per day on our "bulk" routes; this will postpone the problem, but the problem will still be there
- do nothing and consume as much quota as possible per day (we may need to check the back end first to be sure that a depleted quota will not prevent new videos to be added)

In the future we may want to consider using a task queue, to spread the API calls over the whole day while saving some credits for our daily routines like the playlist updates.

**(2) about users that will import several times their history**

It will possible for a user to import several times his/her history to Tournesol. It will also be possible that subsequent imports will re-add videos to the rate-later list that have been automatically removed after 4 comparisons. This will, for instance, make the suggestions made by the Auto button less relevant, because it could suggest videos the users don't want to compare.

In my opinion it's not a blocking point to merge this PR, but it could be useful to have a solution for that.

Ideas:
- by default do not add to the rate-later list videos that have been compared more that X times (X being the setting `rate_later__auto_remove`).
- by default import everything but allow the users to "clean" their rate-later lists by removing all videos that have been compared more than X times
- make the import process remember the range of imported videos, to not imported them again (this looks tricky, especially to define what is a "range", two dates?, two UIDs?)
- do nothing

I have a preference for the solution "by default import everything but allow the users to "clean" their rate-later..."

backend/tournesol/views/rate_later.py
sigmike: Good catch. Done in d4f78c4ce7fabd94e0ba97c651ac65cc9b40b6a6.

backend/tournesol/views/rate_later.py
sigmike: Done in d4f78c4ce7fabd94e0ba97c651ac65cc9b40b6a6.

GresilleSiffle:
We created a branch to address the point **(1)** of https://github.com/tournesol-app/tournesol/pull/2059#issuecomment-2703524894.

Regarding the point **(2)** we haven't fully decided yet.

browser-extension/src/background.js
```diff
@@ -115,6 +122,11 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
     return true;
   }
 
+  if (request.message == 'addRateLaterBulk') {
+    addRateLaterBulk(request.videoIds).then(sendResponse).catch(sendResponse);
```
amatissart, line 126: Using `sendResponse()` to send an Error object does not behave as expected on Chrome.
Contrary to Firefox, Chrome uses JSON serialization when passing messages, and the `Error` object is lost.
See [Data Cloning Algorithm](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Chrome_incompatibilities#data_cloning_algorithm) on MDN

browser-extension/src/utils.js
```diff
@@ -89,6 +89,25 @@ export const addRateLater = async (video_id) => {
   };
 };
 
+export const addRateLaterBulk = async (videoIds) => {
+  const ratingStatusReponse = await fetchTournesolApi(
+    'users/me/rate_later/videos/_bulk_create',
```
amatissart, line 94: Add trailing slash to avoid unnecessary redirection:
```suggestion
    'users/me/rate_later/videos/_bulk_create/',
```

browser-extension/src/browserAction/menu.js
```diff
@@ -108,6 +108,12 @@ function openOptionsPage() {
   chrome.runtime.openOptionsPage();
 }
 
+function rateLaterHistory() {
+  browser.tabs.create({
```
amatissart, line 112: Chrome does not support the `browser` namespace, whereas Firefox supports `chrome` :upside_down_face: 
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API#browser_api_differences

```suggestion
  chrome.tabs.create({
```

GresilleSiffle:
Let's not forget this comment https://github.com/tournesol-app/tournesol/pull/2068#discussion_r2016342373

I'll have time tomorrow.

browser-extension/src/background.js
amatissart: I opened https://github.com/tournesol-app/tournesol/pull/2069 to fix this.

GresilleSiffle (approved):

PR #2063, opened 2025-02-27
[ext] feat: display a feedback to the user when the end of history is reached

GresilleSiffle:
target PR #2059

---

### Description

Display a visual feedback to the users when the import is complete.  

capture

### Checklist

### DIFF
--- a/browser-extension/src/rateLaterHistory.js
+++ b/browser-extension/src/rateLaterHistory.js
@@ -35,6 +35,13 @@ const addOverlay = () => {
   loader.classList.add('lds-dual-ring');
   statusBox.append(loader);
 
+  const importComplete = () => {
+    loader.classList.remove('lds-dual-ring');
+    title.textContent = chrome.i18n.getMessage(
+      'rateLaterHistoryStatusBoxTitleFinish'
+    );
+  };
+
   const addCounter = ({ label, initialValue = 0 }) => {
     const container = document.createElement('div');
 
@@ -91,6 +98,7 @@ const addOverlay = () => {
     displayHistoryVideoCount,
     displaySentVideoCount,
     displayFailedVideoCount,
+    importComplete,
     stopButton,
   };
 };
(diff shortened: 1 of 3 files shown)


browser-extension/src/rateLaterHistory.js
```diff
@@ -162,12 +178,13 @@ const startHistoryCapture = async () =>
...
       displaySentVideoCount,
       displayFailedVideoCount,
+      importComplete,
       stopButton,
     } = addOverlay();
 
     let timeout;
 
-    const captureStep = async () => {
+    const captureStep = async (onEndOfHistory) => {
```
sigmike, line 187: Why did you choose to pass this as argument? This function already uses several variables defined in its parent scope.

sigmike (commented):
Thanks for this PR. I'm a little concerned about the way we detect the end of the loading process. If there's a network lag or anything slowing down the loading for some time, we will assume the loading is complete. It won't stop the process though, so the user will still see more videos loading.

My YouTube watch history is too big to reach the end in a reasonable time frame, so I can't really test it.

If it works for you and you don't mind having some false completions detected, I'm fine with merging this.

GresilleSiffle:
> I'm a little concerned about the way we detect the end of the loading process. If there's a network lag or anything slowing down the loading for some time, we will assume the loading is complete. 

You're right. The check is not robust enough, I didn't think about possible network slowdowns sorry.

> If it works for you and you don't mind having some false completions detected, I'm fine with merging this

I'd prefer to avoid introducing erroneous user feedback on this feature. We can totally close this PR.

My initial goal was to add visual cues, to invite the users to click on Stop when the numbers of visible and imported videos are equal. Maybe a small improvement could be to display those number in bold, to make them sand out from the rest of the text. What do you think?

sigmike:
> My initial goal was to add visual cues, to invite the users to click on Stop when the numbers of visible and imported videos are equal. Maybe a small improvement could be to display those number in bold, to make them sand out from the rest of the text. What do you think?

Sounds good. I made the change in the main PR.

If you want to let the user know that nothing has changed for a while, we could display the elapsed time since the last change of the video count. But it may require a small explanation text, so I'm not sure it's actually easier for the user.

GresilleSiffle:
> But it may require a small explanation text, so I'm not sure it's actually easier for the user.

You're right. An extra explanation would make the UI more "heavy" I think. Let's keep it as is for now.

PR #2064, opened 2025-03-06
[back][ext] feat: add rate limit to the bulk create route of the rate-later API

GresilleSiffle:
**target PR**: #2059

---

### Description

- limit the call to the rate-later bulk create API to 20 requests/day
- handle the HTTP 429 status in the extension
- display the number of new videos imported and the number of skipped videos

### Checklist

### DIFF
--- a/backend/settings/settings.py
+++ b/backend/settings/settings.py
@@ -313,6 +313,7 @@ REST_FRAMEWORK = {
         "api_video_post": "50/min",
         "api_users_me_export": "10/min",
         "api_export_comparisons": "10/min",
+        "api_rate_later_bulk_create": "20/day",
         # global throttle on account registrations and password reset
         "email": server_settings.get("THROTTLE_EMAIL_GLOBAL", "15/min"),
     },
--- a/backend/tournesol/views/rate_later.py
+++ b/backend/tournesol/views/rate_later.py
@@ -82,6 +82,7 @@ class RateLaterBulkCreate(RateLaterQuerysetMixin, generics.CreateAPIView):
     """
 
     permission_classes = [IsAuthenticated]
+    throttle_scope = "api_rate_later_bulk_create"
     serializer_class = RateLaterSerializer
 
     def get_serializer(self, *args, **kwargs):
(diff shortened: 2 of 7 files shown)


amatissart (approved):

GresilleSiffle:
An idea for the future, it could be nice to replace the loading animation by a text (or something else) when the user reaches the maximum number of allowed calls to the API. Currently, when the animation disappears, the import status box becomes less visible, less eye catching than when it was loading.

capture

sigmike:
Looks good to me. I have a few remarks:

1. I don't know where the 20/day limit comes from, but couldn't we instead have a limit on a shorter time frame and make the extension wait a bit before it reaches the limit? This would allow users to keep this tab open if they want to import a large history (if it makes sense to do that). But it only makes sense if we can increase the average limit, because waiting 50 minutes doesn't seem reasonable.
2. I agree that the changes happening when the limit is reached are barely noticeable. Maybe we could instead replace the texts and loader with something that looks like a Material UI warning Alert. The user would notice there's something new. And Maybe adding a fixed title to the box would make it look better.
3. There may be too many numbers displayed. We can probably remove some.
4. It's not a good time to do it, but adding React support to the extension would be make this kind of interface much easier to do.

amatissart:
>  1. I don't know where the 20/day limit comes from, but couldn't we instead have a limit on a shorter time frame and make the extension wait a bit before it reaches the limit? This would allow users to keep this tab open if they want to import a large history (if it makes sense to do that). But it only makes sense if we can increase the average limit, because waiting 50 minutes doesn't seem reasonable.

Our quota of requests to the YouTube API is limited daily. So we need a safeguard to avoid the platform being blocked from using the YouTube API when users try to import a large number of videos.

One more improvement would be to store the ids of imported videos in the extension storage, so that they can be ignored on the next run, to let older videos in the history to be processed too. I think @GresilleSiffle plan to work on this tomorrow.

>  4. It's not a good time to do it, but adding React support to the extension would be make this kind of interface much easier to do.

Yes we definitely see the limits of a "Vanilla" js approach on this kind of features. We were initially reluctant to add a framework that would require a separate build environment from the main "dev-env" to develop the extension. It may be easier now that "configure" scripts have been defined. A possible approach would also be to use "preact" which allows no-build workflows: https://preactjs.com/guide/v10/no-build-workflows

GresilleSiffle:
I'm going to merge this branch in `rate-later-history`, and add a system to ignore previously imported videos. With see the vast majority of users will not be blocked by the daily limit.

PR #2068, opened 2025-03-21
[ext] feat: keep imported videos in cache to not import them again

GresilleSiffle:
### Description

The history imports made by the browser extension now remembers the YouTube id of imported videos. This allows the users to import new batch of their history daily, without wasting their limited Tournesol API requests by trying to import already imported videos.

To simplify the UI, and to hide the implementation details, the users will see a single number of successfully imported videos, named `imported videos:`. This number was previously split in two distinct labels `new videos` and `already in your list`. A value of 10 means 10 videos from the history are now in your rate later list (some were already present, some are new, but in the end all videos are in the rate-later list).

### Preview

During the import 👇️

cap0

When the Tournesol API returns a HTTP 429 response 👇️

capture

#### Questions

Do we want to allow the users to clear their import cache, to allow them to import a second time their history? 

### Checklist

### DIFF
--- a/browser-extension/src/rateLaterHistory.js
+++ b/browser-extension/src/rateLaterHistory.js
@@ -217,9 +217,6 @@ const saveImportedToLocalStorage = async (importedVideosSet) => {
     videosStr = videosStr.substring(1);
   }
 
-  console.log('found ', historyInStorage);
-  console.log('new ', videosStr);
-
   await chrome.storage.local.set({
     youtubeHistoryImported: historyInStorage
       ? historyInStorage + videosStr
@@ -346,7 +343,7 @@ const startHistoryCapture = async () =>
       try {
         newVideosToSend = await skipPreviouslyImported(videosToSend);
       } catch (error) {
-        console.log(error);
+        console.error(error);
         newVideosToSend = videosToSend;
       }
 
--- a/browser-extension/src/rateLaterHistory.js
+++ b/browser-extension/src/rateLaterHistory.js
@@ -57,10 +57,8 @@ const addOverlay = () => {
     loader.classList.add('display-none');
     loader.classList.remove('lds-dual-ring');
     alertBox.classList.remove('display-none');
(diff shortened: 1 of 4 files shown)


sigmike:
Looks good, works well.

I think the text "You'll be able to start a new import in 24 hours, after having watched new videos" would be better positioned inside the "alert-box". It also probably needs updating (EN and FR) since it's now also valid to start it again the next day even if you haven't watched any new video. Maybe something like "You'll be able to start a new import in 24 hours to continue importing your previous history, or to import any videos that you've watched in the meantime."

The English title of the status box "Videos are being imported, please wait" needs to be updated as it remains there when the import is stopped.

> Do we want to allow the users to clear their import cache, to allow them to import a second time their history?

The user can remove and reinstall the extension to clear the local storage. It's probably good enough.

browser-extension/src/rateLaterHistory.js
```diff
@@ -178,6 +186,42 @@ const addRateLaterBulk = async (videoIds) =>
...
+
+const saveImportedToLocalStorage = async (importedVideosSet) => {
+  const localStorage = await chrome.storage.local.get('youtubeHistoryImported');
+  const historyInStorage = localStorage?.youtubeHistoryImported;
+
+  let videosStr = '';
+  for (const video of importedVideosSet) {
+    if (!historyInStorage?.includes(video)) {
+      videosStr += `,${video}`;
+    }
```
amatissart, line 211: Loading `youtubeHistoryImported` in a `Set` could perform better, instead of using `String.includes()`. I suppose that is not issue even for thousands of videos?

amatissart (approved):

browser-extension/src/rateLaterHistory.js
GresilleSiffle: Yes I thought about it. I tried with a string of 10,000 videos and the performance were good enough, but it's probably better to use a `Set`.

PR #2078, opened 2025-04-18
Upgrade mui to v6

GresilleSiffle:
### Description

Upgrade Material UI to version 6.

### Checklist

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,9 +6,9 @@
   "dependencies": {
     "@emotion/react": "^11.7.1",
     "@emotion/styled": "^11.6.0",
-    "@mui/icons-material": "^5.14.3",
-    "@mui/material": "^5.14.5",
-    "@mui/styles": "^5.14.5",
+    "@mui/icons-material": "5.15.0",
+    "@mui/material": "5.15.0",
+    "@mui/styles": "5.15.0",
     "@react-hook/resize-observer": "^1.2.5",
     "@reduxjs/toolkit": "^1.7.1",
     "@types/react": "^17",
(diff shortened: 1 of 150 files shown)


amatissart (approved):
Great :+1: Works without issue on my side.

2 small observations that we may want to address at some point:

- A fresh `yarn install` shows this warning:
    ```
    warning " > notistack@2.0.8" has incorrect peer dependency "@mui/material@^5.0.0".
    ```
  It seems **notistack** has a v3 which is no longer coupled with MUI. Perhaps we should use that, or something else?

- Unit tests are a bit slower.   
After some debugging, it seems vitest wastes quite a lot of time to process all icons in `@mui/icons-material`. That's because some import statements are written like `import { ArrowBack } from '@mui/icons-material'` which requires Vite and Vitest to process the entire package with all icons. That also affects the dev-env where packages are loaded dynamically without optimization (for hot replacement, etc). It does not affect production where javascript bundles are optimized on build. Not sure why it appears now, are there more icons in MUIv6?
In practice, except ~30s more in unit tests, it's not very noticeable. So we can live with that, but we may want to add a linter rule to enforce importing icons individually, like  `import ArrowBack from '@mui/icons-material/ArrowBack'`

GresilleSiffle:
Thanks for the review :+1: 

> It seems notistack has a v3 which is no longer coupled with MUI. Perhaps we should use that, or something else?

Exactly, I was thinking of upgrading `notistack` just after this pull request.

> Unit tests are a bit slower.

Good catch, I didn't see that. I'd be ok with enforcing the syntax `import ArrowBack from ...` if it speeds up the development.

PR #2087, opened 2025-05-23
[front] chore: prepare upgrade of react-router-dom

GresilleSiffle:
### Description

The upgrade of React from version 17 to 18 also requires upgrading several dependencies. I'm starting with `react-router-dom`. I'm not sure the our current version of `react-router-dom` is incompatible with React 18, but its version looks old enough to upgrade it.

This branch is a refactor to prepare the upgrade of `react-router-dom` from v5 to v6. If later we upgrade React to 18, we will be able to also upgrade `react-router-dom` to version 7.

### Checklist

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { Switch, Redirect, useLocation } from 'react-router-dom';
+import { Switch, Redirect, useLocation, Route } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import { i18n as i18nInterface } from 'i18next';
 
@@ -17,6 +17,7 @@ import Frame from './features/frame/Frame';
 import { StatsLazyProvider } from './features/statistics/StatsContext';
 import PublicRoute from './features/login/PublicRoute';
 import PrivateRoute from './features/login/PrivateRoute';
+import PublicRouteWrapper from './features/login/PublicRouteWrapper';
 
 import ActionsPage from './pages/actions/ActionsPage';
 import ForgotPassword from './pages/login/ForgotPassword';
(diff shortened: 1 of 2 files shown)


GresilleSiffle:
Hi @amatissart 

I was not able to determine why the test "describe:Setting - exclude compared entities / it:handles the value false (include)" is constantly failing on GitHub. Curiously the tests pass correctly on my local environment when I run them headed or headless.

The tests pass on others branches, so I think the problem is specific to this branch.

This API tests pass, so the "exclude already compared" feature is working, at least in the API. The problem could be related to the test itself, but if so, why it doesn't fail in others branches? Could it be related to the packages upgrade? If yes, this doesn't explain why I cannot see the problem locally. There is something I'm missing.

frontend/src/app/PollRoutes.tsx
```diff
@@ -1,11 +1,10 @@
 import React, { useEffect } from 'react';
-import { Switch, useRouteMatch } from 'react-router-dom';
+import { Route, Routes } from 'react-router-dom';
 
 import { Box, CircularProgress } from '@mui/material';
 
 import { useCurrentPoll } from 'src/hooks/useCurrentPoll';
-import PublicRoute from 'src/features/login/PublicRoute';
-import PrivateRoute from 'src/features/login/PrivateRoute';
```
amatissart, line 8: I suppose `PublicRoute.tsx` and `PrivateRoute.tsx` could be deleted?

tests/cypress/e2e/frontend/settingsPreferencesPage.cy.ts
```diff
@@ -368,7 +368,9 @@ describe('Settings - preferences page', () => {
         });
       });
 
-      describe('Setting - exclude compared entities', () => {
+      // TOFIX: these tests fail nearly everytime. This could be related to
+      // the API cache, or a bad usage of Cypress + React.
```
amatissart, line 372: Possibly fixed by #2112

amatissart (approved):
Looks very good :+1:

Issue #2103, opened 2025-07-01
feat: Stats & gamification for video watched


lfaucon:
**Context:**

In order to make the platform more attractive, we want to highlight the positive aspect of watching the videos recommended by Tournesol. Gamifying how many videos users have watch could help for making the platform more attractive.

**Acceptance criteria:**

- System to measure whether a video is watched by a user or not
- Stats page highlighting count of video watched (as a primary statistic more prominent than other statistics)
- (optional) badge/mission-like system to reward users

GresilleSiffle:
To brainstorm, some examples of watched videos.

#### displayed with a badge

- pros:
- cons:
  - depending on the chosen symbol, the video looks "certified" or "more recommended than the others", this could be confusing
  - more symbols make the entity card harder to read

<img width="1916" height="873" alt="Image" src="https://github.com/user-attachments/assets/7d21bda8-c493-4bba-9034-cfde7e0eb890" />

#### displayed with a chip

- pros: the message is clear
- cons: it takes a bit of space

<img width="1919" height="876" alt="Image" src="https://github.com/user-attachments/assets/0e1ddf1e-bcd4-4cc9-b731-6ddc61e6afae" />

or

<img width="1919" height="872" alt="Image" src="https://github.com/user-attachments/assets/8ae3201a-47dc-473b-9a03-8ff34f3a978b" />

#### displayed with a different color

- pros:
  - no UI overload with symbols
  - it doesn't take space
  - it makes me want to watch everything 
- cons:
  - a colour without any symbol doesn't work for colour-blind people (but we could display a kind of watermark in the background, like the letter W as Watched) 

<img width="1919" height="876" alt="Image" src="https://github.com/user-attachments/assets/def92ef4-30ce-4189-ac39-ecf05f327756" />

amatissart:
### Decisions during brainstorm:

* button on the video card to mark/unmark video as watched
* different style/icon depending on state watched / non-watched
* in addition to the button, add a marker on all video cards to show video as watched (bottom border? badge?)
* extension: add right click "mark as watched"
* new page: watched videos

PR #2122, opened 2025-08-08
[back] feat: add notion of "seen/read/understood" for ContributorRating

GresilleSiffle:
**related issues** #2103 

---

### Description

This branch adds a new field `entity_seen` to the model `ContributorRating`. The value `True` means the entity has been seen (for a video), or read (for a document), or understood, etc.

After the merge, the value of this field will be set to `True` for all `ContributorRating`s referencing compared entities.

This field will be automatically updated by the API each time a comparison is made.

The ratings API will be updated in the next PR, to allow the users to manually set the value of  `entity_seen`.

Feedback is welcome : ) especially on the name of the field.

### Todo

model `ContributorRating`
- [x] add new field `entity_seen` with `help_text` attr and migration
- [x] update the admin interface, with a new column and a new filter
  -  the new field is read only in the edition form, to avoid unexpected manipulation from the ui

model `Comparison` and API
- [x] automatically mark compared videos as seen by the users 

existing data
- [x] add a new migration for the exiting data
   - mark all compared entities as seen by the users

tests
- [x] add new tests
- [x] update the existing tests

more
- [x] update the script `load_public_dataset`?
  - mark all compared entities as seen by the users

### Checklist


backend/tournesol/models/comparisons.py
```diff
@@ -110,6 +110,23 @@ def get_comparison(user, poll_id, uid_a, uid_b):
...
+            entity__in=[self.entity_1, self.entity_2],
+        )
+
+        for rating in ratings:
+            rating.entity_seen = True
+
+        ContributorRating.objects.bulk_update(
+            ratings,
+            ["entity_seen"],
+        )
```
amatissart, line 128: Can be simplified into a single query with:
```suggestion
        ContributorRating.objects.filter(
            poll=self.poll,
            user=self.user,
            entity__in=[self.entity_1, self.entity_2],
        ).update(entity_seen=True)
```

backend/tournesol/views/comparison.py
```diff
@@ -116,6 +116,8 @@ def perform_create(self, serializer):
         comparison.entity_2.inner.refresh_metadata()
         comparison.entity_2.auto_remove_from_rate_later(poll=poll, user=self.request.user)
 
+        comparison.mark_compared_entities_as_seen()
```
amatissart, line 119: Note that the signal `initialize_ratings_on_comparison_creation` in "signals.py" handles creating related `ContributorRating` objects if necessary, after a new comparison is saved. 
Your version is more explicit, but modifying the signal would ensure that `entity_seen` is set even when comparisons are created outside of this specific view. Not sure that's important.

amatissart (approved):
Great :+1:

backend/tournesol/views/comparison.py
GresilleSiffle: Should be fixed by https://github.com/tournesol-app/tournesol/pull/2122/commits/1e48977ce19ef8cb55b6dcc87875fed242db8ac6

PR #2139, opened 2025-10-10
[ext] fix: obsolete selectors breaking YT history import...

GresilleSiffle:
**related issues** #2134

---

### Description

The DOM of the YT history page has changed, and the current version of the browser extension cannot import the history to the rate-later list anymore.

This branch fixes that, but our code will probably break again in the future if YT keeps changing the import page.

### Checklist

- [x] I added the related issue(s) id in the related issues section (if any)
  - if not, delete the related issues section
- [x] I described my changes and my decisions in the PR description
- [x] I read the development guidelines of the [CONTRIBUTING.md][development-guidelines]
- [x] The tests pass and have been updated if relevant
- [x] The code quality check pass

[development-guidelines]: https://github.com/tournesol-app/tournesol/blob/main/CONTRIBUTING.md#development-guidelines

PR #2140, opened 2025-10-10
[ext] feat: import YT history mark entities as seen

GresilleSiffle:
**related issue**: #2103

This branch is based on `fix-import_yt_history`, the PR #2139 should be merged first.

---

### Description

This branch updates the `RateLaterBulkCreate` behaviour. When the URL query parameter `entity_seen=true` is provided, a contributor rating will be created for each submitted entity. If a contributor rating already exists, only the field `entity_seen` is updated.

Note that the field `is_public` will be set to true for new contributor ratings. This matches the behaviour of the comparison UI which also set `is_public` to true by default. There is a difference still: the comparison UI set explicitly the value true in the UI (but not in the database), while the import process set it implicitly to true in the database. I think it's important to remember that now, when we see a value `True` for an entity that has not been compared, we can't be sure it matches the user preferences. Ultimately, the users will always have the opportunity to change these values in the comparison UI before submitting their first comparison, the same way they can do it when they compared new videos. 

**todo**
- [x] use a bulk operation instead of a loop + update_or_create 
- [x] add new API tests
  - [x] `is_public` should be True by default, but not updated
  - [x] `entity_seen` should be True by default, and updated for existing contributor ratings

### Checklist

### DIFF
--- a/backend/tournesol/views/rate_later.py
+++ b/backend/tournesol/views/rate_later.py
@@ -82,7 +82,8 @@ class RateLaterList(RateLaterQuerysetMixin, generics.ListCreateAPIView):
         parameters=[
             OpenApiParameter(
                 name="entity_seen",
-                description='If "true", also mark the entities as seen/watched/read/understood by the user',
+                description='If "true", also mark the entities as '
+                "seen/watched/read/understood by the user",
                 required=False,
             ),
         ],


backend/tournesol/views/rate_later.py
```diff
@@ -93,6 +107,29 @@ def get_serializer(self, *args, **kwargs):
...
+            ContributorRating.objects.bulk_create(
+                contributor_ratings,
+                update_conflicts=True,
+                update_fields=["entity_seen"],
+                unique_fields=["poll_id", "user_id", "entity_id"],
+            )
+
+        # TOFIX: rate_later_instances seems to always be an empty list
         for rate_later in rate_later_instances:
             self.prefetch_entity(rate_later)
```
amatissart: Hum, how did you observe that  `rate_later_instances` is an empty list?

Not necessarily a topic for this PR, by maybe we could add a test to validate the number of SQL queries when submitting a batch, and the prefetch works as expected?

PR #2184, opened 2026-03-19
[front] feat: allow hiding the homepage intro+manifesto

venturqx:
### Issue
#2183

### Description

The homepage intro/manifesto is great and the visuals look nice.

However, it takes up all the vertical space on the mobile homepage. This means mobile users often have to scroll to reach the actual content.
Making it collapsible for users comes at no cost, since the manifesto remains easy to find via the hamburger menu anyway (only two clicks away).

### Screenshot - Mobile

| Before | After (not hidden) | After (hidden) |
|--------|--------|--------|
| old | new_phone_nothidden | new_phone_hidden  | 

### Screenshot - Desktop

| After (not hidden) | After (hidden) |
|--------|--------|
| new_desktop_nothidden | new_desktop_hidden | 

### Decisions
- Added a "Hide intro" button on the homepage.
- Added persistent user preference using "general.home__intro_hidden" (mirrored in a username-scoped `localStorage` cache key)
- Anonymous users still use the existing browser `localStorage` cache key.
- Updated the backend serializer, OpenAPI schema, tests, and translations to support the setting.

### Commits:
- [front] feat: allow hiding the homepage intro+manifesto

### Checklist

| yarn test | yarn lint |
|--------|--------|
| 2026-03-19-233520_hyprshot | 2026-03-19-233559_hyprshot |

### DIFF
--- a/frontend/src/features/settings/userSettingsSlice.ts
+++ b/frontend/src/features/settings/userSettingsSlice.ts
@@ -30,19 +30,6 @@ export const userSettingsSlice = createSlice({
       state.settings = action.payload;
       state.loaded = true;
     },
-    // Apply a partial settings update (e.g. from a PATCH response).
-    mergeSettings: (state, action) => {
-      const payload: TournesolUserSettings = action.payload ?? {};
-      (Object.keys(payload) as (keyof TournesolUserSettings)[]).forEach(
-        (scope) => {
-          state.settings[scope] = {
-            ...state.settings[scope],
-            ...payload[scope],
-          } as TournesolUserSettings[typeof scope];
-        }
-      );
-      state.loaded = true;
-    },
   },
   extraReducers: (builder) => {
     // Replace all user's settings of all polls by new ones.
@@ -58,7 +45,6 @@ export const userSettingsSlice = createSlice({
 
 export const selectSettings = (state: RootState) => state.settings;
 
-export const { clearSettings, replaceSettings, mergeSettings } =
-  userSettingsSlice.actions;
+export const { clearSettings, replaceSettings } = userSettingsSlice.actions;
 
 export default userSettingsSlice.reducer;
--- a/frontend/src/features/settings/userSettingsSlice.ts
+++ b/frontend/src/features/settings/userSettingsSlice.ts
@@ -2,7 +2,10 @@ import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
 import { RootState } from 'src/app/store';
 import { TournesolUserSettings, UsersService } from 'src/services/openapi';
 
-export const userSettingsInitialState: { settings: TournesolUserSettings } = {
+export const userSettingsInitialState: {
+  settings: TournesolUserSettings;
+  loaded?: boolean;
+} = {
   settings: {},
 };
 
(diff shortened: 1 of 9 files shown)


frontend/src/pages/home/videos/sections/RecommendationsSection.tsx
```diff
@@ -24,48 +71,142 @@ const RecommendationsSection = () => {
...
+                  boxShadow: 'none',
+                },
+                '&:focus-visible': {
+                  outline: `2px solid ${theme.palette.primary.main}`,
+                  outlineOffset: 2,
+                },
+              })}
+            >
+              {t('home.hideIntro')}
+            </Button>
```
amatissart, line 198: Thanks for the suggestion @venturqx !

When discussing this internally, we were a bit concerned that the "hide" action action is currently irreversible and its purpose might not be immediately clear to users.  We’d be happy to merge this if the button is converted into a toggle so the intro can be expanded again. For example with something like this before the `<TitleSection>`:
```jsx
      <Box textAlign="right" mt={1}>
        <ToggleButton
          size="small"
          sx={{
            border: "none",
            color: "rgba(255, 255, 255, 0.8)",
          }}
          value={isIntroHidden}
          onClick={(_event, value) => setIntroHidden(!value)}
        >
          {isIntroHidden ? <ExpandMore/> : <Close />}
        </ToggleButton>
      </Box>
``` 

@venturqx What do you think? Would that work for you, even though the toggle takes up a bit of space at the top of the page?

On a broader note, this highlights that our current home page is primarily designed for people who are not yet familiar with Tournesol. It could make sense to serve different content to logged-in users, similarly to what is implemented in `<PwaEntryPoint>`  for app users. This could be considered as a separate task.

GresilleSiffle: @amatissart @venturqx 

Do we really want to allow anonymous users to hide the introduction? What are our motivations? The first paragraph on the home page is one of our best place to explain what the project is about and to raise the awareness on the four pillars of the manifesto.

I think it's okay to let authenticated users customise the UI, as long as they can revert their actions. But regarding anonymous users, we could miss an opportunity to clarify what are the intentions behind the Tournesol project.

If a user really don't want to create an account, and is really interested by the top recommendations, and is often going to on the home page to see them, we have two alternatives that I think are better:
- installing the app, for anonymous users the home page is the top recommendations;
- or bookmark the top recommendations using a mobile web browser

I think hiding the introduction to anonymous users when the website is displayed by a mobile web browser will benefit only the persons who took the time to search tournesol.app in their search bar. I guess that people don't do that often, except if it's the first time they are looking for Tournesol, in this case we want to display the introduction. And if they actually do that often, what about bookmarking the top recommendations instead.

Depending on your opinion @amatissart @venturqx , I'd suggest to always display the introduction for anonymous users. This will also simplify a bit the code : )

venturqx: Hi @amatissart and @GresilleSiffle,
Thank you for taking the time to review my suggestion!
Indeed, hiding the intro for anonymous users doesn't seem justified, and if added, it should be a reversible preference.

Here 8099e22 :
- I converted the button to a collapse/expand button (following @amatissart's suggestion).
- I restricted the collapse/expand to authenticated users only.

It looks like this
| Expanded | Collapsed |
|--------|--------|
| expanded_m | collapsed_m note: the collapsed text is a link to the `/manifesto` page. |

What do you think?

GresilleSiffle: The visuals look very good :+1: 

We will try to merge this work by the end of the week. If we find any "problem" we will fix them on the go.

frontend/src/features/settings/userSettingsSlice.ts
```diff
@@ -20,22 +23,42 @@ export const userSettingsSlice = createSlice({
...
       state.settings = {};
+      state.loaded = false;
     },
     // Replace all user's settings of all polls by new ones.
     replaceSettings: (state, action) => {
       state.settings = action.payload;
+      state.loaded = true;
+    },
+    // Apply a partial settings update (e.g. from a PATCH response).
+    mergeSettings: (state, action) => {
```
amatissart: Is it really necessary to introduce `mergeSettings()`? 

The API response for `PATCH /users/me/settings` contains the complete settings (even if a partial object was given in the request), so I would say `replaceSettings()` could be used in all cases.

amatissart (approved):
Looks very good to me. Thanks for integrating our suggestions.

Approved with a minor question.

frontend/src/features/settings/userSettingsSlice.ts
GresilleSiffle: I'm about to test `replaceSettings`...

GresilleSiffle: I confirm that `mergeSettings` is not required in our case : )

I strongly suggest to keep `replaceSettings` and delete `mergeSettings` to reduce the complexity of `userSettingsSlice.ts`.

Also, I think it's a good practice in our case to always consider the API as the main source of trust, and to use its responses to fully replace the local settings. A user connected on several devices may have updated its preferences elsewhere, so I think it's safer to avoid manually rebuilding the settings.

GresilleSiffle: Done in: https://github.com/tournesol-app/tournesol/pull/2184/commits/b339325618d851bfd2db1a2bcf7ec4c8014d42e7

PR #2202, opened 2026-04-04
[ext] Autodetect YouTube watch

sigmike:
**Related issue:** #2103

---

### Description

This PR makes the browser extension detect when a video has been watched on YouTube. It mimics the behavior of the frontend when a video is watched there (i.e. the video has been playing for 60% of its duration).

It detects when an ad is playing and doesn't count that play time.

Unlike the front, there's no button to mark as (un)watched. I initially added one, but I found it to not be very useful and taking space in an already crowded area.

Should we add an option in the preferences to disable (or enable) this feature?

You've certainly considered that already, but if it's always enabled or enabled by default it might result in a lot more videos added to the database.

### Checklist


GresilleSiffle:
> Should we add an option in the preferences to disable (or enable) this feature?

That's a good question. Also, should the option be enabled or disabled by default?

If no option is provided, every YouTube video watched creates a new `ContributorRating` entry in the database. From a privacy perspective, this will allow the back end administrators to guess with more or less precision, the users' activity on YouTube. Well, in fact it's not much different than users watching videos on Tournesol, or manually marking them as watched.

I'd say, it looks safe to have this option enabled by default.

Now should we have this option? As a precaution, I'd say yes. I'm fine with adding that in a second pull request. An opinion @amatissart ?

> You've certainly considered that already, but if it's always enabled or enabled by default it might result in a lot more videos added to the database.

That's right. I think if we provide this option, it's because we want the users to enable it. So we should be ready to have an increase of videos added to the database.

I'm fine with that. I'm not sure our infrastructure can handle a huge increase in database requests, but we could monitor that by tracking the number of newly added videos. In case of need, we can always temporarily turn off the option, make some performance adjustments, and turn it on again.

GresilleSiffle:
It works well :+1: thanks for your work.

Let's wait for eventual inputs from @amatissart regarding the option and its default behaviour before merging.

GresilleSiffle:
I think we should update our privacy policy before this change right? If so, maybe we should first update our privacy policy, notify the users, and then merge the feature.

GresilleSiffle:
Hello @sigmike , we talked about this pull request during our last weekly meeting, here is the plan:
1. the implementation is good, we would like to merge it, **but with the feature disabled for now**
2. then I'll update the privacy policy
3. then we will add an option in the Preferences page to allow the users to enable/disable this feature (disabled by default)
4. we will make this parameter discoverable in the YT and Tournesol UI

If you are ok with that, I can take care of (1) and (2) this Friday.

sigmike:
Hi @GresilleSiffle. Sure, no problem. I should have some free time soon, so let me know if you'd like me to work on any of those steps (or anything else).

GresilleSiffle:
Hi @sigmike. Sorry I had to work on something else last week, I haven't made any progress since.

Feel free to start working on point (3) and (2), could be directly in this PR.

About the new setting, I suggest the following implementation. The reason behind the choice field instead of a boolean field is to have the possibility in the future to add more options, such as "Mark as watched and rate  later". 

```python
class VideosPollUserSettingsSerializer(GenericPollUserSettingsSerializer):
    #...
    ON_VIDEO_WATCHED_CHOICES = [
        ("DO_NOTHING", "do dothing"),
        ("MARK_AS_WATCHED", "mark as watched"),
    ]

    extension__on_video_watched = serializers.ChoiceField(
        choices=ON_VIDEO_WATCHED_CHOICES, allow_blank=True, required=False
    )
    # ...
```

You can see how we have implemented a choice field setting in this diff (the code probably evolved since, so it's good to also check the current files):
- https://github.com/tournesol-app/tournesol/commit/e56f088d4c65cab04df33bba059b6ec553c279da

The important files are the those in the back end, `VideosPollUserSettingsForm.tsx` and `VideosPollUserSettingsForm.spec.tsx`.

sigmike:
Hi @GresilleSiffle. I added the user setting and made an attempt to update the privacy policy. I tried to keep it short. It may be too short.

GresilleSiffle:
Thank you, I'll have time this week end for the review.

GresilleSiffle (approved):

PR #2223, opened 2026-05-22
[front] ui: add password strength meter to help users set a strong passwd

GresilleSiffle:
### Description

Add a password strength meter to invite users to use a strong password. And increase the minimum length from 8 to 9.

To not mix the responsibilities of the API and the front end, only the API can reject a password, while the front end only displays an indication without blocking the submission.

The library used is suggested by: https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls

### Preview

(Weak and Very weak are swapped in the image, but not in the code)

cap

### Checklist

### DIFF
--- a/frontend/src/components/PasswordStrengthIndicator.tsx
+++ b/frontend/src/components/PasswordStrengthIndicator.tsx
@@ -1,0 +1,61 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { Box, Chip, Typography } from '@mui/material';
+import zxcvbn from 'zxcvbn-ts';
+
+const PasswordStrengthIndicator = ({ pwd }: { pwd: string }) => {
+  const { t } = useTranslation();
+
+  const result = zxcvbn(pwd);
+
+  const passwordStrengths = [
+    {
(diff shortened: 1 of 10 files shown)


frontend/src/components/PasswordStrengthIndicator.tsx
```diff
@@ -0,0 +1,69 @@
...
+          return (
+            <Chip
+              key={strength.score}
+              size="small"
+              label={strength.label}
+              color={strength.color}
+              disabled={disabled}
+              data-disabled={disabled}
+              data-testid={strength.testId}
+            />
```
amatissart, line 62: The list of Chips gets wrapped over 2 lines on small mobiles.
So I would suggest to use a smaller font-size:

```suggestion
              data-testid={strength.testId}
              sx={(t) => ({ [t.breakpoints.down('sm')]: { fontSize: '11px' } })}
            />
```

Alternatively, a `<LinearProgress>` component along with a single label, could be more compact.

amatissart (approved):
nice 👍

PR #2222, opened 2026-05-22
ATproto feed server with feed mentioning Tournesol-approved videos

amatissart:
### Description

Very early draft

### Checklist

- [ ] I added the related issue(s) id in the related issues section (if any)
  - if not, delete the related issues section
- [ ] I described my changes and my decisions in the PR description
- [ ] I read the development guidelines of the [CONTRIBUTING.md][development-guidelines]
- [ ] The tests pass and have been updated if relevant
- [ ] The code quality check pass

[development-guidelines]: https://github.com/tournesol-app/tournesol/blob/main/CONTRIBUTING.md#development-guidelines

PR #2238, opened 2026-06-20
Solidago refactor optimizations

sigmike:
Optimization suggestions for the ChronoFair feed in #2222.

### DIFF
--- a/solidago/src/solidago/functions/post_process/sum_criteria.py
+++ b/solidago/src/solidago/functions/post_process/sum_criteria.py
@@ -72,4 +72,4 @@ class GlobalSumCriteria(PollFunction):
             sum_per_entity(lefts),
             sum_per_entity(rights),
         ))
-        return ScoringModel(directs=DirectScores(rows, columns=["entity_name", "criterion", "value", "left", "right"]))
+        return ScoringModel(directs=DirectScores(rows, columns=["entity_name", "criterion", "value", "left_unc", "right_unc"]))
--- a/solidago/src/solidago/functions/post_process/sum_criteria.py
+++ b/solidago/src/solidago/functions/post_process/sum_criteria.py
@@ -2,6 +2,9 @@ from copy import deepcopy
 from functools import reduce
 from typing import Hashable
 
+import numpy as np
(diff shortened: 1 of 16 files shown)


solidago/src/solidago/functions/post_process/sum_criteria.py
```diff
@@ -40,15 +43,33 @@ def __init__(self, weights: dict[str, float] | None = None, aggregated_name: str
         self.aggregated_name = aggregated_name
 
     def fn(self, global_model: ScoringModel) -> ScoringModel:
-        scores = list()
-
-        for (e,), subscores in global_model().iter("entity_name"):
-            v, l, r = self._add(subscores).to_triplet()
-            scores.append((e, self.aggregated_name, v, l, r))
-
-        return ScoringModel(directs=DirectScores(scores, columns=["entity_name", "criterion", "value", "left", "right"]))
```
sigmike, line 49: Possible issue: the `left`/`right` column names here might not be right — the rest of the code seems to use `left_unc`/`right_unc`, so these columns may not be getting read back, which I think would make the global model's uncertainties come out as 0. Could be wrong, so the behavior is left unchanged here — just flagging in case it's worth a look.

lenhoanglnh: The issue was in my code. Indeed it should be left_unc / right_unc

lenhoanglnh (changes requested):
As sigmike said I made an error with left/right as opposed to left_unc/right_unc.
Otherwise it looks good to me.

solidago/src/solidago/functions/post_process/sum_criteria.py
amatissart: fixed in 1a6c910b71f94

amatissart:
Solidago's test suite is green. Other failures are expected as the "backend" on this branch is very old and incompatible with Python 3.13. Let's merge this to its base branch, to keep https://github.com/tournesol-app/tournesol/pull/2079 as the reference for the algorithms used in #2222.

PR #2239, opened 2026-07-07
[ext] Add YouTube banner to choose the on-video-watched setting

sigmike:
**related issues** #2103, #2239

---

### Description

This PR adds a banner on YouTube to make users decide what the extension should do when a video is watched. The banner is placed above the Tournesol buttons and will appear on each YouTube video watch page until the setting is set.

I chose a banner because a button+popup would be confusing if we don't have enough room to display text on the button (and even if we did, the text also seemed difficult to choose). The banner is not very invasive, it only takes some vertical space, and it's not meant to stay. The user can ignore it as long as they want, and it disappears the moment they make a decision.

On a small screen:
Screenshot of the banner on a small screen

On a wider screen:
Screenshot of the banner on a wider screen

The banner only appears if:

- the user is logged in on Tournesol
- the settings fetch is successful
- the extension setting "When a video is watched on YouTube" is not set (i.e. the user has never saved their preferences since the setting was added)

As soon as the user clicks on one of the buttons:

- the setting is set,
- the banner disappears (and will never appear again for this user),
- and unless the user chose "Do nothing" the extension starts monitoring the watch time on the current video (it misses the previously watched time, though).

So the "Do nothing" button acts as a "dismiss" and "don't ask me again" button. We could also add an "X" button to just dismiss and hide the banner temporarily, but the condition under which it would appear again is not obvious.

### Checklist


GresilleSiffle (approved):
Good job 👍

PR #2243, opened 2026-07-17
[back] feat: support Bilibili videos in the videos poll

samuelnihoul:
Related issue: none — happy to open one if you prefer tracking this feature in an issue first.

## Description

This PR adds **backend support for Bilibili videos** in the `videos` poll, as a first step towards supporting video platforms other than YouTube.

Bilibili is the leading video platform for Chinese-speaking audiences, and its catalog (notably science / education content) is largely disjoint from YouTube's, so this would open Tournesol to a large new pool of content and contributors.

### Design

Rather than introducing a new entity type, this PR leans on the existing namespace dispatch built into `EntityType.get_uid_regex(namespace)`: the `video` entity type now accepts a second UID namespace, `bili:<BV-id>` (e.g. `bili:BV1GJ411x7h7`), next to `yt:<video-id>`. The `source` metadata field distinguishes the platforms (`youtube` / `bilibili`).

This keeps Bilibili videos inside the existing `videos` poll — comparable with YouTube videos — with **no DB migration and no API schema change** (the generated OpenAPI spec is unchanged, since entity `metadata` is a free-form object in the schema).

### Changes

- `utils/api_bilibili.py` (new): metadata fetching through the public Bilibili web API (`api.bilibili.com/x/web-interface/view`). No API key required. Maps the response to the existing `VideoMetadata` shape (`name`, `description`, `uploader`, `channel_id`, `publication_date`, `duration`, `views`, …).
- `entities/video.py`: `VideoEntity` accepts the `bili` namespace, and `update_metadata_field()` dispatches to the YouTube or Bilibili client based on the UID namespace.
- `serializers/metadata.py`: `video_id` accepts either a YouTube id or a Bilibili BV-id; new optional `thumbnail_url` field, since Bilibili thumbnail URLs (unlike YouTube's) cannot be derived from the video id and must be stored at fetch time.
- `models/entity.py`: `create_from_video_id()` / `get_from_video_id()` take an optional `namespace` (default `yt`, so all existing callers are unaffected); `link_to_tournesol()` now builds the URL from `self.uid` instead of hardcoding `yt:`.
- `serializers/entity.py`: the dynamic entity creation on comparison/rate-later now passes the validated UID namespace through, so comparing a `bili:` UID in a video poll creates the entity from the Bilibili API.

### Tests

- `entities/test_video_bilibili.py`: UID regex validation (base58 BV-ids, non-interchangeability of namespaces), `VideoMetadata` video_id validation, and metadata refresh dispatch (Bilibili UID → Bilibili API, YouTube UID → YouTube API).
- `tests/test_utils_api_bilibili.py`: response mapping, video-not-found error codes, unexpected error codes, request errors.
- `tests/test_api_comparison.py`: a comparison with a non-existing `bili:` UID dynamically creates the entity with metadata fetched from the Bilibili API (mocked).

### Out of scope / follow-ups

This PR is intentionally backend-only so it stays reviewable. If the approach suits you, I'd follow up with:

1. **Front end**: parse `bilibili.com/video/BVxxxx` URLs in the entity selector / rate-later form, render Bilibili embeds (`player.bilibili.com`) and thumbnails (`metadata.thumbnail_url`) in `EntityImagery`, and stop hardcoding the `yt:` prefix.
2. **Previews**: use `metadata.thumbnail_url` in the preview endpoints instead of `img.youtube.com` when the source is not YouTube (currently a Bilibili entity would fall back to YouTube's grey placeholder thumbnail).
3. Possibly language handling: Bilibili content is mostly `zh`, which currently maps to the `simple` Postgres search config.

Notes for reviewers:
- The 11-character edge case: an 11-char string starting with `BV` is indistinguishable from a YouTube video id, but since the platform is determined by the UID namespace (not the id shape), this is only a theoretical concern for the shared `video_id` metadata regex.
- The Bilibili API needs a browser-like `User-Agent` and `Referer` header, hence the hardcoded values in `api_bilibili.py`.

## Checklist

- [ ] I referenced the issue id, if any.
- [x] I added a description of the problem and its solution.
- [x] I read the [guidelines about contributing](https://github.com/tournesol-app/tournesol/blob/main/CONTRIBUTING.md), including the code quality requirements.
- [x] The automated tests pass, and I updated them if required.
- [x] The automated code quality checks pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/tournesol/entities/test_video_bilibili.py
+++ b/backend/tournesol/entities/test_video_bilibili.py
@@ -1,0 +1,76 @@
+import re
+from unittest.mock import patch
+
+from django.test import SimpleTestCase, TestCase
+
+from tournesol.serializers.metadata import VideoMetadata
+from tournesol.tests.factories.entity import BilibiliVideoFactory, VideoFactory
+
+from .video import BILIBILI_UID_NAMESPACE, YOUTUBE_UID_NAMESPACE, VideoEntity
+
+
+class BilibiliUidRegexTestCase(SimpleTestCase):
+
+    def test_uid_regex_matches_bilibili_uid(self):
+        regex = VideoEntity.get_uid_regex(BILIBILI_UID_NAMESPACE)
+        self.assertTrue(re.fullmatch(regex, "bili:BV1GJ411x7h7"))
+
+    def test_uid_regex_rejects_invalid_bilibili_id(self):
+        regex = VideoEntity.get_uid_regex(BILIBILI_UID_NAMESPACE)
+        # "0", "I", "O" and "l" are not part of the base58 alphabet
+        self.assertIsNone(re.fullmatch(regex, "bili:BV0GJ411x7hI"))
+        # A YouTube video id is not a valid Bilibili video id
+        self.assertIsNone(re.fullmatch(regex, "bili:NeADlWSDFAQ"))
+
+    def test_uid_namespaces_are_not_interchangeable(self):
+        yt_regex = VideoEntity.get_uid_regex(YOUTUBE_UID_NAMESPACE)
+        bilibili_regex = VideoEntity.get_uid_regex(BILIBILI_UID_NAMESPACE)
+        self.assertIsNone(re.fullmatch(yt_regex, "yt:BV1GJ411x7h7"))
+        self.assertIsNone(re.fullmatch(bilibili_regex, "yt:NeADlWSDFAQ"))
+
+
+class VideoMetadataVideoIdTestCase(SimpleTestCase):
+
+    def test_video_id_accepts_youtube_id(self):
+        serializer = VideoMetadata(data={"video_id": "NeADlWSDFAQ"})
+        self.assertTrue(serializer.is_valid(), serializer.errors)
+
+    def test_video_id_accepts_bilibili_id(self):
+        serializer = VideoMetadata(data={"video_id": "BV1GJ411x7h7"})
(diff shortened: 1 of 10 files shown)


GresilleSiffle:
Hi @samuelnihoul 

Thank you for taking the time to contribute to the project, we really appreciate that.

Unfortunately Bilibili is not a platform we can afford to support right now. Still, your work is very valuable as it can help us in the future to add new platforms.

On our side, we are still focusing on the key features tournesol.app is lacking, such as channel subscriptions and feed customization. We don't have the resources to completely support a new platform, especially one that is not as documented as the YouTube API.

I suppose adding Bilibili also implies updating the front end to support all related languages and having the capacity to communicate with the Bilibili community to promote tournesol.app. For now, this is clearly out of scope.

> Related issue: none — happy to open one if you prefer tracking this feature in an issue first.

Usually [we prefer to discuss a feature](https://github.com/tournesol-app/tournesol/blob/main/CONTRIBUTING.md#missing-feature) before implementing it. Do not hesitate to reach us before submitting pull requests of that scale ; )

> 🤖 Generated with Claude Code

We don’t have a specific policy for pull requests that are written mostly (or entirely) by AI. For now, we expect developers to fully understand the submitted code and its implications. That's also why we prefer to give a feedback on the feasibility of the feature before it is implemented.

Thanks again. I'm gonna close the PR so that we can still refer to this code in the future.

PR #2249, opened 2026-07-24
[front] Vite upgrade

sigmike:
This PR upgrades Vite to version 8, Vitest to version 4 and `@vitejs/plugin-react` to version 5.

The unit tests pass, the e2e tests pass, the dev-env works, and an ansible deployment to a vm works.

I changed the way we import `redux-persist` because they dropped a compatibility with some CJS imports are made. As an alternative we can enable the legacy option `inconsistentCjsInterop` instead. It seems `redux-persist` is the only affected import.

I removed `rollupOptions.output.manualChunks` option because it doesn't seem to be necessary anymore (no warning about circular dependencies).

I rewrote the `fix-jsx-runtime` plugin for Rolldown. The previous one still worked but it's cleaner this way.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -58,7 +58,7 @@
     "@types/redux-mock-store": "^1.0.3",
     "@typescript-eslint/eslint-plugin": "^5.8.1",
     "@typescript-eslint/parser": "^5.8.1",
-    "@vitejs/plugin-react": "^4.7",
+    "@vitejs/plugin-react": "^5",
     "eslint": "^8",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^4.0.0",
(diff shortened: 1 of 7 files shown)


sigmike:
The frontend tests keeps failing with a timeout. They pass on my system. I don't think the upgrade slowed it down because on my system the test takes about the same time on main. Maybe we should increase the timeout?

amatissart:
> The frontend tests keeps failing with a timeout. They pass on my system. I don't think the upgrade slowed it down because on my system the test takes about the same time on main. Maybe we should increase the timeout?

I increased the timeout in https://github.com/tournesol-app/tournesol/pull/2249/commits/68fb6211fa3b142058d9817d1f2e28925ac6a4f9

I also moved the custom "fix-jsx-runtime" plugin to the "test" section to clarity this is only required for testing. I guess the proper fix would be to upgrade to a newer React soon.

amatissart (approved):

PR #2252, opened 2026-07-30
Add Safari extension build support

nbonamy:
## What changed

- add Safari as a supported manifest-generation target
- generate a Safari-compatible Manifest V2 build without unsupported permissions or options
- use full-path Safari host patterns so background API requests are authorized
- fall back to tabs.onUpdated because Safari does not expose the webNavigation namespace
- handle failed API fetches without crashing the background page
- add a macOS/Xcode packaging script using Apples Safari Web Extension packager
- document local packaging, signing, and enablement in Safari

## Why

Tournesol publishes Chrome and Firefox extensions, but Safari users currently have no build target. Reusing the existing Manifest V2 background page preserves its ECMAScript module imports; the Safari packager does not support the Manifest V3 type: module service-worker declaration used by the Chrome build.

Safari also lacks the webNavigation namespace exposed to the existing build and requires complete scheme, host, and path match patterns for website access. The compatibility fallbacks preserve navigation and API behavior without crashing the background page.

## User impact

A developer on macOS can run ./build-safari.sh, select a signing team in the generated Xcode project, and build the Tournesol Safari Web Extension from the same source as Chrome and Firefox.

## Validation

- yarn lint
- Firefox MV2, Chrome MV3, and Safari MV2 manifest-generation assertions
- Safari MV3 rejection assertion
- Safari manifest assertions for wildcard host permissions and omitted webNavigation permission
- live Tournesol API reachability checks for anonymous settings and recommendations
- Safari Web Extension packaging with no manifest warnings on Xcode 26.6
- unsigned Xcode Release build of the generated host app and extension

### DIFF
--- a/browser-extension/.gitignore
+++ b/browser-extension/.gitignore
@@ -3,3 +3,4 @@ node_modules/
 src/manifest.json
 src/config.js
 src/importWrappers/
+safari/
--- a/browser-extension/README.md
+++ b/browser-extension/README.md
@@ -48,6 +48,36 @@ By default, the script creates an extension that connects to the production Tour
 
 The default is also to build a Manifest V2 extension for Firefox. To build a Manifest V3 extension for Chrome, run `EXTENSION_BROWSER=chrome MANIFEST_VERSION=3 yarn configure`.
 
+To prepare a Manifest V2 extension compatible with Safari, run:
+
+```shell
+yarn configure:safari
+```
+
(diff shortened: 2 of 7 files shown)


GresilleSiffle:
Hello @nbonamy 

Thanks for the pull request :+1: We will test that as soon as possible. We are waiting for our macOS developer to be available ;)

@tournesol-app/maintainers do we actually want to support Safari (vs pushing free software like Firefox)?

PR #2266, opened 2026-09-01
[front] Upgrade to React 18

sigmike:
This PR upgrades React from 17 to 18.

The upgrade itself was not really a problem: it was mostly type fixes and using `createRoot` (2039cede73fdf4d2ee28519fc6423461c2f4d8e1).

The difficulty was the [StrictMode](https://18.react.dev/reference/react/StrictMode) because of the [effects re-running](https://18.react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-effects-in-development). This `StrictMode` change triggers problems only in development so we can also choose to disable it.

I identified 3 problems and fixed them as described below. I'm pretty confident there are no other problems, but I might have missed some.

## The "back" button on the search page

The back parameters were stored in redux when the feed pages mount, and were cleared when the "back" button was unmounted. The mount/unmount/remount while preserving state of the new `StrictMode` made the "back" button clear the back parameters when it appeared.

There was no easy fix for that. The options I examined were (there may be others):

1. keep the back parameters in redux and clear them on each event that moves the user to the search page. Error prone, we might forget about that.
2. move the back parameters into the URL. Would be kind of ugly and it's probably not something users would want to carry if they share the URL.
3. move the back parameters into the location.state. It's hidden from the user and attached to the location which seems like a good fit. But since the search page rewrites the URL when the search parameters change, we have to propagate the state during these operations.

I chose option 3, and added a new hook to "update the query string while preserving the state" (`useUpdateSearchParams`).

Done in 79a02c7fff4430388b6bcb5d26d35030b8b0f77e.

## The comparison sliders' mount check

`ComparisonSliders` had a guard against changing the component state while unmounted because of React 17's warning about that. It broke in StrictMode (because the mount/unmount/remount kept `isMounted` false) and React 18 removed this warning so I just removed the guard.

Done in 09809b09a040aea24da3a015051aa56eead5f9be.

## Comparison auto-fill

The mount/unmount/remount didn't work well with the `initializeWithSuggestions` guard. The component was unmounted before the auto-fill finished so a second auto-fill was triggered. So I set the "auto-fill already started" flag earlier, and added a new state that let the comparison fetch effect run only when the auto-fill is complete.

Done in 4848d0044a4a758c15947f37f1a02e3cdcdbfcbb.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,11 +11,11 @@
     "@mui/styles": "^6.4.11",
     "@react-hook/resize-observer": "^1.2.5",
     "@reduxjs/toolkit": "^1.7.1",
-    "@types/react": "^17",
-    "@types/react-dom": "^17.0.11",
+    "@types/react": "^18",
+    "@types/react-dom": "^18",
     "@types/react-redux": "^7.1.7",
     "@use-gesture/react": "^10.3.1",
-    "i18next": "^22.4.0",
+    "i18next": "^23",
     "i18next-browser-languagedetector": "^8.0.0",
     "i18next-http-backend": "^3.0.5",
     "linkify-string": "^4.1",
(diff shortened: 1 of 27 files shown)


amatissart (approved):
Looks good to me. Thanks @sigmike !

Regarding e2e tests, I could also run them locally with success. As discussed, they can perform poorly when using the dev-env, as Cypress introduces some overhead on network requests. It seems this is particularly visible against Vite dev server (see https://github.com/cypress-io/cypress/issues/22968) .

That's why we are using `FRONTEND_START_SCRIPT=start:e2e` to start the dev-env in the CI, in order to run tests agains a production bundle instead of Vite's server. That's something we should document better in tests/README.md

Issue #2264, opened 2026-08-23
Design of the subscriptions API


GresilleSiffle:
## API

A subscription is not poll-specific.

`POST /users/me/subscriptions/`
- the logged-in user subscribes to a new channel

`GET /users/me/subscriptions/`
- list all my subscriptions

`GET /users/me/subscriptions/`
- list all my subscriptions

To define, solution 1

```json
[
  {
     "entity_source": { ... }
     "volume":
     "etc.":
  }
]
```

Or solution 2

```json
[
  {
     "entity_source": { ... }
     "subscription_data": { ... }
  }
]
```

`GET /users/me/subscriptions/{source_uid}`
- retrieve a user subscription 
- example for the poll "videos" `yt:UCs2UszeXZlQ-DsHI667DONA`
  - `yt:` is a const in the back end

`PUT /users/me/subscriptions/{source_uid}`

`PATCH /users/me/subscriptions/{source_uid}`

`DELETE /users/me/subscriptions/{source_uid}`

In the future
- `POST /users/me/subscriptions/_bulk_create` to import YT subscriptions from the browser extension

## Database tables

We prefer the solution A.

Models without poll (solution A)

```
subscription
- user_id:
- entity_source_uid:


entity_source
- uid
- entity_type
- metadata -> JSON to define, like the metadata field of the entity table
```

Models with poll (solution B)

```
subscription
- poll_id:
- user_id:
- entity_source_uid:


entity_source
- uid
- metadata
```

## Questions

Is it relevant to have != subscriptions per poll?

PR #2276, opened 2026-09-16
[back] Channel subscription

sigmike:
**related issue** #2264

---

### Description

The PR implements the channel subscription API.

Choices I made compared to the issue:

- Subscriptions are not poll specific because I think it would not make much sense to the user.
- No `PUT/PATCH /users/me/subscriptions/{source_uid}` because we don't have anything to change on the subscriptions (yet).
- No `POST /users/me/subscriptions/_bulk_create` because this will likely be done internally with the YouTube API instead.
- On the subscription model the entity source is linked by its internal id, not its uid (the issue said `entity_source_uid`).
- I didn't add `entity_type` to `entity source` because I'm not sure how it'll be filled or used so I left it to a future change when it's needed.
- The subscription data (only `created_at` for now) are returned at the root of the subscription object, instead of a nested `subscription_data`. I think it's more simple like that and I don't see where it would be a problem.

Other choices I made:

- Channel metadata are fetched on subscription if the source doesn't exist. If the YouTube channel doesn't exist then a 400 error is returned. If there's any other error then the source is created with empty metadata.
- Source metadata are never refreshed. It will likely be the process that fetches the new videos of every channel that will also update the metadata.
- I didn't create a serializer for the source metadata (there's one on the videos). I'm not sure we need one, and it can be added later if we do.

### Checklist


GresilleSiffle:
> Subscriptions are not poll specific because I think it would not make much sense to the user.

Good.

> No PUT/PATCH /users/me/subscriptions/{source_uid} because we don't have anything to change on the subscriptions (yet).

OK

> No POST /users/me/subscriptions/_bulk_create because this will likely be done internally with the YouTube API instead.

OK

> On the subscription model the entity source is linked by its internal id, not its uid (the issue said entity_source_uid).

Perfect.

> I didn't add entity_type to entity source because I'm not sure how it'll be filled or used so I left it to a future change when it's needed.

I think it could be useful in the front end, but I'm not 100% sure.

When an authenticated user wants to list its subscriptions while navigating the poll  `videos`, it could be expected to display only the sources that exist in this poll, on our case only YouTube channels. I said I'm not 100% sure because we can still filter the sources based on the UID namespace.

For a given poll, we know the related namespaces, so we know which are the related sources.

The drawback of that, is the fact that guessing is less explicit than filtering on field `entity_type`. We rely on an implementation convention: the namespace of an entity type is included in the namespace of an entity source, and we expect this convention to last. It could be seen a less Pythonic than using a dedicated field. The SQL filtering could also be a little less performant, but I'm not sure.

Well, this is not blocking at all for this pull request. Let's talk about it during a meeting.

> The subscription data (only created_at for now) are returned at the root of the subscription object, instead of a nested subscription_data. I think it's more simple like that and I don't see where it would be a problem.

OK

> Channel metadata are fetched on subscription if the source doesn't exist. If the YouTube channel doesn't exist then a 400 error is returned. If there's any other error then the source is created with empty metadata.

This is how the rate-later API works. It may not be ideal, but it's good to have a coherent behaviour.

> Source metadata are never refreshed. It will likely be the process that fetches the new videos of every channel that will also update the metadata.

OK, let's create an issue to address that.

> I didn't create a serializer for the source metadata (there's one on the videos). I'm not sure we need one, and it can be added later if we do.

It can be useful. Especially if we add the subscriptions in the Django admin. It will ensure that we don't insert malformed data in the database.

See the method `clean` of the model `Entity` https://github.com/tournesol-app/tournesol/blob/36b6b805ae6b626a0aaf9067c908d3fb0b6591c2/backend/tournesol/models/entity.py#L407

Yes, let's do it later.

GresilleSiffle (approved):
Looks good for me.

@amatissart do you also want to have a look?

